PR #1, opened 2017-06-23
Adding BQ OMOP tools

danqrodney:


jmandel:
Sweet! From the SQL files, this is all CDM v5. Might be good to capture this in file or dir names for our BigQuery schema, too.

PR #2, opened 2017-07-01
Set datetime fields to nullable rather than required

mark-velez:
* As defined by OMOP spec (see https://github.com/OHDSI/CommonDataModel)
 * Datetimes in observation_period will likely be removed (see https://github.com/OHDSI/CommonDataModel/pull/64)

### DIFF
--- a/tools/bq/schemas/condition_occurrence.json
+++ b/tools/bq/schemas/condition_occurrence.json
@@ -22,7 +22,7 @@
     {
         "type": "timestamp", 
         "name": "condition_start_datetime", 
-        "mode": "required"
+        "mode": "nullable"
     }, 
     {
         "type": "date", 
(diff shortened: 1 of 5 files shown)


mark-velez:
Were `tools/bq/schemas/*.json` auto-genned from a DDL in https://github.com/OHDSI/CommonDataModel? If so, where from (so I might report bug)?

danqrodney:
Yeah, they were generated from e.g.  https://github.com/OHDSI/CommonDataModel/blob/master/PostgreSQL/OMOP%20CDM%20ddl%20-%20PostgreSQL.sql#L347

where start timestamps are not nullable.

If that's wrong, it seems to be wrong with the DDL, too.

What's the point of adding these columns but allowing them to be null when the date version isn't? (Shouldn't they always have the date value's timestamp in there?)

danqrodney:
(I do understand allowing them to be nullable at least temporarily when upgrading an existing database, but it seems like for analysis purposes you'd want all the non-nullable date fields to have corresponding non-nullable datetime fields.)

mark-velez:
No longer relevant

PR #5, opened 2017-08-29
changes to tests. bq bad table name works now.

aahladmanas:
Added ValueError to bq_utils.load_table_from_bucket when the table name is not recognized. removed the test_cron_run.


aahladmanas:
Bad request.

mark-velez:
@aahladmanas I will close this so you can do the rebase before the PR like we discussed.

PR #6, opened 2017-08-29
bq bad_table_name test passes + error message. added warning-files test.

aahladmanas:


data_steward/bq_utils.py
```diff
@@ -36,7 +36,13 @@ def load_table_from_bucket(hpo_id, cdm_table_name):
     :param cdm_table_name: name of the CDM table
     :return: an object describing the associated bigquery job
     """
-    assert (cdm_table_name in common.CDM_TABLES)
+    try:
+        if cdm_table_name not in common.CDM_TABLES:
+            raise ValueError
```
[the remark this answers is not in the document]
aahladmanas: That should happen. The err message is added in the next line.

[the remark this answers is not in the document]
aahladmanas: I improved that line. Check the latest commit.

aahladmanas:
So I added the new test that checks the creation of a warning_unknown_files.csv that contains the list of files that we don't handle with the header 'file_name'. For example : 

> "file_name"
"person_xyz.csv"
"condition_occurrence_NO.csv"
"visit_occurrence_aaaaaah.csv"
"procedure_occurrence_best.csv"

data_steward/test/unit_test/validation_test.py
```diff
@@ -93,6 +95,27 @@ def test_all_files_unparseable_output(self, mock_check_cron):
                 expected = f.read()
                 self.assertEqual(expected, actual_result)
 
+    @mock.patch('api_util.check_cron')
+    def test_bad_file_names(self, mock_check_cron):
+        exclude_file_dic_list = resources._csv_to_list(WARNING_UNKNOWN_FILES)
```
mark-velez: Good idea to test bad file names! I suggest we use hard-coded strings here. Hard-coded strings will make the cases we are testing much more obvious and more independent (i.e. dropping our dependency on `_csv_to_list`). This great, I'm glad we are thinking seriously about handling the bad input we will inevitably get.

data_steward/test/unit_test/validation_test.py
```diff
@@ -9,13 +9,15 @@
 from validation import main
 import resources
 import common
+import copy
```
mark-velez: Is this used?

aahladmanas: No. I had that for a test I wrote that became redundant. Will remove it.

data_steward/test/unit_test/validation_test.py
aahladmanas: Awesome! hardcoding will be readable yes. will do this also.

PR #16, opened 2017-10-02
Adding value_source_concept_id, value_source_value, questionnaire_res…

danqrodney:
…ponse_id to observation schema.


mark-velez:
A reminder that we should ensure that any copying around of data during curation preserves the columns added to observation here.

PR #17, opened 2017-10-06
5.2 schema update

RobertJCarroll:
Do not merge for now- this is the VUMC team's update of the schema files to support OMOP v5.2

### DIFF
--- a/data_steward/resources/fields/note.json
+++ b/data_steward/resources/fields/note.json
@@ -55,13 +55,13 @@
         "mode": "nullable"
     }, 
     {
-        "type": "integer", 
-        "name": "visit_occurrence_id", 
-        "mode": "nullable"
-    }, 
-    {
         "type": "string", 
         "name": "note_source_value", 
         "mode": "nullable"
-    }
+    },
+    {
+        "type": "integer", 
+        "name": "visit_occurrence_id", 
+        "mode": "nullable"
+    } 
 ]
--- a/data_steward/resources/fields/note.json
+++ b/data_steward/resources/fields/note.json
@@ -25,8 +25,28 @@
         "mode": "required"
     }, 
     {
+        "type": "integer", 
+        "name": "note_class_concept_id", 
+        "mode": "required"
+    }, 
+    {
+        "type": "string", 
+        "name": "note_title", 
+        "mode": "nullable"
+    }, 
+    {
         "type": "string", 
         "name": "note_text", 
+        "mode": "required"
+    }, 
+    {
+        "type": "integer", 
+        "name": "encoding_concept_id", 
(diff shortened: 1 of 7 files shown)


data_steward/resources/fields/note.json
```diff
@@ -25,12 +25,32 @@
         "mode": "required"
     }, 
     {
+        "type": "integer", 
+        "name": "note_class_concept_id", 
+        "mode": "required"
```
danqrodney: Dumb question: how can we add required fields / make existing fields required? Won't old versions of OMOP data lack these?

(Can we force all data going forward to have these fields?)

RobertJCarroll: Good question. Short answer is the data is not directly portable up in that fashion. We can fill 0s as an interim solution, but it's not really the "right" solution.

We may end up with different OMOP versions over time in the CDR; we probably need a bigger picture answer to this as well. They try and keep new specs compatible with old code, but with a big version increment coming it may not be entirely true.

RobertJCarroll: (And as far as I am concerned, they don't have to be required; I think this is just following spec)

jmandel:
Mostly I wanted to get these listed here so we could have this discussion :-)

Do we know what version of OMOP we're targeting for:

* EHR site uploads
* First-pass CDR publication

and are they different?

cukarthik:
Sorry if I missed something. Is there a reason we are targeting 5.2 vs 5.1?

RobertJCarroll:
This was more due to needing to get a working set for 5.2 posted for the sample data.

I'm inclined to push to 5.2 as the biggest changes are more around notes which we aren't capturing. The rest could be left missing, but could add value if sites can implement.

cukarthik:
The other change is an additional column in the condition occurrence table. We should decide if we want to just do 5.2 for everything including HPOs.

data_steward/resources/fields/note_nlp.json
```diff
@@ -0,0 +1,72 @@
...
+        "mode": "nullable"
+    }, 
+    {
+        "type": "string", 
+        "name": "snippet", 
+        "mode": "nullable"
+    }, 
+    {
+        "type": "string", 
+        "name": "\"offset\"",
```
jmandel: @RobertJCarroll is this a bu with quotes?

RobertJCarroll:
@cukarthik One can just leave those 2 fields empty re status if needed. It's an optional field.

Are you internally at 5.2? I just presume folks try and keep up to date!

cukarthik:
I think we have most of this now since we are on 5.2 and had pulled @danqrodney's PR with the descriptions.

PR #24, opened 2017-11-08
run_achilles_and_export script added to repo

aahladmanas:
This script is able to run Achilles, Achilles heel and export on a data-set which contains some cdm tables. It requires setting environment variables

APPLICATION_ID, BIGQUERY_DATASET_ID, BUCKET_NAME_{hpo_id}

where, --hpo_id is an argument to the script.


mark-velez:
In the name of keeping things [DRY](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself), can we re-use the functions in `validation.main`? See comments above.

mark-velez:
Also, I was wrong: GitHub still shows on this PR commits that have already been merged into the target branch. Nonetheless, it shouldn't be a problem when we rebase and merge this feature branch.

aahladmanas:
So my thought process was that we might do different things with this because we run this  from command line and so I wanted the functions explicit. But as logging is done for the functions anyway, it's fine. I changed it to use the corresponding functions from validation.main

PR #28, opened 2017-12-20
Fix branch/environment configs

aahladmanas:
Changes to run for all hpos.

1. hpo_csv() now returns all hpos as listed in spec/_data/hpo.csv.
2. spec/_data/hpo.csv. is the latest as of 12/19/2017 5 pm.
2. wait added after queries in achilles and achilles_heel so as to not hit rate limits in bigquery
3. validation/main now has an all run which only looks at the hpos whose buckets exist.
4. spec/main now only looks at the hpos whose buckets exist instead of throwing error.
5. *ignore folder config and files generate_app_yaml.sh and cron_generate.sh for now.*


data_steward/spec/main.py
```diff
@@ -102,7 +102,11 @@ def get_full_result_log():
         hpo_id = hpo['hpo_id']
         hpo_bucket = gcs_utils.get_hpo_bucket(hpo_id)
 
-        obj_metadata = gcs_utils.get_metadata(hpo_bucket, RESULT_CSV)
+        try:
+            obj_metadata = gcs_utils.get_metadata(hpo_bucket, RESULT_CSV)
+        except:
```
mark-velez: It might be better to catch the specific exception raised when the bucket doesn't exist and wrap the remainder of this code block. Don't worry about it, I may look into it on merge.

data_steward/validation/main.py
```diff
@@ -107,67 +107,83 @@ def _upload_achilles_files(hpo_id):
...
-
-    # separate cdm from the unknown (unexpected) files
-    found_cdm_files = []
-    unknown_files = []
-    for bucket_item in bucket_items:
-        if _is_cdm_file(bucket_item):
-            found_cdm_files.append(bucket_item)
-        else:
-            if bucket_item['name'].lower() in common.IGNORE_LIST + common.CDM_FILES:
+    hpo_ids = [hpo_id]
```
mark-velez: Having several changed lines in a critical section like this makes me a bit nervous. Perhaps we can avoid it. What do you think about creating a separate endpoint/function which just invokes the intact function for each `hpo_id`?

aahladmanas: there's flask app +  cron problem. The way it's structured now, we can only call the endpoint and not a function and we can't call the endpoint because that can only be done by cron.

mark-velez: Ack, I thought that might be the case. Thanks for the info. Could we create a dumb function just for the endpoint?

    @api_util.auth_required_cron
    def validate_hpo_files_endpoint(hpo_id):
      return validate_hpo_files(hpo_id)

aahladmanas: possible.

PR #44, opened 2018-04-20
[DC-97] Enable concurrent testing

mark-velez:


data_steward/ci/setup.sh
```diff
@@ -0,0 +1,19 @@
+#!/bin/bash
+
+# Create buckets
+gsutil mb -c nearline -p ${APPLICATION_ID} gs://${DRC_BUCKET_NAME}
```
aahladmanas: We'll need a variable for `chs` also now. Using it for the `merge_test` for now.

data_steward/init_env.sh
```diff
@@ -0,0 +1,47 @@
...
+    exit 1;
+fi
+
+# Dataset IDs must be alphanumeric (plus underscores) and <= 1024 characters
+# See https://cloud.google.com/bigquery/docs/reference/rest/v2/datasets#datasetReference.datasetId
+PROJECT_PREFIX=`echo ${APPLICATION_ID} | sed 's/-/_/g'`
+USERNAME_PREFIX=`echo ${USERNAME} | sed 's/-/_/g'`
+
+# GCS buckets are globally unique, so we prefix with project id and username
+export DRC_BUCKET_NAME=${PROJECT_PREFIX}_${USERNAME_PREFIX}_drc
```
aahladmanas: Same as above. Need a bucket for `chs`

data_steward/ci/setup.sh
```diff
@@ -0,0 +1,19 @@
...
+gsutil mb -c nearline -p ${APPLICATION_ID} gs://${BUCKET_NAME_FAKE}
```
aahladmanas: Should we make all our testing happen in buckets that are nearline?

circle.yml
```diff
@@ -10,63 +10,25 @@ dependencies:
     - cd ~/curation/data_steward/ && pip install -t lib/ -r requirements.txt 
    
 machine:
+  pre:
+    - ~/curation/data_steward/init_env.sh
+    - cat $HOME/.circlerc
```
aahladmanas: Does this create .circlerc?

data_steward/ci/setup.sh
mark-velez: Didn't think about that. Now I'm thinking of changing it to `us-east4` (see [Bucket Locations](https://cloud.google.com/storage/docs/bucket-locations)). What do you think?

circle.yml
mark-velez: No, just shows the contents. It's useful for troubleshooting.

data_steward/ci/setup.sh
mark-velez: Ok I'll add it in advance.

PR #51, opened 2018-05-02
[DC-100] Disable/hide achilles reports that are not yet implemented

cukarthik:
Cleaned up achilles report to only display the three reports we generate and default to the person dashboard.


mark-velez:
This has been merged. Thanks @cukarthik .

PR #54, opened 2018-05-04
[DC-106] Create tables before attempting to load hpo data

mark-velez:


data_steward/validation/main.py
```diff
@@ -174,33 +192,50 @@ def run_validation(hpo_id, error_ignore_flag=False):
         errors = []
         results = []
         found_cdm_file_names = found_cdm_files
+
+        # Create all tables first to simplify downstream processes
+        # (e.g. ehr_union doesn't have to check if tables exist)
         for cdm_file_name in common.CDM_FILES:
+            cdm_table_name = cdm_file_name.split('.')[0]
+            table_id = bq_utils.get_table_id(hpo_id, cdm_table_name)
+            bq_utils.create_standard_table(cdm_table_name, table_id, drop_existing=True)
```
aahladmanas: I don't think we want to drop existing. If this load fails, then we've lost the data we had also.

data_steward/validation/main.py
```diff
@@ -130,31 +139,40 @@ def validate_all_hpos():
     """
     for item in resources.hpo_csv():
         hpo_id = item['hpo_id']
-        run_validation(hpo_id, True)
+        try:
+            run_validation(hpo_id, True)
```
aahladmanas: What happens if some other error happens? We want to continue no matter what happens yes?

data_steward/validation/main.py
```diff
@@ -174,33 +192,50 @@ def run_validation(hpo_id, error_ignore_flag=False):
...
                 incomplete_jobs = bq_utils.wait_on_jobs([load_job_id], retry_count=BQ_LOAD_RETRY_COUNT)
 
                 if len(incomplete_jobs) == 0:
                     job_resource = bq_utils.get_job_details(job_id=load_job_id)
                     job_status = job_resource['status']
                     if 'errorResult' in job_status:
-                        error_messages = ['{}'.format(item['message'], item['location']) for item in job_status['errors']]
-                        errors.append((cdm_file_name, ' || '.join(error_messages)))
+                        # These are issues (which we report back) as opposed to internal errors
+                        issues = ['{}'.format(item['message'], item['location']) for item in job_status['errors']]
```
aahladmanas: (also a codacy issue) we need a '{} at {}' instead of '{}'.

data_steward/validation/main.py
aahladmanas: (Also a codacy issue) Remove the extra argument.

data_steward/validation/main.py
mark-velez: Will this still work? I agree we would lose data which may have been saved in previous load, but if we don't drop existing, we may run into an error when it tries to create a table which already exists. (This may be something to refactor at a later time.)

data_steward/validation/main.py
mark-velez: thank you will do

data_steward/validation/main.py
aahladmanas: So shouldn't we only create tables that don't exist. The table names don't change. So we should be fine.

data_steward/bq_utils.py
```diff
@@ -299,9 +301,8 @@ def query(q, use_legacy_sql=False, destination_table_id=None, retry_count=BQ_DEF
                 'datasetId': get_dataset_id()
             },
             'query': q,
-            'timeoutMs':60000,
-            'useLegacySql': use_legacy_sql,
-            "maxResults": 1000
```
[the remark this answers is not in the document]
aahladmanas: This is what I was looking at. The api library's pydoc page. https://developers.google.com/resources/api-libraries/documentation/bigquery/v2/python/latest/bigquery_v2.jobs.html#query

PR #57, opened 2018-05-15
chart.css modfied to reduce the size of legend, axis text and labels.

juh7007:
@mark-velez - Please review this pull request.. Thanks, Mark!!!

### DIFF
--- a/data_steward/resources/curation_report/css/chart.css
+++ b/data_steward/resources/curation_report/css/chart.css
@@ -5,7 +5,7 @@ div.chartcontainer {
 
 .legend {
 	padding: 5px;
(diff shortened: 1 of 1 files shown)


mark-velez:
Thanks @juh7007. The fonts were somewhat small to begin with on some views and appear even smaller with the new styles (see before and after screenshot below). Does this fix a higher priority issue? Sorry if I'm missing it. Copying @cukarthik for feedback.
image

juh7007:
@mark-velez and @cukarthik - Is this better? I adjusted the size so that it looks better on both person and data density reports.

image

image

PR #58, opened 2018-05-21
Changing import_rdr_omop to set up any table with person_id 

danqrodney:
as clustered on it. (This currently requires the project to be whitelisted.)

Dropping RDR-specific observation table logic.

Changing concept date columns to be strings (as this is more convenient when exporting data to Cloud SQL later.)

### DIFF
--- a/data_steward/bq_utils.py
+++ b/data_steward/bq_utils.py
@@ -104,16 +104,8 @@ def load_csv(schema_path, gcs_object_path, project_id, dataset_id, table_id, wri
             'writeDisposition': 'WRITE_TRUNCATE',
             'allowJaggedRows': allow_jagged_rows
             }
-    field_names = [field['name'] for field in fields]
-    if 'person_id' in field_names:
-        load['clustering'] = {
-            'fields': ['person_id']
-        }
-        load['timePartitioning'] = {
-            'type': 'DAY'
-        }
     job_body = {'configuration': {
-        'load': load
+            'load': load
         }
     }
     insert_job = bq_service.jobs().insert(projectId=project_id, body=job_body)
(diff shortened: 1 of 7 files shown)


data_steward/bq_utils.py
```diff
@@ -41,6 +41,24 @@ def get_table_id(hpo_id, table_name):
...
+    bq_service = create_service()
+    if project_id is None:
+        project_id = app_identity.get_application_id()
+    if dataset_id is None:
+        dataset_id = get_dataset_id()
+    job = bq_service.tables().get(projectId=project_id, datasetId=dataset_id, tableId=table_id)
+    return job.execute(num_retries=BQ_DEFAULT_RETRY_COUNT)
+
+
 def load_csv(schema_path, gcs_object_path, project_id, dataset_id, table_id, write_disposition='WRITE_TRUNCATE',
```
danqrodney: Is this used to create new tables, or just load CSV into existing ones?

(If it's the latter, you don't need to bother with the clustering stuff... clustering only happens at creation time.)

mark-velez: Yes, I think you're right. I will confirm and fix. Thanks @danqrodney! Btw, you may already know this, but TIL from the [circle build](https://circleci.com/gh/all-of-us/curation/59?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link#action-21) that querying tables with clustering is not supported in Legacy SQL. I do not believe we need Legacy SQL in this project (will check).

danqrodney: Yeah, we don't use legacy SQL for anything. It's probably better to avoid it entirely.

danqrodney:
@mark-velez Your changes LGTM. As this is my PR, I can't officially "approve". :)

Can you approve? Do you want to resolve the conflicts, or shall I?

PR #59, opened 2018-05-22
New achilles and export script can use bucket/folder without specifying HPO

aahladmanas:
The following functions were changed to handle hpo_id being None.
When necessary, they were also given a target_bucket argument which
defaults to None
 - bq_utils.get_table_id
 - run_export, run_achilles, upload_achilles_files and save_datasources_json in main
 - export.export_from_path
 - sql_wrangle.qualify_tables


data_steward/validation/main.py
```diff
@@ -52,49 +52,65 @@ def all_required_files_loaded(hpo_id, folder_prefix):
     return True
 
 
-def save_datasources_json(hpo_id, folder_prefix=""):
-    hpo_bucket = gcs_utils.get_hpo_bucket(hpo_id)
+def save_datasources_json(hpo_id=None, folder_prefix="", target_bucket=None):
+    if hpo_id is None and target_bucket is None:
```
mark-velez: As discussed, if all three args are supplied, report should be in `target_bucket` and use `hpo_id` as the name of the data source.

PR #63, opened 2018-05-31
Feature/recategorize heel

cukarthik:


mark-velez:
@cukarthik this has [many of?] the same commits from #61 which was merged as of 5/27. Are there any additions? Perhaps rebasing the branch on develop would help..?

PR #65, opened 2018-06-04
[DC-109] Store EHR person data in observation table

aahladmanas:
Thes PR has the new ehr_person_move_to_observation function and it's test.

### DIFF
--- a/data_steward/test/unit_test/combine_ehr_rdr_test.py
+++ b/data_steward/test/unit_test/combine_ehr_rdr_test.py
@@ -9,9 +9,11 @@ import logging
 import sys
 
 from tools.combine_ehr_rdr import copy_rdr_table, ehr_consent, main, mapping_table_for, create_cdm_tables
+from tools.combine_ehr_rdr import copy_ehr_table, move_ehr_person_to_observation
 from tools.combine_ehr_rdr import DOMAIN_TABLES, EHR_CONSENT_TABLE_ID, RDR_TABLES_TO_COPY
 from google.appengine.ext import testbed
 from tools.combine_ehr_rdr import logger
+from validation.export import
 
 
 class CombineEhrRdrTest(unittest.TestCase):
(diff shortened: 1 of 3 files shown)


data_steward/tools/combine_ehr_rdr.py
```diff
@@ -170,6 +172,88 @@ def copy_rdr_table(table):
...
+              FROM {ehr_dataset_id}.person
+
+              UNION ALL
+
+              --DOB
+              SELECT person_id, 4083587 as observation_concept_id, 38000280 as observation_type_concept_id, 
+              birth_datetime as observation_datetime,
+              NULL as value_as_concept_id,
+              birth_datetime as value_as_string,
+              '' as observation_source_value,
```
mark-velez: @cukarthik is there a reason to use empty string instead of `NULL`?

data_steward/test/unit_test/combine_ehr_rdr_test.py
```diff
@@ -124,23 +126,84 @@ def test_copy_rdr_tables(self):
             self.assertEqual(rdr_count, combined_count,
                              msg_fmt.format(table=table, rdr_count=rdr_count, combined_count=combined_count))
 
+    def test_ehr_person_to_observation(self):
+        # ehr person table converts to observation records
+        self.load_dataset_from_files(self.combined_dataset_id, test_util.NYC_FIVE_PERSONS_PATH)
```
mark-velez: @aahladmanas is it sufficient for this test to start with empty tables instead? Loading from files takes a long time. I will make the change if you think that'd be ok.

data_steward/tools/combine_ehr_rdr.py
cukarthik: I thought it created an error for me when using NULL, but I can't reproduce it. so no. I'll make it NULL.

PR #66, opened 2018-06-04
Added hpo_id for the UC consortium

cukarthik:


danqrodney:
I noticed this PR, and thought I'd mention: RDR maintains a list of awardees, orgs, and sites here:

https://github.com/all-of-us/raw-data-repository/tree/master/rest-api/data

I don't think it's using the same IDs you're using here; and I'm not sure if it would make sense to change that in future. Just figured I should bring it up.

cukarthik:
thanks @danqrodney! we weren't sure the rdr had it at the site level, but it looks like [this file](https://github.com/all-of-us/raw-data-repository/blob/master/rest-api/data/organizations.csv) is what we want to use. we'll work on using this list of codes to stay consistent.

mark-velez:
@cukarthik do you want to continue with this pr for now and handle using the rdr file in a subsequent pr?

cukarthik:
Yes. Making the change of the codes will require more testing.

mark-velez:
This has been merged. Since it is a hotfix, I cherry-picked the relevant commits and merged it to `master`.

PR #75, opened 2018-07-26
DC-155 Deid 

lnyemba:
The issue is a misconfiguration on line 13 that causes City, Zip WorkAddress to be skipped in the observation table.


mark-velez:
Thanks @lnyemba this has been merged to develop

PR #91, opened 2018-09-20
[DC-177] Conservative check on submission data retention before processing

ksdkalluri:


mark-velez:
Merged. I squashed the commits related to data retention check into one.

PR #92, opened 2018-09-21
[DC-179] Attribute HPO site at the row-level for each domain table for any dataset

cukarthik:


data_steward/test/unit_test/combine_ehr_rdr_test.py
```diff
@@ -35,34 +35,42 @@ def setUpClass(cls):
         rdr_dataset_id = bq_utils.get_rdr_dataset_id()
         test_util.delete_all_tables(ehr_dataset_id)
         test_util.delete_all_tables(rdr_dataset_id)
-        cls.load_dataset_from_files(ehr_dataset_id, test_util.NYC_FIVE_PERSONS_PATH)
+        cls.load_dataset_from_files(ehr_dataset_id, test_util.NYC_FIVE_PERSONS_PATH, True)
```
mark-velez: @cukarthik yay to test data! But I'm not sure how it is used. Can you please explain? In-person is ok.

mark-velez: @cukarthik thanks for clarifying. To summarize, for the `combine_ehr_rdr` tests to pass you had to load the mappings tables in the input EHR dataset. The queries that propagate `hpo_id` to the combined mapping tables depend on them.

PR #96, opened 2018-10-18
[DC-178] Create script to automate curation pipeline

ksdkalluri:


data_steward/run_curation_pipeline.sh
```diff
@@ -0,0 +1,207 @@
...
+source deid_env/bin/activate
+
+# install the requirements in the virtualenv
+pip install -r requirements.txt
+#------------------------------------------------
+
+cd src
+set -e
+
+python deid.py --i_dataset ${cdr} --table person --o_dataset ${cdr_deid} --config ../test_config.json --log
```
mark-velez: We need a way for deid's config to be based on user input. We can just pass it as another parameter or let me know any other suggestions; the only thing that changes in `deid/config.json` is the [path to the keyfile](https://github.com/all-of-us/curation/blob/develop/data_steward/deid/config.json#L3) which this script already accepts as input.

```suggestion
python deid.py --i_dataset ${cdr} --table person --o_dataset ${cdr_deid} --config ${deid_config} --log
```

PR #100, opened 2018-11-09
[DC-252] Upgrade requests to 2.20.0 in data_steward and deid

mark-velez:


data_steward/deid/requirements.txt
```diff
@@ -54,7 +54,7 @@ python-dateutil==2.7.2
 pytz==2018.4
 pyzmq==17.0.0
 qtconsole==4.3.1
-requests==2.18.4
+requests==2.20.0
```
mark-velez: hey @lnyemba, we will be upgrading requests due to security vulnerabilities in the old version. we tested it on the main pipeline and have not had any issues. i don't expect deid will have any issues. i'm just putting this on your radar.

PR #103, opened 2018-11-12
DC-249

lnyemba:
Bug fix with date-shift (computation error)


ksdkalluri:
This is a duplicate of #106  pull request [DC-249] which is merged.

PR #122, opened 2019-01-18
Adjust date shift inverval

RobertJCarroll:
This change alters the implicit "round()" performed by the cast to an explicit "ceil()" so that the result is a uniform [1,365] date_shift instead of the mostly uniform [0,365] date_shift.


nishanthpp93:
Minor change and looks like it should work as expected. However, in the future it would be good to have a test to make sure it works as expected.

PR #129, opened 2019-02-13
[DC-288]Show the aggregate heel errors for PPI and PM data

ksdkalluri:


mark-velez:
Rebased and merged to `develop`

PR #143, opened 2019-04-15
[DC-363] Adds Script to get the latest results.html from all the site buckets.

ksdkalluri:
* Adds script to get latest results.html from the site buckets
* Updates the query to get the achilles reports.


data_steward/tools/consolidated_achilles_report/curation_report/server.py
```diff
@@ -1,15 +1,16 @@
...
 class Handler(SimpleHTTPServer.SimpleHTTPRequestHandler):
     pass
 
+
 Handler.extensions_map['.json'] = 'application/json'
 
 httpd = SocketServer.TCPServer(("", PORT), Handler)
 
-print "serving at port", PORT
+print("serving at port {port}".format(port=PORT))
```
lrwb-aou: Some static analysis (like pylint) will complain about the unnecessary parenthesis.  But I think if we're serious about moving to 3.X, we should leave them.  This will make that a little easier.

data_steward/tools/consolidated_achilles_report/get_all_results_html.py
```diff
@@ -0,0 +1,63 @@
...
+        report_path = p.replace('person.csv', 'results.html')
+        result = dict(hpo_id=hpo_id, updated=dte, report_path=report_path)
+        results.append(result)
+    return results
+
+
+def download_report(s):
+    """
+    Download most recent report files
+    :param s:
```
lrwb-aou: Can we use a more descriptive variable name than 's'?  And please add the documentation definition for 's' here.

data_steward/tools/consolidated_achilles_report/get_all_results_html.py
```diff
@@ -0,0 +1,63 @@
...
+    :return:
+    """
+    # Save it to curation_report/data/<hpo_id>
+    cdir = os.getcwd()
+    if not os.path.exists('%s/result_data' % (cdir)):
```
lrwb-aou: To prevent race conditions and simplify your code, use try/except instead of checking for existence.  I think of this type of try/except as asking forgiveness rather than permission.

```
try:
    os.makedirs('%s/result_data/%s' % (cdir, s['hpo_id']))
except OSError:
    # log the exception but keep moving because it doesn't hurt your code.
    LOGGER.debug("The file %s/result_data/%s already exists", cdir, s['hpo_id'])
```

Then you can remove the if/else statements on lines 42-50 and replace them with `cmd`, `print`, and `os` calls once.
```
cmd = 'gsutil -m cp -r %s ./result_data/%s/' % (s['report_path'], s['hpo_id'])
print('Downloading {} rpt with cmd: `{}`...'.format(s['hpo_id'], cmd))
os.system(cmd)
```

I think there should be a way to do this without issuing a system command, but I'm still looking into it.  Codacy is complaining about the system command lines.

data_steward/tools/consolidated_achilles_report/get_all_results_html.py
```diff
@@ -0,0 +1,63 @@
+from __future__ import print_function
```
lrwb-aou: This is a completely new file, but I don't see any associated unit tests.  We should build our unit tests as we add our code.  Otherwise, we won't know when we've made a breaking change.  There should be at least one test method for each public function in this module.  Ideally, there will be multiple unit tests for each function:  a minimum of one "happy-path" test and other tests for possible bad inputs and expected behaviours.

data_steward/tools/consolidated_achilles_report/get_all_results_html.py
```diff
@@ -0,0 +1,63 @@
...
+import gcs_utils
```
lrwb-aou: According to [pep 8](https://www.python.org/dev/peps/pep-0008/#imports), your imports should be ordered as:

```
import json
import os

import gcs_utils
import query_reports
```

I like to add comments in to help remind me to do this.  Something like:
```
# Python imports
import json
import os

# Third party imports

# Project imports
import gcs_utils
import query_reports
```

data_steward/tools/consolidated_achilles_report/get_all_results_html.py
```diff
@@ -0,0 +1,63 @@
...
+
+def get_hpo_id(p):
+    rel_path = p[len(DRC_BUCKET_PATH):]
+    return rel_path[:rel_path.index('/')]
+
+
+def transform_bq_list(uploads):
+    """
+    Get paths to all most recent report files
+
```
lrwb-aou: Need to add the uploads parameter to the doc string.  I can guess that it is an iterable, but it would be nice to know if you expect a specific iterable.

data_steward/tools/consolidated_achilles_report/get_all_results_html.py
```diff
@@ -0,0 +1,63 @@
...
+    :return:
+    """
+    results = []
+    for upload in uploads:
+        dte, p = upload['upload_timestamp'], upload['file_path']
+        hpo_id = get_hpo_id(p)
+        report_path = p.replace('person.csv', 'results.html')
+        result = dict(hpo_id=hpo_id, updated=dte, report_path=report_path)
```
lrwb-aou: I haven't used this method of setting up a dictionary before.  It is elegant.  There is a small performance penalty, but it likely won't be noticed in this case.  I tested it using:
```
import timeit
timeit.timeit("a = dict(alpha='alphabetic', beta='betty', zeta='zebedee')", number=1000000)
timeit.timeit("a = {'alpha': 'alphabetic', 'beta': 'betty', 'zeta': 'zebedee'}", number=1000000)
```
The second method ran in a little less than half the same amount of time as the first.

data_steward/tools/consolidated_achilles_report/query_reports.py
```diff
@@ -11,35 +11,61 @@
   MAX(timestamp) AS upload_timestamp,
   CONCAT('gs://{drc_bucket}', MAX(SUBSTR(protopayload_auditlog.resourceName, 49))) as file_path
 FROM
-  `{app_id}.GcsBucketLogging.cloudaudit_googleapis_com_data_access_2018*` l
+  `{app_id}.GcsBucketLogging.cloudaudit_googleapis_com_data_access_2019*` l
```
lrwb-aou: Maybe we can change the date portion to be a variable.  Then this can be more easily updated in the future.

```
`{app_id}.GcsBucketLogging.cloudaudit_googleapis_com_data_access_{year}*` l
```

data_steward/tools/consolidated_achilles_report/query_reports.py
```diff
@@ -11,35 +11,61 @@
...
-  _TABLE_SUFFIX > '0801'
-  AND protopayload_auditlog.resourceName LIKE '%aou%'
+protopayload_auditlog.resourceName LIKE '%aou%'
   AND protopayload_auditlog.methodName = 'storage.objects.create'
   AND resource.labels.bucket_name LIKE '{drc_bucket}'
   AND protopayload_auditlog.resourceName LIKE '%datasources.json'
 GROUP BY
   REGEXP_EXTRACT(protopayload_auditlog.resourceName, r".+\/aou[0-9]+")"""
 
+LATEST_RESULTS_JSON = 'latest_results.json'
```
lrwb-aou: Seems like a good opportunity to start using a constants file for this module.

data_steward/tools/consolidated_achilles_report/query_reports.py
```diff
@@ -11,35 +11,61 @@
...
+GROUP BY
+  REGEXP_EXTRACT(protopayload_auditlog.resourceName, r".+\/aou[0-9]+")"""
 
-def get_most_recent(app_id=None, drc_bucket=None):
+
+def get_most_recent(app_id=None, drc_bucket=None, report_for=None):
     """
     Query audit logs for paths to the most recent datasources.json files in the DRC bucket.
 
     Note: Results are cached in a local json file to avoid unnecessary queries.
```
lrwb-aou: If your queries are only select statements, there isn't a limit to how many times you can query.  If you're concerned about concurrent query limitations, if you set this to use `batch`, the queries will execute when there is available bandwidth and won't count toward your concurrent query quota.  The `batch` parameter is added to the query function in #148.  Your json file may be preferable here, but you do have an option.

data_steward/tools/consolidated_achilles_report/query_reports.py
```diff
@@ -11,35 +11,61 @@
...
         drc_bucket = gcs_utils.get_drc_bucket()
-    if not os.path.exists(LATEST_REPORTS_JSON):
-        query = LATEST_REPORTS_QUERY.format(app_id=app_id, drc_bucket=drc_bucket)
-        query_job = bq_utils.query(query)
-        result = bq_utils.response2rows(query_job)
-        with open(LATEST_REPORTS_JSON, 'w') as fp:
-            json.dump(result, fp, sort_keys=True, indent=4)
-    with open(LATEST_REPORTS_JSON, 'r') as fp:
-        return json.load(fp)
+        if report_for == 'achilles':
```
lrwb-aou: The values that are valid for `report_for` should be stored in a constants file.  Then other pieces of the software can use the code correctly by using the constants.  When the constants are updated, you will only need to update in one location (the constants file) rather than 2 or more (here and any file that uses the `get_most_recent` function.

data_steward/tools/consolidated_achilles_report/query_reports.py
```diff
@@ -11,35 +11,61 @@
...
-def get_most_recent(app_id=None, drc_bucket=None):
+
+def get_most_recent(app_id=None, drc_bucket=None, report_for=None):
     """
     Query audit logs for paths to the most recent datasources.json files in the DRC bucket.
 
     Note: Results are cached in a local json file to avoid unnecessary queries.
     :param app_id: identifies the GCP project
     :param drc_bucket: identifies the DRC bucket
+    :param report_for: denotes which query to use
```
lrwb-aou: Please add the valid values for this parameter and what happens if the parameter is not specified.

data_steward/tools/consolidated_achilles_report/curation_report/server.py
ksdkalluri: Going forward, this usage makes our life easier while upgrading our platform. Thats the only reason I am following this convention.

lrwb-aou: Agreed.  Going forward, it makes life easier.  I didn't mean to imply you should change this.

data_steward/tools/consolidated_achilles_report/get_all_results_html.py
ksdkalluri: you are right, this was created in very short time and I was only looking if it works. I will try to make it a single script.

data_steward/tools/consolidated_achilles_report/get_all_results_html.py
ksdkalluri: yes, this is a great idea, i'll update it.

data_steward/tools/consolidated_achilles_report/query_reports.py
ksdkalluri: These are just select queries, and we don't run these queries very often, so I think this should do.

PR #144, opened 2019-04-23
[DC-281] Create new Concept IDs for Generalization Rules for De-ID

mark-velez:


data_steward/resources.py
```diff
@@ -125,3 +139,15 @@ def cdm_schemas(include_achilles=False, include_vocabulary=False):
             if include_table:
                 result[table_name] = schema
     return result
+
+
+def hash_dir(in_dir):
+    """
+    Generate an MD5 digest from the contents of a directory
+    """
+    fs = os.listdir(in_dir)
```
lrwb-aou: This is one of the areas some static analysis tools complain about and others do not.  Variable names that are too short are picked up by pylint, but not pycharm.  The idea is that if we are writing self documenting code (few comments), then our variable names should reflect the data they hold.  So, do we want to implement something at a team level that points out areas where we have variable names that are possibly too short?  If so, I would suggest making this segment:

`filelist = os.listdir(in_dir)`
`hash_obj = hashlib.md5()`
`for filename in filelist:`
`    filepath = os.path.join(in_dir, filename)`
`    hash_obj.update(open(filepath, 'rb').read())`
`return hash_obj.hexdigest()`

data_steward/resources.py
```diff
@@ -125,3 +139,15 @@ def cdm_schemas(include_achilles=False, include_vocabulary=False):
...
+    hash_obj = hashlib.md5()
```
lrwb-aou: According to codacy, this is an insecure hash function.  Is this just to generate a version string?

data_steward/test/unit_test/vocabulary_test.py
```diff
@@ -0,0 +1,97 @@
...
+        out_fp = StringIO.StringIO()
+        _transform_csv(in_fp, out_fp)
+        return out_fp.getvalue()
+
+    def test_transform_file(self):
+        header = ['concept_id', 'valid_start_date', 'valid_end_date', 'invalid_reason']
+        row1 = ['1', '2017-05-17', '2099-12-31', '']
+        row2 = ['2', '2019-01-01', '2099-12-31', '']
+        rows = [header, row1, row2]
+        lines = map(lambda r: DELIMITER.join(r), rows)
```
lrwb-aou: Codacy is indicating this lamba expression could be replaced by:
`lines = [DELIMITER.join(row) for row in rows]`

For this small example, it isn't much of a performance hit, but generally, the list comprehension is faster than the map function.  You can test it yourself with a python interpreter.

`import timeit`
`timeit.timeit('map(lambda r: "\t".join(r), rows)', setup="rows=[['concept_id', 'valid_start_date', 'valid_end_date', 'invalid_reason'], ['1', '2017-05-17', '2099-12-31', ''], ['2', '2019-01-01', '2099-12-31', '']]", number=5000)`
`timeit.timeit('["\t".join(row) for row in rows]', setup="rows=[['concept_id', 'valid_start_date', 'valid_end_date', 'invalid_reason'], ['1', '2017-05-17', '2099-12-31', ''], ['2', '2019-01-01', '2099-12-31', '']]", number=5000)`

As mentioned, for this small example, there isn't a noticable change, but it is there.

data_steward/vocabulary.py
```diff
@@ -1,11 +1,179 @@
...
+        raise ValueError('Cannot parse value {v} as date'.format(v=s))
+
+
+def _transform_csv(in_fp, out_fp, err_fp=None):
+    if not err_fp:
+        err_fp = sys.stderr
+    csv_reader = csv.reader(in_fp, delimiter=DELIMITER)
+    header = next(csv_reader)
+    date_indexes = []
+    for i in range(0, len(header)):
```
lrwb-aou: Codacy is complaining about this loop.  It could be changed to:

`for index, item in enumerate(header):`
`    if item.endswith('_date'):`
`        date_indexes.append(index)`

data_steward/vocabulary.py
```diff
@@ -1,11 +1,179 @@
...
+        if header[i].endswith('_date'):
+            date_indexes.append(i)
+    csv_writer = csv.writer(out_fp, delimiter=DELIMITER, lineterminator=LINE_TERMINATOR)
+    csv_writer.writerow(header)
+    for row_index, row in enumerate(csv_reader):
+        try:
+            for i in date_indexes:
+                row[i] = format_date_str(row[i])
+            csv_writer.writerow(row)
+        except Exception, e:
```
lrwb-aou: Catching an exception that is too broad.  Explicitly catch your expected `ValueError`.

data_steward/vocabulary.py
```diff
@@ -1,11 +1,179 @@
...
+        err_fp = sys.stderr
+    csv_reader = csv.reader(in_fp, delimiter=DELIMITER)
+    header = next(csv_reader)
+    date_indexes = []
+    for i in range(0, len(header)):
+        if header[i].endswith('_date'):
+            date_indexes.append(i)
+    csv_writer = csv.writer(out_fp, delimiter=DELIMITER, lineterminator=LINE_TERMINATOR)
+    csv_writer.writerow(header)
+    for row_index, row in enumerate(csv_reader):
```
lrwb-aou: This loop isn't using enumeration.  You can change it and achieve the same effect with:

`for row in csv_reader:`

data_steward/vocabulary.py
```diff
@@ -1,11 +1,179 @@
...
+
+def transform_file(file_path, out_dir):
+    """
+    Format file date fields and standardize line endings a local csv file and save result in specified directory
+
+    :param file_path: Path to the csv file
+    :param out_dir: Directory to save the transformed file
+    """
+    file_name = os.path.basename(file_path)
+    table_name, _ = os.path.splitext(file_name.lower())
```
lrwb-aou: This line isn't used.  You can remove it.

data_steward/test/unit_test/resources_test.py
```diff
@@ -34,7 +34,7 @@ def test_cdm_schemas(self):
         self.assertListEqual([], result_achilles_tables,
                              msg='Achilles tables %s should not be in result of cdm_schemas()' % result_achilles_tables)
 
-        result_vocab_tables = [table_name for table_name in table_names if table_name in vocabulary.VOCABULARY_TABLES]
+        result_vocab_tables = [table_name for table_name in table_names if table_name in resources.VOCABULARY_TABLES]
```
lrwb-aou: Since you have removed the dependence on the vocabulary module, you should remove the module import statement at the top of the file.

`import vocabulary`

data_steward/resources.py
```diff
@@ -1,12 +1,24 @@
...
 import inspect
 import os
 import csv
 import cachetools
 import json
-import vocabulary
 
 base_path = os.path.dirname(os.path.abspath(inspect.getfile(inspect.currentframe())))
 
+CONCEPT = 'concept'
```
lrwb-aou: This segment could benefit from moving these items into a constants file.  I've started a constants package structure on my id-validation-report branch.  We should do something like this, especially to clean up the possibility of circular imports that I saw in other code reviews.

data_steward/test/unit_test/vocabulary_test.py
```diff
@@ -0,0 +1,97 @@
...
+        # after transform should output valid csv with date format yyyy-mm-dd
+        expected_text = LINE_TERMINATOR.join(lines) + LINE_TERMINATOR
+        input_text = expected_text.replace('-', '')
+        actual_text = VocabularyTest.do_transform_file(input_text)
+        self.assertEqual(expected_text, actual_text)
+
+        # windows line endings are replaced with linux ones
+        input_text = expected_text.replace(LINE_TERMINATOR, '\r\n')
+        actual_text = VocabularyTest.do_transform_file(input_text)
+        self.assertEqual(expected_text, actual_text)
```
lrwb-aou: Because you are testing two functionalities in the same function, I would suggest adding a message in case one of your assertions fails.  This will help make it clearer to the maintainer what has changed if this test fails.  

`self.assertEqual(expected_text, actual_text, "The date formatted text did not format as expected.")`
and
`self.assertEqual(expected_text, actual_text, "The Windows line endings were not replaced as expected.")`

data_steward/test/unit_test/vocabulary_test.py
```diff
@@ -0,0 +1,97 @@
...
+
+    def test_format_date_str(self):
+        self.assertEqual('2019-01-23', format_date_str('20190123'))
```
lrwb-aou: Same as above, I would suggest adding failure messages because this is testing the same function in multiple scenarios.

data_steward/test/unit_test/vocabulary_test.py
```diff
@@ -0,0 +1,97 @@
...
+        self.assertEqual('2019-01-23', format_date_str('2019-01-23'))
+        with self.assertRaises(ValueError):
+            format_date_str('201901234')
+
+    def test_append_vocabulary(self):
+        in_path = test_util.TEST_VOCABULARY_VOCABULARY_CSV
+        out_path = os.tempnam()
```
lrwb-aou: What are the benefits of using `os.tempnam()` over `tempfile.mkdtemp()`?  I haven't run into this function before.

Also, you should wrap this whole thing in a `try/except/finally` block.  In the `finally` block remove everything you've made there.  That way, if this fails due to an assertion or for any other reason, those files get cleaned up.

data_steward/test/unit_test/vocabulary_test.py
```diff
@@ -0,0 +1,97 @@
...
+        out_path_2 = os.tempnam()
+        with mock.patch('warnings.warn') as warn_call:
+            append_vocabulary(out_path, out_path_2)
+            warn_call.assert_called()
+
+        os.remove(out_path)
+        os.remove(out_path_2)
+
+    def test_append_concepts(self):
+        in_path = test_util.TEST_VOCABULARY_CONCEPT_CSV
```
lrwb-aou: I would wrap this in a `try/except/finally` block too.

data_steward/tools/load_vocab.sh
```diff
@@ -0,0 +1,95 @@
+#!/usr/bin/env bash
```
lrwb-aou: Given that this is a 95 line bash script and only likely to get longer, why is it a bash script?  What is it doing that can't be done in python?

data_steward/vocabulary.py
```diff
@@ -1,11 +1,179 @@
...
-CONCEPT_RELATIONSHIP = 'concept_relationship'
-CONCEPT_SYNONYM = 'concept_synonym'
-DOMAIN = 'domain'
-DRUG_STRENGTH = 'drug_strength'
-RELATIONSHIP = 'relationship'
-VOCABULARY = 'vocabulary'
-VOCABULARY_TABLES = [CONCEPT, CONCEPT_ANCESTOR, CONCEPT_CLASS, CONCEPT_RELATIONSHIP, CONCEPT_SYNONYM, DOMAIN,
-                     DRUG_STRENGTH, RELATIONSHIP, VOCABULARY]
+import os
+import csv
```
lrwb-aou: I would move `import csv` before `import os`.  While not technically required to be in alphabetic order in the import groups, I think it's easier to peruse the list later when maintaining the module.

data_steward/vocabulary.py
```diff
@@ -1,11 +1,179 @@
...
+import re
+import sys
+import warnings
+
+from resources import CONCEPT, VOCABULARY, AOU_GENERAL_PATH, AOU_GENERAL_CONCEPT_CSV_PATH, hash_dir
+
+csv.field_size_limit(sys.maxsize)
+
+DELIMITER = '\t'
```
lrwb-aou: I would move lines 11-27 to a constants file.

data_steward/vocabulary.py
```diff
@@ -1,11 +1,179 @@
...
+
+AOU_GEN = 'AoU_General'
+AOU_GEN_VOCABULARY_CONCEPT_ID = '2000000000'
+AOU_GEN_VOCABULARY_REFERENCE = 'https://docs.google.com/document/d/10Gji9VW5-RTysM-yAbRa77rXqVfDfO2li2U4LxUQH9g'
+OMOP_VOCABULARY_CONCEPT_ID = '44819096'
+
+ERROR_APPENDING = 'Appending to {in_path} which already contains rows for ' + AOU_GEN
+
+
+def format_date_str(s):
```
lrwb-aou: I would change this function to use the parse function from dateutil.parser .  I would note in the documentation string that I expect this to raise a ValueError.

```
:raises:  ValueError if a valid date object cannot be parsed from the string
"""
date_obj = parse(date_str)
formatted_date_str = date_obj.strftime('%Y-%m-%d')
return formatted_date_str
```

This will let you get rid of those DATE regular expressions.

data_steward/vocabulary.py
```diff
@@ -1,11 +1,179 @@
...
+
+    :param file_path: Path to the csv file
+    :param out_dir: Directory to save the transformed file
+    """
+    file_name = os.path.basename(file_path)
+    table_name, _ = os.path.splitext(file_name.lower())
+    out_file_name = os.path.join(out_dir, file_name)
+    err_dir = os.path.join(out_dir, ERRORS)
+    err_file_name = os.path.join(err_dir, file_name)
+    if not os.path.exists(err_dir):
```
lrwb-aou: In python, it's generally preferred to ask forgiveness than permission.  This conditional is subject to a race condition.  It is unlikely that we will notice this run as a single process, it could become an issue if it is run by multiple processes.  What is probably a better approach is to use:

```
try:
    os.makedirs(err_dir)
except OSError:
    LOGGER.debug("Error directory:\t%s\t already exists", err_dir)
```
We log it already exists and move on.

data_steward/vocabulary.py
```diff
@@ -1,11 +1,179 @@
...
+
+
+def transform_files(in_dir, out_dir):
+    """
+    Transform vocabulary files in a directory and save result in another directory
+
+    :param in_dir: Directory containing vocabulary csv files
+    :param out_dir: Directory to save the transformed file
+    """
+    fs = os.listdir(in_dir)
```
lrwb-aou: I suggest changing `f` and `fs` to longer variable names to make them support self documentation.

data_steward/vocabulary.py
```diff
@@ -1,11 +1,179 @@
...
+
+
+def get_aou_general_version():
+    return hash_dir(AOU_GENERAL_PATH)
+
+
+def get_aou_general_vocabulary_row():
+    aou_gen_version = get_aou_general_version()
+    # vocabulary_id vocabulary_name vocabulary_reference vocabulary_version vocabulary_concept_id
+    return DELIMITER.join([AOU_GEN, AOU_GEN, AOU_GEN_VOCABULARY_REFERENCE, aou_gen_version, AOU_GEN_VOCABULARY_CONCEPT_ID])
```
lrwb-aou: It's fine for vocabulary_id and vocabulary_name to be the same, but this assumes they are always the same and uses the same variable to set them.  Would it be better to go ahead and separate them with a `AOU_GEN_ID` and `AOU_GEN_NAME` variables?

data_steward/vocabulary.py
```diff
@@ -1,11 +1,179 @@
...
+            for row in in_fp:
+                if AOU_GEN in row:
+                    # skip it so it is appended below
+                    warnings.warn(ERROR_APPENDING.format(in_path=in_path))
+                else:
+                    out_fp.write(row)
+
+        # append new rows
+        with open(AOU_GENERAL_CONCEPT_CSV_PATH, 'rb') as aou_gen_fp:
+            _ = next(aou_gen_fp)  # skip header
```
lrwb-aou: You can also skip headers in csv files with or without knowing they exist by:
```
has_header = csv.Sniffer.has_header(aou_gen_fp.read(1024))
aou_gen_fp.seek(0)
if has_header:
    next(aou_gen_fp)
```

This way allows you to not care if the header does or doesn't exist for your input file.

data_steward/vocabulary.py
```diff
@@ -1,11 +1,179 @@
...
+        out_fp.write(new_row)
+
+
+def add_aou_general(in_dir, out_dir):
+    fs = os.listdir(in_dir)
+    concept_in_path = None
+    vocabulary_in_path = None
+    # Case-insensitive search for concept and vocabulary files
+    for f in fs:
+        t, _ = os.path.splitext(f.lower())
```
lrwb-aou: What is `t`?

data_steward/vocabulary.py
```diff
@@ -1,11 +1,179 @@
...
+        raise IOError('VOCABULARY.csv was not found in %s' % in_dir)
+
+    concept_out_path = os.path.join(out_dir, os.path.basename(concept_in_path))
+    append_concepts(concept_in_path, concept_out_path)
+
+    vocabulary_out_path = os.path.join(out_dir, os.path.basename(vocabulary_in_path))
+    append_vocabulary(vocabulary_in_path, vocabulary_out_path)
+
+
+if __name__ == '__main__':
```
lrwb-aou: I've noticed we have a lot of command line setups like this throughout the pipeline.  Have we ever thought about consolidating/refactoring our environment variables into configuration files, command line options, and the few necessary environment variables?

data_steward/test/unit_test/vocabulary_test.py
```diff
@@ -0,0 +1,97 @@
+import unittest
```
lrwb-aou: According to [pep 8],(https://www.python.org/dev/peps/pep-0008/#imports) imports should be grouped as, Python Standard Imports, Third party imports, and Project imports.  This should be changed to:

```
import os
import StringIO
import unittest

import mock

import test_util
import vocabulary
```

Also, Pep 8 discourages importing via `*` and importing a function who is considered private.  If it's for testing purposes, the functions that use `_transform_csv` should handle this information and checks.  If `_transform_csv` may be used elsewhere, rename it and drop the underscore.

data_steward/test/unit_test/vocabulary_test.py
```diff
@@ -0,0 +1,97 @@
...
+            expected_lines = original_lines + all_lines[1:]
+            with open(out_path, 'rb') as out_fp:
+                actual_lines = out_fp.readlines()
+                self.assertSequenceEqual(actual_lines, expected_lines)
+
+        # should warn when concepts already in input
+        out_path_2 = os.tempnam()
+        with mock.patch('warnings.warn') as warn_call:
+            append_concepts(out_path, out_path_2)
+            warn_call.assert_called()
```
lrwb-aou: Unless we've increased our python version without my knowledge, these should fail.  `assert_called()` isn't available until Python 3.6.  You should be able to do ```self.assertTrue(warn_call.called)``` or ```self.assertEqual(warn_call.call_count, # of expected calls)```.

data_steward/resources.py
mark-velez: I changed it to sha1

data_steward/resources.py
mark-velez: You're right about variable names. I tried to address as many as I could find in this PR. Please feel free to fix any I may have missed. Moving forward I'll make them more descriptive.

data_steward/test/unit_test/vocabulary_test.py
mark-velez: Thanks! I forgot about that library. I changed it to use `tempfile.mkdtemp()`. Does this try/finally block look ok?

data_steward/vocabulary.py
mark-velez: This is great advice! Done.

data_steward/vocabulary.py
mark-velez: Yes, it has become rather unwieldy. I'd definitely appreciate if you could guide us in our thinking about config files / env vars / options so we can refactor!

lrwb-aou: I think it is something we should plan to improve.  I don't expect us to get it done here and now.  Just wanted to bring it up.

data_steward/test/unit_test/vocabulary_test.py
lrwb-aou: I feel better about it removing the files when the test is done.  From experience, it's a pain tracking down which failing test is littering your code base with temporary files.

data_steward/vocabulary.py
nishanthpp93: I wanted to note that the performance can take a hit here:
`timeit.timeit('format_date_str_regex(s)', setup="from __main__ import format_date_str_regex; s='20170517'", number=5000)`
0.0126650333405
`timeit.timeit('format_date_str_strftime(s)', setup="from __main__ import format_date_str_strftime; s='20170517'", number=5000)`
0.318553924561

I think we should generally stick to standard libraries for dates but in this case regex is much faster and we could make an exception. I ran into this issue while trying to load the new vocabulary. The millions of entries make a big difference.

lrwb-aou: Good catch Nishanth.  Thanks!  I had not run a time comparison on this.  With you suggestion I ran the following:
```
timeit.timeit("from dateutil.parser import parse\ndate_obj = parse(date_str)\nformatted = date_obj.strftime('%Y-%m-%d')", setup = 'date_str = "20170101"', number = 1000000)
timeit.timeit("if bq_pattern.match(date_str):\n    formatted = date_str\nelif raw_pattern.match(date_str):\n    parts = date_str[0:4], date_str[4:6], date_str[6:8]\n    formatted = '-'.join(parts)", setup="import re\ndate_str='20170101'\nRAW_DATE_REGEX=r'\d{8}$'\nBQ_DATE_REGEX = r'\d{4}-\d{2}-\d{2}$'\nraw_pattern = re.compile(RAW_DATE_REGEX)\nbq_pattern=re.compile(BQ_DATE_REGEX)", number = 1000000)
timeit.timeit("if re.match(BQ_DATE_REGEX, date_str):\n    formatted = date_str\nelif re.match(RAW_DATE_REGEX, date_str):\n    parts = date_str[0:4], date_str[4:6], date_str[6:8]\n    formatted = '-'.join(parts)", setup="import re\ndate_str='20170101'\nRAW_DATE_REGEX=r'\d{8}$'\nBQ_DATE_REGEX = r'\d{4}-\d{2}-\d{2}$'\n", number = 1000000)
```
I rant this with a pre-compiled regular expression and one that is compiled every time.  The string format one in 52.6 seconds, the regular expression one ran in 1.7 seconds, and the pre-compiled regular expression one in 0.7 seconds.  For performance, regular expressions are better.  It's a personal preference, but I prefer to pre-compile my regular expressions when I use them..

lrwb-aou: I've been thinking about this.  Yes, the `dateutil.parser` is much slower than regular expressions.  It was the `parse` function that raised the exception with the bad date objects in the vocabulary files (00170101 dates).  If we use a regular expression instead of the `parse` function, we should consider making the regular expression less open ended.  

```
RAW_DATE_REGEX=r'(19\d{6}$)|(20\d{6}$)'
BQ_DATE_REGEX = r'(19\d{2}-\d{2}-\d{2}$)|(20\d{2}-\d{2}-\d{2}$)'
```

This will still match some bad dates, like 2015-48-67, but it will limit the years to starting with 19 or 20.

If you wanted to make this more specific, you could use the following:

```
RAW_DATE_REGEX = r'(19\d{2}(01|02|03|04|05|06|07|08|09|10|11|12)([0-3][0-9])$)|(20\d{2}(01|02|03|04|05|06|07|08|09|10|11|12)([0-3][0-9])$)'
BQ_DATE_REGEX = r'(19\d{2}-(01|02|03|04|05|06|07|08|09|10|11|12)-([0-3][0-9])$)|(20\d{2}-(01|02|03|04|05|06|07|08|09|10|11|12)-([0-3][0-9])$)'
```
This will keep the month and years correct, but some months could have the wrong number of days.  What I'm getting at is, we shouldn't trust the data just because it comes from another project.  We need to verify it ourselves.  If we want to get really fancy, we can map out the correct months and days, too.  Then pre-compiling the regular expression becomes more important.

data_steward/test/unit_test/vocabulary_test.py
mark-velez: @lrwb-aou do you mean that you'd prefer for [this](https://github.com/mark-velez/curation/blob/add_gen_vocab/data_steward/test/unit_test/vocabulary_test.py#L115) to go to `tearDown()`?

lrwb-aou: No, I don't.  You may not have files to remove every time a function executes.  I was just noticing the assert statements were in a context manager.  I think if something fails inside there, the remove statements execute because they are outside the context manager, but I haven't tried it locally to verify it.   I do know the `try/except/finally` block, will work every time.  This should work.  I was just saying I've done it wrong in the past and tried to find which test was littering my work environment with temp files.  That's a pain.  What you've done should work.

data_steward/test/unit_test/vocabulary_test.py
mark-velez: @lrwb-aou this is what I'm seeing- please let me know where I may be going wrong. My understanding is that the mock library eventually became part of Python standard. Based on [requirements.txt](https://github.com/mark-velez/curation/blob/add_gen_vocab/data_steward/requirements.txt#L17) we are using mock 2.0.0  which I understand is a backport for Python 2.7. The [docs say](https://mock.readthedocs.io/en/latest/changelog.html#id1) `assert_called()` had just been added in 2.0.0. Is it possible you're referring to when the method became available in [Python standard](https://docs.python.org/3/library/unittest.mock.html#unittest.mock.Mock.assert_called)?

lrwb-aou: Sorry, I didn't see this comment earlier.  I was basing this off the documentation [here](https://cpython-test-docs.readthedocs.io/en/latest/library/unittest.mock.html#unittest.mock.Mock.assert_called).  So yes, it seems this documentation is referring to when it became available in Python Standard.  If it works, that's fantastic.  Leave this as is.

mark-velez:
Re-issuing this PR from another branch add_gen_vocab_2

PR #146, opened 2019-05-03
[DC-275] Add Drug Checks to the results html

ksdkalluri:


data_steward/validation/main.py
```diff
@@ -286,36 +286,54 @@ def process_hpo(hpo_id, force_run=False):
...
+    if not heel_result:
+        heel_header_list = ['Record Count', 'Heel Error', 'Analysis ID', 'Rule ID']
     else:
-        for dict_item in list_of_dicts:
-            dict_values = tuple(dict_item.values())
-            result_list.append(dict_values)
-        return result_list
+        heel_header_list = _get_query_result_header(heel_result)
+    return heel_errors, heel_header_list
+
```
nishanthpp93: Both `get_heel_errors_in_results_html()` and `get_heel_errors_in_results_html` perform the same function except for the headers. They can be generalized to `add_table_in_results_html()` which takes in an additional input for headers.

data_steward/validation/main.py
```diff
@@ -340,14 +349,21 @@ def get_heel_result(hpo_id, app_id=None, dataset_id=None):
     return result
 
 
-def get_heel_error_header(list_of_dicts):
-    fake_header_list = ['Record Count', 'Heel Error', 'Analysis ID', 'Rule ID']
-    if not list_of_dicts:
-        return fake_header_list
+def _convert_query_result_to_row_list(list_of_dicts):
+    row_list = list()
+    if list_of_dicts is []:
```
nishanthpp93: This could be changed according to [PEP 8](https://www.python.org/dev/peps/pep-0008/):
```md
For sequences, (strings, lists, tuples), use the fact that empty sequences are false.

Yes: if not seq:
     if seq:

No:  if len(seq):
     if not len(seq):
```

nishanthpp93: Also interesting to note that
```python
l = []
print(l == []) # True
print(l is []) # False
```

data_steward/validation/main.py
```diff
@@ -340,14 +349,21 @@ def get_heel_result(hpo_id, app_id=None, dataset_id=None):
...
+        return row_list.append(tuple())
```
nishanthpp93: This returns the result of the _action of appending itself_, not `row_list`, so the return type will be `None`. 
`row_list` needs to be returned.

data_steward/validation/main.py
nishanthpp93: Also this not necessary, since an empty list of iterables can still be iterated through, but the inner loop won't be traversed. So the following code should suffice:
```python
def _convert_query_result_to_row_list(list_of_dicts):
    row_list = list()
    for dict_item in list_of_dicts:
        dict_values = tuple(dict_item.values())
        row_list.append(dict_values)
    return row_list
```

PR #148, opened 2019-05-04
[DC-356] Id match report

lrwb-aou:
This makes the following changes:

- Creates a constants package at the top level. The folder structure mimics the existing project structure and constants for a module are placed in a correspondingly named module in the constants package.
- Adds the .coveragerc file to help with unit testing. It ignores files we don't care about seeing results for.
- Adds the echo statement to the circleci config file for testing purposes.
- Creates a new dataset to hold the validation results. This dataset is described by the ehr and combined datasets analyzed in the matching process. It is named by whatever variable is passed in for it. The date of the combined dataset is appended to it or if no date is available, the date it is created.
- The new dataset contains a table with the observation data we are using to match from the rdr. This gives us a smaller table to query throughout the matching process.
- Each site gets its own validation table. Tables are named <site_identifier>_identity_match.
- An insert statement is created for each comparison field. It contains all the match data for the site. This helps avoid the 1000 updates per table per day limit.
- The fields are merged from sparse records to full records. 1 query per table.
- Sparse records are removed. 1 query per table.
- Records with null values have the nulls replaced with Missing values. 1 query per table.
- Writes a report to the internal curation bucket at <bucket_name>/<validation_dataset_name>/id-validation.csv. This report covers all files.
- Writes a report to each site bucket for each site. The report is found at <site_bucket>/drc-validations-/id-validation.csv. This covers only the data for a single site.
- Unit tests are written using mock.
- Integration test data is available but not included.


data_steward/validation/participants/identity_match.py
```diff
@@ -485,82 +511,318 @@ def _compare_birth_dates(person_id_set, rdr_birthdates, ehr_birthdates, match_va
...
+    return datetime.now().strftime(consts.DRC_DATE_FORMAT)
 
 
-def match_participants(dataset):
+def match_participants(project, rdr_dataset, ehr_dataset, dest_dataset_id):
     """
     Entry point for performing participant matching of PPI, EHR, and PII data.
 
-    :param dataset:  The string name of the dataset to use when matching participants
+    :param project: a string representing the project name
```
ksdkalluri: It will be informative if the parameters rdr_datasetr, ehr_dataset, dest_dataset_id  were added to the doc string.

data_steward/bq_utils.py
```diff
@@ -511,3 +546,69 @@ def _transform_row(row, schema):
...
+        bq_consts.DESCRIPTION: description,
+    }
+
+    if friendly_name:
+        job_body.update({bq_consts.FRIENDLY_NAME: friendly_name})
+
+    insert_dataset = bq_service.datasets().insert(projectId=app_id, body=job_body)
+    try:
+        insert_result = insert_dataset.execute(num_retries=bq_consts.BQ_DEFAULT_RETRY_COUNT)
+    except HttpError:
```
ksdkalluri: We wanted to make sure that the HTTP error is raised only if the dataset is already available in the project, because we should not be trying to delete the dataset if the HTTP error is due to other reason.

data_steward/validation/participants/identity_match.py
```diff
@@ -1,379 +1,336 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
 """
 A module to perform initial participant identity matching based on PII data.
 
 Compares site PII data to values from the RDR, looking to identify discrepancies.
 """
+# Python imports
+from datetime import datetime
+import json
```
ksdkalluri: This can be removed if we use the resources.

data_steward/validation/participants/identity_match.py
```diff
@@ -1,379 +1,336 @@
...
+        concept_id,
+        pii_field
+    ):
     """
-    Get email addresses from the site's PII table.
+    For an hpo, compare all first, middle, and last name fields to omop settings.
 
-    :param hpo:  hpo string to use when identifying table names for lookup.
+    This compares a site's name field values found in their uploaded PII
+    tables with the values in the OMOP observation table.
```
ksdkalluri: Doc string should be updated with the  project, rdr_dataset, pii_dataset, concept_id, pii_field parameters descriptions.

data_steward/validation/participants/identity_match.py
```diff
@@ -1,379 +1,336 @@
...
 
-    :param hpo:  hpo string to use when identifying table names for lookup.
 
-    :return:  A list of tuples with the first tuple element as the person_id,
-    the second tuple element as the address line one, the third tuple element is
-    address line 2, the fourth tuple element is city, the fifth tuple element is
-    state, and the sixth tuple element is zip code.
-    [(1, 'street 1', '', 'nashville', 'tn', 33333),
-     (2, 'street_8', 'street_10', 'yuma', 'az', 22222)]
+def _compare_email_addresses(
```
ksdkalluri: doc strings for all the functions in this script needs to be updated with parameter descriptions.

data_steward/validation/participants/identity_match.py
```diff
@@ -485,82 +511,318 @@ def _compare_birth_dates(person_id_set, rdr_birthdates, ehr_birthdates, match_va
...
+    validation_dataset = validation_dataset.get(bq_consts.DATASET_ID, '')
+
+    # create intermediate observation table in new dataset
+    readers.create_match_values_table(project, rdr_dataset, dest_dataset_id)
+
+    hpo_sites = readers.get_hpo_site_names()
+
+    field_list = []
+    #TODO:  create a proper config file to store this path
+    with open('resources/fields/identity_match.json') as json_file:
```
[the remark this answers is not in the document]
lrwb-aou: I have made this change, but have some ideas to discuss.

Getting the correct basename in the resources directory depends on where the resources.py file exists.  If this file is ever moved, the base_path line will not return the same location, and everything that relies on the base_path will fail.  I still think the pipeline project will benefit from using configuration files and handling paths properly from there.

data_steward/validation/participants/identity_match.py
lrwb-aou: Good catch.  Done.

data_steward/validation/participants/readers.py
```diff
@@ -0,0 +1,273 @@
...
+    This retrieves all possible matching values from the observation table
+    used in participant matching.  Returned query data is limited to the
+    person_id, observation_concept_id, and the value_as_string fields.
+
+    :param project: The name of the project to query for ehr values
+    :param dataset:  The name of the dataset to query for ehr values
+    :param table_name:  The name of the table to query for ehr values
+    :param concept_id: the id of the concept to identify with ehr data
+    :param id_list:  list of person_id values to limit the result set to
+
```
ksdkalluri: unknown parameters in doc string.

data_steward/bq_utils.py
lrwb-aou: Updated the code to log when it creates the dataset successfully, log when it deletes an existing dataset and creates a new one, log/raise when it tries to delete an existing dataset but the overwrite parameter is set to false so it fails, and log/raise when any other HttpError is caught.

data_steward/validation/participants/identity_match.py
lrwb-aou: I still need the datetime to provide the date for the new match dataset name.  I have removed the json import.

data_steward/validation/participants/readers.py
lrwb-aou: concept_id and id_list removed.  column_name added.

data_steward/validation/participants/identity_match.py
lrwb-aou: All doc strings in the module are updated.  Thanks for catching it.

data_steward/constants/bq_utils.py
```diff
@@ -0,0 +1,32 @@
+# bq_utils default values
```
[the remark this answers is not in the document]
lrwb-aou: This is a really nice question, and I've put off answering it until it was the last known thing to address in this PR.  It is a personal preference to have this top level constants package whose directory structure mirrors the directory structure of the project components it is being created for.  There are pros and cons:

Pros:

1.  I like to use the pep 8 import rules structure of Python imports, blank line, third party imports, blank line, and project imports.  I add the comments denoting this too.  Within each block I like to alphabetize the imported components based.  This will visually group all the constants modules imported by another module together.
2.  Having separate constants modules named identical to the module they are used in aids traceability.
3.  A package may contain multiple modules.  If you have only one constants file per package, that can become hard to easily read and maintain, especially if you store a lot of long statements or large structures in them.  This package originally used one constants file.  It grew to over 300 lines.  That is hard to read and maintain. 
4.  Splitting them according to module, means each module only loads what it needs.  So the namespace isn't polluted with a lot of data it isn't using.
5.  If the constants package and structure becomes adopted, it will support inheritance.  We can still define something in one place instead of multiple.

Cons:
1.  Your constants file isn't in the same package you are currently working in.  This might cause a mental shift.
2.  Until the inheritance is worked out, you may still define the same thing multiple times (such as concept ids defined in multiple constants files).
3.  Constants are split into correspondingly named modules.  If it doesn't exist, you have to create it.
4.  If you need constants from more than one module, it is two or more imports.  You will have to use the `import package.module.name as name` convention to distinguish between them.
5.  The mental load of having multiple constants files.

I prefer the package arrangement because it keeps the namespace much less polluted with stuff that a particular module does not need, it is easy to determine what module is using the constant based on the module name, if you are using multiple constants files this gives you an extra idea of how the pieces are working together, and this will make it easier to identify opportunities to use inheritance in the constants files.

I'm available for a discussion on this whenever we want to have one.  I think it would benefit us all to move toward that architectural style.

data_steward/validation/participants/identity_match.py
```diff
@@ -485,82 +572,333 @@ def _compare_birth_dates(person_id_set, rdr_birthdates, ehr_birthdates, match_va
...
+        project, validation_dataset, hpo_sites, bucket, filename
+    )
 
     return results
 
 
 if __name__ == '__main__':
-    DATASET = 'combined20190211'
-    match_participants(DATASET)
+    RDR_DATASET = 'lrwb_combined20190415'   # the combined dataset
```
ksdkalluri: these hard coded data set names should not be uploaded to the repository.

lrwb-aou: I have removed the test variable names.

PR #149, opened 2019-05-14
[DC-281] Create new Concept IDs for Generalization Rules for De-ID

mark-velez:
#144 with rebase


data_steward/tools/load_vocab.sh
```diff
@@ -0,0 +1,88 @@
...
+fi
+echo "Creating and loading dataset ${DATASET}..."
+bq mk --project_id ${APP_ID} --dataset_id ${DATASET} --description "Vocabulary ${OMOP_VOCABULARY_VERSION} loaded from ${GCS_PATH}"
+for file in $(gsutil ls ${GCS_PATH})
+do
+ filename=$(basename ${file})
+ # bash3 friendly lowercase
+ filename_lower=$(echo ${filename} | tr '[:upper:]' '[:lower:]')
+ table_name="${filename_lower%.*}"
+ gsutil cp ${file} .
```
lrwb-aou: Why are you uploading the files to cloud storage, then downloading the files, and then uploading to bq?  Why not use the local files you already uploaded to send to bq?  Also, we are explicitly converting filenames to all lower case.  Is there a reason we aren't uploading them the way we intend to use them?

lrwb-aou: This bash script will be replaced by a python script in another pull request.

PR #157, opened 2019-06-06
[DC-397] Validation cron

lrwb-aou:
Create cron validated endpoints to allow us to 
1. Create a validation dataset.
2. Create a results file in the DRC internal bucket
3.  Create result files in all the site buckets.

A few things I would like to have covered.  Is it possible to assume we are setting environment variable names for the datasets?  This patch makes an assumption that we can set those names.  Is this a safe assumption?


lrwb-aou:
Closing PR in favor of new branch PR.

PR #158, opened 2019-06-10
[DC-390] ID Columns in each domain should be unique

ksdkalluri:
Refactors Cleaner package and adds cleaning rule 4

### DIFF
--- a/data_steward/cleaners/clean_cdr.py
+++ b/data_steward/cleaners/clean_cdr.py
@@ -20,6 +20,7 @@ LOGGER = logging.getLogger(__name__)
 def _gather_ehr_queries(project, dataset):
     """
     gathers all the queries required to clean ehr dataset
+
     :param project: project name
     :param dataset: ehr dataset name
     :return: returns list of queries
@@ -32,6 +33,7 @@ def _gather_ehr_queries(project, dataset):
 def _gather_rdr_queries(project, dataset):
     """
     gathers all the queries required to clean rdr dataset
+
     :param project: project name
     :param dataset: rdr dataset name
     :return: returns list of queries
@@ -44,6 +46,7 @@ def _gather_rdr_queries(project, dataset):
 def _gather_ehr_rdr_queries(project, dataset):
     """
     gathers all the queries required to clean ehr_rdr dataset
+
     :param project: project name
     :param dataset: ehr_rdr dataset name
     :return: returns list of queries
@@ -56,6 +59,7 @@ def _gather_ehr_rdr_queries(project, dataset):
 def _gather_ehr_rdr_de_identified_queries(project, dataset):
     """
     gathers all the queries required to clean de_identified dataset
+
     :param project: project name
     :param dataset: de_identified dataset name
     :return: returns list of queries
@@ -68,6 +72,7 @@ def _gather_ehr_rdr_de_identified_queries(project, dataset):
 def _gather_unioned_ehr_queries(project, dataset):
     """
     gathers all the queries required to clean unioned_ehr dataset
+
     :param project: project name
     :param dataset: unioned_ehr dataset name
     :return: returns list of queries
@@ -135,6 +140,7 @@ def clean_ehr_rdr_de_identified_dataset(project=None, dataset=None):
 def get_dataset_and_project_names():
     """
     :return: A dictionary of dataset names and project name
+
     """
     project_and_dataset_names = dict()
     project_and_dataset_names[clean_cdr_consts.EHR_DATASET] = bq_utils.get_dataset_id()
--- a/data_steward/cleaners/clean_cdr.py
+++ b/data_steward/cleaners/clean_cdr.py
@@ -1,145 +1,81 @@
 """
 A module to serve as the entry point to the cleaners package.
 """
-# Python imports
-import logging
 
-# Third party imports
-import googleapiclient
-from google.appengine.api import app_identity
-import oauth2client
+# Python imports
+import argparse
 
 # Project imports
 import bq_utils
-#import constants.bq_utils as bq_consts
+import clean_cdr_engine as clean_engine
+import clean_rule_4 as rule_4
+
+# import constants.bq_utils as bq_consts
 import constants.cleaners.combined as combined_consts
 import constants.cleaners.combined_deid as deid_consts
 import constants.cleaners.ehr as ehr_consts
 import constants.cleaners.rdr as rdr_consts
 import constants.cleaners.unioned as unioned_consts
 
-LOGGER = logging.getLogger(__name__)
-
-def _add_console_logging(add_handler):
-    # this config should be done in a separate module, but that can wait
-    # until later.  Useful for debugging.
-    logging.basicConfig(level=logging.DEBUG,
-                        filename='/tmp/cleaner.log',
-                        filemode='a')
-
-    if add_handler:
-        handler = logging.StreamHandler()
-        handler.setLevel(logging.INFO)
-        formatter = logging.Formatter('%(asctime)s - %(name)s - %(levelname)s - %(message)s')
-        handler.setFormatter(formatter)
-        LOGGER.addHandler(handler)
-
-
-def _clean_dataset(project=None, dataset=None, statements=None):
-    if project is None or project == '' or project.isspace():
-        project = app_identity.get_application_id()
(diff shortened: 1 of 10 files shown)


data_steward/cleaners/clean_cdr_engine.py
```diff
@@ -0,0 +1,70 @@
...
+
+    if add_handler:
+        handler = logging.StreamHandler()
+        handler.setLevel(logging.INFO)
+        formatter = logging.Formatter('%(asctime)s - %(name)s - %(levelname)s - %(message)s')
+        handler.setFormatter(formatter)
+        LOGGER.addHandler(handler)
+
+
+def clean_dataset(project=None, dataset=None, rule_query=None):
```
lrwb-aou: To make this work with an iterable list of SQL statements, `rule_query` needs to be an iterable as it used to be.  We want to send a list of work to clean the dataset, not to run a single query.  If we really want to run a single query, we can achieve that by sending a list of one item.  I had suggested deleting lines 73 to the end because it was redundant.  By keeping lines 73 to the end and deleting the original function, we have lost the iterable quality of this function.

data_steward/cleaners/clean_rule_4.py
```diff
@@ -0,0 +1,52 @@
...
+rule_4_query = """
+select {columns}
+from (select m.*,
+ROW_NUMBER() OVER (PARTITION BY m.{domain_table}_id) AS row_num
+from `{project_id}.{dataset_id}.{table_name}` as m) as t
+where row_num = 1
+"""
+
+
+def run_clean_rule_4(project_id, dataset_id):
```
lrwb-aou: Names are important.  `run_clean_rule_4` is vague and doesn't provide good information about what is happening.  Try something more descriptive.  Like `run_ensure_unique_ids`.

data_steward/cleaners/clean_rule_4.py
```diff
@@ -0,0 +1,52 @@
...
+Rule: 4
+ID columns in each domain should be unique
+"""
+
+import bq_utils
+import cdm
+import clean_cdr_engine
+import resources
+
+rule_4_query = """
```
lrwb-aou: Make this a descriptive name.  For example, `ensure_unique_ids`.   I suggest updating the name of the module too.  When looking at files in a directory, no one is going to know what `clean_rule_4.py` is supposed/actually doing.

data_steward/cleaners/clean_rule_4.py
```diff
@@ -0,0 +1,52 @@
...
+            fields = resources.fields_for(table)
+            # Generate column expressions for select
+            col_exprs = [field['name'] for field in fields]
+            cols = ',\n        '.join(col_exprs)
+            query = rule_4_query.format(columns=cols,
+                                        project_id=project_id,
+                                        dataset_id=dataset_id,
+                                        domain_table=table,
+                                        table_name=table_name)
+            clean_cdr_engine.clean_dataset(project_id, dataset_id, query)
```
lrwb-aou: Have this return your query.  It doesn't need to execute it.

data_steward/cleaners/clean_rule_4.py
```diff
@@ -0,0 +1,52 @@
...
+    parser = argparse.ArgumentParser(
+        description='Parse project_id and dataset_id',
+        formatter_class=argparse.RawDescriptionHelpFormatter)
+    parser.add_argument('project_id',
+                        help='Project associated with the input and output datasets')
+    parser.add_argument('dataset_id',
+                        help='Dataset where cleaning rules are to be applied')
+    args = parser.parse_args()
+    if args.dataset_id:
+        run_clean_rule_4(args.project_id, args.dataset_id)
```
lrwb-aou: If you want to execute this as a one-off, use:
```
import clean_cdr_engine

if args.dataset_id:
    query_string = run_clean_ruole_4(args.project_id, args.dataset_id)
    clean_cdr_engine.clean_dataset(project_id, dataset_id, [query])
```

data_steward/cleaners/clean_cdr.py
```diff
@@ -88,7 +20,7 @@ def clean_rdr_dataset(project=None, dataset=None):
         LOGGER.info('Dataset is unspecified.  Using default value of:\t%s', dataset)
 
     LOGGER.info("Cleaning rdr_dataset")
-    _clean_dataset(project, dataset, rdr_consts.SQL_QUERIES)
+    rule_4.run_clean_rule_4(project, dataset)
```
lrwb-aou: This change only allows you to run a single query instead of a set of queries.  There will need to be a function to gather the SQL statements.  Then the statements are sent to the `clean_dataset(project_id, dataset_id, statements_list)` function.  As this is set up, this file will have to be modified every time we want to add a rule.

data_steward/cleaners/clean_cdr.py
```diff
@@ -97,33 +29,34 @@ def clean_ehr_dataset(project=None, dataset=None):
...
     LOGGER.info("Cleaning unioned_dataset")
-    _clean_dataset(project, dataset, unioned_consts.SQL_QUERIES)
+    rule_4.run_clean_rule_4(project, dataset)
 
 
 def clean_ehr_rdr_dataset(project=None, dataset=None):
     if dataset is None or dataset == '' or dataset.isspace():
         dataset = bq_utils.get_ehr_rdr_dataset_id()
-        LOGGER.info('Dataset is unspecified.  Using default value of:\t%s', dataset)
+        clean_engine.LOGGER.info('Dataset is unspecified.  Using default value of:\t%s', dataset)
```
lrwb-aou: remove this and just use `LOGGER.info`.  It's already set up.  You don't have to funnel it through `clean_engine.LOGGER`.

data_steward/cleaners/clean_cdr.py
```diff
@@ -97,33 +29,34 @@ def clean_ehr_dataset(project=None, dataset=None):
...
     LOGGER.info("Cleaning ehr_rdr_dataset")
-    _clean_dataset(project, dataset, combined_consts.SQL_QUERIES)
+    clean_engine.clean_dataset(project, dataset)
 
 
 def clean_ehr_rdr_unidentified_dataset(project=None, dataset=None):
     if dataset is None or dataset == '' or dataset.isspace():
         dataset = bq_utils.get_combined_deid_dataset_id()
-        LOGGER.info('Dataset is unspecified.  Using default value of:\t%s', dataset)
+        clean_engine.LOGGER.info('Dataset is unspecified.  Using default value of:\t%s', dataset)
```
lrwb-aou: remove this and just use `LOGGER.info`.  It's already set up.  You don't have to funnel it through `clean_engine.LOGGER`.

data_steward/cleaners/clean_cdr.py
```diff
@@ -1,85 +1,17 @@
...
-
-    if add_handler:
-        handler = logging.StreamHandler()
-        handler.setLevel(logging.INFO)
-        formatter = logging.Formatter('%(asctime)s - %(name)s - %(levelname)s - %(message)s')
-        handler.setFormatter(formatter)
-        LOGGER.addHandler(handler)
-
-
-def _clean_dataset(project=None, dataset=None, statements=None):
```
lrwb-aou: I really don't like this being made a public function in its own module.  I don't want anyone to call this outside of the restrictions we place on it because I think clean rules should be run as a group.  That's my opinion and it is open for discussion.

data_steward/cleaners/clean_cdr.py
```diff
@@ -1,145 +1,171 @@
...
-                       project, dataset)
 
-    if failures > 0:
-        print("Failed to apply %d clean rules for %s.%s",
-                       failures, project, dataset)
-        LOGGER.warning("Failed to apply %d clean rules for %s.%s",
-                       failures, project, dataset)
+def _gather_ehr_queries(project, dataset):
+    """
+    gathers all the queries required to clean ehr dataset
```
lrwb-aou: This is the only new comment.  When writing docstrings, write a single line summary statement, followed by a blank line.  You may then add more descriptions and parameter definitions.  Just separate sections with blank lines.  This is from [pep-257](https://www.python.org/dev/peps/pep-0257/).  [Pep-8 docstrings documentation](https://www.python.org/dev/peps/pep-0008/#documentation-strings) references pep-257.

PR #161, opened 2019-06-13
[DC-391] Person id validation

lrwb-aou:
This implements a check to ensure that the person_id found in multiple tables actually exist in the person table.  If a person_id does not exist in the person table, the row is deleted.  It performs an additional check on these tables to ensure we only keep EHR consented records.  This is done via the mapping tables.


lrwb-aou:
Closing this in favor of a clean pull request.

PR #166, opened 2019-06-18
[DC-388] Invalid foreign keys will be nulled out and the remainder of the row persists.

ksdkalluri:
Adds a rule (cleaning_rule_1) which generates queries to nullify invalid foreign keys.

### DIFF
--- a/data_steward/cleaners/clean_cdr.py
+++ b/data_steward/cleaners/clean_cdr.py
@@ -11,6 +11,7 @@ import bq_utils
 import clean_cdr_engine as clean_engine
 import constants.cleaners.clean_cdr as clean_cdr_consts
 import id_deduplicate as id_dedup
+import null_invalid_foreign_keys as null_foreign_key
 
 # import constants.bq_utils as bq_consts
 
@@ -26,7 +27,6 @@ def _gather_ehr_queries(project, dataset):
     :return: returns list of queries
     """
     query_list = []
-    query_list.extend(id_dedup.get_id_deduplicate_queries(project, dataset))
     return query_list
 
 
@@ -53,6 +53,7 @@ def _gather_ehr_rdr_queries(project, dataset):
     """
     query_list = []
     query_list.extend(id_dedup.get_id_deduplicate_queries(project, dataset))
+    query_list.extend(null_foreign_key.null_invalid_foreign_keys(project, dataset))
     return query_list
 
 
--- a/data_steward/cleaners/clean_cdr_engine.py
+++ b/data_steward/cleaners/clean_cdr_engine.py
@@ -7,6 +7,8 @@ import oauth2client
 from google.appengine.api import app_identity
 
 import bq_utils
+import constants.bq_utils as bq_consts
+import constants.cleaners.clean_cdr as cdr_consts
 
 LOGGER = logging.getLogger(__name__)
 FILENAME = '/tmp/cleaner.log'
@@ -32,6 +34,13 @@ def add_console_logging(add_handler):
 
 
 def clean_dataset(project=None, dataset=None, statements=None):
+    """
+       Run the assigned cleaning rules.
+
+       :param project:  the project name
+       :param dataset:  the dataset to clean
+       :param statements:  a list of dictionary objects to run the query
+       """
     if project is None or project == '' or project.isspace():
         project = app_identity.get_application_id()
         LOGGER.debug('Project name not provided.  Using default.')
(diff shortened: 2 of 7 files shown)


data_steward/cleaners/null_invalid_foreign_keys.py
```diff
@@ -0,0 +1,116 @@
...
+FOREIGN_KEYS_FIELDS = ['person_id', 'visit_occurrence_id', 'location_id', 'care_site_id', 'provider_id']
+
+INVALID_FOREIGN_KEY_QUERY = '''
+            SELECT {cols} 
+            FROM {project}.{dataset_id}.{table_name} t 
+            {join_expr}
+            '''
+
+
+def mapping_table_for(domain_table):
```
lrwb-aou: It seems like you are only using this function in this module.  You could give it python's version of a "protected" status by prefixing the name with an underscore, `_mapping_table_for`.  Then it will be available to your module, but will never be imported if someone were to use a wildcard import.  Single and double leading underscores are a pep 8 convention.  This [page](https://dbader.org/blog/meaning-of-underscores-in-python) does a nice job discussing this.

data_steward/cleaners/null_invalid_foreign_keys.py
```diff
@@ -0,0 +1,116 @@
...
+    """
+    Get name of mapping table generated for a domain table
+
+    :param domain_table: one of the domain tables (e.g. 'visit_occurrence', 'condition_occurrence')
+    :return: mapping table name
+    """
+    return '_mapping_' + domain_table
+
+
+def null_invalid_foreign_keys(project_id, dataset_id):
```
lrwb-aou: This generates a list of SELECT queries.  When is it actually updating the value to null?

data_steward/cleaners/clean_cdr.py
```diff
@@ -40,6 +40,7 @@ def _gather_rdr_queries(project, dataset):
     """
     query_list = []
     query_list.extend(id_dedup.get_id_deduplicate_queries(project, dataset))
+    query_list.extend(null_foreign_key.null_invalid_foreign_keys(project, dataset))
```
lrwb-aou: According to the response on the ticket, you can remove this.

data_steward/cleaners/clean_cdr.py
```diff
@@ -66,6 +68,7 @@ def _gather_ehr_rdr_de_identified_queries(project, dataset):
...
```
lrwb-aou: According to the response on the ticket, you can remove this.

data_steward/cleaners/clean_cdr.py
```diff
@@ -79,6 +82,7 @@ def _gather_unioned_ehr_queries(project, dataset):
...
```
lrwb-aou: According to the response on the ticket, you can remove this.

data_steward/cleaners/null_invalid_foreign_keys.py
ksdkalluri: I am using this strategy because, we can simply achieve that using the Left join to the tables where those keys are primary keys.

data_steward/cleaners/null_invalid_foreign_keys.py
ksdkalluri: Done

data_steward/cleaners/null_invalid_foreign_keys.py
lrwb-aou: You're right, the left join will return all the records from the left table and will fill null values for any record in the right table.  However, I thought that BigQuery created temporary result sets.  SELECT statements won't actually update the table values.   Or have I missed something in the BQ documentation?

ksdkalluri: the query function in bq_utils will update the original table if we pass, destination table and destination dataset_id.

lrwb-aou: Then that change needs to be made as part of this PR.  But it can't break the existing queries.

data_steward/cleaners/clean_cdr_engine.py
```diff
@@ -42,11 +51,28 @@ def clean_dataset(project=None, dataset=None, statements=None):
...
     successes = 0
     for statement in statements:
-        rule_query = statement.format(project=project, dataset=dataset)
+        query = statement.get(cdr_consts.QUERY, '')
+        rule_query = query.format(project=project, dataset=dataset)
+
+        legacy_sql = statement.get(cdr_consts.LEGACY_SQL, False)
+        destination_table = statement.get(cdr_consts.DESTINATION_TABLE, None)
+        retry = statement.get(cdr_consts.RETRY_COUNT, bq_consts.BQ_DEFAULT_RETRY_COUNT)
+        disposition = statement.get(cdr_consts.DISPOSITION, 'WRITE_EMPTY')
```
lrwb-aou: I've created a patch from your branch of my suggested changes.  Let me know if you want to look at it so you can apply it yourself to see how the code looks.  I'll discuss my suggestions here.  

I've noticed we use the write disposition strings frequently and now we're doing it in multiple places.  I think we should add WRITE_EMPTY, WRITE_TRUNCATE, and WRITE_APPEND to our constants/bq_utils.py module.  You've used the lowercase variants for the spellings in both query files.  BigQuery defines these as upper cased constants.  BQ probably won't recognize lower cased variants.  By adding these definitions to the constants file, we can import them and use them instead of writing each one and potentially introducing an error through misspellings.  

The patch addresses the write dispositions in bq_utils.py, constants/bq_utils.py, cleaners/id_deduplicate.py, and cleaners/null_invalid_foreign_keys.py

data_steward/cleaners/null_invalid_foreign_keys.py
```diff
@@ -0,0 +1,119 @@
+"""
+
```
lrwb-aou: See the [pep-257](https://www.python.org/dev/peps/pep-0257/#multi-line-docstrings) description of multi-line docstrings.  Specifically, it should be a summary line, followed by a blank line, followed by a description.  I would change this to:
```
"""
Cleaning Rule 1 : Foreign key references (i.e. visit_occurrence_id in the condition table) should be valid.

(Existing Achilles rule - validating for foreign keys include provider_id, care_site_id, location_id,
person_id,  visit_occurrence_id)
"""

data_steward/cleaners/clean_cdr_engine.py
ksdkalluri: can you push the changes to this branch

lrwb-aou: I can.  But I wasn't sure if you would want me to push the changes or let you review my suggestions first.  I will push them if you want.

lrwb-aou: I pushed my suggestions to the branch.  Look them over and let me know if you've got questions about why a change was suggested.

ksdkalluri:
> thoughts

well I used it so that it will be easy to debug when we get a query formatted in logs, but if it makes the api call leaner, i would probably switch back.

lrwb-aou:
The code base uses this block quotes for SQL a lot.  This might be something we should talk about.  The pros and cons of the SQL block string vs the SQL concatenated string.  The block string does make it more human readable in the logs...

PR #172, opened 2019-06-28
[DC-404] Ensure no Date or DTTM data point has a value > 30 days from the death date

ksdkalluri:
* Adds Clean rule to remove data points 30 days after the death date.
* refactored other cleaning rules.

### DIFF
--- a/data_steward/cdr_cleaner/cleaning_rules/no_data_30_days_after_death.py
+++ b/data_steward/cdr_cleaner/cleaning_rules/no_data_30_days_after_death.py
@@ -27,23 +27,23 @@ TEMPORAL_TABLES_WITH_NO_END_DATES = {'person': 'birth',
 
 REMOVE_DEATH_DATE_QUERY_WITH_END_DATES = (
     "SELECT * "
-    "FROM `{dataset}.{table_name}` "
+    "FROM `{project}.{dataset}.{table_name}` "
     "WHERE {table_name}_id NOT IN ("
     "  SELECT ma.{table_name}_id "
-    "  FROM `{dataset}.{table_name}` ma"
-    "  JOIN `{dataset}.death` d"
+    "  FROM `{project}.{dataset}.{table_name}` ma"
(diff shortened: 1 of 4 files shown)


data_steward/cdr_cleaner/cleaning_rules/no_data_30_days_after_death.py
```diff
@@ -0,0 +1,112 @@
...
+
+# Join Death to domain_table ON person_id
+# check date field is not more than 30 days after the death date
+# select domain_table_id from the result
+# use the above generated domain_table_ids as a list
+# select rows in a domain_table where the domain_table_ids not in above generated list of ids
+
+REMOVE_DEATH_DATE_QUERY_WITH_END_DATES = (
+    "SELECT * "
+    "FROM `{dataset}.{table_name}` "
```
lrwb-aou: I suggest adding the `{project_id}.` qualifier to lines 30, 33, 34, 45, and 46.

data_steward/cdr_cleaner/cleaning_rules/no_data_30_days_after_death.py
```diff
@@ -0,0 +1,112 @@
...
+    "WHERE {table_name}_id NOT IN ("
+    "  SELECT ma.{table_name}_id "
+    "  FROM `{dataset}.{table_name}` ma"
+    "  JOIN `{dataset}.death` d"
+    "  ON ma.person_id = d.person_id"
+    "  WHERE (date_diff(ma.{start_date},d.death_date, DAY) > 30"
```
lrwb-aou: This `WHERE` clause is my only other concern.  The issue says, 

> ensure that no records for that person_id exist more than 30 days after the death date.

This clause will allow records with a start_date of 29 or less days and an end_date of 30 or more days to remain in the CDR.  Is this desired?  Should the `AND` in the clause become an `OR`?  Allowing if either start or end date time is more than 30 past the death date, the entire row should be identified and dropped.

PR #187, opened 2019-08-01
[DC-379] Add completeness statistics to HPO report

mark-velez:
A data completeness section is added to the site report. For all columns of the CDM it displays how many records had null values and, if it is a concept field, how many had zero in the submission. It also displays the percentage of all records where the field is populated.

Special alerts are displayed on the top of the report. If the submission is invalid an error message will appear. If validation of a submission fails due to an internal error (and will therefore be re-run) a notice will appear.

image

* To ease testing/maintenance HPO report is refactored to use a jinja template
* Related validation tests are refactored to use mock and to reduce network overhead


data_steward/constants/validation/metrics/completeness.py
```diff
@@ -0,0 +1,39 @@
...
+FROM {dataset_id}.INFORMATION_SCHEMA.COLUMNS c
+ JOIN {dataset_id}.__TABLES__ t on c.table_name = t.table_id
+  WHERE t. table_id NOT LIKE '\\\\_%' 
+  AND c.IS_HIDDEN = 'NO'
+ ORDER BY table_name, c.ORDINAL_POSITION
+"""
+COMPLETENESS_SUBQUERY_FMT = """
+(SELECT *,
+ CASE 
+  WHEN table_row_count=0 THEN NULL
```
nishanthpp93: `null` here may not appear in the report as intended

PR #188, opened 2019-08-05
[DC-402] Ensure alignment of the domain_id in the concept table for each concept_id

ChaoPang:


ChaoPang:
The branch name needs to lowercased in order to create buckets on google cloud

PR #200, opened 2019-08-14
[DC-467] Id match query length

lrwb-aou:
Updates complex/too long queries for Big Query.  Updates required unit tests.


data_steward/constants/validation/participants/writers.py
```diff
@@ -1,7 +1,16 @@
+from constants.bq_utils import (WRITE_APPEND, WRITE_TRUNCATE)
+from constants.validation.participants.identity_match import (
```
nishanthpp93: This statement is sufficient to define the constants below

data_steward/constants/validation/participants/writers.py
```diff
@@ -1,7 +1,16 @@
...
+    MATCH, MISMATCH, MISSING,
+    PERSON_ID_FIELD, FIRST_NAME_FIELD, MIDDLE_NAME_FIELD, LAST_NAME_FIELD,
+    EMAIL_FIELD, PHONE_NUMBER_FIELD, SEX_FIELD, ZIP_CODE_FIELD, STATE_FIELD,
+    CITY_FIELD, ADDRESS_ONE_FIELD, ADDRESS_TWO_FIELD, BIRTH_DATE_FIELD, VALIDATION_FIELDS)
+
+WRITE_APPEND = WRITE_APPEND
```
nishanthpp93: All of these assignments can be removed

lrwb-aou: Yes, but static analysis tools complain when you import something and don't use it.  And I can see the point.  How do you know if something should actually be imported if it isn't used in the module?  The tools don't look outside of the module to see how the constants are used elsewhere and therefore don't know that they are used elsewhere.  This is a quirk that makes sense to me.

data_steward/validation/participants/identity_match.py
```diff
@@ -741,6 +785,23 @@ def write_results_to_drc_bucket(project, validation_dataset=None):
                     )
 
 
+def _add_matches_to_results(results, matches, field):
```
nishanthpp93: There is one more case that can occur here: The individual does not exist in `matches` but exists in `results`. The value for the field should be assigned to `consts.MISSING` for such individuals.

data_steward/constants/validation/participants/writers.py
nishanthpp93: I see. I would consider this a bug in the analysis tools. For Codacy, we can ignore issues for unused imports here.

data_steward/validation/participants/identity_match.py
lrwb-aou: You are right.  This case is handled by this function.  For example, when this gets triggered the first time, every individual's entry in the matches dictionary will run through the initialization condition for an individual.  This initialization condition sets every value in the individual's dictionary to the `consts.MISSING` value.  So, on the first name comparison, everything is set to `consts.MISSING` then the first name match values are assigned.  Now if person person 1 has first name data but no last name data, they will not be part of the matches dictionary sent to this function with the last name data.  This is fine because when the person dictionary definition was created all their fields were initialized to `consts.MISSING`.

nishanthpp93: Yes, that makes sense. Thanks for clarifying that, Lauren.
I now have a question which need not be addressed in this pull request. If an individual does not exist in any of the `pii` tables, but does exist in the `participant_match` table, should they be included in the `identity_match` table with all fields marked as missing? Right now, it looks like they do not exist in the `identity_match` table, which I think is also fine. It just needs to be communicated to the sites and anyone querying the two match tables.

PR #225, opened 2019-09-15
Adds script to run the deid with environment settings

ksdkalluri:
* Updates run_curation_pipeline.sh to use new deid_runner script


data_steward/deid_runner.sh
```diff
@@ -0,0 +1,108 @@
...
+    ;;
+  --)
+    shift
+    break
+    ;;
+  *) break ;;
+  esac
+done
+
+if [[ -z "${key_file}" ]] || [[ -z "${cdr_id}" ]] || [[ -z "${cdr_id}" ]]; then
```
mark-velez: @ksdkalluri cdr_id appears twice

data_steward/deid_runner.sh
```diff
@@ -0,0 +1,108 @@
...
+
+cp -R "${GAE_SDK_APPENGINE}" "${VENV_LIB_GOOGLE}"
+cp -R "${GAE_SDK_NET}" "${VENV_LIB_GOOGLE}"
+
+export BIGQUERY_DATASET_ID="${cdr_deid}"
+export PYTHONPATH="${PYTHONPATH}:${DEID_DIR}:${DIR}"
+
+# Get Git version tag
+tag=$(git describe --abbrev=0 --tags)
+version=${tag}
```
mark-velez:
```suggestion
# Version is the most recent tag accessible from the current branch
version=$(git describe --abbrev=0 --tags)
```

ksdkalluri: added this change.

ksdkalluri:
> @ksdkalluri this looks good! Please just address the two minor fixes.
> 
> I feel like `deid_runner.sh` should be collocated with `run_curation_pipeline.sh`. What do you think about moving `deid_runner.sh` and `run_deid.py` to `tools/`? It wouldn't necessarily have to be part of this PR. I could take it on if you think it's a good idea.

I have made these changes. Please review.

PR #257, opened 2019-10-03
[DC-518]  create questionnaire_response_id mappings

lrwb-aou:
Creates a mapping table.  Mapping will be done manually for the time being.


mark-velez:
Seems to generate the response mappings as intended. In the future it should include updating the deid observation table. Doing manually for now.

PR #268, opened 2019-10-18
Add pop_retract notebook [ci skip]

mark-velez:
Determines number of records per table to be retracted and compares
expected vs actual counts


mark-velez:
Hey @nishanthpp93 do you think we could just use `jinja`  rather than `jinjasql` and would it be easier?

mark-velez:
@nishanthpp93 I was able to template queries using `jinja` only (no need for `jinjasql`).

PR #271, opened 2019-10-18
[DC-518]/[DC-544]  integrating questionnaire_response_id permutation

lrwb-aou:
Allows deid to use multiple sql updates
Previously deid created one large SQL statement.  This architecture eventually failed on BigQuery.  This was changed to use multiple SQL statements with WRITE_APPEND instead of a UNION ALL approach.
This also permutes the value of questionnaire_response_ids.


mark-velez:
Hey @lrwb-aou this looks good to me. One thing I'm not clear about- the signature of base method `Press.submit(self, **args)` seems to me to differ from derived method `AOU.submit(self, sql, create)`. I'm not sure, is it possible this might cause issues if ever we were using `Press` references to call `submit()`..? I don't think this ever comes up so I will go ahead and merge this.

lrwb-aou:
Hey @mark-velez , you're right, it's odd.  I was trying to make codacy happy.  I originally had `Presss.submit(self, sql, create)` but codacy was still complaining.  I will switch back to that command line in another PR.

PR #279, opened 2019-10-22
Use binary read for calls to gcs_utils.upload_object

mark-velez:
* Remove assertDictContainsSubset keyword args


mark-velez:
Merged with different PR

PR #294, opened 2019-10-28
[DC-553] Add sandbox datasets as part of the cdr_cleaner

ChaoPang:


mark-velez:
Put `get_sandbox_dataset_id` and `create_sandbox_dataset` in a separate module (i.e. `data_steward.sandbox`)

PR #295, opened 2019-10-29
Moved staging.UNION from the wrong function call to clean_dataset [ci skip]

ChaoPang:


mark-velez:
@ChaoPang please let me know how you'd like to address the failing circleci build- you might have multiple options here.

mark-velez:
@ChaoPang it looks like this change is in the repo already; can we close this?

PR #306, opened 2019-11-06
[DC-500] Removes operational PII fields from RDR export

ksdkalluri:


data_steward/cdr_cleaner/clean_cdr.py
```diff
@@ -35,6 +33,9 @@
 import cdr_cleaner.cleaning_rules.valid_death_dates as valid_death_dates
 import cdr_cleaner.manual_cleaning_rules.negative_ppi as negative_ppi
 import cdr_cleaner.manual_cleaning_rules.ppi_drop_duplicate_responses as ppi_drop_duplicates
+import cdr_cleaner.manual_cleaning_rules.remove_operational_pii_fields as operational_pii_fields
+import sandbox
+from constants.cdr_cleaner.clean_cdr import DataStage as stage
```
lrwb-aou: I know this isn't part of this code review, but why did we create a class, `DataStage`, to represent a list of constant values?

data_steward/cdr_cleaner/manual_cleaning_rules/remove_operational_pii_fields.py
```diff
@@ -0,0 +1,101 @@
...
+  SELECT
+    *
+  FROM
+    {dataset}.observation
+  WHERE
+    observation_id IN (
+    SELECT
+      observation_id
+    FROM
+      `{project}.{dataset}_sandbox.{pii_fields_table}` as pii
```
lrwb-aou: This line hard codes the `_sandbox` dataset naming convention.  This is repeated on line 48.  Consider changing `get_remove_operational_pii_fields_query(project_id, dataset_id)` to `get_remove_operational_pii_fields_query(project_id, dataset_id, sandbox_dataset_id)`

data_steward/cdr_cleaner/clean_cdr.py
```diff
@@ -72,6 +73,7 @@ def _gather_rdr_queries(project_id, dataset_id, sandbox_dataset_id):
         ppi_drop_duplicates.get_remove_duplicate_set_of_responses_to_same_questions_queries(project_id,
                                                                                             dataset_id,
                                                                                             sandbox_dataset_id))
+    query_list.extend(operational_pii_fields.get_remove_operational_pii_fields_query(project_id, dataset_id))
```
lrwb-aou: Consider adding the `sandbox_dataset_id` to this signature and using it in your intermediate queries.

data_steward/cdr_cleaner/manual_cleaning_rules/remove_operational_pii_fields.py
```diff
@@ -0,0 +1,101 @@
...
+
+def get_remove_operational_pii_fields_query(project_id, dataset_id):
+    """
+
+    :param project_id: Name of the project
+    :param dataset_id: Name of the dataset where the queries should be run
+    :return:
+    """
+
+    bq_utils.load_table_from_csv(project_id=ARGS.project_id, dataset_id=ARGS.dataset_id + '_sandbox',
```
lrwb-aou: You need to remove the `ARGS` references from this line.  This function doesn't have an `ARGS` argument.  You also don't need to explicitly set the `csv_path` and `fields` parameters, because they already default to `None`.  You should use the `sandbox_dataset_id` here.  With these changes, this line becomes:
`
bq_utils.load_table_from_csv(
    project_id=project_id,
    dataset_id=sandbox_dataset_id,
    table_name=OPERATIONAL_PII_FIELDS_TABLE
)
`

data_steward/cdr_cleaner/manual_cleaning_rules/remove_operational_pii_fields.py
```diff
@@ -0,0 +1,101 @@
+# coding=utf-8
+"""
+Some new operational fields exists, that were not blacklisted in the RDR export. These rows needs to be dropped in the
```
lrwb-aou: Possibly add a short summary and blank line before this paragraph.  https://www.python.org/dev/peps/pep-0257/#multi-line-docstrings

data_steward/cdr_cleaner/manual_cleaning_rules/remove_operational_pii_fields.py
```diff
@@ -0,0 +1,101 @@
...
+    """
+
+    bq_utils.load_table_from_csv(project_id=ARGS.project_id, dataset_id=ARGS.dataset_id + '_sandbox',
+                                 table_name=OPERATIONAL_PII_FIELDS_TABLE, csv_path=None, fields=None)
+
+    queries_list = []
+
+    # Save operational pii records being deleted in sandbox `dataset.intermediary_table` .
+    query = dict()
+    query[cdr_consts.QUERY] = OPERATION_PII_FIELDS_INTERMEDIARY_QUERY.format(dataset=dataset_id,
```
lrwb-aou: Have you tried running this using just the query with the API?  Does it work successfully without explicitly setting the `destination_dataset_id` parameter?  If it does, no change is needed.

data_steward/cdr_cleaner/manual_cleaning_rules/remove_operational_pii_fields.py
```diff
@@ -0,0 +1,113 @@
...
+import logging
+import os
+
+import bq_utils
+import constants.cdr_cleaner.clean_cdr as cdr_consts
+import sandbox
+
+LOGGER = logging.getLogger(__name__)
+
+module_name = os.path.basename(file[:-3])
```
lrwb-aou: This won't work.  It has to be `__file__[:-3]`

data_steward/cdr_cleaner/manual_cleaning_rules/remove_operational_pii_fields.py
ksdkalluri: I tried running it with the API, and it worked.

data_steward/cdr_cleaner/clean_cdr.py
ksdkalluri: I am not sure, I created data_stage in the Parameter to make clean_cdr run on particular stage. But Datastage is added by Chao, i'll find out.

data_steward/resources/fields/operational_pii_fields.json
```diff
@@ -0,0 +1,17 @@
+[
+  {
+    "type": "string",
+    "name": "observation_source_value",
+    "mode": "nullable"
```
lrwb-aou: Do we really want or expect these fields to be nullable?  It's easier to relax a `mode` than it is to strengthen one.  

Also, if you wouldn't mind adding descriptions that would be really helpful for future maintenance efforts.  Out of context, I'm not sure what `total_count` represents.

ksdkalluri: These were the tables given to us, even I am not sure about the total_count too. we are not using it in the query. I can make fields required.

lrwb-aou: Interesting.  Who gave us the tables?

PR #308, opened 2019-11-07
[DC-499] Update survey questions and answers that arew not mapped to OMOP concepts.

ksdkalluri:


data_steward/cdr_cleaner/manual_cleaning_rules/update_questiona_answers_not_mapped_to_omop.py
```diff
@@ -0,0 +1,167 @@
...
+
+import bq_utils
+import constants.cdr_cleaner.clean_cdr as cdr_consts
+import sandbox
+
+LOGGER = logging.getLogger(__name__)
+
+OLD_MAP_SHORT_CODES_TABLE = 'old_map_short_codes'
+
+OLD_MAP_SHORT_CODES_TABLE_FIELDS = [
```
lrwb-aou: Should we define this table in the `resources/fields` section?  Maybe consider changing mode to required.  This will be easier to relax to nullable in the future if it is needed.  Please add a `description` string for each field.  This will make maintenance easier for everyone.

data_steward/cdr_cleaner/manual_cleaning_rules/update_questiona_answers_not_mapped_to_omop.py
```diff
@@ -0,0 +1,167 @@
...
+        ON
+            (standard_c.concept_id=c_r.concept_id_2)
+        WHERE
+            source_c.vocabulary_id='PPI'
+            AND c_r.relationship_id LIKE 'Maps to%'
+        ) map
+    WHERE
+        map.observation_source_value=obs.observation_source_value"""
+
+UPDATE_ANSWERS_MAP_QUERY = """
```
lrwb-aou: This is duplicating the query above.  Suggest removing one query, renaming the other to something like 'UPDATE_MAP_QUERY' and parameterizing the type.

data_steward/cdr_cleaner/manual_cleaning_rules/update_questiona_answers_not_mapped_to_omop.py
```diff
@@ -0,0 +1,167 @@
...
+        WHERE
+            source_c.vocabulary_id='PPI'
+        AND c_r.relationship_id LIKE 'Maps to%') map
+    WHERE
+        map.value_source_value=obs.value_source_value"""
+
+
+def get_update_questions_answers_not_mapped_to_omop(project_id, dataset_id, sandbox_dataset_id):
+    """
+
```
lrwb-aou: Please add a short description.

data_steward/cdr_cleaner/manual_cleaning_rules/update_questiona_answers_not_mapped_to_omop.py
```diff
@@ -0,0 +1,167 @@
...
+    :param project_id: Name of the project
+    :param dataset_id: Name of the dataset where the queries should be run
+    :param sandbox_dataset_id: Name of the sandbox dataset
+    :return:
+    """
+
+    bq_utils.load_table_from_csv(project_id=project_id,
+                                 dataset_id=sandbox_dataset_id,
+                                 table_name=OLD_MAP_SHORT_CODES_TABLE,
+                                 csv_path=None,
```
lrwb-aou: Explicitly setting csv_path to None is not necessary.  None is the default value.

data_steward/cdr_cleaner/manual_cleaning_rules/update_questiona_answers_not_mapped_to_omop.py
```diff
@@ -0,0 +1,167 @@
...
+                                 fields=OLD_MAP_SHORT_CODES_TABLE_FIELDS)
+
+    queries_list = []
+
+    # Update concept_ids to questions using OLD_MAP_SHORT_CODES_TABLE.
+    query = dict()
```
lrwb-aou: If you use one query instead of two separate queries, this can become
```
for concept_type in ['Question', 'Answer']:
    query = {}
    query[cdr_consts.QUERY] = UPDATE_MAP_QUERY.format(
        dataset=dataset_id,
        project=project_id,
        old_map=OLD_MAP_SHORT_CODES_TABLE,
        sandbox=sandbox_dataset_id
    )
queries_list.append(query)
```

data_steward/cdr_cleaner/manual_cleaning_rules/update_questiona_answers_not_mapped_to_omop.py
ksdkalluri: there are also some changes in the fields, thats the reason i used different queries.

data_steward/cdr_cleaner/manual_cleaning_rules/update_questiona_answers_not_mapped_to_omop.py
```diff
@@ -0,0 +1,167 @@
...
+    UPDATE
+        `{project}.{dataset}.observation` obs
+    SET
+        observation_concept_id=new_observation_concept_id,
+        observation_source_concept_id=new_observation_source_concept_id
+    FROM
+        (SELECT
+        DISTINCT short_pmi_code AS observation_source_value,
+        source_c.concept_id AS new_observation_source_concept_id,
+        FIRST_VALUE(standard_c.concept_id) OVER (PARTITION BY source_c.concept_id ORDER BY c_r.relationship_id DESC ) AS new_observation_concept_id
```
ksdkalluri: @lrwb-aou these fields are also separate in each type of query.1

lrwb-aou: ok.  The structure is so similar, I thought it would be easy to create a single query and just provide parameters for the differences.

data_steward/cdr_cleaner/manual_cleaning_rules/update_questiona_answers_not_mapped_to_omop.py
lrwb-aou: You're right.  Parameterizing this would add 4-5 extra string parameters.  This is fine.

data_steward/cdr_cleaner/manual_cleaning_rules/update_questiona_answers_not_mapped_to_omop.py
lrwb-aou: No longer relevant.

PR #314, opened 2019-11-11
[DC-551] Update app_base.yaml and gcloudignore for GAE py3

mark-velez:
* Set runtime to python37
 * Remove obsolete directives: api_version, threadsafe and skip_files
 * Add .gcloudignore file to replace skip_files directive
 * Remove admin handler because multiple script keys not permitted,
   still must set up handler for it

### DIFF
--- a/data_steward/.gcloudignore
+++ b/data_steward/.gcloudignore
@@ -80,13 +80,7 @@ ipython_config.py
 .python-version
 
 # Environments
-.env
-.venv
-env/
-venv/
-ENV/
-env.bak/
-venv.bak/
+*_env/
 
 # Ignored by the build system
 /setup.cfg
--- a/data_steward/.gcloudignore
+++ b/data_steward/.gcloudignore
@@ -93,3 +93,6 @@ config/
 test/
 *.sh
 dev_requirements.txt
+
+logs/
+LOGS/
--- a/data_steward/.gcloudignore
+++ b/data_steward/.gcloudignore
@@ -93,6 +93,5 @@ venv.bak/
 ci/
 config/
 test/
-tools/
 *.sh
 dev_requirements.txt
--- a/data_steward/.gcloudignore
+++ b/data_steward/.gcloudignore
@@ -1,0 +1,98 @@
+# This file specifies files that are *not* uploaded to Google Cloud Platform
+# using gcloud. It follows the same syntax as .gitignore, with the addition of
+# "#!include" directives (which insert the entries of the given .gitignore-style
+# file at that point).
+#
+# For more information, run:
+#   $ gcloud topic gcloudignore
+#
+.gcloudignore
+.git
+.gitignore
+
+# Byte-compiled / optimized / DLL files
+__pycache__/
+*.py[cod]
+*$py.class
+
+# C extensions
+*.so
+
+# Distribution / packaging
+.Python
+build/
+develop-eggs/
+dist/
+downloads/
+eggs/
+.eggs/
+lib/
+lib64/
+parts/
+sdist/
+var/
+wheels/
+pip-wheel-metadata/
+share/python-wheels/
+*.egg-info/
+.installed.cfg
+*.egg
+MANIFEST
+
+# PyInstaller
+#  Usually these files are written by a python script from a template
+#  before PyInstaller builds the exe, so as to inject date/other infos into it.
+*.manifest
+*.spec
+
+# Installer logs
+pip-log.txt
+pip-delete-this-directory.txt
+
+# Unit test / coverage reports
+.tox/
+.nox/
+.coverage
+.coverage.*
+.cache
+nosetests.xml
+coverage.xml
+.pytest_cache/
+
+# Django stuff:
+*.log
+local_settings.py
+db.sqlite3
+db.sqlite3-journal
+
+# Flask stuff:
+instance/
+.webassets-cache
+
+# Jupyter Notebook
+.ipynb_checkpoints
+
+# IPython
+profile_default/
+ipython_config.py
+
+# pyenv
+.python-version
+
+# Environments
+.env
+.venv
+env/
+venv/
+ENV/
+env.bak/
+venv.bak/
+
+# Ignored by the build system
+/setup.cfg
+ci/
+config/
+test/
+tools/
+*.sh
+dev_requirements.txt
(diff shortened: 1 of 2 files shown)


data_steward/.gcloudignore
```diff
@@ -0,0 +1,98 @@
+# This file specifies files that are *not* uploaded to Google Cloud Platform
+# using gcloud. It follows the same syntax as .gitignore, with the addition of
+# "#!include" directives (which insert the entries of the given .gitignore-style
+# file at that point).
+#
+# For more information, run:
+#   $ gcloud topic gcloudignore
+#
+.gcloudignore
```
lrwb-aou: There are multiple redundancies between `.gitignore` and this `.gcloudignore` file.  Why not make use of the include directive?
```
#!include:.gitignore
```
Then you can remove lines 13-33, 37-39, 42-53, 55-59, 63, 64, 68-73, 79-80.

data_steward/.gcloudignore
```diff
@@ -0,0 +1,98 @@
...
+profile_default/
+ipython_config.py
+
+# pyenv
+.python-version
+
+# Environments
+.env
+.venv
+env/
```
lrwb-aou: I'm not sure what the virtual environment naming convention used by the CUMC team is, but I tend to use `<some_pattern>_env`.  So a directive like `*_env/` would exclude all virtual environment files like this without needing to explicitly add each one to this ignore file.

data_steward/app_base.yaml
```diff
@@ -1,29 +1,16 @@
...
-    secure: always
-    script: auto
+#  TODO set up handler for Admin API in py3
+#  - url: /admin/.*
+#    script: admin.admin_api.app
+#    secure: always
+#    script: auto
 
   - url: /.*
     script: validation.main.app
```
ChaoPang: The attribute script is duplicated with the `script: auto` below

data_steward/.gcloudignore
mark-velez: I like the idea of removing redundancy, but I avoided the include for fear of repeating a prior experience. I did something like this previously and thereby excluded a file from deployment unintentionally. The resulting error I encountered in prod `module 'xyz' not found` only occurred several hours later and was not easy to figure out at the time or to reproduce in tests (because gcloudignore affects the runtime environment only). I believe files like `{app,cron}.yaml` we **should** include in gitignore but not in gcloudignore. Does this make sense?

data_steward/.gcloudignore
mark-velez: Full disclosure: I just copy/pasted this ignore file from elsewhere. At Columbia we might be following different practices with venv names. I personally keep mine out of my working directory altogether so I'm happy to use whatever pattern as long as it works for everyone. Will ping everyone.

mark-velez: @lrwb-aou I have consulted others. I will update to `*_env/` like you suggested. Thanks! :)

mark-velez:
Note: The fix for the circle failure was merged to develop already; squash/merge should incorporate it

data_steward/.gcloudignore
lrwb-aou: I understand your hesitation.  But we are not currently ignoring `{app,cron}.yaml` files in our `.gitignore` file.  Even if we were, we can explicitly add those files to be included with the deployment by adding these:
```
!app.yaml
!cron.yaml
```
to the `.gcloudignore` file.  Also, if we need to list the files that will be uploaded we can use the `gcloud meta list-files-for-upload` command.

Out of curiousity, why should those files be excluded from git?  I've been on projects where configuration files were templated and version controlled to make sure we knew what variables had to be set and how.  Then each dev had a functional copy of the file.  To make sure we didn't store passwords/keys/secrets, we ran `git update-index --assume-unchanged <filepath>` to prevent us from accidentally uploading credentials.  https://stackoverflow.com/questions/10755655/git-ignore-tracked-files

If you prefer to err on the side of caution, I understand.  It just seems like both files are adding a lot of boilerplate to ignore packages/imports/files that we aren't even using in the project.

lrwb-aou: Final decision is yours.  Please don't forget to update the file to include the line `*_env/`.

mark-velez: Thanks @lrwb-aou! I didn't know about that meta command. This makes me more comfortable about including `.gitignore` in `.gcloudignore` so I will look into it. :)

Prior DRC discussions suggested we should be cautious about any secrets or identifiers for cloud resources in our repositories. I think we agree our approach has to strike a balance between confidentiality and ease of deployment- paranoia could mean terribly incovenient/error-prone deployment. Yes to templating the yaml files! I am [using jinja to template cron.yaml](https://github.com/all-of-us/curation/blob/develop/data_steward/resources/templates/cron.tpl.yaml) and would like for us to do something similar for `app.yaml`. Please **do** share any suggestions you may have on how engineers can better share app secrets.

lrwb-aou: I'm in complete agreement about keeping secrets to ourselves.  We definitely don't want to version any of that.

I love the cron template.  I haven't personally used jinja often, but I know it's a powerful tool and the template is very clean and understandable.  I'll look around to see if there is anything better than telling git to ignore file changes.  I would think there is.

mark-velez:
@lrwb-aou please check that I integrated the changes we discussed. Note I had to rebase onto py3 so commits are different.

data_steward/.gcloudignore
```diff
@@ -0,0 +1,95 @@
...
+
+# IPython
+profile_default/
+ipython_config.py
+
+# pyenv
+.python-version
+
+# Environments
+*_env/
```
lrwb-aou: One other suggestion.  You might consider ignoring logging directories (`logs/` or `LOGS/`), rather than just files that end in `.log`.  The deid software saves the SQL it creates in `*.sql` files in the `LOGS` directory.  I don't think we have a conventional logging directory location, but a directory name like `LOGS` is fairly common.  Since deid logs to it, it contains `*.log` and `*.sql` files, none of which are needed by GAE.  It would be nice to start logging all of this stuff (when it runs locally) to a common location.

Everything else looks fine.

mark-velez: Good idea @lrwb-aou. We wouldn't want to deploy these files and be misled by log files we may have uploaded.

PR #323, opened 2019-11-19
[DC-579]  Update deid with python 3 required changes

lrwb-aou:
* Add deid required packages so unit tests will execute.
* Allow existing unit test to execute.
* Change defunct StandardError to a tuple of probable Error types.
* Throw RuntimeError if the GOOGLE_CLOUD_PROJECT environment variable is not set.  Fast failure and better error message.
* Remove long type from code, (all int types became longs in python 3).
* Update run_deid script to create LOGS directory if not previously created.
* Store deid logs in LOGS.
* Add forgotten table to deid config files.  Required for the observation de-identifying execution.


.circleci/config.yml
```diff
@@ -46,7 +46,10 @@ jobs:
           command: python3 -m venv ~/curation_env && echo "source ~/curation_env/bin/activate" >> $BASH_ENV
       - run:
           name: Upgrade pip and install requirements
-          command: pip install --upgrade pip setuptools && pip install -r requirements.txt
+          command: |
+            pip install --upgrade pip setuptools
+            pip install -r requirements.txt
+            pip install -r deid/requirements.txt
```
nishanthpp93: I think it's best if we include versions in `deid/requirements.txt`
However this doesn't need to be addressed now and it may not be as simple, since some package versions might conflict with other packages in `requirements.txt`. We need to do a `pip freeze` with everything working during our deid test run and add the right versions.

PR #333, opened 2019-11-26
Py3 key notification

ChaoPang:


data_steward/admin/admin_api.py
```diff
@@ -3,16 +3,19 @@
...
 
+import slack
 import api_util
 from admin import key_rotation
 
+SLACK_TOKEN = 'SLACK_TOKEN'
+SLACK_CHANNEL = 'SLACK_CHANNEL'
+DEFAULT_SLACK_CHANNEL = 'test_channel'
+
 LOGGER = logging.getLogger(__name__)
```
mark-velez: @ChaoPang it doesn't look like `LOGGER` is used- should it be?

PR #337, opened 2019-11-26
Fix path and dataset_name in aou.py

ksdkalluri:
[ci skip]

### DIFF
--- a/data_steward/deid/aou.py
+++ b/data_steward/deid/aou.py
@@ -72,26 +72,25 @@ DESIGN:
                         submit      will create an output table
                         debug       will just print output without simulation or submit (runs alone)
 """
-# Python imports
-from copy import copy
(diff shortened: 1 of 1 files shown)


data_steward/deid/aou.py
```diff
@@ -335,7 +334,7 @@ def map_questionnaire_response_ids(self, lower_bound):
 
         :param lower_bound:  The smallest number that may be used as an identifier.
         """
-        map_tablename = "_deid_questionnaire_response_map"
+        map_tablename = self.idataset + "._deid_questionnaire_response_map"
```
lrwb-aou: Interesting that this failed.  It was supposed to use the dataset name passed by the config dictionary.  This same type of query and config is used in other places in aou.py.  I wonder why this one failed.

PR #339, opened 2019-11-27
[DC-587] Added the check in the log handler to check if the environment is GAE [skip CI]

ChaoPang:


ChaoPang:
The log messages are just printed out if the environment is not in GAE_ENV, that’s why those log messages are printed to the CircleCI console.

lrwb-aou:
This is much better.  The only test that looks odd is the CombineEhrRdrTest.  That may be due to setting debug messages to info level messages.?.?  Thanks for the changes Chao.

PR #343, opened 2019-12-03
[DC-595] Participant match datetime bug

lrwb-aou:
The unit test is decorated as an expectedFailure.  When the
bug is fixed, the test will write out as '(unexpected_successes=1)'.
The expectedFailure decorator should be removed then.


lrwb-aou:
Careful!  That's a slippery slope to test driven development.  LOL.

PR #348, opened 2019-12-07
[DC-584] Drop participants lacking both "The Basics" and any EHR data

calbach:
Putting this up for initial feedback, will add tests once I've gotten directional confirmation.


data_steward/common.py
```diff
@@ -30,6 +30,13 @@
 AOU_REQUIRED = [CARE_SITE, CONDITION_OCCURRENCE, DEATH, DEVICE_EXPOSURE, DRUG_EXPOSURE,
                 FACT_RELATIONSHIP, LOCATION, MEASUREMENT, NOTE, OBSERVATION, PERSON,
                 PROCEDURE_OCCURRENCE, PROVIDER, SPECIMEN, VISIT_OCCURRENCE]
+# TODO: Document.
```
calbach: @ChaoPang I could use help understanding how to document this set of tables accurately - or is there a better name I should choose instead? Or are these really just specific to the person_id_validator tool, and I shouldn't have pulled them out?

For context: my goal was to get a list of all tables which have a person ID in them, where the rows have provenance (indicating EHR vs RDR).

data_steward/cdr_cleaner/cleaning_rules/drop_participants_without_ppi_or_ehr.py
```diff
@@ -0,0 +1,92 @@
+"""
+Run the drop_participants_without_ppi_or_ehr validation clean rule.
```
calbach: @mark-velez please review the approach, let me know if there is any better way to proceed here you can think of.

data_steward/cdr_cleaner/cleaning_rules/drop_participants_without_ppi_or_ehr.py
```diff
@@ -0,0 +1,92 @@
...
+import resources
+
+LOGGER = logging.getLogger(__name__)
+
+BASICS_MODULE_CONCEPT_ID = 1586134
+
+SELECT_PERSON_WITH_BASICS_OR_EHR = """
+SELECT {fields}
+FROM `{project}.{dataset}.{table}` person
+LEFT JOIN (
```
calbach: @mark-velez please also review this - my strategy is to perform an outer join across a TheBasics subquery, and every EHR table subquery to gather the criteria for the decision.

data_steward/common.py
ChaoPang: @calbach this is a good question. The **person_id_validator** rule checks if the person_id is valid in all the domain tables that contain patient data, which is a subset of the [OMOP Standardized Clinical Data Tables](https://github.com/OHDSI/CommonDataModel/wiki/Standardized-Clinical-Data-Tables). 

1. This is just my personal opinion, I don't think the set of tables in your cleaning rule is the same as **person_id_validator** because it should inform us what AOU required tables contain patient data, whereas **person_id_validator** separated them out into 2 sets. 

2.  In addition, HPO sites will start submitting note data next year so we will have to add **note** and **note_nlp** to this set and MAPPED_VALIDATION_TABLES (or whatever this list should be called) needs to be updated.

data_steward/cdr_cleaner/cleaning_rules/drop_rows_for_missing_persons.py
```diff
@@ -0,0 +1,39 @@
...
+
+        delete_query = SELECT_EXISTING_PERSON_IDS.format(
+            project=project, dataset=dataset, table=table, fields=fields
+        )
+
+        query_list.append({
+            clean_consts.QUERY: delete_query,
+            clean_consts.DESTINATION_TABLE: table,
+            clean_consts.DESTINATION_DATASET: dataset,
+            clean_consts.DISPOSITION: bq_consts.WRITE_TRUNCATE,
```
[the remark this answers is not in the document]
calbach: Thanks, fixed.

data_steward/cdr_cleaner/cleaning_rules/drop_rows_for_missing_persons.py
```diff
@@ -0,0 +1,39 @@
+import clean_consts
```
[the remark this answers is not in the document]
calbach: Fixed

data_steward/cdr_cleaner/cleaning_rules/drop_participants_without_ppi_or_ehr.py
```diff
@@ -0,0 +1,92 @@
+"""
```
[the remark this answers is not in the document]
calbach: Thanks for the example, done

data_steward/common.py
calbach: Thanks, this is very helpful. I've renamed the tables and added documentation. Please let me know if this makes sense to you.

data_steward/cdr_cleaner/cleaning_rules/drop_participants_without_ppi_or_ehr.py
mark-velez: @calbach I think the left join filtering strategy is good! In terms of syntax- we often find it difficult to understand and troubleshoot queries (SQL) interpersed with procedural logic (Python). To address this we just started to template our SQL (see [pop_retract.py](https://github.com/all-of-us/curation/blob/develop/data_steward/analytics/cdr_ops/pop_retract.py#L72)). To be sure, templates come with challenges of their own, but I'm wondering your experience, if you think using them here might improve intelligibility/testability. Either way, approach looks good. :+1:  Happy to discuss if you like.

data_steward/cdr_cleaner/cleaning_rules/drop_participants_without_ppi_or_ehr.py
```diff
@@ -0,0 +1,114 @@
...
+        clean_consts.DESTINATION_TABLE: 'person',
+        clean_consts.DESTINATION_DATASET: dataset,
+        clean_consts.DISPOSITION: bq_consts.WRITE_TRUNCATE,
+    }] + drop_rows_for_missing_persons.get_queries(project, dataset)
+
+
+if __name__ == '__main__':
+    import cdr_cleaner.clean_cdr_engine as clean_engine
+
+    parser = argparse.ArgumentParser(formatter_class=argparse.RawDescriptionHelpFormatter)
```
mark-velez: @calbach I get unresolved reference `argparse` here- is an import missing?

data_steward/cdr_cleaner/cleaning_rules/drop_participants_without_ppi_or_ehr.py
calbach: I like it - added. Thanks for the pointer.

data_steward/cdr_cleaner/cleaning_rules/drop_participants_without_ppi_or_ehr.py
calbach: Added. Interestingly when I try to execute this against my bq test dataset I see:
```
Traceback (most recent call last):
  File "/usr/local/google/home/calbach/aou/curation/data_steward/cdr_cleaner/clean_cdr_engine.py", line 72, in clean_dataset
    batch=batch)
  File "/usr/local/google/home/calbach/aou/curation/data_steward/bq_utils.py", line 427, in query
    return bq_service.jobs().insert(projectId=app_id, body=job_body).execute(num_retries=retry_count)
  File "/usr/local/google/home/calbach/.virtualenvs/curation/lib/python3.7/site-packages/oauth2client/_helpers.py", line 133, in positional_wrapper
    return wrapped(*args, **kwargs)
  File "/usr/local/google/home/calbach/.virtualenvs/curation/lib/python3.7/site-packages/googleapiclient/http.py", line 842, in execute
    raise HttpError(resp, content, uri=self.uri)
googleapiclient.errors.HttpError: <HttpError 400 when requesting https://bigquery.googleapis.com/bigquery/v2/projects/aou-res-curation-test/jobs?alt=json returned "Required parameter is missing">

```

data_steward/test/unit_test/data_steward/cdr_cleaner/cleaning_rules/drop_participants_without_ppi_or_ehr_test.py
```diff
@@ -0,0 +1,28 @@
...
+    @classmethod
+    def setUpClass(cls):
+        print('**************************************************************')
+        print(cls.__name__)
+        print('**************************************************************')
+
+    def test_get_queries(self):
+        results = drop_participants_without_ppi_or_ehr.get_queries('foo', 'bar')
+
+        self.assertEquals(len(results), 1 + len(common.CLINICAL_DATA_TABLES),
```
calbach: The testing story looks bleak here, I didn't see any clear way of testing my code that wasn't effectively a change-detector test. This may be a larger discussion, but things I might hope to have in a hermetic local test:

1. SQL validation
2. Execution against an in-memory DB (with workarounds where needed)
3. (maybe - if I can't have 1/2) golden file SQL output tests

data_steward/test/unit_test/data_steward/cdr_cleaner/cleaning_rules/drop_participants_without_ppi_or_ehr_test.py
```diff
@@ -0,0 +1,123 @@
...
+
+
+# Participant 1: no data (to be removed)
+# Participant 2: has the basics only
+# Participant 3: has EHR data only
+# Participant 4: has both basics and EHR
+# Participant 5: has only RDR consent (to be removed)
+# Participant 6: has EHR observation only
+INSERT_FAKE_PARTICIPANTS_TMPLS = [
+    # XXX: Ideally these tests should not manipulate concept table, not currently hermetic.
```
calbach: @mark-velez  Do we have a way to ensure that the concept/concept_ancestor tables exist in the environment under testing? Do we have a better way to inject test data (here I used a forced nullable approach to make test data insertion easier)?

data_steward/test/unit_test/data_steward/cdr_cleaner/cleaning_rules/drop_participants_without_ppi_or_ehr_test.py
mark-velez: I agree, this is a larger discussion and also on the desired local test features. If I understand 3 correctly, I think one problem with golden files is they may be sensitive to changes that are not functional. For example, I'm not sure we want solutions confined to be particular about whitespace, casing, or even whether a query uses a CTE vs subqueries (if both can do the job just as well).

data_steward/test/unit_test/data_steward/cdr_cleaner/cleaning_rules/drop_participants_without_ppi_or_ehr_test.py
mark-velez: @calbach we do not have a clear way to ensure the vocabulary tables exist in the test environment. It's not the best solution, but we typically load the vocabulary tables once in all our test datasets (e.g. using `table_copy.sh`). Thereafter we do not touch the vocabulary tables. The function `test_util.delete_all_tables` which is often used in our `tearDown`s does **not** delete the vocabulary tables.

data_steward/cdr_cleaner/cleaning_rules/drop_participants_without_ppi_or_ehr.py
mark-velez: Thanks, I think this is quite a bit more intelligible!

data_steward/test/unit_test/data_steward/cdr_cleaner/cleaning_rules/drop_participants_without_ppi_or_ehr_test.py
calbach: Is there a way to invoke this setup from an integration test, or do the integration tests generally assume that these tables have already been created? (and will this be true for someone who is running Circle for their own PR?).

data_steward/cdr_cleaner/cleaning_rules/drop_participants_without_ppi_or_ehr.py
mark-velez: I think this is related to a missing environment variable (guessing maybe `GOOGLE_CLOUD_PROJECT` or `BIGQUERY_DATASET_ID`)?

calbach: I resolved this by iterating on my test. I believe GOOGLE_CLOUD_PROJECT was missing in this case, but there were subsequent failures due to syntax issues in the SQL which I don't think were reported when I ran this. I think using a standard BigQuery library will help immensely with this kind of issue.

data_steward/test/unit_test/data_steward/cdr_cleaner/cleaning_rules/drop_participants_without_ppi_or_ehr_test.py
mark-velez: The integration tests generally and unfortunately assume that these tables have already been created. One possible solution might be to add a step in an init script which loads the vocabulary tables if they do not already exist. A more hermetic solution though might need to ensure the tables are untampered with (schema is correct, rows are as expected). Another idea might be to create a view for each vocabulary table onto some shared (somehow trusted, maybe immutable?) dataset which contains the actual vocabulary tables.

data_steward/test/unit_test/data_steward/cdr_cleaner/cleaning_rules/drop_participants_without_ppi_or_ehr_test.py
calbach: Yes, golden file testing has many problems. Setting aside whitespace changes (these would hopefully be formatted uniformly by your golden data test / generation), they do not actually do any semantic testing or validation of the SQL. This is only a last resort if 1,2 are too high effort; the reason I even mention the option is that I see a bunch of string-based testing in some of the cleaning rule units tests - these have the same deficiencies as golden file tests but with higher development overhead.

I am interested in exploring whether https://github.com/verilylifesciences/purplequery can work for hermetic fake BQ testing. It was recently created by Verily, and I can't fully vouch for it yet - but if it works as advertised it would be a really nice solution for testing.

calbach:
"too many updates" achilles_test failure, rerunning. @mark-velez let me know if this looks reasonable as-is. The integration test I've added would need to be reworked when transitioning to a common integration test package, but it can be run manually locally and it passes.

PR #349, opened 2019-12-09
Fix the unit test names

ChaoPang:
renamed test_admin_api.py and test_key_rotation.py to follow the unit test naming convention


lrwb-aou:
The string on line 99 of `curation/data_steward/test/unittest/data_steward/admin/key_rotation_test.py` needs to be a byte string, `b'not found'`.  Then it should pass.

PR #351, opened 2019-12-11
[DC-593] Improve robustness of dev setup scripts

calbach:
Fixing several issues I encountered when running these scripts.

See: https://docs.google.com/document/d/16sDcgjfA6NvBbCYLJr3lsA7mtZlf9x6uJZnOQvgBv0o/edit#

- Slightly reduce assumptions about globals
- Fix CIRCLECI check
- Reduce assumptions about working directory for scripts
- Fix stale docs / flags in run_tests.sh
- Fix CircleCI SA key upload generation


lrwb-aou:
The updates to `run_tests.sh` look fine.  They should work as expected.  I'd still be happy scrapping the script all together.

data_steward/ci/upload_cloud_creds.sh
```diff
@@ -46,4 +46,4 @@ echo "----- Created vars:"
 curl https://circleci.com/api/v1.1/project/$VCS_TYPE/$CI_USERNAME/$CI_PROJECT/envvar?circle-token=$CI_TOKEN
 
 echo "----- Decryption command:"
-echo 'echo $GCLOUD_CREDENTIALS | openssl enc -d -aes-256-cbc -base64 -A -k $GCLOUD_CREDENTIALS_KEY'
+echo "echo \$GCLOUD_CREDENTIALS | openssl enc -d -aes-256-cbc -base64 -A -md md5 -k \$GCLOUD_CREDENTIALS_KEY"
```
amcgrenera-vumc: Should the openssl in line 49 match line 32? They are slightly different.

data_steward/test/run_tests.sh
```diff
@@ -8,17 +8,13 @@ subset="all"
```
amcgrenera-vumc: @calbach circleci build is failing, not sure if these edits in this file is the reason? If not since these are all shell scripts (with no unit testing needed) I can go ahead and squash and merge. Just wanted to double check.

data_steward/ci/upload_cloud_creds.sh
lrwb-aou: @amcgrenera-vumc I don't think so.  I believe Line 32 is creating the encrypted SSL key.  Line 49 is echoing the command to decrypt the key.  So they should be different.

data_steward/test/run_tests.sh
calbach: Rerunning now, I think this broke on account of some local integration testing I was doing.

calbach:
Keeps failing on update flakes. I will merge since this seems unrelated to the PR, and the other tests pass fine.

calbach:
Sorry, I thought I had approval - I broke the new process rules already :) I will look at the "require PR approval" bit.

PR #352, opened 2019-12-11
[DC-547] updated the transformation rule for min to beats/min for heart rate

ChaoPang:


ChaoPang:
@mark-velez I was gonna assign this to Noah but realized he was not one of the people who have merging power

PR #353, opened 2019-12-11
[DC-595] participant match datetime parsing failure

lrwb-aou:
A fix for transforming float values whose field types are defined as date, datetime, or timestamp to a string.
Updates unit tests.
Adds a unit test to verify byte string conversions.


amcgrenera-vumc:
@lrwb-aou this all looks good to me, it does look like it blocked bc of a fail in Codacy.

lrwb-aou:
@amcgrenera-vumc  I've updated the constants file to satisfy codacy.

PR #354, opened 2019-12-12
[DC-586] Import RDR rules should support null fields

mark-velez:


data_steward/bq_utils.py
```diff
@@ -739,6 +739,31 @@ def create_dataset(
     return insert_result
 
 
+def get_sql_row_expr(row: dict, fields: list):
+    """
+    Get appropriate SQL expression for a row based on fields spec
```
calbach: nit: could you add a little more here, is this meant to be used in a SELECT statement, or is this for an insertion? opt: an example output format might be useful

data_steward/bq_utils.py
```diff
@@ -739,6 +739,31 @@ def create_dataset(
...
+
+    :param row: csv row with untyped values
```
calbach: CSV row as in, a parsed line from a CSV file? Probably worth making this clearer maybe in the function name e.g. maybe `csv_line_to_sql_row_expr`

data_steward/bq_utils.py
```diff
@@ -761,25 +786,11 @@ def load_table_from_csv(project_id, dataset_id, table_name, csv_path=None, field
...
-        for k, v in mapping_dict.items():
-            for field in fields:
-                if field['name'] == k and field['type'] == 'integer':
-                    converted_dict[k] = int(v)
-                else:
-                    converted_dict[k] = v
-        pair_expr = insert_format_vals.format(**converted_dict)
-        pair_exprs.append(pair_expr)
-    formatted_mapping_list = ', '.join(pair_exprs)
+        pair_expr = get_sql_row_expr(mapping_dict, fields)
```
calbach: opt: maybe a list comprehension here `row_exprs = [get_sql_row_expr(t, fields) for t in table_list]`

data_steward/bq_utils.py
```diff
@@ -739,6 +739,31 @@ def create_dataset(
...
+    for field_name, val in row.items():
+        field = next(filter(lambda f: f['name'] == field_name, fields), None)
+        if field is None:
+            raise InvalidOperationError(f'Unable to marshal {val}: field "{field_name}" was not found')
+        if not val and field['mode'] == 'nullable':
+            val_expr = "NULL"
+        elif field['type'] in ['string', 'date', 'timestamp']:
+            val_expr = f"'{val}'"
+        else:
+            val_expr = f"{val}"
```
calbach: opt: existing issue, but are these input CSV files trusted in some way? What happens if someone uploads a CSV with value "; DROP TABLES" (insert actually meaningful BigQuery injection)

data_steward/bq_utils.py
calbach: What does untyped mean here - are they arbitrary types?

data_steward/bq_utils.py
```diff
@@ -739,6 +739,31 @@ def create_dataset(
...
+    :param fields: bigquery fields spec
+    :return: format string for row object
+    """
+    val_exprs = []
+    # TODO refactor for all other types or use external library
+    for field_name, val in row.items():
+        field = next(filter(lambda f: f['name'] == field_name, fields), None)
+        if field is None:
+            raise InvalidOperationError(f'Unable to marshal {val}: field "{field_name}" was not found')
+        if not val and field['mode'] == 'nullable':
```
calbach: Can `val` be a numeric type here? If so you would be inserting NULL instead of 0 for nullable int fields which may not be desired

Either way, I would add a test case for this

data_steward/bq_utils.py
mark-velez: Not nitpicking- thanks for mentioning this. I will add an example.

data_steward/bq_utils.py
mark-velez: yes- I typically only use list comprehensions when they are as simple as this. will do.

data_steward/bq_utils.py
mark-velez: @calbach this is good to think about. I have examined all usages of this function and they are all associated with our own internal use (e.g. loading lookup tables from resource files)- but I would like to prevent injection issues. Any suggestions? For now I can put a warning notice in the docstring.

calbach: Thanks for checking. The usual way I would address this is by using a parameterized query. That's probably a big change, and I don't know whether that's something you're currently using. i.e. https://cloud.google.com/bigquery/docs/parameterized-queries, use `@param` in the query and then pass an extra parameter map when you execute the query.

data_steward/bq_utils.py
mark-velez: I think `val` shouldn't be numeric here. I will add a test case.

data_steward/bq_utils.py
mark-velez: I agree, parameterized queries makes sense. That _would_ require changes to `bq_utils`- which does not currently support parameterized queries and which we are hoping to obliterate entirely now that `py3` opens us up to better bq libraries; is it ok by you that we leave this as-is and file a ticket to refactor this with parameterized queries?

calbach: Yep, sounds good - thanks.

data_steward/bq_utils.py
mark-velez: Thanks, untyped was the wrong word- values are represented as strings. I made the docstring more explicit.

mark-velez: I have renamed the function and clarified the docstring

data_steward/bq_utils.py
```diff
@@ -739,6 +740,43 @@ def create_dataset(
...
+def csv_line_to_sql_row_expr(row: dict, fields: list):
+    """
+    Translate a dict to a SQL row expression based on a fields spec
+
+    :param row: dict whose values are all strings
+    :param fields: bigquery fields spec with keys {name, type, mode, description}
+    :return: SQL expression for row object
+    :rtype: str
+    :example:
+    >>> fields = [{ 'name': 'int_col',  'type': 'integer', 'mode': 'required', 'description': ''},
```
calbach: Nice thanks.

FYI: you may already be aware of this but there's a python library which can execute your code example snippets as unit tests. https://docs.python.org/3/library/doctest.html. Food for thought.

data_steward/bq_utils.py
```diff
@@ -25,7 +25,8 @@ class InvalidOperationError(RuntimeError):
     """Raised when an invalid Big Query operation attempted during the validation process"""
 
     def __init__(self, msg):
-        super(InvalidOperationError, self).__init__(msg)
+        self.msg = msg
```
calbach: Is this change related? I don't quite follow why it's needed and I don't see it referenced below.

mark-velez: Ah forgot to mention it in my commit message- I put this for the assertRaises check [here](https://github.com/all-of-us/curation/pull/354/commits/c6cf95c3c936e4f1f53797ac9a3b74c6e9ef58be#diff-6cb8929a23157cc412569417bb428efaR457).

PR #355, opened 2019-12-16
[DC-530] Remove PIIBirthInformation_BirthDate

lrwb-aou:
* Suppresses rows based on observation_source_concept_id 1585259.
* Changes row suppression from string value suppression on observation_source_value to concept_id suppression on observation_source_concept_id.
* Configuring the path to deid instead of hard coding.  Possibly change this to a config file later.
* Adding deid requirement for tqdm.  Used by pandas but not installed as a dependency of pandas.


data_steward/deid/aou.py
```diff
@@ -212,7 +213,9 @@ def create_allowed_states_table(input_dataset, credentials):
     """
 
     map_tablename = input_dataset + "._mapping_src_hpos_to_allowed_states"
-    data = pd.read_csv('../deid/config/internal_tables/src_hpos_to_allowed_states.csv')
+    data = pd.read_csv(
+        os.path.join(DEID_PATH, 'config/internal_tables/src_hpos_to_allowed_states.csv')
```
nishanthpp93: This isn't very significant since we use unix based systems for development, but if we were ever to use Windows, the forward slashes would need be removed so something like `os.path.join(DEID_PATH, 'config', 'internal_tables', 'src_hpos_to_allowed_states.csv')` would make it compatible.

data_steward/tools/run_deid.py
```diff
@@ -186,20 +186,21 @@ def main(raw_args=None):
     args = parse_args(raw_args)
     add_console_logging(args.console_log)
     known_tables = get_known_tables(fields_path)
-    configured_tables = get_known_tables('../deid/config/ids/tables')
+    deid_tables_path = os.path.join(DEID_PATH, 'config/ids/tables')
```
nishanthpp93: Same here, `os.path.join(DEID_PATH, 'config', 'ids', 'tables')`

data_steward/tools/run_deid.py
```diff
@@ -186,20 +186,21 @@ def main(raw_args=None):
...
         tablepath = None
         if table in configured_tables:
-            tablepath = '../deid/config/ids/tables/' + table + '.json'
+            tablepath = os.path.join(deid_tables_path, table + '.json')
         else:
             tablepath = table
 
         parameter_list = [
-            '--rules', '../deid/config/ids/config.json',
+            '--rules', os.path.join(DEID_PATH, 'config/ids/config.json'),
```
nishanthpp93: This would be 
`'--rules', os.path.join(deid_ids_path, 'config.json'),`

data_steward/deid/aou.py
lrwb-aou: Nice catch.  Thank you!

lrwb-aou: I force pushed the changes.  So now these comments have an outdated status.

PR #357, opened 2019-12-18
[DC-550] Apply yapf to whole codebase

calbach:
This PR achieves the following:

- Defines a yapf style
- Requires yapf to pass on CircleCI
- Applies the yapf format to the codebase
- Documents the option of installing a yapf git pre-commit hook (this will autoformat if needed)

Sample yapf usage:

```
# lint whole codebase
yapf -drp .

# format whole codebase
yapf -rip .
```


.yapfignore
```diff
@@ -0,0 +1,3 @@
+_pb2.py
+data_steward/analytics
+data_steward/lib
```
mark-velez: If `lib` here refers to vendor libs then maybe we do not need this ignore as this may be GAE py2 specific.

calbach: Hmm, dev requirements still list this:
3d. Change directory to curation/data_steward Run `pip install -r requirements.txt -t lib/`

https://docs.google.com/document/d/16Q_F5ecN1hnKviuPTbXxy914FUGFuTXO1F4zvYbjDmc/edit#

Should that be updated?

Either way, I think people probably still have this directory, so it would make sense to clean up .gitignore and .yapfignore at the same time

calbach:
- Switched to a Google base style, to address @lrwb-aou 's concern around dictionary indention. Please take a look and verify this updated style looks OK.
- Bumped line limit to 100 per @mark-velez and @lrwb-aou 
- Added to dev_requirements.txt
- RE precommit, we can look at adding this later; I will ask at the eng sync. In general I think it's useful, but some folks want to maintain autonomy of their own local dev workflows.

Still TODO:
- Will integrate this into CircleCI to enforce conformance. Will wait for #367.

calbach:
Codacy issues are all existing problems. I can send a separate PR to address all of these before continuing with this PR, if preferred.

.yapfignore
lrwb-aou: The documentation is updated.  I'm fine with leaving this in because it probably does still exist in someone's personal repo.

data_steward/cdr_cleaner/cleaning_rules/populate_route_ids.py
```diff
@@ -18,35 +18,37 @@
...
+    "description": "The dose_form_concept_id of the dose form."
+}, {
+    "type":
+        "integer",
+    "name":
+        "route_concept_id",
+    "mode":
+        "required",
+    "description":
+        "The route_concept_id indicating the typical route used for administering the drug."
```
lrwb-aou: FYI:  this looks much nicer than before.

data_steward/constants/validation/participants/readers.py
```diff
@@ -1,46 +1,41 @@
...
-    'WHERE observation_source_concept_id={field_value} '
-    'ORDER BY person_id'
-)
+PPI_OBSERVATION_VALUES = ('SELECT person_id, observation_source_concept_id, value_as_string '
+                          'FROM `{project}.{dataset}.{table}` '
+                          'WHERE observation_source_concept_id={field_value} '
+                          'ORDER BY person_id')
 
 # Select values from ehr person table
-EHR_PERSON_VALUES = (
```
lrwb-aou: nit:  So, this is a difference in the google formater and the pep-8 formater.  With the google style, lines 12 - 15 
```
EHR_PERSON_VALUES = (
    'SELECT person_id, {field} '
    'FROM `{project}.{dataset}.{table}` '
)
```
become line 10, like below.  
```
EHR_PERSON_VALUES = ('SELECT person_id, {field} ' 'FROM `{project}.{dataset}.{table}` ')
```
I personally prefer to have the query separated on multiple lines because it's a little easier to read "formatted SQL".  And it didn't join the two strings like a dev would have.  

This is the pep-8 version:
```
EHR_PERSON_VALUES = ('SELECT person_id, {field} '
                     'FROM `{project}.{dataset}.{table}` ')
```
This seems to be a rare occurrence, it just caught my eye.

data_steward/curation_logging/curation_gae_handler.py
```diff
@@ -111,11 +112,7 @@ def setup_log_line(record: logging.LogRecord, resource=None, method=None):
         funcname = record.funcName if record.funcName else ''
         file = record.pathname if record.pathname else ''
         lineno = record.lineno if record.lineno else 0
-        source_location = {
-            "file": file,
-            "functionName": funcname,
-            "line": lineno
-        }
```
lrwb-aou: This is another instance where I prefer the pep-8 style to google style.  The pep-8 style left lines 114-118 unchanged.  I feel like line 115 is cluttered and harder on the eyes.

data_steward/curation_logging/curation_gae_handler.py
```diff
@@ -137,10 +134,7 @@ def get_highest_severity_level_from_lines(lines):
     """
     if lines:
 
-        s = sorted(
-            [line['severity'] for line in lines],
-            key=lambda severity: -severity
-        )
+        s = sorted([line['severity'] for line in lines], key=lambda severity: -severity)
```
lrwb-aou: Again, google style is more aggressive with filling up the line length.  pep-8 style transforms lines 140-143 to:
```
s = sorted([line['severity'] for line in lines],
                   key=lambda severity: -severity)
```
I feel like the pep-8 style is more readable.  each argument is on its own line.

lrwb-aou:
At first glance, I was ok with it.  But it feels like the google style is more aggressive about filling up line lengths before continuing to the next line.  I noted some instances of differences in pep-8 vs google style guides.  

It seems like google style forces as much on one line as possible before moving to the next line.  This can make reading some code segments a little more difficult.

Of the choices, I prefer the pep-8 with a small style file requiring dictionary values to be indented.

calbach:
What I'd strongly prefer to do here is just take one of the default styles, add the sole override around line length, and then treat the style rules as immutable (never touching them again). I don't feel strongly about whether that's google, pep-8, or another one of the defaults. This is one of the primary benefits of an autoformatter like this is that you can entirely ignore subjective style issues which are covered by the formatter - and you don't need to even have a discussion about whether you should change the style rules.

To me, the dictionary style rule is too minor to make an exception and it sets a precedent of treating the style spec as "mutable" going forward. Does this sound reasonable? If you agree, do you have a preference on one of the preset styles?

lrwb-aou:
We both disagree with the idea that a style formatter works perfectly out of the box (i.e. we both want to increase the maximum line lengths).  I have fewer objections to the pep-8 style than the google style.  That said, what is the point of making the formatter configurable if you don't use it to make minor tweaks to a style?

calbach:
To be honest, my preference would be to not override anything including line length, but I'm pragmatic and realize that would be a tough sell, and that there are legitimate dev grievances with an 80char line limit.

I don't think it clearly follows that because we **can** configure the style rules, we **should**. In my experience, autoformatters which don't allow for configuration (e.g. gofmt) tend to be more transformative in terms of developer productivity (specifically, avoiding debates around what minor style tweaks should be adopted), and in my opinion this outweighs the subjective loss of readability in certain cases. This can be simulated in yapf by treating the style format as immutable, even though the tool supports style tweaks.

Lets discuss more tomorrow at the eng sync.

data_steward/curation_logging/curation_gae_handler.py
lrwb-aou: This is a by product of extending the line length.  pep8 makes the same changes with an extended line.

data_steward/curation_logging/curation_gae_handler.py
lrwb-aou: This is also a by product of extending the line length.

lrwb-aou:
The things I noticed that I disliked are issues in pep8 and google with an extended line length.  I'd personally want to use the pep8 style just to be in conformance with the python community at large.  The additional readability of libraries in the google style is nice.  

I don't mind an 80 character line limit if everyone else can manage.  I think line length should be what we address at the engineers meeting.  

But I do want to say, pep8 is a style **guide**.  It makes sense to define a style for the team, **possibly** with minor modifications that make sense for the team and project, and move on.  It's not something we'll be revisiting often so why not make sure we get it right to begin with.

calbach:
OK - updated per discussion at the eng sync. This now just applies default base "google" style. I've also added instructions in the dev guide for setting up yapf autoformatting both in PyCharm and via commit hook.

PR #361, opened 2019-12-19
[DC-588] Fix unresolved module_name in remove_operational_pii_fields cleaning rule

ksdkalluri:
[ci skip]

### DIFF
--- a/data_steward/cdr_cleaner/manual_cleaning_rules/remove_operational_pii_fields.py
+++ b/data_steward/cdr_cleaner/manual_cleaning_rules/remove_operational_pii_fields.py
@@ -74,7 +74,7 @@ def load_operational_pii_fields_lookup_table(project_id, sandbox_dataset_id):
                                  fields=None)
 
 
-def parse_intermediary_table_query(dataset_id, project_id, sandbox_dataset_id):
+def get_intermediary_table_query(dataset_id, project_id, sandbox_dataset_id):
     """
     parses the intermediary query used to store records being deleted.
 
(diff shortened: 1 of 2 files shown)


data_steward/test/unit_test/data_steward/cdr_cleaner/manual_cleaning_rules/remove_operational_pii_fields_test.py
```diff
@@ -0,0 +1,78 @@
...
+        q = SELECT_RECORDS.format(project_id=self.project_id,
+                                  dataset_id=self.sandbox_dataset_id,
+                                  table_id=remove_operational_pii_fields.OPERATIONAL_PII_FIELDS_TABLE)
+        r = bq_utils.query(q)
+        actual = bq_utils.response2rows(r)
+
+        for i, _ in enumerate(expected):
+            self.assertCountEqual(expected[i], actual[i])
+
+    def test_parse_intermideary_table_query(self):
```
calbach: The integration test is good, but I would just  remove these two unit tests. Simple internal helper methods generally shouldn't need their own unit tests. In this case the test is really just duplicating the original method, so it isn't really adding value and you'll need to keep it exactly synchronized.

data_steward/cdr_cleaner/manual_cleaning_rules/remove_operational_pii_fields.py
```diff
@@ -61,6 +58,61 @@
...
+                                                           intermediary_table=INTERMEDIARY_TABLE,
+                                                           pii_fields_table=
+                                                           OPERATIONAL_PII_FIELDS_TABLE,
+                                                           sandbox=sandbox_dataset_id
+                                                           )
+
+    return query
+
+
+def parse_delete_query(dataset_id, project_id, sandbox_dataset_id):
```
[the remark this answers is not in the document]
ksdkalluri: I am calling these in unittest, that's the reason I used the public.

data_steward/cdr_cleaner/manual_cleaning_rules/remove_operational_pii_fields.py
```diff
@@ -74,7 +74,7 @@ def load_operational_pii_fields_lookup_table(project_id, sandbox_dataset_id):
                                  fields=None)
 
 
-def parse_intermediary_table_query(dataset_id, project_id, sandbox_dataset_id):
+def get_intermediary_table_query(dataset_id, project_id, sandbox_dataset_id):
```
calbach: Now that these are not called from tests, should they be made private?

ksdkalluri: yeah i can make them private.

PR #362, opened 2019-12-19
[DC-598] Improve RDR import script

calbach:
- Rename flags for clarity
- Include `-x` for script logging, no longer need to echo the exact command (adds risk of divergence)
- Add error checks for existence of the input RDR directory
- Remove requirement for `gcloud config set project xxx`
- Ignore virtualenv dir

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -82,6 +82,7 @@ celerybeat-schedule
 .venv/
 venv/
 ENV/
+curation_venv/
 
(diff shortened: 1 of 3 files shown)


calbach:
All passed except flake due to https://precisionmedicineinitiative.atlassian.net/browse/DC-621 , will ignore as this is a known issue.

PR #363, opened 2019-12-20
Cpt modifiers fix notebook

amcgrenera-vumc:


data_steward/analytics/cdr_ops/ac120_cpt_modifiers.py
```diff
@@ -124,14 +124,53 @@ def cstr(s, color='black'):
...
+# # g) What procedure_concept_ids are associated? (should these records be deleted or updated)
+
+# +
+#list procedure_concept_id's and counts to help determine if these records procedure_concept_id's are valid and if
+#these records should be updated rather than deleted
+
+procedure_concept = '''
+
+SELECT DISTINCT p.procedure_concept_id, c.concept_name, COUNT(*) AS count
+FROM unioned_ehr20191004.procedure_occurrence p
```
mark-velez: Please use a placeholder for dataset names {DATASET}

data_steward/analytics/cdr_ops/ac120_cpt_modifiers.py
```diff
@@ -103,9 +103,9 @@ def cstr(s, color='black'):
 
 modifier_concept_id_null = '''
 
-select distinct modifier_concept_id from {DATASET_ID}.procedure_occurrence
+select distinct modifier_concept_id, COUNT(*) as count from unioned_ehr20191004.procedure_occurrence
```
mark-velez: Please use placeholder `{DATASET_ID}` here.

data_steward/analytics/cdr_ops/ac120_cpt_modifiers.py
```diff
@@ -71,7 +71,7 @@ def cstr(s, color='black'):
...
 
 
-'''.format(DATASET_ID=parameters.DEID_DATASET_ID)
+'''.format(DATASET_ID=parameters.UNION_DATASET_ID)
```
mark-velez: You and I found it helpful to swap the dataset to run queries on. Rather than specifying a particular dataset in each query, it might make it easier to use a variable which is declared in the beginning of the notebook (e.g. `dataset_id = parameters.DEID_DATASET_ID`).

amcgrenera-vumc: yes definitely, will make that adjustment.

amcgrenera-vumc:
> @amcgrenera-vumc this looks good! It helps us identify violating sites and determine if the `procedure_concept_id`s for any bad records are ok. Please address my comments below and then draft a cleaning rule.
> 
> Based on your findings I think the cleaning rule should remove records where:
> 
> 1. `procedure_concept_id` is not a standard concept in the procedure domain
>    AND
> 2. `procedure_source_concept_id` is not in the procedure domain (they ARE allowed to be non-standard).
>    There may be better solutions. Please let me know what you think.

@mark-velez I have the code set up for the cleaning rule. I am still playing around with the query-- when I suppress `procedure_source_concept_id` is not in the procedure domain, it takes out the 315 records that needed to be removed initially from this bug as they are 'Procedure' domains and standard. I am pushing this additional query to the pycharm notebook.

data_steward/analytics/cdr_ops/ac120_cpt_modifiers.py
```diff
@@ -172,7 +172,26 @@ def cstr(s, color='black'):
...
-# # h) How would you fix the issue?
+# # h) What is the domain and concept of the affected concept ids
+
+# +
+#list procedure_concept_id's and counts to help determine if these records procedure_concept_id's are valid and if
+#these records should be updated rather than deleted
+
+domain = '''
+
+SELECT DISTINCT c.domain_id, standard_concept
```
mark-velez: Please let me know when you've drafted a cleaning rule to address records containing cpt modifier codes. While our rule will focus on addressing only the records with cpt modifier codes, it may be that these are just a subset of coding issues that could be discovered using a query like below.
```
SELECT 
 c.domain_id,
 c.concept_class_id,
 c.standard_concept,
 COUNT(1) n
FROM `{DATASET_ID}.procedure_occurrence` p
JOIN {DATASET_ID}.concept c 
 ON p.procedure_source_concept_id = c.concept_id
GROUP BY c.domain_id, c.concept_class_id, c.standard_concept
ORDER BY n DESC, c.standard_concept, c.concept_class_id, c.domain_id
```

data_steward/deid/config/ids/tables/procedure_occurrence.json
```diff
@@ -3,6 +3,10 @@
         {
```
mark-velez: We said we would apply a cleaning rule further upstream. Please revert this change to `procedure_occurrence.json` when you get a chance.

amcgrenera-vumc: removed!

PR #365, opened 2019-12-20
Visualizations notebooks updated

noahgengel:
Minor updates to visualization notebooks. Mostly formatting improvements for output purposes.


ksdkalluri:
does this need a review or should I merge it.

noahgengel:
> does this need a review or should I merge it.

I would double-check that my version of the transformation for heartbeat does not mess up with the solution we addressed (I think it is right given that, in my memory, we no longer have the 1/x transformation). Otherwise, all of the files should be fine as-is.

nishanthpp93:
Rolled up into https://github.com/all-of-us/curation/pull/395

PR #368, opened 2019-12-26
[DC-621] Rate limit exceeded on Achilles test and Achilles heel test

nishanthpp93:
Creates randomized tables so that achilles tests do not hit rate limits.
Also makes the following improvements:
- Initialize table achilles_analysis by loading from csv instead of bq VALUES
- Update queries in `achilles_dml.sql` and `achilles_heel_dml.sql` so that the output fields conform to table fields (name and type)
- Use `select *` instead of bq `insert into`, which still count towards the table load quota, but highlights issues which could hit table schema/metadata update quota. (This is probably the limit we keep hitting, which is 5 updates every 10 seconds)

### DIFF
--- a/data_steward/resources/achilles_dml.sql
+++ b/data_steward/resources/achilles_dml.sql
@@ -1,515 +1,3 @@
---populate the tables with names of analyses
-insert into synpuf_100.achilles_analysis (
-    analysis_id,
-    analysis_name,
-    stratum_1_name,
-    stratum_2_name,
-    stratum_3_name,
-    stratum_4_name
-)
-VALUES
-(0, 'Source name', NULL, NULL, NULL, NULL),
---000. PERSON statistics
-(1, 'Number of persons', NULL, NULL, NULL, NULL),
-(2, 'Number of persons by gender', 'gender_concept_id', NULL, NULL, NULL),
-(3, 'Number of persons by year of birth', 'year_of_birth', NULL, NULL, NULL),
-(4, 'Number of persons by race', 'race_concept_id', NULL, NULL, NULL),
-(5, 'Number of persons by ethnicity', 'ethnicity_concept_id', NULL, NULL, NULL),
-(7, 'Number of persons with invalid provider_id', NULL, NULL, NULL, NULL),
-(8, 'Number of persons with invalid location_id', NULL, NULL, NULL, NULL),
-(9, 'Number of persons with invalid care_site_id', NULL, NULL, NULL, NULL),
-(10, 'Number of all persons by year of birth by gender', 'year_of_birth', 'gender_concept_id', NULL, NULL),
-(11, 'Number of non-deceased persons by year of birth by gender', 'year_of_birth', 'gender_concept_id', NULL, NULL),
-(12, 'Number of persons by race and ethnicity', 'race_concept_id', 'ethnicity_concept_id', NULL, NULL),
---100. OBSERVATION_PERIOD (joined to PERSON)
-(101, 'Number of persons by age, with age at first observation period', 'age', NULL, NULL, NULL),
-(102,
-    'Number of persons by gender by age, with age at first observation period',
-    'gender_concept_id',
-    'age',
-    NULL,
-    NULL
-),
-(103, 'Distribution of age at first observation period', NULL, NULL, NULL, NULL),
-(104, 'Distribution of age at first observation period by gender', 'gender_concept_id', NULL, NULL, NULL),
-(105, 'Length of observation (days) of first observation period', NULL, NULL, NULL, NULL),
-(106, 'Length of observation (days) of first observation period by gender', 'gender_concept_id', NULL, NULL, NULL),
-(107, 'Length of observation (days) of first observation period by age decile', 'age decile', NULL, NULL, NULL),
-(108,
-    'Number of persons by length of observation period, in 30d increments',
-    'Observation period length 30d increments',
-    NULL,
-    NULL,
-    NULL
-),
-(109, 'Number of persons with continuous observation in each year', 'calendar year', NULL, NULL, NULL),
-(110, 'Number of persons with continuous observation in each month', 'calendar month', NULL, NULL, NULL),
-(111, 'Number of persons by observation period start month', 'calendar month', NULL, NULL, NULL),
-(112, 'Number of persons by observation period end month', 'calendar month', NULL, NULL, NULL),
-(113, 'Number of persons by number of observation periods', 'number of observation periods', NULL, NULL, NULL),
-(114, 'Number of persons with observation period before year-of-birth', NULL, NULL, NULL, NULL),
-(115, 'Number of persons with observation period end < observation period start', NULL, NULL, NULL, NULL),
-(116,
-    'Number of persons with at least one day of observation in each year by gender and age decile',
-    'calendar year',
-    'gender_concept_id',
-    'age decile',
-    NULL
-),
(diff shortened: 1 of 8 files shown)


data_steward/resources/achilles_dml.sql
```diff
@@ -2262,9 +1703,9 @@ insert into synpuf_100.achilles_results (analysis_id, stratum_1, stratum_2, coun
 --
 -- 903	Number of distinct drug era concepts per person
 insert into synpuf_100.achilles_results_dist (analysis_id, count_value, min_value, max_value, avg_value, stdev_value, median_value, p10_value, p25_value, p75_value, p90_value)
-WITH rawdata as ( select  COUNT(distinct de1.drug_concept_id)  as count_value   from  synpuf_100.drug_era de1
+WITH rawdata as ( select  cast(COUNT(distinct de1.drug_concept_id) as float64) as count_value   from  synpuf_100.drug_era de1
 	 group by  de1.person_id
- ), overallstats  as (select  cast(avg(1.0 * count_value)  as float64)  as avg_value, cast(STDDEV(count_value)  as float64)  as stdev_value, min(count_value)  as min_value, max(count_value)  as max_value, COUNT(*)  as total  from  rawdata
```
calbach: This seems more significant than some of the other changes, is this because `avg_value` is an integer in the schema?

data_steward/validation/achilles.py
```diff
@@ -1,86 +1,143 @@
...
+        # ignoring comments, all commands must start with 'insert into', followed by the table_id
+        if line.strip().lower().startswith(INSERT_INTO):
+            words = line.strip().split()
+            table_id = words[2]
+    if not table_id:
+        raise RuntimeError('Table does not exist for insert query %s' % command)
+    return table_id
+
+
+def convert_insert_to_append(command):
```
calbach: What is the advantage in converting from an insert -> select? And is the insert form still being used elsewhere by other code? The dependency here on the exact formatting of the insertion SQL is a bit brittle and hard to validate, so I would lean against using this approach unless there is a strong motivation for it

data_steward/test/unit_test/achilles_test.py
```diff
@@ -90,7 +92,7 @@ def test_load_analyses(self):
 
     def test_get_run_analysis_commands(self):
         cmd_iter = achilles._get_run_analysis_commands(FAKE_HPO_ID)
-        commands = list(cmd_iter)
+        commands = [achilles.convert_insert_to_append(command)[0] for command in cmd_iter]
```
calbach: grabbing `[0]` seems unnecessary to me given the test you've written, am I missing something?

data_steward/test/unit_test/achilles_heel_test.py
```diff
@@ -27,81 +30,105 @@ def setUpClass(cls):
...
-        cmd = sql_wrangle.qualify_tables(
-            'SELECT COUNT(1) FROM %sachilles_heel_results' % sql_wrangle.PREFIX_PLACEHOLDER, FAKE_HPO_ID)
-        result = bq_utils.query(cmd)
-        self.assertEqual(ACHILLES_HEEL_RESULTS_COUNT, int(result['rows'][0]['f'][0]['v']))
-        cmd = sql_wrangle.qualify_tables(
-            'SELECT COUNT(1) FROM %sachilles_results_derived' % sql_wrangle.PREFIX_PLACEHOLDER, FAKE_HPO_ID)
-        result = bq_utils.query(cmd)
-        self.assertEqual(ACHILLES_RESULTS_DERIVED_COUNT, int(result['rows'][0]['f'][0]['v']))
+
+        # create randomized tables to bypass BQ rate limits
```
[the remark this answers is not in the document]
nishanthpp93: There is no TTL but they are deleted during the `tearDown()` stage of the test.

data_steward/test/unit_test/achilles_test.py
nishanthpp93: You are correct, I need to remove it. I had initially set up `achilles.convert_insert_to_append()` to return two outputs, one for the `select` command and the other for the `table_id`, but later split them into two separate functions.

data_steward/validation/achilles.py
nishanthpp93: I agree, this is very fragile and can break if we change the formatting. I don't think we are using the insert form elsewhere.
BigQuery appears to be lenient with running insert statements with regards to column name mismatches and column type mismatches. I believe we were hitting the `Maximum rate of table metadata update operations` due to this. However, BQ is more stringent with `select/write append` statements, and this allowed me to make changes in the `achilles_dml.sql` and `achilles_heel_dml.sql` to cast to the correct type and rename columns to the correct name, after looking at the failed queries.
Since this process is now complete, I could revert this by removing the functions `convert_insert_to_append` and `extract_table_id_from_query`. Let me know if you believe it's better to do so.

data_steward/resources/achilles_dml.sql
nishanthpp93: `avg_value` would be the same as `count_value`, and we wanted it to be a float since that was the column type. However, `count_value` was an integer previously, requiring the cast to float for `avg_value`. It also required casting for `stdev_value`, `min_value`, `max_value` (they should've been casted too, but BQ insert is lenient in this regard). I cast `count_value` as a float so we don't have to cast for the derivatives.

nishanthpp93:
I also wanted to note that even though we avoid some of the rate limits in this pull request, we might hit at least one limit, which can cause the test to fail at times when multiple devs are running tests at the same time.
@mark-velez and I believe it could be the `Concurrent rate limit for interactive queries — 10 concurrent queries` (though it is listed as 100 in https://cloud.google.com/bigquery/quotas, it is set to 10 internally in the project API limit settings) or the `Maximum requests per second per project: 10` limit.
This makes at least one query fail due to rate limits, but it is reset almost immediately so subsequent queries run. However, this can result in a failed test.

data_steward/test/unit_test/achilles_heel_test.py
```diff
@@ -27,81 +30,105 @@ def setUpClass(cls):
 
     def setUp(self):
         self.hpo_bucket = gcs_utils.get_hpo_bucket(test_util.FAKE_HPO_ID)
+        self.dataset = bq_utils.get_dataset_id()
         test_util.empty_bucket(self.hpo_bucket)
-        test_util.delete_all_tables(bq_utils.get_dataset_id())
+        test_util.delete_all_tables(self.dataset)
```
calbach: nit: same change of `self.dataset` can be used in tearDown @ L147

data_steward/test/unit_test/achilles_heel_test.py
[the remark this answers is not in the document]
calbach: Got it, thanks. I existing issue, but this does introduce some problems when two of these tests run simultaneously since all tests will delete all tables. Something also worth fixing in in the future.

data_steward/validation/achilles.py
calbach: Yes - in my opinion I think reverting to the insert would be better, thank you!

data_steward/test/unit_test/achilles_heel_test.py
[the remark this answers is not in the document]
nishanthpp93: Agree this is an existing issue. Currently, we encounter this when two commits are made to the same branch within a short while, resulting in two circle tests on the same branch.

PR #369, opened 2019-12-26
[DC-616] Exclude Physical Measurements from height and weight cleaning rule

nishanthpp93:
Use `src_id` from `measurement_ext` to exclude RDR data from cleaning rule


nishanthpp93:
Thank you @amcgrenera-vumc!
- Circle looks like it is failing because of rate limits. I have another pull request to resolve it. Other tests seem to pass.
- That is correct. I was initially incorporating it into the query, however, @mark-velez brought up that some of the sites might use that concept by mistake and we don't want to exclude such sites. I will add this to the Jira ticket.

amcgrenera-vumc:
@nishanthpp93 thanks for the clarification!

PR #376, opened 2020-01-06
[DC-618] Add new custom concept to vocabulary for race non-responses

mark-velez:


data_steward/test/unit_test/vocabulary_test.py
```diff
@@ -8,9 +8,9 @@
 
 from test.unit_test import test_util
 from common import DELIMITER, LINE_TERMINATOR
-from resources import AOU_GENERAL_CONCEPT_CSV_PATH
+from resources import AOU_VOCAB_CONCEPT_CSV_PATH
```
lrwb-aou: If you implement the possible suggestions, this test case will have to be updated.

data_steward/common.py
```diff
@@ -138,11 +138,15 @@
 TRANSFORM_FILES = 'transform_files'
 APPEND_VOCABULARY = 'append_vocabulary'
 APPEND_CONCEPTS = 'append_concepts'
-ADD_AOU_GENERAL = 'add_aou_general'
+ADD_AOU_VOCABS = 'add_aou_vocabs'
 ERRORS = 'errors'
 AOU_GEN_ID = 'AoU_General'
 AOU_GEN_NAME = 'AoU_General'
```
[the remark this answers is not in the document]
mark-velez: @lrwb-aou thank you for being so thoughtful on this! I love how the suggested `VOCABULARY_UPDATES` dict reduces code line count. I will implement this.

The only part I'm unsure about is `VERSION_TEMPLATE`. I think we agree that it is an identifier that should change whenever there is any change to the custom vocabulary concepts. Ideally I think we shouldn't rely on our remembering to change this value but I'm also not happy about it being dynamically evaluated as it is now. I will ping you on Slack to talk through this.

data_steward/vocabulary.py
```diff
@@ -89,30 +94,61 @@ def transform_files(in_dir, out_dir):
...
+
+def _vocab_id_match(s):
+    """
+    Get a matching AOU vocabulary ID in the specified string, if any
+
+    :param s: string to search for AOU vocabulary IDs
+    :return: the first vocabulary ID found in the string, otherwise None
+    """
+    vocab_id_in_row_iter = (vocab_id for vocab_id in VOCABULARY_UPDATES if vocab_id in s)
+    return next(vocab_id_in_row_iter, None)
```
lrwb-aou: This is interesting.  I haven't used `next()` before.  I had to run it to see the difference in returning the iterator value vs. the list.  A comment might make what is happening clearer for new developers.

data_steward/tools/load_vocab.sh
```diff
@@ -1,5 +1,5 @@
 #!/usr/bin/env bash
-
+set -ex
```
lrwb-aou: I like `set -x`, but be sure to end the script with `set +x` or the command line will keep echoing everything even after the script has finished.  

What is your user experience with `set -e`?  I really dislike it since I work from the command line.  When it encounters an error it closes my working terminal and I can't debug.  If you leave it in, I would suggest turning it off at the end of the script also `set +e`.

data_steward/tools/load_vocab.sh
```diff
@@ -12,12 +12,14 @@ tools/load_vocab.sh --app_id app_id --in_dir in_dir [--gcs_path gcs_path --datas
  --in_dir   in_dir   directory where vocabulary files are located
  --gcs_path gcs_path full GCS path to save transformed files
  --dataset  dataset  name of BigQuery dataset to create and load vocabulary
+ [--venv_dir venv_dir directory where virtual environment should be created]
```
lrwb-aou: Is this a major difference between command line development and PyCharm development?  From my perspective, it's redundant.

If you create a virtual environment, consider cleaning it up before leaving the script with a good 
```
deactivate
rm -rf ${VENV_DIR}
```

mark-velez: Does it seem redundant because the script can assume a virtual environment will be activated prior to running it? Or because the location of the virtual environment should be assumed? I noticed most of our scripts assume a virtual env in `${CURATION_HOME}/dataset_steward/curation_venv`- should I run on this assumption?

lrwb-aou: It's redundant (for me) from a command line perspective because I always have my virtual environment activated.  That's why I was asking if this is a development environment issue (because I think I'm the only one using command line and noticing this).  If I were going to leave this in here, I wouldn't have it build the virtual environment inside our `data_steward` directory.  This is our package and product.  It's not where I would install third party libraries just because someone might forget to clean it up.

data_steward/tools/load_vocab.sh
```diff
@@ -12,12 +12,14 @@ tools/load_vocab.sh --app_id app_id --in_dir in_dir [--gcs_path gcs_path --datas
...
 "
 SCRIPT_PATH="$(
```
lrwb-aou: Since we're in this script, now is an opportune time to implement CH's previous suggestion and fix our `$BASE_DIR` and `$SCRIPT_PATH` setters so they are not reliant on which directory you are physically located in to execute.

```
BASE_DIR="$(git rev-parse --show-toplevel)"/data_steward
SCRIPT_PATH=${BASE_DIR}/tools
```

Then this should work whether you start if from `curation` or `curation/data_steward/tools`.

data_steward/tools/load_vocab.sh
mark-velez: Thanks, I didn't realize these shell options persisted. I will end the script with `set +x`.

I just used `set -e` because I saw it in many of our scripts. I dislike it because it seems to me to abandon ship even when commands just issue a warning. For example, in this script if `cp` warns it's skipping copying a directory or `bq mk` warns of an existing dataset- I think we'd want to proceed. I will remove that option altogether.

lrwb-aou: I didn't know they persisted either until I started tinkering with them in the `run_tests.sh` script.  Woohoo!  We both learned something new!

lrwb-aou:
I'm no bash script hero, but it looks good with a few minor modifications:

1. Unset the variables before exiting.
2. Deactivate and remove the virtual environment.
3. Please change how `BASE_DIR` and `SCRIPT_PATH` are set so the script is location agnostic at execution.
4. nit:  A brief comment describing `next()` would be nice.

data_steward/tools/load_vocab.sh
mark-velez: I agree- I've had issues in the past when the virtual env was in `data_steward`. Like you I run these from command line myself but yesterday I'd forgotten to activate my environment first. The terminal alerted me there was a problem only after this script had run for several minutes; so I put the virtual env stuff in to prevent wasted time. Perhaps this is a good topic for team discussion. For now I can do the clean up you suggested if that's ok with you.

lrwb-aou: Yes, please do the clean up.

data_steward/tools/load_vocab.sh
```diff
@@ -12,12 +12,11 @@ tools/load_vocab.sh --app_id app_id --in_dir in_dir [--gcs_path gcs_path --datas
...
  --dataset  dataset  name of BigQuery dataset to create and load vocabulary
+ [--venv_dir venv_dir directory where virtual environment should be created]
 "
-SCRIPT_PATH="$(
-  cd "$(dirname "$0")"
-  pwd -P
-)"
-BASE_DIR="$(cd ${SCRIPT_PATH} && cd .. && pwd)"
+BASE_DIR="$(git rev-parse --show-toplevel)"/data_steward
+TOOLS_PATH=${BASE_DIR}/tools
```
lrwb-aou: FYI:  Love changing the name from `SCRIPT_PATH` to `TOOLS_PATH`

PR #377, opened 2020-01-07
[DC-639] Update retraction, fix bugs

nishanthpp93:
* Use try-except for non-numeric column values
* Add docstring for module
* Add notes for non-existent env variables


data_steward/bq_utils.py
```diff
@@ -79,10 +79,13 @@ def get_retraction_sandbox_dataset_id():
 
 def get_retraction_dataset_ids():
     # dataset ids from which to retract, separated by spaces ' '
+    # if retraction needs to be performed on all datasets in the project, remove this env variable
```
lrwb-aou: Since you're updating comments for both `get_retraction_dataset_ids()` and `get_retraction_submission_folder()`, would you mind using the Pep 8 doc string format?
```
one line summary
blank line
more descriptive summary or paragraph
blank line
parameter definitions, if any
expected return values
```

data_steward/tools/retract_data_gcs.py
```diff
@@ -169,7 +193,6 @@ def get_response():
 def get_integer(num_str):
```
lrwb-aou: I looked at `get_integer(num_st)` and I'm curious.  What does placing this as a function offer over an in line `int(num_str)` cast?

data_steward/tools/retract_data_gcs.py
```diff
@@ -124,25 +135,38 @@ def retract(pids, bucket, found_files, folder_prefix, force_flag):
...
-                if input_line != '':
-                    cols = input_line.split(",")
-                    col_1 = cols[0]
-                    col_2 = cols[1]
-                    if (table_name in PID_IN_COL1 and get_integer(col_1) in pids) or \
-                            (table_name in PID_IN_COL2 and get_integer(col_2) in pids):
-                        lines_removed += 1
+                # ensure line is not empty
+                if input_line.strip() != b'':
+                    input_line = input_line.strip()
```
lrwb-aou: Why do the strip() twice?  Suggestion:
```
input_line = input_line.strip()
# ensure line is not empty
if input_line != b'':
    cols = input_lint.split(b',')
```

data_steward/tools/retract_data_gcs.py
```diff
@@ -124,25 +135,38 @@ def retract(pids, bucket, found_files, folder_prefix, force_flag):
...
+                    cols = input_line.split(b',')
+                    # ensure at least two columns exist
+                    if len(cols) > 1:
+                        col_1 = cols[0]
+                        col_2 = cols[1]
+                        # skip if non-integer is encountered and keep the line as is
+                        try:
+                            if (table_name in PID_IN_COL1 and get_integer(col_1) in pids) or \
+                                    (table_name in PID_IN_COL2 and get_integer(col_2) in pids):
+                                lines_removed += 1
```
lrwb-aou: nit:  add a comment here saying why the line isn't written back but the removed count is increased.

data_steward/validation/main.py
```diff
@@ -730,15 +730,20 @@ def run_retraction_cron():
 
     # retract from bq
     dataset_ids = bq_utils.get_retraction_dataset_ids()
-    dataset_ids = dataset_ids.split() if dataset_ids else None
+    # If the env variable RETRACTION_DATASET_IDS does not exist, dataset_ids will be set to None
+    # This allows retraction to be performed on all datasets existing in the project
+    if dataset_ids is not None:
```
lrwb-aou: nit:
```
if dataset_ids:
    dataset_ids = dataset_ids.split()
```

data_steward/validation/main.py
```diff
@@ -730,15 +730,20 @@ def run_retraction_cron():
...
     retract_data_bq.run_retraction(project_id, sandbox_dataset_id, pid_table_id, hpo_id, dataset_ids)
     logging.info('Completed retraction on BQ datasets')
 
     # retract from gcs
     folder = bq_utils.get_retraction_submission_folder()
-    if not folder:
-        folder = None
+    # If the env variable RETRACTION_SUBMISSION_FOLDER does not exist, folder will be set to None
+    # This allows retraction to be performed on all folders submitted by the site that exist in the archive
+    logging.info('Submission folder from env variable: %s of type %s' % (folder, type(folder)))
```
lrwb-aou: nit:  This is an odd log message.  Literally comes out to either:

```
"Submission folder from env variable: alsdkfjaldskfj of type <class 'str'>"
"Submission folder from env variable: None of type <class 'NoneType'>"
```
You could drop the type and you would still have the useful part of the message.

lrwb-aou:
Looking through these test results, /test_results/all/html/_home_circleci_curation_data_steward_tools_retract_data_bq_py.html, it looks like `int_list_to_bq(l)` is dead code.

 nit:  /test_results/all/html/_home_circleci_curation_data_steward_tools_retract_data_bq_gcs.html, seems like we could unit test `get_response()` pretty easily.

My biggest concern is having to redeploy a new app_prod.yaml whenever we want to retract some data.  At that point, the code isn't changing, the config isn't changing, how do we:

1. Differentiate why we're redeploying?
2. Prevent someone from copying the app.yaml file, leaving in retraction environment variables, and redeploying with extraneous environment variables?

This solution requires a lot of technical working knowledge of GAE, our pipeline, and their interaction.

data_steward/bq_utils.py
nishanthpp93: Sure! It was meant to be more like a note rather than a docstring, but I guess its better formatted as a docstring.

data_steward/tools/retract_data_gcs.py
nishanthpp93: It was performing other checks earlier, but we cut it down. I guess it's retained in case we want the functionality back.

data_steward/tools/retract_data_gcs.py
nishanthpp93: I did this so that the original line is intact when writing back to the file if it is not retracted.

data_steward/tools/retract_data_gcs.py
nishanthpp93: Sure, I can add something like `do not write line back since it contains pid to retract`

data_steward/validation/main.py
lrwb-aou: I was finally able to reproduce this behavior in an interpreter with the command line.  Maybe, we should redesign how you get your environment variables.  It seems that:
```
# in bash
$ export FAKE_VAR=

# in GAE app.yaml
FAKE_VAR:
``` 
produces an environment variable that is an empty string.  So it is technically not `None`.

One solution is to change your environment variable getters, (this may be something we need to investigate further for all environment variable getters).  Something like this would work:
```
def _get_env_var(var_name):
    """
    Get either a non-empty string environment variable or None.

    :param var_name:  name of the environment variable to look up
    :returns:  either a non-empty string or None
    """
    env_var = os.environ.get(var_name)
    env_var = env_var.strip()
    return env_var if env_var else None

def get_retraction_submission_folder():
    # submission folder from which to retract
    # if retraction needs to be performed on all submissions by a site, remove this env variable
    return _get_env_var('RETRACTION_SUBMISSION_FOLDER')
```

data_steward/validation/main.py
lrwb-aou: Then this is a logic flaw.  If `dataset_ids` can be any type other than a string and get through this condition, the code will fail on the next line.  `split()` is a string attribute.  If you try to use it on a boolean, integer, or some other type you will get an `AttributeError`.  So you probably want to check:
```
if isinstance(dataset_ids, str):
    dataset_ids = dataset_ids.split()
```

data_steward/tools/retract_data_gcs.py
lrwb-aou: That would be great!  I know someone reading the code could eventually get there, but an explicit comment is so helpful.

data_steward/tools/retract_data_gcs.py
lrwb-aou: I don't think writing back the same amount of white space is necessary on blank lines because line 151 makes it appear that we are expecting comma delimited fields.  But, if you don't like the above suggestion, you could also try:
```
if not input_line.isspace():
    input_line = input_line.strip()
    ...
```
This conditional ensures the `input_line` variable is never changed (even accidentally) and all the blank space can be rewritten as in the original.

nishanthpp93: Yes, that works! Thanks

data_steward/validation/main.py
nishanthpp93: Yes, locally we get an empty string, however in GAE we get the string 'None' for the same conditions rather than an empty string. One solution to this was to remove the environment variable from app.yaml. This returns None as expected both locally and in GAE.

lrwb-aou: I added an additional check in the helper function.  Now, it should return `None` if the environment variable exists but is all spaces or if it is missing.  I feel like encapsulating this GAE specific logic in one location and setting an intuitive value is better than having this GAE logic spread out over multiple functions.

lrwb-aou:
> Do you think it makes sense to consider the environment variable changes as a config change? (i.e. app.yaml considered part of config)

Yes, for now I would consider this a configuration change.  

> I agree it is not safe. Until we can refactor the implementation, we could enforce a rule that a new app.yaml files need to be generated from the script rather than copying over.

Yes, an argument for refactoring (along with the fact that participant match shouldn't be a cron job either).  This is not something you need to address in this PR, but do you know if the script for generating an app.yaml is current?

data_steward/validation/main.py
nishanthpp93: I agree in centralizing it. However, we can do away with this check completely and add in our playbook a note saying 'no empty env variables allowed, remove if necessary'. I was exploring empty vars earlier but I believe the standard way to get `None` via os.environ.get` is to remove it completely. Also, the 'None' string that we get from GAE might be fixed/changed later on since it is unexpected/unexplained behavior, so I don't think we should rely on it.

lrwb-aou: I'm hesitant about that because:  

1. This has caused you issues, costing debugging and development time.  
2.  It's nice to have a line in a developer guide, but I wouldn't rely on people to always read that every time they are going to deploy or develop new code.  
3.  This affects not just your retraction scripts, but any environment variable in app.yaml.
4.  This is not dependent on the environment variable returning `'None'`.  If the variable is missing or an empty string, it also returns `None`.
5.  This is a GAE specific quirk.  On any other system `os.environ.get` will default to `None`, not the string None.

data_steward/tools/retract_data_gcs.py
nishanthpp93: Actually, I'll remove this function since it doesn't serve any additional purpose for now

data_steward/validation/main.py
nishanthpp93: Thanks for the discussion, @lrwb-aou. I have used `all_datasets` and `all_folders` as the default values.

tests/unit_tests/data_steward/tools/retract_data_gcs_test.py
```diff
@@ -1,7 +1,5 @@
 import unittest
 
-from tools import retract_data_gcs as rd
```
lrwb-aou: If you're only testing built-in type conversions, you don't need this.  I think you can remove this entire file.

nishanthpp93: I agree, it's not required.

PR #378, opened 2020-01-08
[DC-624] cleaning rule to remove extreme physical measurements

amcgrenera-vumc:


data_steward/cdr_cleaner/cleaning_rules/drop_extreme_measurements.py
```diff
@@ -0,0 +1,115 @@
...
+# Project imports
+import constants.cdr_cleaner.clean_cdr as cdr_consts
+
+DELETE_HEIGHT_ROWS_QUERY = """
+--drop all height recrods out of bounds
+DELETE
+FROM `{project_id}.{dataset_id}.measurement`
+WHERE measurement_source_concept_id = 903133
+AND value_as_number NOT BETWEEN 90 AND 228
+UNION ALL
```
nishanthpp93: I like this because I would've preferred this way too. However, BigQuery refuses to allow this. I typically check if a query is valid by pasting it into the BQ query editor and populating it with test datasets without actually running it. If I need to see the output of a query, I typically replace `delete/update/insert` statements with `select` and only run it after making sure they are all replaced.
There are other ways to achieve this though. `measurement_id` is a good marker since it is unique to each row, and  there can be a subquery that picks and unions the bad measurement_ids

data_steward/cdr_cleaner/cleaning_rules/drop_extreme_measurements.py
```diff
@@ -0,0 +1,115 @@
...
+--drop BMI row associated with PID
+DELETE
+FROM `{project_id}.{dataset_id}.measurement`
+WHERE person_id IN (SELECT person_id
```
nishanthpp93: I saw that the ticket requests related rows to be dropped (for bad BMI, it says drop the height and weight rows for that pid). However, it does not clarify a few things:
- If one BMI record is bad, should all BMI records for that pid be dropped?
- If one BMI record is bad, should all height and weight rows for that pid be dropped or just one each? Also how to find the height and weight rows used to calculate the BMI row? Are they all related via timestamps?

nishanthpp93: After receiving clarification, it looks like the only link between the bad records in a particular measure (say height) and the corresponding records in the other measures (BMI/weight) is the timestamp.
I believe there are many ways to tackle this. One way I can think of would be to have a subquery which only retrieves the person_id, timestamp and the extreme value for a particular measure (say height), and the main query choosing measurement_ids for the same person_ids and same timestamps but for all measures (height/BMI/weight), and the delete statement deleting these measurement_ids. May not be the most practical way to do it so feel free to solve it your way.

amcgrenera-vumc:
@nishanthpp93 I have updated the queries. Please take a look when you get the chance!

PR #385, opened 2020-01-22
[DC-555] Updates pipeline scripts to use new dataset naming convention.

ksdkalluri:


data_steward/tools/create_ehr_and_rdr_snapshot.sh
```diff
@@ -26,18 +36,18 @@ while true; do
...
-if [[ -z "${key_file}" ]] || [[ -z "${ehr_dataset}" ]]; then
-  echo "Specify the key file location and ehr_dataset ID. $USAGE"
+if [[ -z "${key_file}" ]] || [[ -z "${ehr_dataset}" ]] || [[ -z "${rdr_dataset}" ]] || [[ -z "${dataset_release_tag}" ]]; then
+  echo "Specify the key file location and ehr_dataset ID, rdr_dataset ID and Dataset release tag . $USAGE"
   exit 1
 fi
 
-today=$(date '+%Y%m%d')
-app_id=$(python -c 'import json,sys;obj=json.load(sys.stdin);print(obj["project_id"]);' < "${key_file}")
+app_id=$(python -c 'import json,sys;obj=json.load(sys.stdin);print(obj["project_id"]);' <"${key_file}")
```
lrwb-aou: Did you intentionally remove the space between `<"`?

data_steward/tools/generate_combined_dataset.sh
```diff
@@ -1,6 +1,5 @@
 #!/usr/bin/env bash
 set -ex
```
lrwb-aou: nit:  consider adding `set +ex` at the end of the script.  As is, once set the command line will continue to echo until this is turned off.  

opt:  Do you need `set -e`?  How do you debug when the script stops?  It closes my terminal window and I can't debug when `set -e` is set.

data_steward/tools/deid_runner.sh
```diff
@@ -50,16 +55,12 @@ gcloud auth activate-service-account --key-file="${key_file}"
...
 ROOT_DIR=$(git rev-parse --show-toplevel)
 DATA_STEWARD_DIR="${ROOT_DIR}/data_steward"
 TOOLS_DIR="${DATA_STEWARD_DIR}/tools"
-DEID_DIR="${DATA_STEWARD_DIR}/deid" 
+DEID_DIR="${DATA_STEWARD_DIR}/deid"
 
+#------Create de-id virtual environment----------
 virtualenv -p "$(which python3.7)" "${DATA_STEWARD_DIR}/curation_venv"
 
 source "${DATA_STEWARD_DIR}/curation_venv/bin/activate"
```
lrwb-aou: This skips the part about deid making a dataset and filling it with the data.  If I remember correctly, it's done programmatically via python.  That may have to be an extra jira issue.

data_steward/tools/import_rdr_omop.sh
```diff
@@ -8,7 +8,6 @@ set -ex
 USAGE="tools/import_rdr_omop.sh
     --rdr_project <PROJECT where rdr files are dropped>
     --rdr_directory <DIRECTORY, not including the gs:// bucket name>
-    --output_dataset <DATA SET>
```
lrwb-aou: It seems like this should have become `dataset_release_tag`.  Lines 61-63 are still using the old convention of the day the import was run for naming the dataset instead of the `{YYYY}q{X1}r{X2}` format in the naming convention document.

data_steward/tools/import_rdr_omop.sh
```diff
@@ -53,14 +51,16 @@ DATA_STEWARD_DIR="${ROOT_DIR}/data_steward"
 TOOLS_DIR="${DATA_STEWARD_DIR}/tools"
 CLEANER_DIR="${DATA_STEWARD_DIR}/cdr_cleaner"
 
-app_id=$(python -c 'import json,sys;obj=json.load(sys.stdin);print(obj["project_id"]);' < "${key_file}")
+app_id=$(python -c 'import json,sys;obj=json.load(sys.stdin);print(obj["project_id"]);' <"${KEY_FILE}")
```
lrwb-aou: Why remove the space?

data_steward/tools/import_rdr_omop.sh
```diff
@@ -95,20 +95,28 @@ for file in $cdm_files; do
...
 
 #set BIGQUERY_DATASET_ID variable to dataset name where the vocabulary exists
 export BIGQUERY_DATASET_ID="${VOCAB_DATASET}"
-export RDR_DATASET_ID="${OUTPUT_DATASET}"
+export RDR_DATASET_ID="${RDR_STAGING}"
 echo "Cleaning the RDR data"
-python "${CLEANER_DIR}/clean_cdr.py" -d rdr -s
+data_stage="rdr"
+
+python "${CLEANER_DIR}/clean_cdr.py" --data_stag ${data_stage} -s
```
lrwb-aou: Should be `--data_stage ${data_stage}`

data_steward/tools/run_achilles_report.sh
```diff
@@ -1,15 +1,12 @@
 #!/usr/bin/env bash
 set -ex
```
lrwb-aou: nit:  consider adding `set +ex` to the end of this script.

data_steward/tools/run_curation_pipeline.sh
```diff
@@ -52,68 +52,82 @@ while true; do
...
 
 #-------------------------------------------------------
 # Step 6 Run achilles on de-identified dataset
-echo "-------------------------->Run achilles on identified CDR"
-echo "cdr_deid --> ${cdr_deid}"
-"${TOOLS_DIR}/run_achilles_report.sh" --dataset ${cdr_deid} --key_file ${key_file} --result_bucket ${result_bucket}
+echo "-------------------------->Run achilles on de-identified CDR"
+cdr_deid_base="${cdr_deid}_base"
+cdr_deid_clean="${cdr_deid}_clean"
+export BUCKET_NAME_NYC="test-bucket"
```
lrwb-aou: Is this line necessary?

data_steward/tools/create_ehr_and_rdr_snapshot.sh
ksdkalluri: it is wrong formatting i'll update it.

data_steward/tools/generate_combined_dataset.sh
ksdkalluri: sure added the line.

data_steward/tools/deid_runner.sh
ksdkalluri: It is still being used in the script after setting up the environment.

data_steward/tools/import_rdr_omop.sh
ksdkalluri: this script will be used to load the monthly RDR dumps to BigQuery. we snapshot this dataset and apply the cleaning rules while we start running the pipeline using `create_ehr_and_rdr_snapshot.sh` the first step of running the cdr pipeline

data_steward/tools/import_rdr_omop.sh
ksdkalluri: updated

data_steward/tools/run_curation_pipeline.sh
ksdkalluri: yes, it is a required environment variable. I never got to look deep into why it is required but will definitely look into it.

data_steward/tools/deid_runner.sh
lrwb-aou: I had to look back over the diagram and the deid code.  I was worried about it, but now I don't think it's an issue.

data_steward/tools/generate_combined_dataset.sh
lrwb-aou: Thank you!  It makes running this from the command line a lot nicer.

data_steward/tools/run_curation_pipeline.sh
lrwb-aou: Ok.  We should look into this, but that investigation shouldn't be part of this PR.  Leave it as is to make sure this executes correctly.

lrwb-aou: I created a Jira issue in the backlog for this.  DC-651.

data_steward/tools/run_curation_pipeline.sh
```diff
@@ -52,68 +52,82 @@ while true; do
...
+echo "dataset_release_tag --> ${dataset_release_tag}"
 
 #---------------------------------------------------------
-# Step 1 create EHR snapshot
+# Step 1 create EHR and RDR snapshot
 echo "-------------------------->Take a Snapshot of EHR Dataset (step 1)"
-ehr_snap_dataset="ehr${today}"
-echo "ehr_snap_dataset --> $ehr_snap_dataset"
-"${TOOLS_DIR}/create_ehr_snapshot.sh" --key_file ${key_file} --ehr_dataset ${ehr_dataset}
+"${TOOLS_DIR}/create_ehr_and_rdr_snapshot.sh" --key_file ${key_file} --ehr_dataset ${ehr_dataset} --rdr_dataset ${rdr_datset} --dataset_release_tag ${dataset_release_tag}
```
lrwb-aou: Should be `--rdr_dataset ${rdr_dataset}`.

data_steward/tools/run_curation_pipeline.sh
```diff
@@ -52,68 +52,82 @@ while true; do
...
 
 #---------------------------------------------------------
 # Step 2 Generate Unioned ehr dataset
 echo "-------------------------->Generate Unioned ehr dataset (step 2)"
-source_prefix="unioned_ehr_"
-unioned_ehr_dataset="unioned_ehr${today}"
-echo "unioned_ehr_dataset --> $unioned_ehr_dataset"
-"${TOOLS_DIR}/generate_unioned_ehr_dataset.sh" --key_file ${key_file} --ehr_snap_dataset ${ehr_snap_dataset} --vocab_dataset ${vocab_dataset}
+"${TOOLS_DIR}/generate_unioned_ehr_dataset.sh" --key_file ${key_file} --ehr_snap_dataset ${ehr_snapshot} --vocab_dataset ${vocab_dataset} --dataset_release_tag ${dataset_release_tag}
```
lrwb-aou: You're using `${ehr_snapshot}` without having set it.

data_steward/tools/run_curation_pipeline.sh
```diff
@@ -52,68 +52,82 @@ while true; do
...
 
 #---------------------------------------------------------
 # Step 3 Generate combined dataset
 echo "-------------------------->Generate combined ehr rdr dataset (step 3)"
-cdr="combined${today}"
-tag=$(git describe --abbrev=0 --tags)
-version=${tag}
-echo "cdr --> $cdr"
-"${TOOLS_DIR}/generate_combined_dataset.sh" --key_file ${key_file} --vocab_dataset ${vocab_dataset} --unioned_ehr_dataset ${unioned_ehr_dataset} --rdr_dataset ${rdr_dataset}
+"${TOOLS_DIR}/generate_combined_dataset.sh" --key_file ${key_file} --vocab_dataset ${vocab_dataset} --unioned_ehr_dataset ${unioned_ehr_dataset} --rdr_dataset ${rdr_dataset} --dataset_release_tag ${dataset_release_tag}
```
lrwb-aou: You're using `${unioned_ehr_dataset}` without having set it.

data_steward/tools/import_rdr_omop.sh
lrwb-aou: I think I understand.  It sounds like we are loading RDR data into datasets on a monthly basis, but that is not where the pipeline starts.  
When this runs lines 61 -63, it will create a BigQuery dataset with a name formatted as:  `YYYMMDD_rdr`, `YYYMMDD_rdr_backup`, and `YYYMMDD_rdr_staging`.   Our naming conventions document says they will be named using the convention `{YYYY}q{X1}r{X2}_rdr` and `{YYYY}q{X1}r{X2}_rdr_backup`.  

The staging environment isn't even indicated for the rdr dataset.  Either the code is wrong or the documentation needs updating or both??

ksdkalluri: The dataset naming convention document is for the naming convention that we will be using for the datasets that are being created as part of the pipeline. 
So my understanding from that is - we will load an rdr dump into bigquery and name it as rdr{YYMMDD}. which will be the raw data that we get from the rdr and when we run the CDR generation pipeline we will be snapshotting the dataset which we decide on using and apply the cleaning rules on it which will use the CDR naming convention which is {YYYY}q{X1}r{X2}_rdr.

data_steward/tools/create_ehr_and_rdr_snapshot.sh
```diff
@@ -0,0 +1,118 @@
...
+python "${CLEANER_DIR}/clean_cdr.py" --data_stage ${data_stage} -s
+
+# Create a snapshot dataset with the result
+python "${TOOLS_DIR}/snapshot_by_query.py" -p "${app_id}" -d "${rdr_snapshot_staging}" -n "${rdr_snapshot}"
+
+#copy sandbox dataset
+"${TOOLS_DIR}/table_copy.sh" --source_app_id ${app_id} --target_app_id ${app_id} --source_dataset "${rdr_snapshot_staging}_sandbox" --target_dataset "${rdr_snapshot}_sandbox"
+
+bq rm -r -d "${rdr_snapshot_staging}_sandbox" 
+bq rm -r -d "${rdr_snapshot_staging}"
```
ksdkalluri: @lrwb-aou This is the part I was talking about. This part will snapshot the raw rdr dump dataset that we loaded and applies the cleaning rules on it.

PR #390, opened 2020-01-27
[DC-659] fix query syntax for rdr cleaning

amcgrenera-vumc:
Change delete query to use correct syntax.

### DIFF
--- a/data_steward/cdr_cleaner/cleaning_rules/drop_multiple_measurements.py
+++ b/data_steward/cdr_cleaner/cleaning_rules/drop_multiple_measurements.py
@@ -13,24 +13,22 @@ Relevant measurement_source_concept_ids are listed in query
 import constants.cdr_cleaner.clean_cdr as cdr_consts
 
 REMOVE_MULTIPLE_MEASUREMENTS = """
-DELETE
-FROM
-(
-SELECT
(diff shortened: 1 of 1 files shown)


data_steward/cdr_cleaner/cleaning_rules/drop_multiple_measurements.py
```diff
@@ -13,24 +13,22 @@
...
-ORDER BY
-  person_id, measurement_source_concept_id, row_num
-  )
-WHERE
-  row_num != 1
+DELETE FROM `{project}.{dataset}.measurement`
+WHERE 
+  EXISTS (
+  --subquery to order multiple measurements by measurement_datetime
+    WITH mult_measurements AS(
```
lrwb-aou: FYI:  I haven't used a with clause inside a query like this before.  I'll have to experiment with it sometime.

amcgrenera-vumc: I hadn't before this either. In `SELECT` you're able to add subqueries, but with `DELETE` it isn't allowed, so I found a workaround that allows you to add subqueries inside `EXISTS` within the `WHERE` clause.

PR #398, opened 2020-01-31
[DC-449] Participant match script alert when PII data is missing

nishanthpp93:
* Add new table in results.html for pii missing check
* Add new test files participant match and pii name in five person
* Add check for participant existence in rdr


data_steward/validation/main.py
```diff
@@ -502,6 +477,38 @@ def get_drug_class_counts_query(hpo_id):
...
+    :param hpo_id: identifies the HPO site
+    :return: the query
+    """
+    person_table_id = bq_utils.get_table_id(hpo_id, common.PERSON)
+    pii_name_table_id = bq_utils.get_table_id(hpo_id, common.PII_NAME)
+    pii_wildcard = bq_utils.get_table_id(hpo_id, common.PII_WILDCARD)
+    participant_match_table_id = bq_utils.get_table_id(hpo_id,
+                                                       common.PARTICIPANT_MATCH)
+    rdr_dataset_id = bq_utils.get_rdr_dataset_id()
+    rdr_date = rdr_dataset_id[3:]
```
calbach: These indices are a bit sketchy - I'd probably use a regex or at least make an obvious assertion about the input value, e.g. throw/log an error if it doesn't start with the expected prefix, or isn't the correct length. Also worth adding an example value in a comment here to make this clear

tests/integration_tests/data_steward/validation/main_test.py
```diff
@@ -259,33 +249,20 @@ def test_pii_files_loaded(self, mock_check_cron):
...
-        ]
         main.app.testing = True
         with main.app.test_client() as c:
             c.get(test_util.VALIDATE_HPO_FILES_URL)
             actual_result = test_util.read_cloud_file(
                 self.hpo_bucket, folder_prefix + common.RESULTS_HTML)
-            actual_result_file = self._remove_timestamp_tags_from_results(
-                StringIO(actual_result).getvalue())
-            self.assertIn(actual_result_file, expected_results)
+        # print(actual_result)
```
calbach: nit: better to just remove commented out code

tests/integration_tests/data_steward/validation/main_test.py
```diff
@@ -259,33 +249,20 @@ def test_pii_files_loaded(self, mock_check_cron):
...
-                f.read())
-        expected_results = [
-            expected_result_achilles_success, expected_result_achilles_failure
-        ]
         main.app.testing = True
         with main.app.test_client() as c:
             c.get(test_util.VALIDATE_HPO_FILES_URL)
             actual_result = test_util.read_cloud_file(
                 self.hpo_bucket, folder_prefix + common.RESULTS_HTML)
-            actual_result_file = self._remove_timestamp_tags_from_results(
```
calbach: Did you mean to remove the assertions here? Does this test still do something?

data_steward/constants/validation/main.py
```diff
@@ -133,6 +133,53 @@
...
+participant_records AS
+(SELECT person_id
+FROM `{project_id}.{dataset_id}.{participant_match_table_id}`)
+(SELECT DISTINCT {ehr_no_pii} AS missingness_type,
+    COUNT(person_id) AS count
+FROM (SELECT person_id FROM ehr_persons
+    EXCEPT DISTINCT
+    SELECT person_id FROM pii_names))
+UNION ALL
+(SELECT DISTINCT {ehr_no_rdr}{rdr_date} AS missingness_type,
```
calbach: I would combine these outside of the SQL template, i.e. ehr_no_rdr should already include the date at this point - maybe just use a template for the EHR_NO_RDR message.

tests/integration_tests/data_steward/validation/main_test.py
```diff
@@ -259,33 +249,20 @@ def test_pii_files_loaded(self, mock_check_cron):
...
-                f.read())
-        with open(test_util.FIVE_PERSON_RESULTS_ACHILLES_ERROR_FILE, 'r') as f:
-            expected_result_achilles_failure = self._remove_timestamp_tags_from_results(
-                f.read())
-        expected_results = [
-            expected_result_achilles_success, expected_result_achilles_failure
-        ]
         main.app.testing = True
         with main.app.test_client() as c:
             c.get(test_util.VALIDATE_HPO_FILES_URL)
```
calbach: Is it possible to test the new queries?

data_steward/validation/main.py
nishanthpp93: Yeah, good point. I'll add something here to check for the format and test it.

tests/integration_tests/data_steward/validation/main_test.py
nishanthpp93: I was using this as a debugging spot to see the result in a browser, but I think I should add a html parser to check for the added table.

tests/integration_tests/data_steward/validation/main_test.py
nishanthpp93: This test was earlier comparing the html files for actual and expected, which caused issues will small changes like a new site being added. We disabled this test until we came across a better alternative. I'll modify this to parse html and check for expected tables.

data_steward/constants/validation/main.py
nishanthpp93: Yeah I was a bit unsure of doing this. Will make that change.

tests/integration_tests/data_steward/validation/main_test.py
nishanthpp93: Yeah I'll modify this test. I was using it to display the html file, but I think I can add checks using a parser.

PR #403, opened 2020-02-06
[DC-682] Fix RDR dataset id error in tests

nishanthpp93:
Fixed issue where RDR dataset naming is checked for all tests

### DIFF
--- a/tests/integration_tests/data_steward/validation/ehr_union_test.py
+++ b/tests/integration_tests/data_steward/validation/ehr_union_test.py
@@ -4,10 +4,12 @@ import re
 import unittest
 
 import dpath
+import mock
 import moz_sql_parser
 
 import bq_utils
 import cdm
+import common
 from constants.tools import combine_ehr_rdr
 from constants.validation import ehr_union as eu_constants
 import gcs_utils
@@ -77,12 +79,14 @@ class EhrUnionTest(unittest.TestCase):
     def _load_datasets(self):
         """
         Load five persons data for each test hpo
-        """
         # expected_tables is for testing output
         # it maps table name to list of expected records ex: "unioned_ehr_visit_occurrence" -> [{}, {}, ...]
+        :param pto: Flag indicating whether load is for person_to_observation tests
+        """
         expected_tables = dict()
         running_jobs = []
-        for cdm_table in resources.CDM_TABLES:
+        cdm_tables = resources.CDM_TABLES
+        for cdm_table in cdm_tables:
             output_table = ehr_union.output_table_for(cdm_table)
             expected_tables[output_table] = []
             for hpo_id in self.hpo_ids:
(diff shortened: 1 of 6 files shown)


tests/integration_tests/data_steward/validation/main_test.py
```diff
@@ -127,8 +130,11 @@ def test_bad_file_names(self):
                                      self.folder_prefix)
         self.assertCountEqual(expected_warnings, r['warnings'])
 
+    @mock.patch('validation.main.all_required_files_loaded')
```
ChaoPang: I noticed that `all_required_files_loaded` was not called during `validate_submission`, I think we can remove this mock

tests/integration_tests/data_steward/validation/ehr_union_test.py
```diff
@@ -447,8 +459,17 @@ def test_ehr_person_to_observation(self):
         self.assertEqual(len(expected), len(actual))
         self.assertCountEqual(expected, actual)
```
ChaoPang: Is the second assertion same as the previous one?

tests/integration_tests/data_steward/validation/ehr_union_test.py
```diff
@@ -77,12 +79,14 @@ def _create_hpo_table(self, hpo_id, table, dataset_id):
...
         Load five persons data for each test hpo
-        """
         # expected_tables is for testing output
         # it maps table name to list of expected records ex: "unioned_ehr_visit_occurrence" -> [{}, {}, ...]
+        :param pto: Flag indicating whether load is for person_to_observation tests
+        """
         expected_tables = dict()
         running_jobs = []
-        for cdm_table in resources.CDM_TABLES:
+        cdm_tables = resources.CDM_TABLES
```
ChaoPang: Is there a particular reason that `cdm_tables` was created?

tests/integration_tests/data_steward/validation/main_test.py
nishanthpp93: I agree, looks like it only gets called in `generate metrics`

tests/integration_tests/data_steward/validation/ehr_union_test.py
nishanthpp93: I think it was part of a change we made during migration. One of them was testing for the individual items while the other was testing for the keys. I didn't make this change for this ticket but I could remove one of them for now

tests/integration_tests/data_steward/validation/ehr_union_test.py
nishanthpp93: I had created it to mock `resources.CDM_TABLES` but later realized that was not necessary since I can mock `tables_to_map`. I'll remove it

tests/integration_tests/data_steward/validation/ehr_union_test.py
ChaoPang: Thanks for the explanation. Sorry this is totally not within the scope! I was just trying to run the test from my machine and step through the code, that's when I saw these two assertions. I thought it might be a good idea to remove one if they were checking the same thing.

nishanthpp93: I'll remove the `assertEqual` since `assertCountEqual` handles list of dictionaries

PR #406, opened 2020-02-07
[DC-662] Improve clean engine logging

ChaoPang:
Improves clean engine logging by:
* adding an indicator to log `N out of M rules is running` type message.
* adds module info with a decorator function to log which module generated the SQL that the engine is running.
* adds full query set parameters to log in the event of an error
* alerts to all rules successfully run, no failures, or how many failures occurred.


data_steward/cdr_cleaner/clean_cdr.py
```diff
@@ -46,9 +47,71 @@
...
+DATASET_ID = 'dataset_id'
+SANDBOX_DATASET_ID = 'sandbox_dataset_id'
+
+
+class CleanCdrIllegalArgumentError(ValueError):
+    pass
+
+
+def add_module_info_decorator(query_function,
+                              project_id,
```
lrwb-aou: As defined, this will break.  I tested it with the patch I am attaching.  It raises your `CleanCdrIllegalArgumentError`.

[improved_logging_breaks.txt](https://github.com/all-of-us/curation/files/4182624/improved_logging_breaks.txt)

Essentially, the above patch assumes every query module is using a keyword based approach for parameters.  They are not.  This approach requires all the query modules name their parameters the same way.  Running this patch with `PYTHONPATH=.:$PYTHONPATH python cdr_cleaner/clean_cdr.py -d 'deid_base'`, you can see the module is expecting `sandbox_dataset` but is receiving `sandbox_dataset_id`.  This fails and is not able to seamlessly integrate with each existing query module.

[improved_logging_works.txt](https://github.com/all-of-us/curation/files/4182625/improved_logging_works.txt)
The above patch makes use of positional and keyword parameter arguments.  It uses the `*` and `**` prefixes.  It lets the query modules use whatever naming convention they prefer. It also allows you to remove some unnecessary global variables, dictionaries, and checks.  You can test it with `PYTHONPATH=.:$PYTHONPATH python cdr_cleaner/clean_cdr.py -d 'deid_base'` also.

Of course, fixing this will require updating your tests.  Let me know what you think.

FYI:  to apply a patch from file from the command line:
```
cd <wherever your project is>/curation/data_steward
git apply <path to downloaded patch file>
```
then you can activate your python environment and run the python command on your patched file to see the break and the working file.

data_steward/cdr_cleaner/clean_cdr.py
```diff
@@ -46,9 +47,71 @@
...
+                              dataset_id,
+                              sandbox_dataset_id=None):
+    """
+    A decorator for adding the module information to the list of query dictionaries generated by the cleaning rules
+    :param query_function: a function that generates a list of query dictionaries
```
lrwb-aou: For the purpose of automated documentation, there should be a blank line after the one line summary sting.

data_steward/bq_utils.py
```diff
@@ -320,6 +320,15 @@ def job_status_done(job_id):
     return job_running_status == 'DONE'
 
 
+def job_status_errored(job_id):
```
lrwb-aou: Handy function.  Please provide a pep-8 style documentation string.  https://www.python.org/dev/peps/pep-0257/#multi-line-docstrings

data_steward/cdr_cleaner/clean_cdr_engine.py
```diff
@@ -14,6 +14,21 @@
 LOGGER = logging.getLogger(__name__)
 FILENAME = '/tmp/cleaner.log'
 
+FAILURE_MESSAGE_TEMPLATE = '''
```
lrwb-aou: nit:  consider creating a constants module for this and other globals and moving it there.

data_steward/cdr_cleaner/clean_cdr_engine.py
```diff
@@ -77,28 +92,40 @@ def clean_dataset(project=None, statements=None, data_stage=stage.UNSPECIFIED):
                                      batch=batch)
 
         except (oauth2client.client.HttpAccessTokenRefreshError,
-                googleapiclient.errors.HttpError):
-            LOGGER.exception("FAILED:  Clean rule not executed:\n%s",
-                             rule_query)
+                googleapiclient.errors.HttpError) as exp:
+
+            LOGGER.error(
```
lrwb-aou: I would leave this as an `exception` to give access to the stack trace.   The failure message is going to contain the extra relevant information.  But this did technically fail when the query was issued.

data_steward/cdr_cleaner/clean_cdr_engine.py
```diff
@@ -108,7 +135,30 @@ def clean_dataset(project=None, statements=None, data_stage=stage.UNSPECIFIED):
                        data_stage)
 
     if failures > 0:
-        print("Failed to apply {} clean rules for {}.{}".format(
```
lrwb-aou: good call removing this.

data_steward/cdr_cleaner/clean_cdr_engine.py
```diff
@@ -108,7 +135,30 @@ def clean_dataset(project=None, statements=None, data_stage=stage.UNSPECIFIED):
                        data_stage)
 
     if failures > 0:
```
lrwb-aou: nit:  After running the pipeline with Aimee, it would be nice if this conditional had an else statement that explicitly said there were no failures.

data_steward/cdr_cleaner/clean_cdr_engine.py
```diff
@@ -77,28 +92,40 @@ def clean_dataset(project=None, statements=None, data_stage=stage.UNSPECIFIED):
                                      batch=batch)
```
lrwb-aou: OPTIONAL:  At line 85, it logs which query is executing.  It would be nice to change this to something like:
```
statement_length = len(statements)
for index, statement in enumerate(statements):
    ...
    LOGGER.info("Running query %d out of %d:\n%s", index, statement_length, rule_query)
```

data_steward/cdr_cleaner/clean_cdr.py
ChaoPang: Thanks for providing those patches and instructions on how to apply those! This is really cool and it is much much easier to read and understand once I get around the positional the keyword arguments. I was trying to solve this issue by introducing the check but this was totally unnecessary.

data_steward/cdr_cleaner/clean_cdr_engine.py
ChaoPang: @lrwb-aou just wondering, what is the best practice for maintaining the logging related globals? Shoud they be moved to the constant module as well?

data_steward/cdr_cleaner/clean_cdr.py
ChaoPang: Added

data_steward/bq_utils.py
ChaoPang: Added the docstring

data_steward/cdr_cleaner/clean_cdr_engine.py
ChaoPang: Reverted this back to an exception

data_steward/cdr_cleaner/clean_cdr_engine.py
ChaoPang: Added the else statement.

data_steward/cdr_cleaner/clean_cdr_engine.py
ChaoPang: This is done.

tests/unit_tests/data_steward/cdr_cleaner/clean_cdr_test.py
```diff
@@ -0,0 +1,56 @@
...
+            cdr_consts.MODULE_NAME: self.model_name,
+            cdr_consts.FUNCTION_NAME: self.function_name,
+            cdr_consts.LINE_NO: self.line_no
+        }
+
+    @mock.patch('inspect.signature')
+    @mock.patch('inspect.getsourcelines')
+    @mock.patch('inspect.getmodule')
+    def test_add_module_info_decorator(self, mock_get_module,
+                                       mock_getsourcelines, mock_signature):
```
lrwb-aou: @ChaoPang Codacy is complaining about the `mock_signature`.  It doesn't look like you're using it.  You should be able to remove it and its patch.

data_steward/cdr_cleaner/clean_cdr_engine.py
lrwb-aou: I don't know.  Use your best judgement.  Eventually, I hope all this logging setup will be cleared out of this module and done in its own system wide logging setup module.  I _think_ the logging function doc string says something about needing to be properly placed in a separate module.

tests/unit_tests/data_steward/cdr_cleaner/clean_cdr_test.py
ChaoPang: @lrwb-aou thanks for pointing this out, issue addressed.

data_steward/cdr_cleaner/clean_cdr_engine.py
ChaoPang: @lrwb-aou  thanks for leaving this comment. I would move `FILENAME='/tmp/cleaner.log'` to the constants module and leave `LOGGER = logging.getLogger(__name__)` in `clean_cdr_engine.py` because it will be replaced by the system wide logging setup module eventually.

lrwb-aou: Sorry I wasn't clear earlier.  Yes, you should leave `LOGGER = logging.getLogger(__name__)` here.  It will always have to be here if you want your logs to print from this module with the correct module name.  The other globals may be moved to a constants file.

PR #411, opened 2020-02-11
[EDQ-368][EDQ-369] Outlier measurement script updates

noahgengel:
* Updated outlier measurement script with latest measurement ancestor IDs
* Updated .gitignore to exclude any images that result from the cdr_ops files


nishanthpp93:
Rolled up in #423

PR #414, opened 2020-02-13
[EDQ-372] Create separate folder for table metrics that are designed for the combined datasets (with different table naming conventions) / Update .gitignore

noahgengel:
NOTE PR includes commits that were note yet approved from [pull #411](https://github.com/all-of-us/curation/pull/411): Outlier measurement script updates.

* separates out the table_metrics scripts to two different categories. This allows analysis on both combined and ehr_ops datasets

* changes the .gitignore to exclude the CSV files that are generated from the table metrics scripts


nishanthpp93:
Rolled up in #423

PR #416, opened 2020-02-14
[DC-636] Refactor Circle and split into nightly job

calbach:
New functionality:
- Integration/unit tests run nightly on `develop`
- Integration/unit tests run on `master` merge

Maintained:
- Unit and integration tests run on PRs.
- Unit and integration tests run on remote commits with "all tests" in the message.
- Unit tests run on develop merge.

This also breaks out the tests into two separate jobs with composable `command` steps, and defines the new structure via workflows.


.circleci/config.yml
```diff
@@ -94,16 +128,26 @@ jobs:
...
-      - store_artifacts:
-          path: ~/curation/tests/results/coverage
-          destination: test_results
+      - test_teardown
+
+workflows:
+  version: 2
+  test:
+    jobs:
+      - unit_test
```
lrwb-aou: Should this be `- unit_test:`??  Unit tests didn't run on this PR.

calbach: They ran:

image


Separately however, they did fail. It looks like some of these tests depend on ADC being configured since I am no longer authenticating the unit test process. I will fix those today - but if you're able please review the config in the meantime.

calbach: Also, this is a nice side-effect of not setting up buckets or auth on unit tests, it forcibly reveals the non-hermetic dependencies.

lrwb-aou: You're right.  I skimmed and did not notice that the unit tests and integration tests were reporting separately now.  I like revealing the dependencies.  These shouldn't be needed for unit testing anyway.  Thanks CH.

.circleci/config.yml
```diff
@@ -94,16 +128,26 @@ jobs:
...
+      - integration_test:
+          filters:
+            branches:
+              # This would cause too much noise to run on commit, run nightly
+              # instead per below workflow.
+              ignore: develop
```
lrwb-aou: I take this to mean, integration tests won't run on develop when a PR is merged, but will run nightly on develop.  Correct?

calbach: Yep. When a PR is merged, just the unit tests will run.

PR #417, opened 2020-02-18
[DC-601] suppression to remove ehr dob from observation table

amcgrenera-vumc:


amcgrenera-vumc:
@calbach Just an FYI, I was able to run the deid on test data that Lauren provided me, to confirm that this new suppression worked and it was successful.

data_steward/deid/config/ids/tables/observation.json
```diff
@@ -117,7 +117,7 @@
         {   "comment": "removing ethnic groups, alternative race and fields (data has redundancies)",
              "on": "observation_concept_id in (4013886, 4271761, 4135376)"
```
calbach: Optional: I could see an argument to modify this rule instead (include `4083587` on this line), since these are the other 3 concepts which get copied from the person table and the suppression rationale seems similar in both cases:

https://github.com/all-of-us/curation/blob/62730f89c231b23fc9d370891f25a57dce2f9038/data_steward/validation/ehr_union.py#L541

amcgrenera-vumc: I agree, will adjust and make a second commit!

amcgrenera-vumc: @calbach please verify change and if it looks alright to you I will merge, thanks!

PR #418, opened 2020-02-18
[DC-636] Circle cleanup: cache pip installs, don't run unit tests nightly

calbach:
Minor follow-ups to https://github.com/all-of-us/curation/pull/416 

1. Cache pip installs

This caches the virtualenv folder between runs, providing a starting point for `pip install`. You can see it in action here (~40s -> ~2s): 
image

Note: there seems to be a fair amount of variance in the time required to download the cache. 3s seems typical, but on this PR the unit tests restore_cache step took ~50s (about the same time as just running pip install).

2. Skip unit tests. These already run post-merge into develop, so there's no additional information gained by running them nightly.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -156,5 +156,4 @@ workflows:
             branches:
               only: develop
     jobs:
-      - unit_test
       - integration_test
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -37,10 +37,13 @@ commands:
             sudo apt-get install dpkg
             sudo apt-get install google-cloud-sdk
       # Dependencies
+      - restore_cache:
+          keys:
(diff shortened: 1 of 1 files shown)


calbach:
@lrwb-aou FYI

lrwb-aou:
Just want to add this here for history.  https://circleci.com/docs/2.0/caching/#managing-caches. If I'm understanding this correctly, if we ever need to clear the cache we will have to update a file name and force a new cache build.

calbach:
@lrwb-aou added more context to the PR description. unit tests already run on every merge to develop anyways, so running them nightly as well just adds noise in my opinion.

There's no ticket. It's a follow-up to DC-636 per some slack discussion.

calbach:
Done

PR #419, opened 2020-02-18
[DC-369] Enforce Submission Directory Naming Convention

nishanthpp93:
Generate empty report if submission folder is named incorrectly

### DIFF
--- a/data_steward/validation/main.py
+++ b/data_steward/validation/main.py
@@ -399,22 +399,20 @@ def generate_empty_report(hpo_id, bucket, folder_prefix):
     report_data = dict()
     processed_datetime_str = datetime.datetime.now().strftime(
         '%Y-%m-%dT%H:%M:%S')
-    logging.info(
-        'Folder %s does not follow naming convention %s',
-        folder_prefix,
-    )
     report_data[report_consts.HPO_NAME_REPORT_KEY] = get_hpo_name(hpo_id)
     report_data[report_consts.FOLDER_REPORT_KEY] = folder_prefix
     report_data[report_consts.TIMESTAMP_REPORT_KEY] = processed_datetime_str
     report_data[
         report_consts.
         SUBMISSION_ERROR_REPORT_KEY] = f'Submission folder name {folder_prefix} does not follow the ' \
-                                       f'naming convention YYYY-MM-DD-vN/, where vN represents the version number ' \
-                                       f'for the day, starting at v1 each day. Please resubmit the files in a new ' \
-                                       f'folder with the correct naming convention'
-    logging.info('Processing skipped. Saving timestamp %s to `gs://%s/%s`.',
-                 processed_datetime_str, bucket,
-                 folder_prefix + common.PROCESSED_TXT)
+                                       f'naming convention {consts.FOLDER_NAMING_CONVENTION}, where vN represents ' \
+                                       f'the version number for the day, starting at v1 each day. ' \
+                                       f'Please resubmit the files in a new folder with the correct naming convention'
+    logging.info(
+        'Processing skipped. Reason: Folder %s does not follow naming convention %s. '
+        'Saving timestamp %s to `gs://%s/%s`.', folder_prefix,
+        consts.FOLDER_NAMING_CONVENTION, processed_datetime_str, bucket,
+        folder_prefix + common.PROCESSED_TXT)
     _write_string_to_file(bucket, folder_prefix + common.PROCESSED_TXT,
                           processed_datetime_str)
     results_html = hpo_report.render(report_data)
--- a/data_steward/validation/main.py
+++ b/data_steward/validation/main.py
@@ -386,6 +386,55 @@ def generate_metrics(hpo_id, bucket, folder_prefix, summary):
     return report_data
 
 
+def generate_empty_report(hpo_id, bucket, folder_prefix):
+    """
+    Generate an empty report with a "validation failed" error 
+    Also write processed.txt to folder to prevent processing in the future
+
+    :param hpo_id: identifies the HPO site
+    :param bucket: name of the bucket with the submission
+    :param folder_prefix: folder containing the submission
+    :return: report_data: dict whose keys are params in resources/templates/hpo_report.html
+    """
+    report_data = dict()
+    processed_datetime_str = datetime.datetime.now().strftime(
+        '%Y-%m-%dT%H:%M:%S')
+    logging.info(
+        'Folder %s does not follow naming convention %s',
+        folder_prefix,
+    )
+    report_data[report_consts.HPO_NAME_REPORT_KEY] = get_hpo_name(hpo_id)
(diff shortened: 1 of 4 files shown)


data_steward/validation/main.py
```diff
@@ -386,6 +386,55 @@ def generate_metrics(hpo_id, bucket, folder_prefix, summary):
...
+    :param hpo_id: identifies the HPO site
+    :param bucket: name of the bucket with the submission
+    :param folder_prefix: folder containing the submission
+    :return: report_data: dict whose keys are params in resources/templates/hpo_report.html
+    """
+    report_data = dict()
+    processed_datetime_str = datetime.datetime.now().strftime(
+        '%Y-%m-%dT%H:%M:%S')
+    logging.info(
+        'Folder %s does not follow naming convention %s',
```
lrwb-aou: This is broken.  You must provide a second replacement string if you're going to use this log line.

tests/integration_tests/data_steward/validation/main_test.py
```diff
@@ -295,6 +296,31 @@ def test_html_report_five_person(self, mock_check_cron, mock_rdr_date,
         self.assertIn(main_consts.EHR_NO_PARTICIPANT_MATCH,
                       missing_record_types)
 
+        # test for required labs
```
lrwb-aou: I'm not sure this code section should be merged as part of this PR.

tests/integration_tests/data_steward/metrics/required_labs_test.py
```diff
@@ -189,39 +188,3 @@ def test_get_lab_concept_summary_query(self):
                          msg='Compare the number '
                          'of labs submitted '
                          'in the measurement')
-
```
lrwb-aou: I'm not sure how this relates to the submission directory naming convention tests.

data_steward/validation/main.py
nishanthpp93: Good catch! Will fix this

tests/integration_tests/data_steward/metrics/required_labs_test.py
nishanthpp93: This test was failing when I pushed my changes, and since Chao and I had discussed merging it earlier, I thought I could do it here. However, we now have #421 which does this, and once it is merged, the commits pertaining to this will not be necessary. I'll remove them and push

tests/integration_tests/data_steward/validation/main_test.py
nishanthpp93: Yes, this was to add the failing test. Since we have #421, I'll remove these changes

data_steward/validation/main.py
```diff
@@ -386,6 +386,55 @@ def generate_metrics(hpo_id, bucket, folder_prefix, summary):
...
+
+def is_valid_folder_prefix_name(folder_prefix):
+    """
+    Verifies whether folder name follows naming convention YYYY-MM-DD-vN, where vN is the submission version number,
+    starting at v1 every day
+    
+    :param folder_prefix: folder containing the submission
+    :return: Boolean indicating whether the input folder follows the aforementioned naming convention
+    """
+    folder_name_format = re.compile(consts.FOLDER_NAME_REGEX)
```
lrwb-aou: you can validate this as a date object with
```
datetime.datetime.strptime(folder_prefix[:10], '%Y-%m-%d)
```
This would catch months greater than 12 or days greater than the last day of each month.  For example, a month value of 15 would raise a ValueError.  I think this is worth checking.
It also make using `timedelta` a possibility to make sure we aren't accepting submissions with really old dates.

nishanthpp93: Yeah, this would help limit the dates to a fixed timeframe.

lrwb-aou: You should still do the regex comparison.  This would just be a nice addition.

lrwb-aou:
- This patch now has a conflict.  
- Try changing `@mock.patch('validation.main._write_string_to_file')` to `@mock.patch('validation.main._write_string_to_file', mock.MagicMock())` to get rid of the codacy message.
- Have you rebased Chao's merge to take advantage of DC-676's updated tests?  An integration test is failing now.

nishanthpp93:
I rebased and force pushed just now. How do I check the patch conflict? Is it the "unused argument error" on Codacy?

lrwb-aou:
- It looks like rebasing fixed the conflict.  Yea!
- I suggested `@mock.patch('validation.main._write_string_to_file', mock.MagicMock())` to try to make Codacy's unused argument error go away.  You have to mock that, but Codacy doesn't seem to like it.  This isn't a blocker.  I just wondered if this change would satisfy the Codacy check.

nishanthpp93:
Awesome! The Codacy thing is annoying. Let me try your suggestion

nishanthpp93:
It looks like it doesn't help resolve the Codacy error. I'll remove the commit and push.

lrwb-aou:
One last CircleCI thing.  We can disable the pylint check for that particular mocked argument with `# pylint: disable=unused-argument`.  It needs to be placed on the same line as the argument you want to ignore.  It will only ignore that argument and will continue to check that the rest of the arguments are used.  I personally like checking that all my arguments are used, but pylint is doing a static analysis and is not picking up on the fact that this mocked argument that is used at runtime.  So, as a general rule, I would only be willing to ignore mocked arguments on a case by case basis.  You can view the StackOverflow explanation here for some other ideas.  https://stackoverflow.com/questions/59136064/python-unused-argument-needed-for-compatibility-how-to-avoid-pylint-complainin

To test this from the command line you can:
```
$ # Enable your python development environment
$ source <env_file_location>/bin/activate
$ # Run pylint on unit tests from curation/data_steward directory
$ pylint ../tests/unit_tests/data_steward/validation/main_test.py
```
You can test without the disable comment and then with the disable comment.  When disabled, the complaint should disappear.

data_steward/validation/main.py
```diff
@@ -386,6 +386,59 @@ def generate_metrics(hpo_id, bucket, folder_prefix, summary):
...
+    if not folder_name_format.match(folder_prefix):
+        return False
+    try:
+        datetime.datetime.strptime(folder_prefix[:10], '%Y-%m-%d')
+    except ValueError:
+        return False
```
lrwb-aou: nit: Consider bracketing this section off with opening and closing blank lines.  A blank line before 435 and after 438.

nishanthpp93:
I think the Codacy issue will go away once the PR is merged. It will be considered a known issue and won't show up again. I think there are lots of instances where Pylint considers a mock argument to be "unused" even though they are used during runtime, and we need a comprehensive way to deal with all of them together. Regarding the blank lines, the function is quite small and PEP 8 checks weren't complaining so I thought it's fine with regards to readability. I was also hoping yapf would handle issues like these but looks like it doesn't.

PR #422, opened 2020-02-21
[DC-529] Remove irrelevant observation_source_concept_ids from RDR

lrwb-aou:
[DC-529] Remove irrelevant observation_source_concept_ids from the RDR Export

* Setting up a base class for Cleaning Rules to Inherit from.
* Modifying the current rdr pipeline to accept class or function names.
* Setting the pipeline up to only accept instances of a BaseClass in the future.
* Unit and integration tests needed.

[DC-529]: https://precisionmedicineinitiative.atlassian.net/browse/DC-529


lrwb-aou:
Closing this PR in favor of a separate PR for the base class and a different PR for creating the cleaning rule using the base class.

PR #425, opened 2020-02-25
[DC-705] Add observation_period to list of tables to retract from

nishanthpp93:
Add observation_period to retraction list


amcgrenera-vumc:
@nishanthpp93 looks good to me, only question I have is if `OBSERVATION_PERIOD` needs to be added to `common.py`?

nishanthpp93:
Yes, luckily it already exists in `common.py` so I didn't have to add it separately.

amcgrenera-vumc:
ahh sorry I missed that, thanks for the clarification!

PR #428, opened 2020-02-28
[EDQ-383] Records Per Capita Notebook

noahgengel:
Creates Notebook for Records Per Capita for Each Table/Site Combination


noahgengel:
NOTE: This kind of analysis would admittedly be far more useful if we had also divided the number of records by the amount of average time since the patient's first medication. This analysis would, however, require far more computational power.

If we were to want to make this kind of analysis more robust (pending a discussion with Karthik), I would need to speak with @ChaoPang to see how I could gain access to remote servers with a greater ability to process.

nishanthpp93:
Rolled up in #436

PR #429, opened 2020-02-28
[DC-691] Dataset Spring Cleaning for test project

nishanthpp93:
Add script to delete test datasets, add dataset deletion to bq_utils


data_steward/tools/clean_test_project.py
```diff
@@ -0,0 +1,75 @@
...
+    :return: 
+    """
+    deleted, not_deleted = 0, 0
+    for dataset in datasets_to_delete_list:
+        try:
+            bq_utils.delete_dataset(project_id, dataset)
+            deleted += 1
+            logging.info('Deleted dataset %s' % dataset)
+        except HttpError:
+            logging.info('Could not delete dataset %s' % dataset)
```
lrwb-aou: if you used `logging.exception(` here we could get the stack trace that goes with your error message.

data_steward/tools/clean_test_project.py
```diff
@@ -0,0 +1,75 @@
...
+# Project imports
+import app_identity
+import bq_utils
+
+logging.basicConfig(
+    stream=sys.stdout,
+    level=logging.DEBUG,
+    format="%(asctime)s - %(name)s - %(levelname)s - %(message)s")
+
+GITHUB_USERNAMES = []
```
lrwb-aou: I assume this is intentionally blank.

data_steward/tools/clean_test_project.py
```diff
@@ -0,0 +1,75 @@
...
+    all_datasets = [
+        bq_utils.get_dataset_id_from_obj(dataset_obj)
+        for dataset_obj in all_dataset_objs
+    ]
+    circle_datasets = get_circle_datasets(all_datasets)
+    delete_datasets(project_id, circle_datasets)
+
+
+if __name__ == '__main__':
+    project_id = app_identity.get_application_id()
```
lrwb-aou: A nice thing would be to use `argparse` make this take a command line parameter to set the `project_id` and the list of `GITHUB_USERNAMES`.   Then this wouldn't be dependent on your environment variable and you woudn't have to manually update the script to run it.

I'm wary of delete because of the possibility of deleting datasets from prod.  Right now, if you ran this with your prod credentials, you might not have permissions to delete but it still makes me nervous.  Consider creating a prompt at line 70 that shows a list of proposed datasets that will be deleted, the project it is proposing to delete from, and require the user to input a case insensitive 'Y' or 'yes'.

data_steward/tools/clean_test_project.py
nishanthpp93: Yeah that makes sense.

data_steward/tools/clean_test_project.py
nishanthpp93: Yeah but it's better to change it to an argument as you suggested.

data_steward/tools/clean_test_project.py
nishanthpp93: Makes sense. Thanks for your suggestions, Lauren. I didn't give this script much thought since I didn't have any intention of re-using it, but I believe it's better to have this in our repo in case we need it.

data_steward/tools/clean_test_project.py
```diff
@@ -0,0 +1,109 @@
+# Python imports
```
lrwb-aou: Nice refactor!  I have a question.  This is a relatively small standalone script.  It's a decent candidate for using the BigQuery client library instead of `bq_utils`.  Have you given any thought to that, or do you prefer `bq_utils`?

nishanthpp93: Thanks! I haven't considered it yet. Do we have a consensus on which one to use? I can use it for my next task which is to identify datasets containing specific participants.

lrwb-aou: This may be something to bring up at the standup Monday.  I don't think we want to mix and match inside modules, if we can avoid it.  But I personally think new development should start incorporating the client library when possible.  This is probably something we'll need to get consensus on.  If you're going to leave this with `bq_utils`, I can go ahead and approve it.

data_steward/notebooks/bq.py
```diff
@@ -7,3 +7,36 @@
...
+    :return: 
+    """
+    if client.project == project_id:
+        client.delete_dataset(dataset_id,
+                              delete_contents=delete_contents,
+                              not_found_ok=not_found_ok)
+    else:
+        raise ValueError(
+            'Wrong project set. Please set the GOOGLE_CLOUD_PROJECT env variable correctly'
+        )
```
lrwb-aou: I like the idea of adding a fail safe, especially because this is a delete function and we can't predict if modules using this function are going to properly implement safety checks.  But I don't know if we want to fall back to environment variables or force setting environment variables.  I think we need to create a utility function to get a client object that requires setting the project (possibly credentials).  You can review the client api here:  https://googleapis.dev/python/bigquery/latest/generated/google.cloud.bigquery.client.Client.html#google.cloud.bigquery.client.Client .

Then that would be the first call you would make here, to get the client (just like in `bq_utils` functions).  The problem with a top level assignment is that it can't be mocked so it really messes up unit testing.  I know you didn't add that, but that's why I'm suggesting creating a function similar to `bq_utils.create_service()`.  But I'd like to start breaking our inherent dependency on environment variable settings.

nishanthpp93: Sure, I will rebase onto develop now that we have it

data_steward/notebooks/bq.py
```diff
@@ -0,0 +1 @@
+
```
lrwb-aou: Not sure this file is needed anymore.

data_steward/bq_utils.py
```diff
@@ -297,6 +297,23 @@ def delete_table(table_id, dataset_id=None):
     return delete_job.execute(num_retries=bq_consts.BQ_DEFAULT_RETRY_COUNT)
 
 
+def delete_dataset(project_id, dataset_id):
```
lrwb-aou: Because we are trying to move toward using the client library, I would remove this from the `bq_utils` module.  This will encourage using the new bq client library utils module.

data_steward/tools/clean_test_project.py
```diff
@@ -0,0 +1,107 @@
...
+import argparse
+import logging
+import sys
+
+# Third party imports
+from googleapiclient.errors import HttpError
+
+# Project imports
+from utils import bq
```
lrwb-aou: The lint check fails because `curation/data_steward/utils/__init__.py` doesn't exist.  Can you add that as part of your PR?

data_steward/utils/bq.py
```diff
@@ -15,11 +17,41 @@ def get_client(project_id=None):
...
+                   dataset_id,
+                   delete_contents=True,
+                   not_found_ok=True):
+    """
+    Delete a dataset in a project
+
+    :param project_id: Identifies the project the containing the dataset
+    :param dataset_id: Identifies the dataset to delete
+    :param delete_contents: If set True, deletes all contents within the dataset
+    :param not_found_ok: If set True, does not raise error if dataset cannot be found
```
lrwb-aou: I know we can easily check the signature, but I would specifically comment about the default value for these parameters.  This will be important if/when we start using documentation tools.

data_steward/tools/clean_test_project.py
```diff
@@ -0,0 +1,107 @@
...
+    response = get_response()
+    if response == "Y":
+        delete_datasets(project_id, datasets_with_substrings)
+    elif response.lower() == "n":
+        logging.info("Aborting deletion")
+
+
+# Make sure user types Y to proceed
+def get_response():
+    prompt_text = 'Please press Y/n\n'
```
lrwb-aou: Can you add a short documentation string here?

data_steward/bq_utils.py
nishanthpp93: I agree we need to use the new module. However this involves changing all current instances of bq_utils to use bq. Do you think it would be better in a separate PR? Maybe something that looks at deprecating bq_utils.py completely.

data_steward/tools/clean_test_project.py
```diff
@@ -0,0 +1,107 @@
...
+            logging.exception(f'Could not delete dataset {dataset}')
+            failed_to_delete.append(dataset)
+    logging.info(
+        f'The following datasets could not be deleted: {failed_to_delete}')
+
+
+def run_deletion(project_id, name_substrings):
+    """
+    1. Retrieves all datasets in project,
+    2. Filters the list based on criteria (eg. github username to identify circle datasets)
```
lrwb-aou: Please add a one line summary followed by a blank line.  Then the rest of your doc string is good.

data_steward/notebooks/bq.py
nishanthpp93: There is one instance of bq in analytics (not part of the pipeline) that was not updated to use the utils module. I thought of creating a separate PR for it since it is not part of our pipeline and can cause linting errors which we are not too concerned about.

data_steward/tools/clean_test_project.py
nishanthpp93: Sure, I will add it

data_steward/utils/bq.py
nishanthpp93: Yes, that is a good point

data_steward/bq_utils.py
lrwb-aou: I was referring to the `delete_dataset` function this PR creates in `bq_utils` only.  We do plan to deprecate `bq_utils` completely, in the future.  Right now, I don't think we should be adding functions to `bq_utils`.  Especially if nothing is using that function.  Your script is specifically using the function in the new wrapper module.  So, this is a function that no one is using?.?.?.

nishanthpp93: Ah, my bad! Sorry I was thinking it already existed and was in use.

data_steward/tools/clean_test_project.py
```diff
@@ -0,0 +1,106 @@
...
+        response = input(prompt_text)
+    return response
+
+
+if __name__ == '__main__':
+    parser = argparse.ArgumentParser(
+        description=
+        'Deletes datasets containing specific strings in the dataset_id.',
+        formatter_class=argparse.RawDescriptionHelpFormatter)
+    parser.add_argument('-p',
```
lrwb-aou: One final thing.  This file is called `clean_test_project`.  But it could be applied to any project and its datasets.  Especially because we are allowing the project_id and name substrings to be passed as arguments to the script.  So a more better name might be `clean_project` or `clean_project_datasets`.

nishanthpp93:
Ah, yes! Thanks for reminding me Lauren!

PR #430, opened 2020-02-29
[DC-565] Repopulate person table gender and sex-at-birth post-deid

calbach:
Sending for initial feedback, will add tests after confirming this is directionally correct.

So far: manually verified SQL works via BQ UI.

CC @mark-velez

### DIFF
--- a/tests/integration_tests/data_steward/cdr_cleaner/cleaning_rules/repopulate_person_post_deid_test.py
+++ b/tests/integration_tests/data_steward/cdr_cleaner/cleaning_rules/repopulate_person_post_deid_test.py
@@ -46,36 +46,46 @@ class RepopulatePersonPostDeidTest(unittest.TestCase):
         print(cls.__name__)
         print('**************************************************************')
 
+    def setUp(self):
+        self.project_id = bq_utils.app_identity.get_application_id()
+        self.dataset_id = bq_utils.get_combined_dataset_id()
+        if not self.project_id or not self.dataset_id:
+            # TODO: Fix handling of globals, push these assertions down if they are required.
+            raise ValueError(
+                f"missing configuration for project ('{self.project_id}') " +
+                f"and/or dataset ('{self.dataset_id}')")
+
+        # TODO: Reconcile this with a consistent integration testing model. Ideally each test should
+        # clean up after itself so that we don't need this defensive check.
+        test_util.delete_all_tables(self.dataset_id)
+
+        create_tables = ['person', 'observation']
+        for tbl in ['concept']:
+            if not bq_utils.table_exists(tbl, dataset_id=self.dataset_id):
+                create_tables.append(tbl)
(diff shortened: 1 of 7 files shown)


calbach:
I made some additional changes, added an integration test and updates the person schema JSON file accordingly for the new changes. Please take another look tomorrow - thanks!

@mark-velez as well since we discussed some of this.

tests/integration_tests/data_steward/cdr_cleaner/cleaning_rules/repopulate_person_post_deid_test.py
```diff
@@ -0,0 +1,122 @@
...
+                gender_nonbinary_source_concept_id,
+                sex_at_birth_concept_id=repopulate_person_post_deid.
+                SEX_AT_BIRTH_CONCEPT_ID,
+                sex_female_concept_id=sex_female_concept_id,
+                sex_female_source_concept_id=sex_female_source_concept_id)
+            try:
+                resp = bq_utils.query(query)
+            except HttpError as e:
+                self.fail("failed to execute query '{}': {}".format(
+                    query, e.content))
```
nishanthpp93: Do you think it would be better to add these persons and codes in a setUp() and remove them/delete the tables in a tearDown()?

tests/integration_tests/data_steward/cdr_cleaner/cleaning_rules/repopulate_person_post_deid_test.py
```diff
@@ -0,0 +1,122 @@
...
+        for (k, v) in want.items():
+            self.assertIn(k, person)
+            self.assertEqual(person[k], want[k])
+
+    def test_execute_queries(self):
+        project_id = bq_utils.app_identity.get_application_id()
+        dataset_id = bq_utils.get_combined_dataset_id()
+        self.assertIsNotNone(project_id)
+        self.assertIsNotNone(dataset_id)
+        test_util.delete_all_tables(dataset_id)
```
nishanthpp93: On the same lines, setting these variables could be included as `self.project_id` etc. in setUp() and delete tables could go in tearDown()

calbach: Done

tests/integration_tests/data_steward/cdr_cleaner/cleaning_rules/repopulate_person_post_deid_test.py
calbach: It feels more likely to me that if I wrote a second test case, I'd probably want to use different data rather than the same data. Therefore while I moved table creation to setUp, I prefer to leave the row insertion here.

nishanthpp93:
Looks good @calbach! Thanks for making the changes. It is ready to be merged.

PR #431, opened 2020-03-02
[DC-708] sandbox and remove pids cleaning rule

amcgrenera-vumc:


data_steward/cdr_cleaner/cleaning_rules/sandbox_and_remove_pids.py
```diff
@@ -0,0 +1,109 @@
...
+
+    :param project_id: bq project_id
+    :param dataset_id: bq dataset_id
+    :param pids_query: query from cleaning rule that lists all PIDs that need to be sandboxed and removed
+    :param ticket_number: ticket number from jira that will be appended to the end of the sandbox table names
+    :return: list of CREATE OR REPLACE queries to create tables in sandbox
+    """
+    person_tables_list = get_tables_with_person_id(project_id, dataset_id)
+    # take pids_query and generate list of pids
+    pid_list = bq.query(
```
ChaoPang: It might be a good idea to move `pid_list = bq.query(` out of this function for better separation of function units and easier testing. The purpose of this function is to sandbox all records associated with a list of `person_ids`, but in the current implementation, it does two things (run `pids_query` to dynamically retrieve `person_ids` and sandbox records), we might want to further separate these two basic function units.   

It would be easier for writing tests, for instance, in the integration test for `get_sandbox_queries`, we don't want to bother with the correctness of `pids_query`, we just want to pass in a list of `person_ids`. Other functions should be responsible for running `pids_query`. I would suggest moving `pid_list = bq.query(` to your own cleaning rule. 

Another reason that this might limit us is that we assume`pids_query` is always rendered the same way using `project_id` and `dataset_id` in `pids_query.format(project=project_id,dataset=dataset_id)`, but it may not be the case (e.g. in the cleaning rule I am working on), we don't want to limit ourselves to the way the `pids_query` is rendered and and this function probably should not how the `pids_query` is rendered or constructed.

amcgrenera-vumc: Sounds good, I am going to remove the functionality the `pids_query` and replace the parameter as a list of PIDS that we pass from each separate cleaning rule.

data_steward/cdr_cleaner/cleaning_rules/sandbox_and_remove_pids.py
```diff
@@ -0,0 +1,104 @@
+# Project imports
+from constants.cdr_cleaner import clean_cdr as clean_consts
+import constants.cdr_cleaner.clean_cdr as cdr_consts
+from notebooks import bq
```
ChaoPang: `from notebooks import bq` this import is not correct, you probably need `import bq_utils` and please checkout `bq_utils.query`  for executing queries

tests/unit_tests/data_steward/cdr_cleaner/cleaning_rules/sandbox_and_remove_pids_test.py
```diff
@@ -0,0 +1,101 @@
...
+        ]
+        self.person_table_list = [
+            'observation', 'drug_era', 'observation_period', '_deid_map',
+            'visit_occurrence', 'condition_era', 'measurement', 'person',
+            'procedure_occurrence', 'specimen', 'death', 'device_exposure',
+            'payer_plan_period', '_ehr_consent', 'condition_occurrence',
+            'drug_exposure', 'note', 'dose_era'
+        ]
+
+        mock_bq_query_patcher = patch(
```
ChaoPang: We probably need a different way for mocking the return_value of `bq_util.query`, please check the example https://github.com/all-of-us/curation/blob/3d25aa64d3d76132d0c4289aecd186081e5f023f/tests/unit_tests/data_steward/cdr_cleaner/cleaning_rules/validate_missing_participant_test.py#L163

data_steward/cdr_cleaner/cleaning_rules/sandbox_and_remove_pids.py
amcgrenera-vumc: @ChaoPang this import uses the BigQuery client library. The `bq_utils` is the API, which I think we are eventually moving away from. Another reason I didn't want to use `bq_utils.query`, is because that executes the query and does not return the query results, since I just wanted to get the query results back, I opted for `bq` in notebooks.

tests/unit_tests/data_steward/cdr_cleaner/cleaning_rules/sandbox_and_remove_pids_test.py
amcgrenera-vumc: Since this is calling the client library I needed to call the entire function so that the Client would also be mocked.

PR #432, opened 2020-03-03
[DC-441] Update pylintrc to check for linting errors.

ksdkalluri:


lrwb-aou:
@ksdkalluri 
[deid_patch.txt](https://github.com/all-of-us/curation/files/4294129/deid_patch.txt)

Here is the promised patch file.  I decided to keep the `test_rules.py` and `meta.py` modules.  Neither module is being actively used by deid yet, but could be in the future.  So I fixed the issues with those modules in this patch.

FYI: to apply a patch from file from the command line:

```
cd <wherever your project is>/curation/data_steward
git apply <path to downloaded patch file>
```

then you can activate your python environment and run pylint.

PR #434, opened 2020-03-04
[DC-719] Implement proposal to add HPO site without requiring code changes

nishanthpp93:
* Add function to fetch hpo_id, hpo_name from table
* Change resources.hpo_csv() references to bq_utils.get_hpo_info()
* Remove hpo.csv file and add config folder


nishanthpp93:
Thanks @ksdkalluri. The Codacy issue cannot be fixed because it is a bug in Pylint. Right now they do not consider a mock variable as used because it is not used statically but it is used at runtime.

ksdkalluri:
> Thanks @ksdkalluri. The Codacy issue cannot be fixed because it is a bug in Pylint. Right now they do not consider a mock variable as used because it is not used statically but it is used at runtime.

this can be fixed by simply adding a dummy return value like 
`mock_create_table.return_value = True`

nishanthpp93:
> > Thanks @ksdkalluri. The Codacy issue cannot be fixed because it is a bug in Pylint. Right now they do not consider a mock variable as used because it is not used statically but it is used at runtime.
> 
> this can be fixed by simply adding a dummy return value like
> `mock_create_table.return_value = True`

This is not a fix, it is a workaround because mock_create_table does not return a boolean, but rather job information after creating the table. Using a boolean only serves to obfuscate rather than clarify its purpose. If you can ignore the issue on Codacy, please do. Otherwise it will be ignored after merging.

ksdkalluri:
> > > Thanks @ksdkalluri. The Codacy issue cannot be fixed because it is a bug in Pylint. Right now they do not consider a mock variable as used because it is not used statically but it is used at runtime.
> > 
> > 
> > this can be fixed by simply adding a dummy return value like
> > `mock_create_table.return_value = True`
> 
> This is not a fix, it is a workaround because mock_create_table does not return a boolean, but rather job information after creating the table. Using a boolean only serves to obfuscate rather than clarify its purpose. If you can ignore the issue on Codacy, please do. Otherwise it will be ignored after merging.

yes, I do know it's not a fix but a workaround, but there's no enough access on codacy for us to ignore an issue right now.

PR #436, opened 2020-03-06
[EDQ-383] [EDQ-385] Records Per Capita and E-mail Generator

noahgengel:
* Creates Notebook for Records Per Capita for Each Table/Site Combination

* Streamline the e-mail generator script to cut down on the amount of text and change the resource links.


noahgengel:
@nishanthpp93 Can you merge this PR and delete [this PR](https://github.com/all-of-us/curation/pull/428)? It looks like the original PR for EDQ-383 was never approved. It's a small series of scripts that should not ultimately affect the pipeline.

PR #437, opened 2020-03-09
[DC-698] Fix bug in query string and add unittest to invalid procedure source records

amcgrenera-vumc:


data_steward/cdr_cleaner/cleaning_rules/remove_invalid_procedure_source_records.py
```diff
@@ -16,17 +16,22 @@
...
 import constants.cdr_cleaner.clean_cdr as cdr_consts
+from sandbox import get_sandbox_dataset_id
 
-REMOVE_INVALID_PROCEDURE_SOURCE_CONCEPT_IDS_QUERY = """
-DELETE
+INTERMEDIARY_TABLE_NAME = 'procedure_occurrence_dc583'
+
+INVALID_PROCEDURE_SOURCE_CONCEPT_IDS_QUERY = """
+CREATE OR REPLACE TABLE
+`{project}.{sandbox_dataset}.{intermediary_table}`
```
lrwb-aou: This line needs to be updated to 
```
`{project}.{sandbox_dataset}.{intermediary_table}` AS
```
It's invalid SQL without the `AS`

amcgrenera-vumc: updated!

data_steward/cdr_cleaner/cleaning_rules/remove_invalid_procedure_source_records.py
```diff
@@ -38,12 +43,20 @@
  SELECT
```
lrwb-aou: Lines 41-43 concern me.  The comment says "not in" but the condition being applied is `IN`.  Is this correct?  FYI:  I really don't know.

data_steward/cdr_cleaner/cleaning_rules/remove_invalid_procedure_source_records.py
```diff
@@ -38,12 +43,20 @@
...
 )
 """
 
+VALID_PROCEDURE_SOURCE_CONCEPT_IDS_QUERY = """
+SELECT * FROM
+`{project}.{dataset}.procedure_occurrence`
+WHERE
+procedure_occurrence_id
+IN (SELECT procedure_occurrence_id FROM `{project}.{sandbox_dataset}.{intermediary_table}`)
+"""
```
lrwb-aou: Another question.  Based on the variable name `INVALID_PROCEDURE_SOURCE_CONCEPT_IDS_QUERY`, it seems like the sandbox should be storing the records we don't want to see in the cleaned table.  If that assumption is correct, should line 57 be using a `NOT EXISTS` clause in place of an `IN` clause?

data_steward/cdr_cleaner/cleaning_rules/remove_invalid_procedure_source_records.py
amcgrenera-vumc: this is correct since it's creating the sandbox table. It is writing all the invalid rows to the corresponding sandbox.

data_steward/cdr_cleaner/cleaning_rules/remove_invalid_procedure_source_records.py
amcgrenera-vumc: yep, 100%. this is my bad. changing to NOT EXISTS

amcgrenera-vumc: @lrwb-aou this is fixed, but please double check `NOT EXISTS` this is my first time using, so I'm not sure if it's correct. I would have typically used the following, but as we discussed previously, I think it would exclude all null values as well. 

```
SELECT * FROM
`{project}.{dataset}.procedure_occurrence`
WHERE
procedure_occurrence_id
NOT IN (SELECT procedure_occurrence_id FROM `{project}.{sandbox_dataset}.{intermediary_table}`)
```

data_steward/cdr_cleaner/cleaning_rules/remove_invalid_procedure_source_records.py
```diff
@@ -57,11 +70,25 @@ def get_remove_invalid_procedure_source_queries(project_id, dataset_id):
...
+
+    # queries to delete invalid procedure source records
+    valid_records = dict()
+    valid_records[
+        cdr_consts.QUERY] = VALID_PROCEDURE_SOURCE_CONCEPT_IDS_QUERY.format(
+            project=project_id,
+            dataset=dataset_id,
+            sandbox_dataset=get_sandbox_dataset_id(dataset_id),
+            intermediary_table=INTERMEDIARY_TABLE_NAME)
+    queries_list.append(valid_records)
```
lrwb-aou: Assuming we are using a `SELECT` query to pick the records we want and overwriting the current table, we need to set the dictionary parameters for the destination dataset, destination table, and write disposition.  You can view the ` get_src_concept_id_logging_queries` function in `curation/data_steward/cdr_cleaner/cleaning_rules/replace_standard_id_in_domain_tables.py` for an example.  Otherwise, the query is just selecting and saving to a temporary table.

amcgrenera-vumc: Sorry this is a mess. You're right, this should be `DELETE FROM` but I had `SELECT` for testing purposes and never reverted back in the query. 

So in that case it should then be switched back to `IN`, see below, let me know if that makes sense.

```
DELETE FROM
`{project}.{dataset}.procedure_occurrence`
WHERE
procedure_occurrence_id IN (SELECT procedure_occurrence_id FROM `{project}.{sandbox_dataset}.{intermediary_table}`)
```

data_steward/cdr_cleaner/cleaning_rules/remove_invalid_procedure_source_records.py
lrwb-aou: Unfortunately, this not exists does not work.  I tested by putting three rows in a sandbox table and running the query created to select only the rows you want.  It did not return any results.  I expected 63973754 records.  You can view my test dataset, `lrwb_deid_test_input_20190910`  and its now associated sandbox `lrwb_deid_test_input_20190910_sandbox`.  

This is a working `NOT EXISTS` query using the sandbox dataset.

```
select * from `aou-res-curation-test.lrwb_deid_test_input_20190910.procedure_occurrence` as p0
where not exists (
select p1.procedure_occurrence_id FROM `aou-res-curation-test.lrwb_deid_test_input_20190910_sandbox.procedure_occurrence_dc583` as p1 where p0.procedure_occurrence_id = p1.procedure_occurrence_id)
```
This did return 63973754 results.  There are 63973757 records in the input procedure_occurrence table.  There are 3 records in the sandbox table.

data_steward/cdr_cleaner/cleaning_rules/remove_invalid_procedure_source_records.py
lrwb-aou: Ok.  BigQuery works faster using Select statements.  So If you want to keep the select statement syntax (which I would suggest) you would need to set the extra query parameters to ovewrite the existing table.  Otherwise you can use a Delete statement without the additional query parameters and with the `IN` condition.

amcgrenera-vumc: I've gone ahead and updated the `SELECT` statement with the additional query parameters.

data_steward/cdr_cleaner/cleaning_rules/remove_invalid_procedure_source_records.py
```diff
@@ -57,11 +70,32 @@ def get_remove_invalid_procedure_source_queries(project_id, dataset_id):
...
+            dataset=dataset_id,
+            table=TABLE,
+            sandbox_dataset=get_sandbox_dataset_id(dataset_id),
+            intermediary_table=INTERMEDIARY_TABLE_NAME)
+    queries_list.append({
+        clean_consts.QUERY: valid_records,
+        clean_consts.DESTINATION_TABLE: TABLE,
+        clean_consts.DESTINATION_DATASET: dataset_id,
+        clean_consts.DISPOSITION: bq_consts.WRITE_TRUNCATE
+    })
```
lrwb-aou: This creates this data structure:
```
{
query:       {'query': '\nSELECT * FROM `aou-res-curation-test.lrwb_deid_test_input_20190910.procedure_occurrence` as p0\nWHERE NOT EXISTS (\nSELECT p1.procedure_occurrence_id FROM `aou-res-curation-test.lrwb_deid_test_input_20190910_sandbox.procedure_occurrence_dc583` AS p1\nWHERE p0.procedure_occurrence_id = p1.procedure_occurrence_id)\n'}
destination_table_id:				procedure_occurrence
destination_dataset_id:				lrwb_deid_test_input_20190910
write_disposition:				WRITE_TRUNCATE
}
```

What we want to create is:
```
{
query: 
SELECT * FROM `aou-res-curation-test.lrwb_deid_test_input_20190910.procedure_occurrence` as p0
WHERE NOT EXISTS (
SELECT p1.procedure_occurrence_id FROM `aou-res-curation-test.lrwb_deid_test_input_20190910_sandbox.procedure_occurrence_dc583` AS p1
WHERE p0.procedure_occurrence_id = p1.procedure_occurrence_id)
destination_table_id:				procedure_occurrence
destination_dataset_id:				lrwb_deid_test_input_20190910
write_disposition:				WRITE_TRUNCATE
}
```

This is happening because `valid_records` is a dictionary.  You should add the `DESTINATION_TABLE`, `DESTINATION_DATASET`, and `DISPOSITION`  to the `valid_records` dictionary and then do `queries_list.append(valid_records)`

lrwb-aou: FYI:  The actual query is correct.

amcgrenera-vumc: thank you! just saw this error when updating my unittest, will adjust both and push.

lrwb-aou: great!

amcgrenera-vumc: fixed and pushed.

PR #438, opened 2020-03-10
[DC-468] Conduct ID Validation for sites missing participant match table

ChaoPang:


data_steward/cdr_cleaner/cleaning_rules/validate_missing_participant.py
```diff
@@ -0,0 +1,247 @@
...
+    *,
+    ({criterion_one_expr}) AS criterion_one,
+    ({criterion_two_expr}) AS criterion_two
+FROM `{project_id}.{validation_dataset_id}.{identity_match_table}`
+)
+
+SELECT
+    person_id
+FROM non_match_participants
+WHERE criterion_one IS TRUE OR criterion_two IS TRUE
```
nishanthpp93: nit: It would be more descriptive to use variable names like key_fields_criteria and all_fields_criteria instead of one and two, but this is not a blocker.

data_steward/bq_utils.py
```diff
@@ -124,6 +124,40 @@ def get_validation_results_dataset_id():
...
+
+        if validation_datasets:
+            validation_datasets = sorted(
+                [(dataset['datasetReference']['datasetId'],
+                  int(dataset['creationTime']))
+                 for dataset in validation_datasets],
+                key=lambda x: x[1],
+                reverse=True)
+            dataset_id = validation_datasets[0][0]
+    return dataset_id
```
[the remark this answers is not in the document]
ChaoPang: I think we want to move away from relying on the name of a dataset. For instance, I noticed that a validation set `validation_CDR_20190912` was created probably manually by someone. If you feel strongly about this, I can make the adjustment since this is a temporary solution before we figure out how to link the datasets to each other.

data_steward/cdr_cleaner/clean_cdr.py
```diff
@@ -420,8 +428,15 @@ def clean_combined_dataset(project_id=None, dataset_id=None):
     sandbox_dataset_id = sandbox.create_sandbox_dataset(project_id=project_id,
                                                         dataset_id=dataset_id)
 
+    validation_dataset_id = bq_utils.get_latest_validation_dataset_id(
+        project_id)
+
+    unioned_dataset_id = bq_utils.get_unioned_dataset_id()
```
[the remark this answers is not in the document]
ChaoPang: Not necessarily, we could totally get those directly for sure. I think retrieving those variables inside the cleaning rule seems to be more pythonic than to pass them down. I remember we had a similar discussion about whether we should pass `sandbox_dataset_id` down to the cleaning rules and we agreed to pass it down. But now that we have so many dataset ids that need to be passed, I agree it may be a better idea to access those variables directly. 

However, what's confusing me is that if we can access those variables directly, why pass any variable at all e.g. `project_id` and `dataset_id`?

data_steward/bq_utils.py
```diff
@@ -635,6 +669,20 @@ def list_datasets(project_id):
...
+    Get the dataset information 
+    :param project_id: 
+    :param dataset_id: 
+    :return: 
+    """
+    service = create_service()
+    req = service.datasets().get(projectId=project_id, datasetId=dataset_id)
+    while req:
+        return req.execute()
+    return None
```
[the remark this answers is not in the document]
ChaoPang: I considered using it but thought we would want to migrate to the new google-cloud-bigquery client. I don't think I am up-to-date in terms of the discussion on the migration, are we migrating it incrementally? For instance, I am working on this piece and I should be responsible for refactoring/migrating?

data_steward/bq_utils.py
```diff
@@ -124,6 +124,40 @@ def get_validation_results_dataset_id():
     return dataset_id
 
 
+def is_validation_dataset_id(dataset_id):
+    """
+    Check if  bq_consts.VALIDATION_PREFIX is in the dataset_id
+    :param dataset_id: 
+    :return: a bool indicating whether dataset is a validation_dataset
+    """
+    return bq_consts.VALIDATION_PREFIX in dataset_id
```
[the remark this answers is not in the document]
ChaoPang: This is related to my previous comment. Not all the validation_dataset_ids follow the naming convention, `validation_CDR_20190912` I am not sure if this was created one-off.

data_steward/cdr_cleaner/cleaning_rules/validate_missing_participant.py
ChaoPang: This sounds like a good idea.

data_steward/bq_utils.py
```diff
@@ -124,6 +125,42 @@ def get_validation_results_dataset_id():
...
+
+
+def get_latest_validation_dataset_id(project_id):
+    """
+    Get the latest validation_dataset_id based on most recent creationTime. 
+    :param project_id: 
+    :return: the most recent validatioN_dataset_id
+    """
+
+    pattern = re.compile(bq_consts.VALIDATION_DATASET_REGEX)
```
nishanthpp93: Hey @ChaoPang, I was referring to these. We do not want to add new functions to `bq_utils.py`. Any new bq function has to use `utils/bq.py`

data_steward/constants/bq_utils.py
```diff
@@ -50,7 +50,8 @@
 
 # Dataset name formats
 VALIDATION_DATASET_FORMAT = 'validation_{}'
-VALIDATION_DATASET_REGEX = 'validation_\d{8}'
+VALIDATION_DATASET_REGEX = 'validation_(\d{8})'
```
nishanthpp93: Does the addition of brackets lead to any changes in the function that was previously using this regex?

ChaoPang: This bracket only adds a new group in the pattern match and doesn't affect the pattern match. I will double-check though.

data_steward/bq_utils.py
ChaoPang: I see, I will move those to `utils/bq.py`

nishanthpp93: Thanks for adding these in `utils/bq.py`, @ChaoPang. These can be removed now.

data_steward/utils/bq.py
```diff
@@ -57,3 +60,39 @@ def delete_dataset(project_id,
...
+    if dataset_id == bq_consts.BLANK:
+        validation_datasets = []
+        for dataset in list_datasets(project_id):
+            dataset_id = dataset.dataset_id
+            match = pattern.match(dataset_id)
+            if match:
+                validation_creation_date = datetime.strptime(
+                    match.group(1), bq_consts.VALIDATION_DATE_FORMAT)
+                validation_datasets.append(
+                    (validation_creation_date, dataset_id))
```
nishanthpp93: It looks like we will change the naming convention to exclude dates and use only the CDR release tag in the name. Can you update this to use `dataset.created` instead?

data_steward/utils/bq.py
```diff
@@ -57,3 +69,34 @@ def delete_dataset(project_id,
...
+    :return: the most recent validatioN_dataset_id
+    """
+
+    dataset_id = os.environ.get(bq_consts.MATCH_DATASET, bq_consts.BLANK)
+    if dataset_id == bq_consts.BLANK:
+        validation_datasets = []
+        for dataset in list_datasets(project_id):
+            dataset_id = dataset.dataset_id
+            if is_validation_dataset_id(dataset_id):
+                dataset = get_dataset(project_id, dataset_id)
```
nishanthpp93: Is `get_dataset` necessary? You can use the `dataset` you already have

nishanthpp93: Thanks @ChaoPang, it is a reference to a dataset so this makes sense

PR #440, opened 2020-03-10
[DC-441] Update pylintrc to check for linting errors.

ksdkalluri:


.circleci/config.yml
```diff
@@ -75,14 +75,27 @@ commands:
...
       - test_setup
       - run:
-          name: Check Python lint
+          name: Check Python lint with yapf
           working_directory: ~/curation
           command: tools/yapf_lint.sh
+      - run:
+          name: Pylint on data_steward
+          working _directory: ~/curation
+          command: pylint -E data_steward/
```
mark-velez: Could you please wrap this in a shell script similar to `tools/yapf_lint.sh`? It will allow us to more easily run the same exact command locally and also ensure it could be run from any subdirectory.

.circleci/config.yml
```diff
@@ -75,14 +75,27 @@ commands:
...
+      - run:
+          name: Pylint on tests
+          working_directory: ~/curation
+          command: pylint -E tests/
```
mark-velez: Could this be combined with the previous command e.g. like this?
`pylint -E data_steward tests`

.circleci/config.yml
```diff
@@ -75,14 +75,27 @@ commands:
...
+      - test_teardown
```
mark-velez: Most of the steps in `test_teardown` (handling test and coverage results) are not relevant for linting and will only add noise. I think `save_cache` is still helpful though and we can refactor such that we just do that last.

.circleci/config.yml
```diff
@@ -75,14 +75,27 @@ commands:
           destination: test_results
 
 jobs:
-  unit_test:
+  linting_checks:
     <<: *job_defaults
     steps:
       - test_setup
```
mark-velez: Refactoring this would cut the build time down substantially. Linting does not need all the steps in `test_setup` (e.g. google cloud sdk is not needed).

.circleci/config.yml
ksdkalluri: we can run but I thought it's better if we have a separation.

.circleci/config.yml
mark-velez: We said we would defer wrapping pylint in a script for later

.circleci/config.yml
ksdkalluri: updated

PR #442, opened 2020-03-11
[EDQ-395] Modified the Weekly DQ Checks

noahgengel:
* Changed the name to source_success_rate.csv to make it more informative
* Changed 'No Data' fields to '0' to reflect when a site has no successful 'integration'
* Updated the concept IDs for the ingredient and drug integrations
* Removing the second table metrics script. Going to perform a complete overhaul to ensure that we have a better definition of 'foreign key concept success rate.'
* Changed the datetime fields to date fields for the end dates preceding start dates
* Changed all of the 'succes' to 'success' to ensure consistency and avoid typos
* Changed variables for the number of participants with anomolous birth dates (<18 yo or >120 yo)
* Changed the means of detecting diabetic persons - uses the concept ancestor table now
* Now creates tables with persons with diabetes according to the condition table and the substantiating T2D drugs. Also calculates the number of persons in the persons table who have substantiating drugs
* No longer differentiate between T1 and T2 diabetes
* Now checks to see what proportion of diabetics that have glucose as a measurement. Removed fasting glucose as a measurement dimension
* Added the VA and UA Tuscaloosa to the explicit list of sites
* Finalized the 'diabetics' metrics. Also changed the CSV destination in all of the notebooks to output to the cwd
* Refined definition for the total number of unit/route rows
* Documentation Improvements


data_steward/analytics/table_metrics/weekly_scripts/Table_Metrics_part_1.py
```diff
@@ -70,6 +70,11 @@ def cstr(s, color='black'):
...
 
 print('done.')
+# -
+
+cwd = os.getcwd()
+cwd = str(cwd)
+print(cwd)
 
 # +
 dic = {
```
nishanthpp93: Instead of using this, you might want to look at `bq_utils.get_hpo_info()`

noahgengel: @nishanthpp93 

Are you referring to the cwd variable? This is used for the output file (it just goes to the current directory).

nishanthpp93: I was referring to the hpo dictionary, but this is not a blocker. I'll go ahead and merge.

PR #444, opened 2020-03-12
[DC-529] Remove irrelevant observation_source_concept_ids from the RDR Export

lrwb-aou:
* Contains an application of using the base class to instantiate a rule.
* Adds a unit test of the new cleaning rule.
* adds a cleaning rule to sandbox rows removed and then remove rows where
observation_source_concept_id IN ((43530490, 43528818, 43530333)

### DIFF
--- a/data_steward/cdr_cleaner/clean_cdr.py
+++ b/data_steward/cdr_cleaner/clean_cdr.py
@@ -498,19 +498,31 @@ if __name__ == '__main__':
     parser.add_argument('-s', action='store_true', help='Send logs to console')
     args = parser.parse_args()
     clean_engine.add_console_logging(args.s)
-    if args.data_stage == stage.EHR:
-        clean_ehr_dataset()
-    elif args.data_stage == stage.UNIONED:
-        clean_unioned_ehr_dataset()
-    elif args.data_stage == stage.RDR:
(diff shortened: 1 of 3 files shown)


amcgrenera-vumc:
Everything looks great, I ran locally with my own test data and everything returns as expected; unittests pass locally for me too. I really like the setup and the use of base class.

lrwb-aou:
@amcgrenera-vumc I really only use `SELECT 1` in `NOT EXISTS` subqueries.  It is literally going to select the value 1 when the condition in the subquery is met.  So, in this case, it is going to return 1 when `o.observation_id = n.observation_id AND n.observation_source_concept_id IN (43530490, 43528818, 43530333)`.  1 typically represents True.  The outer query is selecting rows where `NOT EXISTS` , i.e. not true.  Because I don't really care about the records the sub-query is identifying, I can use `SELECT 1` to make sure I don't select them.  I could have also used `SELECT observation_id` or `SELECT *`.   This used to be an SQL optimization.  It may not be a required optimization anymore.

amcgrenera-vumc:
@lrwb-aou thanks for the explanation, was curious on that. Everything else looks great, excited to start using the base class in all cleaning rules!

PR #448, opened 2020-03-13
[DC-735] Create a utility wrapper for the BigQuery's python client library

ksdkalluri:

### DIFF
--- a/data_steward/bq_utils.py
+++ b/data_steward/bq_utils.py
@@ -10,7 +10,6 @@ from datetime import datetime
 import app_identity
 from googleapiclient.discovery import build
 from googleapiclient.errors import HttpError
-from google.cloud import bigquery
 
 # Project imports
 import common
@@ -20,12 +19,6 @@ from constants import bq_utils as bq_consts
 from io import open
 
 socket.setdefaulttimeout(bq_consts.SOCKET_TIMEOUT)
-
-
-def query_to_df(q, use_cache=False):
-    client = bigquery.Client()
-    query_job_config = bigquery.job.QueryJobConfig(use_query_cache=use_cache)
-    return client.query(q, job_config=query_job_config).to_dataframe()
 
 
 class InvalidOperationError(RuntimeError):
--- a/data_steward/utils/bq.py
+++ b/data_steward/utils/bq.py
@@ -1,8 +1,7 @@
 from google.cloud import bigquery
 
-client = bigquery.Client()
-
 
 def query(q, use_cache=False):
+    client = bigquery.Client()
     query_job_config = bigquery.job.QueryJobConfig(use_query_cache=use_cache)
     return client.query(q, job_config=query_job_config).to_dataframe()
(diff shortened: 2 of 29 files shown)


data_steward/bq_utils.py
```diff
@@ -10,7 +10,6 @@
 import app_identity
 from googleapiclient.discovery import build
 from googleapiclient.errors import HttpError
-from google.cloud import bigquery
```
lrwb-aou: Yea!

data_steward/utils/bq.py
```diff
@@ -0,0 +1,8 @@
+from google.cloud import bigquery
+
+client = bigquery.Client()
```
lrwb-aou: This cannot be mocked.  This needs to exist in a function.  preferably something like:

```
def get_client(project_id=None):
    """
    Get a client for a specified project.
    """"
    if project_id:
        return bigquery.Client(project=project_id)

    raise RuntimeError("You must specify a project to create a client.")
```

data_steward/utils/bq.py
```diff
@@ -1,8 +1,7 @@
 from google.cloud import bigquery
 
-client = bigquery.Client()
-
 
 def query(q, use_cache=False):
+    client = bigquery.Client()
```
lrwb-aou: I would suggest changing this to something like `client = get_client(<project_id>)` by defining a `get_client(project_id=None)` function definition.

ksdkalluri: sure will do.

data_steward/utils/bq.py
```diff
@@ -0,0 +1,17 @@
+from google.cloud import bigquery
+
+
+def get_client(project_id=None):
+    """
+    Get a client for a specified project.
+    """
+    if project_id is None:
+        return bigquery.Client()
```
lrwb-aou: I dislike letting the user "forget" to set the project id.  This goes back to CH's concern about overusing environment variables.  I'd rather require a project_id be specified here and raise an error if it isn't.  Then the user has to provide this parameter and can't accidentally use the wrong dataset because they didn't check what the project environment variable is set to.  A developer can grab it from an environment variable if needed, but I'd like to curb this habit.

ksdkalluri: right now I left it because all the notebooks use it, and in notebooks, it only gets the project_id from by using the service account key.  When we update the notebooks we can make this method much stricter.

lrwb-aou: It'll be on my agenda.  They don't do it because we let them.

lrwb-aou: I'll compromise, how about adding a logging statement nagging them to start specifying the project_id before creating the client?
```
if project_id:
    return bigquery.Client(project=project_id)
else:
    log.(naggging statement)
    return bigquery.Client()
```

ksdkalluri: sure, I can add that.

PR #451, opened 2020-03-16
[EDQ-402][EDQ-405] Fixed DQ Metrics Calculations. Separated out notebooks.

noahgengel:
* Changed some column names. 
* Edited the problem with diabetes export file. 
* Changed the 'other' scripts to be more in line with the weekly metrics (in terms of naming conventions.) 
* Got rid of duplicate instances in the different weekly notebooks. 
* Successfully finished updating the weekly scripts. 
* Changed the way in which we generate the site dataframes in the 'other scripts.'


noahgengel:
@nishanthpp93 This is essentially the same as the past PR; it just deals with the `other_scripts` rather than the weekly` scripts. Commits in the comment above.

PR #455, opened 2020-03-18
[EDQ-409] Change the means of calculating 'total rows' for units and routes

noahgengel:
* No major changes - just to ensure consistency for metrics


noahgengel:
@nishanthpp93 Looks like I accidentally missed something on my ticket for [EDQ-402]. I do not anticipate changing these scripts drastically in the near future so I figured I would make the PR before I forgot.

PR #456, opened 2020-03-18
[EDQ-399] Refactored and validated the m_o_t script for the new metrics

noahgengel:
* Fixed the load_files function; FileNotFound error now displays unique message properly.

* Now using the correct interpreter for troubleshooting.

* Improved the valid_cols_total function to ensure that all of the tables without underscores are also logged. Ensures that the 'total' metric on percentage-based metrics is calculated properly.

* Verified for options A, B, and C. Both table and HPO sheets were tested and manually verified.

* Tested and verified option E (for both table and HPO). Assuming D has a similar implementation.

* Started to investigate the errors in the 'G' path of generation. F and G warrant changes to the analytics report script.

* Verified for H -> A and H -> B. This includes the aggregate df for site sheets.

* Tested both pathways for I. Validated including the aggregate df for the site sheets.

* Fixed the load_files function; FileNotFound error now displays unique message properly.

* Now using the correct interpreter for troubleshooting.

* Improved the valid_cols_total function to ensure that all of the tables without underscores are also logged. Ensures that the 'total' metric on percentage-based metrics is calculated properly.

* Verified for options A, B, and C. Both table and HPO sheets were tested and manually verified.

* Reverified A-E, H-I with the newer metrics. Verification not calculation-based but to ensure that no column name changes would affect the script. There were no anticipated changes so this is to be expected.

* Worked on the measurement unit success rate reports. Streamlined to ensure only reporting out percentages. Still need to reprocess how the aggregate_info is calculated. Likely a problem caused by RDR in the earlier datasets.

* Verified F (unit_concept_success_rate) for DQ dimensions 1, 2, 3, and 5. Need to switch to verify 4 with an update to the way we count total measurements and units.

* Added an explicit function: calculate_unit_aggregate_information.

* Verified G (drug routes) for DQ tests 1, 3, 4, and 5. Made minor adjustments to improve the end aggregate sheet.

* Added a calculate_route_aggregate_information function. Used for G (route) dimension #2 (aggregate info on table sheets - manually verified).

* Fixed bug where >1 aggregate information metrics were being added to table sheets (needed elif statements - not just if statements).

* Changed to ensure that the aggregate information sheet was expressed in terms of percentages.

* Removed source concept success rate; not used and the aggregate_info metric was off.

* Tested and validated all 8 for their aggregate information metrics (across their table and HPO sheets.

### DIFF
--- a/data_steward/analytics/table_metrics/metrics_over_time.py
+++ b/data_steward/analytics/table_metrics/metrics_over_time.py
@@ -77,8 +77,7 @@ def get_user_analysis_choice():
         "A. Duplicates\n" \
         "B. Amount of data following death dates\n" \
         "C. Amount of data with end dates preceding start dates\n" \
-        "D. Success rate for source tables\n" \
-        "E. Success rate for concept tables\n" \
+        "D. Success rate for concept_id field\n" \
         "F. Population of the 'unit' field in the measurement table (" \
         "only for specified measurements)\n" \
         "G. Population of the 'route' field in the drug exposure table\n" \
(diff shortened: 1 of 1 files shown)


noahgengel:
@nishanthpp93 Please let me know if you have any questions - PR should bring the m_o_t script up to date with the new metrics.

PR #457, opened 2020-03-19
[EDQ-407] Prevented 'RDR' from Populating; Changed 'No Data' Values to '0'

noahgengel:
* Specifically removed RDR from all of the queries
* Changed fillna to '0' (rather than 'No Data')
* Verified outputs using a combined dataset as a sample


noahgengel:
@nishanthpp93 this *should be* the second to last PR for the week. I successfully updated all of the sites on their DQ so this overhaul is pretty much done.

PR #461, opened 2020-03-23
[DC-690] Implement error handling for copy_files cron jobs in Slack

lrwb-aou:
* refactors sending slack messages into a utility for both services to use.
* refactors unit tests for the admin api to move relevant tests to a slack test case and make needed amends to api test case
* renames unit tests for admin_api service so they are executed automatically
* creates application level error handlers in a separate module using blueprints
* creates tests for the error handler methods and the slack alert utility
* removes unused requirements for flask flat pages and flask restful
* changes EnvironmentErrors to OSErrors because EnvironmentError is deprecated in python 3.7
* adds a message to a silent error raised when finding a bucket fails
* adds additional unit test to ensure contents of error handling methods are firing.


data_steward/gcs_utils.py
```diff
@@ -31,8 +31,12 @@ def get_hpo_bucket(hpo_id):
     # TODO reconsider how to map bucket name
     bucket_env = 'BUCKET_NAME_' + hpo_id.upper()
     hpo_bucket_name = os.getenv(bucket_env)
+
     if hpo_bucket_name is None:
-        raise EnvironmentError()
+        # should not use hpo_id in message if sent to end user.  For now,
+        # only sent to alert messages slack channel.
+        raise OSError('No bucket name defined for hpo_id: {}'.format(hpo_id))
```
lrwb-aou: @mark-velez Previously this was an empty message.  Is there a more generic error message we can report here?  I did say this should only be sent to the slack channel and never to the end user.  Using error handlers should ensure this.  Before checking in, I want your security perspective.

mark-velez: I think principle of least privilege should prevail and your proposal to only send this information to the Slack channel sounds consistent with that. One could argue the service _does not have_ end-users (in the traditional sense), but I believe it makes sense to design conservatively, as though it did. Good thinking.

PR #463, opened 2020-03-26
[EDQ-427] Python package to create 'Data Quality Issue' dashboard

noahgengel:
Package is intended to create unique dashboards that detail all of the current data quality issues for the sites. The main script in this file should ultimately generate Excel files that can easily be distributed to an online project management platform.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -141,4 +141,3 @@ data_steward/analytics/table_metrics/*.xlsx
 # that result from site DQ issues
 data_steward/analytics/tools/dq_issue_tracker/__init__.py
 data_steward/analytics/tools/dq_issue_tracker/*.xlsx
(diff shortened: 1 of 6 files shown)


noahgengel:
@nishanthpp93 Please merge this when you get the chance. 

I do notice that there is an overlapping commit with something that was already merged (#462, which GitHub seems to already recognize). This is my bad; I had to use the command line since I was using Pycharm on macOS (rather than the WindowsOS, which I am more familiar with) so it looks like I messed up on this front. Please let me know if this is an issue and I can re-open a new PR.

nishanthpp93:
Hey @noahgengel, yes it looks like an already existing commit is in the PR. A new PR with only the latest commit would resolve this.

noahgengel:
@nishanthpp93 anger. I created a new PR that only has one commit. My apologies.

PR #468, opened 2020-04-07
[DC-614] Ensure no null or incorrect date times

dcampbell-vumc:
Modified existing ensure_date_datetime_consistency cleaning rule to make sure no null datetimes.


data_steward/cdr_cleaner/cleaning_rules/ensure_date_datetime_consistency.py
```diff
@@ -46,62 +62,92 @@
...
+            query[cdr_consts.DISPOSITION] = bq_consts.WRITE_TRUNCATE
+            query[cdr_consts.DESTINATION_DATASET] = self.get_dataset_id()
+            queries.append(query)
+        return queries
+
+    def setup_rule(self):
+        """
+        Function to run any data upload options before executing a query.
+        """
+        pass
```
lrwb-aou, line 150: @dcampbell-vumc you may want to add

```
        """
        Returns an empty list because this rule does not use sandbox tables.
        """
       return []
```

to your class.  I added this as an abstract method to the base class.  so when the patch for dc-734 gets added, this will be needed.  you can add it now as part of this patch without any bad side effects.

dcampbell-vumc: Okay, I added it.

lrwb-aou: Great!  Thank you!

data_steward/cdr_cleaner/cleaning_rules/ensure_date_datetime_consistency.py
```diff
@@ -46,62 +62,98 @@
...
+        """
+        Initialize the class with proper information.
+
+        Set the issue numbers, description and affected datasets. As other tickets may affect this SQL,
+        append them to the list of Jira Issues.
+        DO NOT REMOVE ORIGINAL JIRA ISSUE NUMBERS!
+        """
+        desc = (
+            'Ensure date and datetime consistency where nulls exist in nullable fields'
+            'by setting date in datetime to the date in date field and time as midnight (00,00,00)'
```
lrwb-aou, line 91: @dcampbell-vumc you should consider updating this description as well.  This description does accurately reflect what your changes to the rule are doing.  But it doesn't reflect the full extent of what this rule is doing.  This description may be used by reporting when we want to describe how the data was cleaned (i.e. may be seen by people who are not developers).

PR #477, opened 2020-04-16
[EDQ-414] Create 'erroneous date' data quality metric notebook

noahgengel:
* Includes a notebook that calculates the percentage of 'erroneous dates' for the following tables: condition_occurrence, procedure_occurrence, visit_occurrence, drug_exposure, measurement, observation. Erroneous dates are those that are prior to 1900 for the observation table and prior to 1980 for all of the other tables.

* Changed the metrics_over_time library to accommodate for this new metric (part of EDQ-425)

* Validated the output for the resulting metrics_over_time script

### DIFF
--- a/data_steward/analytics/table_metrics/metrics_over_time/dictionaries_and_lists.py
+++ b/data_steward/analytics/table_metrics/metrics_over_time/dictionaries_and_lists.py
@@ -70,7 +70,8 @@ thresholds = {
     'unit_success_min': 85,
     'route_success_min': 85,
 
-    'date_datetime_disparity_max': 0
+    'date_datetime_disparity_max': 0,
+    'erroneous_dates_max': 0
 }
 
 
@@ -84,7 +85,8 @@ choice_dict = {
     'g': 'drug_success',
     'h': 'sites_measurement',
     'i': 'visit_date_disparity',
-    'j': 'date_datetime_disparity'}
+    'j': 'date_datetime_disparity',
+    'k': 'erroneous_dates'}
 
 percentage_dict = {
     'duplicates': False,
@@ -96,7 +98,8 @@ percentage_dict = {
     'drug_success': True,
     'sites_measurement': True,
     'visit_date_disparity': True,
-    'date_datetime_disparity': True
+    'date_datetime_disparity': True,
+    'erroneous_dates': True
 }
 
 target_low_dict = {
@@ -109,7 +112,8 @@ target_low_dict = {
     'drug_success': False,
     'sites_measurement': False,
     'visit_date_disparity': False,
-    'date_datetime_disparity': True
+    'date_datetime_disparity': True,
+    'erroneous_dates': True
 }
 
 columns_to_document_for_sheet = {
@@ -153,6 +157,12 @@ columns_to_document_for_sheet = {
         'measurement_success_rate', 'visit_success_rate'],
 
     'date_datetime_disparity': [
+        'visit_occurrence', 'condition_occurrence',
+        'drug_exposure', 'measurement',
+        'procedure_occurrence', 'observation'
+    ],
+
+    'erroneous_dates': [
         'visit_occurrence', 'condition_occurrence',
         'drug_exposure', 'measurement',
         'procedure_occurrence', 'observation'
@@ -213,7 +223,8 @@ data_quality_dimension_dict = {
     'drug_success': 'Completeness',
     'drug_routes': 'Completeness',
     'measurement_units': 'Completeness',
-    'date_datetime_disparity': 'Conformance'
+    'date_datetime_disparity': 'Conformance',
+    'erroneous_dates': 'Plausibility'
 }
 
 metric_type_to_english_dict = {
@@ -232,14 +243,16 @@ metric_type_to_english_dict = {
 
     # other metrics
     'concept': 'Concept ID Success Rate',
-    'duplicates': 'Duplicate Records'
+    'duplicates': 'Duplicate Records',
+    'erroneous_dates': 'Erroneous Dates'
 }
 
 metrics_to_weight = [
     'measurement_units', 'drug_routes',
     'end_before_begin', 'data_after_death',
     'concept', 'duplicates',
-    'date_datetime_disparity']
+    'date_datetime_disparity',
+    'erroneous_dates']
 
 full_names = {
     "saou_uab_selma": "UAB Selma",
--- a/data_steward/analytics/table_metrics/metrics_over_time/hpo_class.py
+++ b/data_steward/analytics/table_metrics/metrics_over_time/hpo_class.py
@@ -34,6 +34,7 @@ class HPO:
             end_before_begin, data_after_death,
             route_success, unit_success, measurement_integration,
             ingredient_integration, date_datetime_disp,
+            erroneous_dates,
 
             # number of rows for the 6 canonical tables
             num_measurement_rows=0,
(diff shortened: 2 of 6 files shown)


noahgengel:
@nishanthpp93 #2 for the three new NIH metrics.

data_steward/analytics/table_metrics/metrics_over_time/hpo_class.py
```diff
@@ -173,6 +180,7 @@ def print_attributes(self):
         \t Measurement Integration: {measurement_integration}\n
         \t Ingredient Integration: {ingredient_integration}\n
         \t Date/Datetime Disagreement: {date_datetime} \n
+        \t Erroneous Dates: {erroneous_dates}\n
```
nishanthpp93, line 183: You typically don't need the extra `\n` since triple quotes print as-is. Also try to look into `f-strings`, we're starting to use them more frequently.

data_steward/analytics/table_metrics/weekly_scripts/erroneous_dates.py
```diff
@@ -0,0 +1,583 @@
...
+    'src_hpo_id': [
+        "saou_uab_selma", "saou_uab_hunt", "saou_tul", "pitt_temple",
+        "saou_lsu", "trans_am_meyers", "trans_am_essentia", "saou_ummc",
+        "seec_miami", "seec_morehouse", "seec_emory", "uamc_banner", "pitt",
+        "nyc_cu", "ipmc_uic", "trans_am_spectrum", "tach_hfhs", "nec_bmc",
+        "cpmc_uci", "nec_phs", "nyc_cornell", "ipmc_nu", "nyc_hh",
+        "ipmc_uchicago", "aouw_mcri", "syhc", "cpmc_ceders", "seec_ufl",
+        "saou_uab", "trans_am_baylor", "cpmc_ucsd", "ecchc", "chci", "aouw_uwh",
+        "cpmc_usc", "hrhc", "ipmc_northshore", "chs", "cpmc_ucsf", "jhchc",
+        "aouw_mcw", "cpmc_ucd", "ipmc_rush", "va", "saou_umc"
```
nishanthpp93, line 82: For these, try to create your own query to fetch data from the table `lookup_tables.hpo_site_id_mappings`

data_steward/analytics/table_metrics/weekly_scripts/erroneous_dates.py
```diff
@@ -0,0 +1,583 @@
...
+  mde.src_hpo_id, COUNT(*) as num_rows
+  FROM
+  `{DATASET}.unioned_ehr_drug_exposure` de
+  JOIN
+  `{DATASET}._mapping_drug_exposure` mde
+  ON
+  de.drug_exposure_id = mde.drug_exposure_id
+  GROUP BY 1
+  ORDER BY num_rows DESC) total
+
```
nishanthpp93, line 464: These look similar for `visit_occurrence` above. Look into Jinja for templating queries, then you can work off of an existing template and substitute visit_occurrence, drug_exposure etc. to generate the desired queries. An example of this is in the notebook `data_steward/analytics/cdr_ops/pop_retract.py`.

data_steward/analytics/table_metrics/weekly_scripts/erroneous_dates.py
```diff
@@ -0,0 +1,583 @@
...
+    SELECT
+            DISTINCT(src_hpo_id) as src_hpo_id
+    FROM
+         `{DATASET}._mapping_care_site`
+         
+    UNION ALL
+    SELECT
+            DISTINCT(src_hpo_id) as src_hpo_id
+    FROM
+         `{DATASET}._mapping_condition_occurrence`
```
nishanthpp93, line 136: Jinja can be extremely helpful to generate this entire query from a 5 line template

data_steward/analytics/table_metrics/weekly_scripts/erroneous_dates.py
```diff
@@ -0,0 +1,583 @@
...
+def cstr(s, color='black'):
+    return "<text style=color:{}>{}</text>".format(color, s)
+
+
+print('done.')
+# -
+
+cwd = os.getcwd()
+cwd = str(cwd)
+print("Current working directory is: {cwd}".format(cwd=cwd))
```
nishanthpp93, line 69: with f-strings this would be
```suggestion
print(f"Current working directory is: {cwd}")
```

data_steward/analytics/table_metrics/metrics_over_time/dictionaries_and_lists.py
```diff
@@ -96,7 +98,8 @@
     'drug_success': True,
     'sites_measurement': True,
     'visit_date_disparity': True,
-    'date_datetime_disparity': True
+    'date_datetime_disparity': True,
```
nishanthpp93, line 101: It would be good to have both of these in a constants file as well, like 
```suggestion
    consts.DATE_TIME_DISPARITY: consts.TRUE,
```

data_steward/analytics/table_metrics/metrics_over_time/dictionaries_and_lists.py
```diff
@@ -213,7 +223,8 @@
     'drug_success': 'Completeness',
     'drug_routes': 'Completeness',
     'measurement_units': 'Completeness',
```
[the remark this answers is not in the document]
noahgengel: Good idea; I think I'll make a separate EDQ ticket with some of the changes that you are suggesting.

data_steward/analytics/table_metrics/metrics_over_time/hpo_class.py
noahgengel: Also a good idea. I'll be sure to remove the \n; didn't actually think critically about the implications of triple quotes.

data_steward/analytics/table_metrics/weekly_scripts/erroneous_dates.py
noahgengel: This script does not leverage Google bigquery in any sense (I don't even run it in Jupyter noetbooks). I could hypothetically refactor it accordingly but that would fall in line with EDQ-438, which is 'low priority'.

data_steward/analytics/table_metrics/weekly_scripts/erroneous_dates.py
noahgengel: I see - I take it that it implicitly places in the variable that is in the {}?

noahgengel:
NOTE: Nishanth's suggested edits are encapsulated in EDQ-440. This will be used in future iterations of the metrics_over_time script.

nishanthpp93:
Sounds good, I will merge as-is for now.

data_steward/analytics/table_metrics/weekly_scripts/erroneous_dates.py
noahgengel: @nishanthpp93 

Point to note: I intentionally refrain from creating 'general' queries and instead separate them out by table because it enables sites to more easily copy and paste our queries directly (and therefore decrease the amount of emails directed to the DRC). I see, however, how the shorter queries makes for a more efficient/legible notebook as a whole.

Do you think it would be worthwhile to 'consolidate' the queries even if doing so may ultimately confuse the HPOs?

nishanthpp93: I see. I would then restrict using general queries for your own use. If a site uses it, you can keep it as-is, or maybe use f-strings to make them more readable.

PR #479, opened 2020-04-17
[DC-757] Create functionality to tag a pipeline dataset

dcampbell-vumc:
Modified bq.py by adding create_dataset function. Created bq_test.py to test create_dataset functionality.


data_steward/utils/bq.py
```diff
@@ -240,3 +242,43 @@ def get_latest_validation_dataset_id(project_id):
...
+        raise RuntimeError("Cannot create a dataset without a name")
+
+    if not description:
+        raise RuntimeError("Please provide a description to create a dataset")
+
+    if not label or not tag:
+        raise RuntimeError("Label and/or tag is required to create a dataset")
+
+    client = get_client(project_id)
+    client.create_dataset(dataset_id, exists_ok=exists_ok)
```
lrwb-aou, line 282: This line creates a dataset with an id.  It doesn't apply any labels, tags, or descriptions.  Have you looked at incorporating the `define_dataset` function?

dcampbell-vumc: Yes, I will look at the `define_dataset` function, it was suggested to me to follow the `delete_dataset` function at first. I will pivot.

lrwb-aou: Well, deleting is easy.  It may not even need a wrapper function defined here.

data_steward/utils/bq.py
```diff
@@ -211,19 +211,21 @@ def delete_dataset(project_id,
 
 def is_validation_dataset_id(dataset_id):
     """
-    Checks if dataset_id is a validation dataset
+    Check if  bq_consts.VALIDATION_PREFIX is in the dataset_id
+
+    :param dataset_id:
 
-    :param dataset_id: identifies the dataset
```
lrwb-aou, line 216: This patch is removing valid documentation strings.  Did you mean to do this?

dcampbell-vumc: Can you elaborate?

lrwb-aou: your patch removed line 216 in pink which at least has a vague description of the dataset_id parameter.  it adds `:param dataset_id:`, with no description at all.  Did you mean to remove this description?

dcampbell-vumc: No, I didn't mean to remove it. However, since it was initially vague, should I change it to be more descriptive?

data_steward/utils/bq.py
```diff
@@ -240,3 +242,70 @@ def get_latest_validation_dataset_id(project_id):
...
+        dataset = bigquery.Dataset(dataset_id)
+
+        try:
+            client.get_dataset(dataset)
+            return True
+        except NotFound:
+            return False
+
+
+def create_dataset(dataset_id, description, label_or_tag, project_id):
```
ksdkalluri, line 275: I think we should maintain the distinction between labels and tags.

data_steward/utils/bq.py
```diff
@@ -240,3 +242,70 @@ def get_latest_validation_dataset_id(project_id):
...
+    :raises: RuntimeError if the dataset does not have dataset_id
+    :raises: RuntimeError if the dataset does not have a description
+    :raises: RuntimeError if the dataset does not have a label
+    """
+    if dataset_exists(dataset_id, project_id) == False:
+        if description.isspace() or not description:
+            raise RuntimeError(
+                "Please provide a description to create a dataset")
+
+        if not label_or_tag:
```
ksdkalluri, line 294: when the label and tag are separated. I think it is better to check if either one of them both is specified. Not all the datasets might have both labels and tags.

data_steward/utils/bq.py
```diff
@@ -240,3 +242,84 @@ def get_latest_validation_dataset_id(project_id):
...
+
+        if label is None:
+            dataset.labels = tag
+        elif tag is None:
+            dataset.labels = label
+        else:
+            label_and_tag = {}
+
+            for item in [label, tag]:
+                label_and_tag.update(item)
```
lrwb-aou, line 315: Conceptually, is there a difference in a label and a tag?  I thought a tag was a key with a blank value (empty string).

data_steward/utils/bq.py
```diff
@@ -240,3 +242,84 @@ def get_latest_validation_dataset_id(project_id):
...
+    else:
+        client = get_client(project_id)
+        dataset_id = f"{project_id}.{dataset_id}"
+        dataset = bigquery.Dataset(dataset_id)
+
+        try:
+            client.get_dataset(dataset)
+            return True
+        except NotFound:
+            return False
```
lrwb-aou, line 272: What is this providing that `create_dataset(dataset_ref, exists_ok=True/False)` doesn't?

data_steward/utils/bq.py
dcampbell-vumc: Yes, it is so conceptually, there is no difference between the two. However, @ksdkalluri suggested to separate the two to check if either one or both are specified.

data_steward/utils/bq.py
lrwb-aou: That's not exactly what this parameter is doing.  It is not validating the dataset.  It is checking if the validation prefix is in the dataset name.

data_steward/utils/bq.py
dcampbell-vumc: Nothing, it was suggested by @ksdkalluri to create this function for reasons I can't quite remember. @ksdkalluri can you elaborate on why I was to create this function instead of relying on the built in check in the client library `create_dataset`

data_steward/utils/bq.py
dcampbell-vumc: I changed the description of `dataset_id` to say `ID (name) of the dataset to validate` should I change it again?

data_steward/utils/bq.py
dcampbell-vumc: However, I have now added a check to make sure that `client` has been specified or else a RunTimeError is returned.

data_steward/utils/bq.py
```diff
@@ -240,3 +242,84 @@ def get_latest_validation_dataset_id(project_id):
...
+
+            for item in [label, tag]:
+                label_and_tag.update(item)
+
+            dataset.labels = label_and_tag
+
+        client = get_client(project_id)
+        client.create_dataset(dataset, exists_ok=False)
+
+        return client
```
lrwb-aou, line 322: why return the client?

data_steward/utils/bq.py
dcampbell-vumc: This is how I had it before as a parameter called `label_or_tag`, @ksdkalluri will have to elaborate on his reasoning.

data_steward/utils/bq.py
dcampbell-vumc: Should it be dataset instead? Or do I even need to return anything?

lrwb-aou: For this, you wouldn't return anything.

data_steward/utils/bq.py
ksdkalluri: @lrwb-aou I asked to make the distinction so that a developer passing the arguments(label/key) will get to read the distinction b/w both. If that's too lenient we can still make it one argument and have doc string specify the distinction.

data_steward/utils/bq.py
ksdkalluri: since the dataset is being created inside the function we might not need to return anything. But as a best practice, I would suggest you return Dataset instead.

data_steward/utils/bq.py
ksdkalluri: @lrwb-aou @dcampbell-vumc sorry, I have missed the `exists_ok=False` in the `create_dataset()`. If that's available in the `create_dataset` function this function will be obsolete.

data_steward/utils/bq.py
dcampbell-vumc: @lrwb-aou and @ksdkalluri, which would be best in this case?

data_steward/utils/bq.py
lrwb-aou: @ksdkalluri .  this does exist and, if not explicitly provided, defaults to `exists_ok=False`

lrwb-aou:
From the point of view of our meeting Friday, I would remove the functions `create_dataset` and `dataset_exists`.  I would alter the `define_dataset` function to meet the dataset criteria defined in 757.  Then the responsibility is in each module to create a proper client, use the function you create/alter/modify to define a dataset, and use the client library `create_dataset` function.

Also, your title needs to use the hyphen between the DC and 757.  Without the hyphen, it isn't linking back to Jira correctly.  It should be listed as a PR link and is not.
Screen Shot 2020-04-27 at 12 20 45 PM

data_steward/utils/bq.py
ksdkalluri: @dcampbell-vumc Since Lauren said we need not return anything. let's follow that.

lrwb-aou:
I think it makes sense based on our discussion Friday.  What do you two think?  @dcampbell-vumc and @ksdkalluri ?

ksdkalluri:
> I think it makes sense based on our discussion Friday. What do you two think? @dcampbell-vumc and @ksdkalluri ?

I agree.

data_steward/utils/bq.py
```diff
@@ -155,7 +155,7 @@ def get_dataset(project_id, dataset_id):
     return client.get_dataset(dataset_id)
 
 
-def define_dataset(project_id, dataset_id, description, labels):
+def define_dataset(project_id, dataset_id, description, label_or_tag):
```
ksdkalluri, line 158: please add `label_or_tag` param to doc_string and please explain the label/tag distinction.

tests/integration_tests/data_steward/utils/bq_test.py
```diff
@@ -0,0 +1,61 @@
...
+                          self.dataset_id, self.label_or_tag)
+
+        # Tests if no label or tag is given
+        self.assertRaises(TypeError, define_dataset, self.project_id,
+                          self.dataset_id, self.description)
+
+        # Pre-conditions
+        results = define_dataset(self.project_id, self.dataset_id,
+                                 self.description, self.label_or_tag)
+
```
ksdkalluri, line 59: we can also add a check comparing if results.label == self. label_or_tag

ksdkalluri:
@dcampbell-vumc The changes look good to me. @lrwb-aou If you take a final look at this and think it's good, I will approve the pull request.

lrwb-aou:
> @dcampbell-vumc The changes look good to me. @lrwb-aou If you take a final look at this and think it's good, I will approve the pull request.

I think it looks good.

dcampbell-vumc:
> > @dcampbell-vumc The changes look good to me. @lrwb-aou If you take a final look at this and think it's good, I will approve the pull request.
> 
> I think it looks good.

Great! I will squash and merge now.

PR #482, opened 2020-04-20
[DC-764] Build a tool to create _deid_map for all combined datasets

amcgrenera-vumc:
…rresponding deid datasets

### DIFF
--- a/data_steward/retraction/create_deid_map.py
+++ b/data_steward/retraction/create_deid_map.py
@@ -36,7 +36,7 @@ CREATE_DEID_MAP_TABLE_QUERY = """
 CREATE OR REPLACE TABLE `{project}.{dataset}._deid_map` AS (
 SELECT c.person_id, d.person_id AS research_id
 FROM `{project}.{dataset}.observation` c
-JOIN `{project}.{dataset}.observation` d
+FULL OUTER JOIN `{project}.{dataset}.observation` d
 ON c.observation_id = d.observation_id
 )
 """
@@ -104,6 +104,8 @@ def check_if_deid_map_exists(project_id, dataset):
     if 'deid_map' in column_list:
         return 'rename required'
     if '_deid_map' in column_list:
+        return True
+    if 'deid_map' and '_deid_map' in column_list:
         return True
     if ['deid_map', '_deid_map'] not in column_list:
         return False
--- a/data_steward/retraction/create_deid_map.py
+++ b/data_steward/retraction/create_deid_map.py
@@ -168,5 +168,5 @@ if __name__ == '__main__':
                         required=True)
     args = parser.parse_args()
     query_list = create_deid_map_table_queries(args.project_id)
-    # run_queries(query_list)
+    run_queries(query_list)
     logging.info('Creation of _deid_maps complete')
--- a/data_steward/retraction/create_deid_map.py
+++ b/data_steward/retraction/create_deid_map.py
@@ -1,0 +1,180 @@
+"""
+Background
+
+DC-764 - Rename or create _deid_map for datasets without
+
+Currently, some older datasets do not contain the _deid_map table in the combined dataset. This is needed for
+retractions to obtain the correct research_id for the deid datasets.
+
+Create a module in the retraction package that generates all _deid_map tables for each release in the combined dataset
+if there is an associated deid_dataset. This will be created by querying the observation table in combined and joining
+to the observation table in deid on observation_id. Before the table creation, check to see if _deid_map or deid_map
+exists, if deid_map exists rename for consistency to _deid_map.
+"""
+
+# Python imports
+import re
+import argparse
+import logging
+
+# Third party imports
+
+# Project imports
+import utils.bq
+import bq_utils
+from retraction.retract_utils import DEID_REGEX
+
+CURRENT_RELEASE_REGEX = re.compile('R\d{4}q\dr\d')
+
+RENAME_DEID_MAP_TABLE_QUERY = """
+CREATE OR REPLACE TABLE `{project}.{dataset}._deid_map` AS (
+SELECT *
+FROM `{project}.{dataset}.deid_map`
+)
+"""
+CREATE_DEID_MAP_TABLE_QUERY = """
+CREATE OR REPLACE TABLE `{project}.{dataset}._deid_map` AS (
+SELECT c.person_id, d.person_id AS research_id
+FROM `{project}.{dataset}.observation` c
+JOIN `{project}.{dataset}.observation` d
+ON c.observation_id = d.observation_id
+)
+"""
+
+
+def get_combined_datasets_for_deid_map(project_id):
+    """
+    List all datasets in given project_id, filter out datasets that contain a deid dataset and collect the corresponding
+    combined dataset
+    :param project_id: bq name of project_id
+    :return: list of combined_datasets that should contain a _deid_map table
+    """
+    all_datasets_obj = utils.bq.list_datasets(project_id)
+    all_datasets = [d.dataset_id for d in all_datasets_obj]
+    deid_datasets = []
+
+    # Keep only deid datasets with standard cdr release naming convention
+    release_datasets = [
+        d for d in all_datasets if CURRENT_RELEASE_REGEX.match(d)
+    ]
+    # Filter for deid datasets
+    for dataset in release_datasets:
+        if bool(re.match(DEID_REGEX, dataset)) is True:
+            deid_datasets.append(dataset)
+    # Find corresponding combined dataset for each deid dataset
+    combined_datasets = get_corresponding_combined_dataset(
+        all_datasets, deid_datasets)
+    return combined_datasets
+
+
+def get_corresponding_combined_dataset(all_datasets, deid_datasets):
+    """
+    Takes in all datasets in a project and a list of all the deid datasets in that project and returns the corresponding
+    combined dataset to that deid dataset
+    :param all_datasets: list of all datasets in the given bq project
+    :param deid_datasets: list of all deid datasets in the given bq project
+    :return: list of combine datasets
+    """
+    combined_datasets = []
+    # Get release from deid dataset for both older and current release tags and search for combined dataset
+    for d in deid_datasets:
+        release = d.split('_')[0]
+        older_release = release[1:]
+        combined, older_combined = release + '_combined', older_release + '_combined'
+
+        if combined in all_datasets:
+            combined_datasets.append(combined)
+        if older_combined in all_datasets:
+            combined_datasets.append(older_combined)
+    # Remove duplicates
+    combined_datasets = list(set(combined_datasets))
+    return combined_datasets
+
+
+def check_if_deid_map_exists(project_id, dataset):
+    """
+    Checks if _deid_map table exists in the given dataset.
+    :param project_id: bq name of project_id
+    :param dataset: bq name of dataset
+    :return: returns True, False or 'rename required' which had 'deid_map' table instead of '_deid_map'
+    """
+    table_info_df = utils.bq.get_table_info_for_dataset(project_id, dataset)
+    column_list = table_info_df['table_name'].tolist()
+
+    if 'deid_map' in column_list:
+        return 'rename required'
+    if '_deid_map' in column_list:
+        return True
+    if ['deid_map', '_deid_map'] not in column_list:
+        return False
+
+
+def rename_deid_map_table_query(project_id, dataset):
+    return RENAME_DEID_MAP_TABLE_QUERY.format(project=project_id,
+                                              dataset=dataset)
+
+
+def create_deid_map_table_query(project, dataset):
+    return CREATE_DEID_MAP_TABLE_QUERY.format(project=project, dataset=dataset)
+
+
+def create_deid_map_table_queries(project):
+    """
+    Creates a query list to run to create or rename _deid_map tables in each combined dataset that has a deid dataset
+    :param project: bq name of project_id
+    :return: list of queries to run
+    """
+    combined_datasets = get_combined_datasets_for_deid_map(project)
+    queries = []
+    for dataset in combined_datasets:
+        check = check_if_deid_map_exists(project, dataset)
+        if check is True:
+            continue
+        if check is False:
+            queries.append(create_deid_map_table_query(project, dataset))
+        if check == 'rename required':
+            queries.append(rename_deid_map_table_query(project, dataset))
+    return queries
+
+
+def run_queries(queries):
+    """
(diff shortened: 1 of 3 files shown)


tests/unit_tests/data_steward/retraction/create_deid_map_test.py
```diff
@@ -0,0 +1,125 @@
...
+        ]
+
+    @mock.patch('retraction.create_deid_map.utils.bq.list_datasets')
+    def test_get_combined_datasets_for_deid_map(self, mock_list_datasets):
+        result = create_deid_map.get_combined_datasets_for_deid_map(
+            self.project_id)
+
+        expected_all_datasets = mock_list_datasets.return_value = utils.bq.list_datasets(
+            self.project_id)
+        expected_deid_datasets = self.deid_datasets
```
nishanthpp93, line 40: It would be good to see more values here. I cannot see the combined datasets being generated. I would suggest all_datasets to be combined, deid and other types of datasets, and the assert checks if combined datasets are output as expected. 
From my other suggestion, it would also help to see a mapping so we can easily tell which deid datasets have a combined and which don't. The test case could have unmapped deid and combined datasets.

tests/unit_tests/data_steward/retraction/create_deid_map_test.py
```diff
@@ -0,0 +1,125 @@
...
+            release = d.split('_')[0]
+            older_release = release[1:]
+            combined, older_combined = release + '_combined', older_release + '_combined'
+
+            if combined in self.all_datasets:
+                expected.append(combined)
+            if older_combined in self.all_datasets:
+                expected.append(older_combined)
+            # Remove duplicates
+        expected = list(set(expected))
```
nishanthpp93, line 61: This looks like the same logic as the function itself. This can cause the test to miss possible errors in the function since they will always be in agreement. I would suggest maybe hardcoding the expected combined datasets for now and checking if the function generates what you expect.

tests/unit_tests/data_steward/retraction/create_deid_map_test.py
```diff
@@ -0,0 +1,125 @@
...
+
+        expected_table_info_df = mock_get_table_info.return_value = pd.DataFrame(
+            columns=[
+                'table_catalog', 'table_schema', 'table_name', 'column_name',
+                'ordinal_position', 'is_nullable', 'data_type', 'is_generated',
+                'generation_expression', 'is_stored', 'is_hidden',
+                'is_updatable', 'is_system_defined', 'is_partitioning_column',
+                'clustering_ordinal_position'
+            ])
+        expected_column_list = expected_table_info_df['table_name'].tolist()
```
nishanthpp93, line 78: You don't have to define all the columns here. The only column of interest is `table_name`, so the data frame could be instantiated like 
```suggestion
        expected_table_info_df = mock_get_table_info.return_value = pd.DataFrame(
            data={'table_name': ['_deid_map', '_ehr_consent', 'deid_map', common.PERSON, common.OBSERVATION]})
        expected_column_list = expected_table_info_df['table_name'].tolist()
```

tests/unit_tests/data_steward/retraction/create_deid_map_test.py
```diff
@@ -0,0 +1,125 @@
...
+        expected = create_deid_map.CREATE_DEID_MAP_TABLE_QUERY.format(
+            project=self.project_id, dataset=self.dataset_id)
+        self.assertEquals(result, expected)
+
+    @mock.patch('retraction.create_deid_map.utils.bq.list_datasets')
+    def test_create_deid_map_table_queries(self, mock_list_datasets):
+        result = create_deid_map.create_deid_map_table_queries(self.project_id)
+        mock_list_datasets.return_value = utils.bq.list_datasets(
+            self.project_id)
+
```
nishanthpp93, line 106: You could supply this with a list like 
```suggestion
        mock_list_datasets.return_value = self.all_datasets
```
and see what queries it generates. If it is as expected depending on the existence of combined datasets for each deid datasets, it is successful. Also, you might want to mock `get_table_info_for_dataset` such that it returns a data frame with tables as in the test above. Since it will be called as many times as the number of combined datasets with corresponding deid datasets, you could return it different dataframes based on the call count. This can be achieved by something like
```suggestion
        mock_list_datasets.return_value = self.all_datasets
        mock_table_info.side_effect = [dataframe_1, dataframe_2, dataframe_3]
```
where one could have no deid map, another could have `deid_map` and another `_deid_map`.

data_steward/retraction/create_deid_map.py
```diff
@@ -0,0 +1,180 @@
...
+CREATE OR REPLACE TABLE `{project}.{dataset}._deid_map` AS (
+SELECT *
+FROM `{project}.{dataset}.deid_map`
+)
+"""
+CREATE_DEID_MAP_TABLE_QUERY = """
+CREATE OR REPLACE TABLE `{project}.{dataset}._deid_map` AS (
+SELECT c.person_id, d.person_id AS research_id
+FROM `{project}.{dataset}.observation` c
+JOIN `{project}.{dataset}.observation` d
```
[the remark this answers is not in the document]
amcgrenera-vumc: Good call, I will update this!

data_steward/retraction/create_deid_map.py
```diff
@@ -0,0 +1,180 @@
...
+    """
+    table_info_df = utils.bq.get_table_info_for_dataset(project_id, dataset)
+    column_list = table_info_df['table_name'].tolist()
+
+    if 'deid_map' in column_list:
+        return 'rename required'
+    if '_deid_map' in column_list:
+        return True
+    if ['deid_map', '_deid_map'] not in column_list:
+        return False
```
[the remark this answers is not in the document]
amcgrenera-vumc: I initially used all `if`'s because I wanted it to check all possibilities, which now I'm blanking on why, but with adding if 'deid_map' and '_deid_map' both exist I will need to keep the `if`'s. I will make this adjustment!

data_steward/retraction/create_deid_map.py
```diff
@@ -0,0 +1,180 @@
...
+    return combined_datasets
+
+
+def get_corresponding_combined_dataset(all_datasets, deid_datasets):
+    """
+    Takes in all datasets in a project and a list of all the deid datasets in that project and returns the corresponding
+    combined dataset to that deid dataset
+    :param all_datasets: list of all datasets in the given bq project
+    :param deid_datasets: list of all deid datasets in the given bq project
+    :return: list of combine datasets
```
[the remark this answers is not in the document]
amcgrenera-vumc: Do you think it should be a dataframe with two columns, one for the combined dataset and the other for the deid dataset? Or how are in envisioning this?

tests/unit_tests/data_steward/retraction/create_deid_map_test.py
amcgrenera-vumc: will make this change

tests/unit_tests/data_steward/retraction/create_deid_map_test.py
amcgrenera-vumc: will update this, thanks!

tests/unit_tests/data_steward/retraction/create_deid_map_test.py
amcgrenera-vumc: thank you, I think I've updated this correctly!

amcgrenera-vumc:
I have made adjustments from your notes, all except for your notes on `get_corresponding_combined_dataset` and the unit test for that. I would like some more clarification, but in the mean time please take a look and my changes and let me know if that is what you were thinking!

data_steward/retraction/create_deid_map.py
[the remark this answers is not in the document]
nishanthpp93: I think a dictionary with each deid having the corresponding combined as the value should suffice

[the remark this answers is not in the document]
amcgrenera-vumc: Ok great, I'll make this adjustment and update the unit test and let you know when i've committed.

data_steward/retraction/create_deid_map.py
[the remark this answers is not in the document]
nishanthpp93: You could change the order to make this possible: If _deid exists return 'skip', elif deid exists return 'rename', else return 'create'. If both exist, we don't have to worry here since it would be skipped.
Booleans might be confusing since three possibilities exist here. It would help to use constants similar to enum types, like `consts.RENAME = 'rename', consts.SKIP = 'skip', consts.CREATE = 'create'` and returning them for each of these conditions.

tests/unit_tests/data_steward/retraction/create_deid_map_test.py
nishanthpp93: I saw the changes and they make sense. However while you are checking for the assertEquals, you don't need any `if` statements since the conditions are never met for others (no deid_map tables exist in the table name df). It could simply be `self.assertEquals(result, 'CREATE')` and that would be it. 
To test for all the conditions, you'll need to create new dfs with table names having `_deid_map` or `deid_map` or both and use them in side_effects for mock_table info, testing `self.assertEquals(result, action)` separately for each df.

nishanthpp93:
Thanks Aimee, the updates look good. I've added a couple more comments. Let me know if they clarify things.

data_steward/retraction/create_deid_map.py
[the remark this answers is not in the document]
amcgrenera-vumc: Ah I see, thank you! I will create a constants file like I see you've done previously with other retraction files and use those for that if statement.

data_steward/retraction/create_deid_map.py
```diff
@@ -0,0 +1,172 @@
...
+
+def rename_deid_map_table_query(project_id, dataset):
+    return consts.RENAME_DEID_MAP_TABLE_QUERY.format(project=project_id,
+                                                     dataset=dataset)
+
+
+def create_deid_map_table_query(project, dataset):
+    return consts.CREATE_DEID_MAP_TABLE_QUERY.format(project=project,
+                                                     dataset=dataset)
+
```
nishanthpp93, line 108: These functions and their tests can be removed since they're not really doing anything much

data_steward/retraction/create_deid_map.py
```diff
@@ -0,0 +1,172 @@
...
+        combined = prefix + '_combined'
+
+        if combined in all_datasets:
+            new_row = pd.Series({
+                'deid_dataset': d,
+                'combined_dataset': combined
+            })
+            deid_and_combined_datasets_df = deid_and_combined_datasets_df.append(
+                new_row, ignore_index=True)
+
```
nishanthpp93, line 78: It would be informative to have a log message here if a combined dataset is not found for a particular deid dataset.

data_steward/retraction/create_deid_map.py
```diff
@@ -0,0 +1,172 @@
...
+    :param queries: list of queries
+    """
+    query_job_ids = []
+    for query_dict in queries:
+        logging.info('Creating or renaming _deid_map table with query: %s' %
+                     (query_dict['query']))
+        job_results = bq_utils.query(q=query_dict['query'], batch=True)
+        rows_affected = job_results['numDmlAffectedRows']
+        logging.info('%s rows written to _deid_map table' % rows_affected)
```
nishanthpp93, line 144: I'm not sure if `numDmlAffectedRows` is returned when a `CREATE OR REPLACE TABLE` is used, let me know if you've tested it

tests/unit_tests/data_steward/retraction/create_deid_map_test.py
```diff
@@ -0,0 +1,173 @@
...
+                'table_name':
+                    ['_ehr_consent', common.PERSON, common.OBSERVATION]
+            })
+        expected_create_column_list = expected_create_df['table_name'].tolist()
+        if '_deid_map' in expected_create_column_list:
+            expected_create_return = consts.SKIP
+        elif 'deid_map' in expected_create_column_list:
+            expected_create_return = consts.RENAME
+        else:
+            expected_create_return = consts.CREATE
```
nishanthpp93, line 81: Lol I think you might need to call the function `check_if_deid_map_exists` here to test it. Also you can get rid of all the `if` statements, since the function itself should be doing that, and replace them with just `self.assertEqual(result, consts.CREATE)` etc. since you know what to expect based on the input table names you supplied.

data_steward/retraction/create_deid_map.py
```diff
@@ -0,0 +1,172 @@
...
+    release_datasets = [
+        d for d in all_datasets if consts.CURRENT_RELEASE_REGEX.match(d)
+    ]
+    # Filter for deid datasets
+    for dataset in release_datasets:
+        if bool(re.match(DEID_REGEX, dataset)) is True:
+            deid_datasets.append(dataset)
+    # Find corresponding combined dataset for each deid dataset
+    deid_and_combined_df = get_corresponding_combined_dataset(
+        all_datasets, deid_datasets)
```
nishanthpp93, line 51: I think it would be helpful to log rows in the resulting dataframe here

data_steward/retraction/create_deid_map.py
amcgrenera-vumc: I'm using these functions in `create_deid_map_table_queries` should I just move the logic down there and remove the functions?

data_steward/retraction/create_deid_map.py
amcgrenera-vumc: added!

data_steward/retraction/create_deid_map.py
amcgrenera-vumc: added!

data_steward/retraction/create_deid_map.py
amcgrenera-vumc: you're correct, this is removed.

tests/unit_tests/data_steward/retraction/create_deid_map_test.py
amcgrenera-vumc: lol oops! would I need to create multiple results then for each assertEqual?

data_steward/retraction/create_deid_map.py
nishanthpp93: Yeah that should suffice

tests/unit_tests/data_steward/retraction/create_deid_map_test.py
nishanthpp93: Not really, one should suffice

amcgrenera-vumc:
@nishanthpp93 I've made adjustments, please take a look when you get a chance!

PR #485, opened 2020-04-21
[DC-756] Update RDR and Unioned_EHR cleaning rules lists

lrwb-aou:
* adds missing ObservationSourceConceptIDRowSuppression class to RDR rules
* restores commented invalid_procedure_source.get_remove_invalid_procedure_source_queries function
* properly formats CleanMappingExtTables class as a tuple
* restores commented classes that cause sql listing failures.  comments are maintained to indicate changes are needed.
* adds comments above CleanMappingExtTables usage identifying why it breaks on going reporting capabilities.  this should be fixed in a separate PR.
* changes the name of the function the base classes use from `get_query_dictionary_list` to `get_query_specs`.
* formatting changes were introduced by yapf

### DIFF
--- a/data_steward/cdr_cleaner/clean_cdr.py
+++ b/data_steward/cdr_cleaner/clean_cdr.py
@@ -48,16 +48,23 @@ import cdr_cleaner.manual_cleaning_rules.remove_operational_pii_fields as operat
 import cdr_cleaner.manual_cleaning_rules.update_questiona_answers_not_mapped_to_omop as map_questions_answers_to_omop
 import cdr_cleaner.cleaning_rules.remove_aian_participants as remove_aian_participants
 import cdr_cleaner.cleaning_rules.remove_non_matching_participant as validate_missing_participants
-from cdr_cleaner.cleaning_rules.clean_mapping import CleanMappingExtTables
 from cdr_cleaner.cleaning_rules.base_cleaning_rule import BaseCleaningRule
+from cdr_cleaner.cleaning_rules.clean_mapping import CleanMappingExtTables
 from cdr_cleaner.cleaning_rules.ensure_date_datetime_consistency import EnsureDateDatetimeConsistency
+from cdr_cleaner.cleaning_rules.rdr_observation_source_concept_id_suppression import ObservationSourceConceptIDRowSuppression
 from constants.cdr_cleaner.clean_cdr import DataStage as stage
 from constants.cdr_cleaner import clean_cdr as cdr_consts
 
 LOGGER = logging.getLogger(__name__)
 
-EHR_CLEANING_CLASSES = [(id_dedup.get_id_deduplicate_queries,),
-                        CleanMappingExtTables]
+EHR_CLEANING_CLASSES = [
+    (id_dedup.get_id_deduplicate_queries,),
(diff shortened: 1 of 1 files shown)


data_steward/cdr_cleaner/clean_cdr.py
```diff
@@ -48,16 +48,23 @@
...
 
-EHR_CLEANING_CLASSES = [(id_dedup.get_id_deduplicate_queries,),
-                        CleanMappingExtTables]
+EHR_CLEANING_CLASSES = [
+    (id_dedup.get_id_deduplicate_queries,),
+    # trying to query a table while creating query strings,
+    # can't work with mocked strings.  should use base class
+    # setup_query_execution function to load dependencies before query execution
+    (
+        CleanMappingExtTables,)
```
nishanthpp93, line 66: Is this an easy fix? Also, what considerations does one need to take before running the CleanMappingExtTables cleaning rule? Does it fail as-is?

lrwb-aou: The rule itself does not fail.  Like the comment states, a reporting function (that hasn't been built yet), can't provide mocked strings for the project, dataset, and sandbox dataset to get back sql statements because it is trying to use those mocked strings to read table names from projects that don't exist.  If I try to run, 
```
PYTHONPATH=.:$PYTHONPATH python cdr_cleaner/cleaning_rules/clean_mapping.py -p foo -d bar -b baz -s -l
```
A bigquery error is returned and the rules can't be listed.  If I provided good parameters, I don't think this is an issue.  But providing mocked parameters in reporting allows us to be absolutely sure we don't run a rule when we just want a listing of the sql or leak project/dataset names.

I'm also open to other suggestions.
I didn't suggest a fix, because I think it is out of the scope of this issue.  But I think this is a rule that would make use of a `setup()` function to get the table names, store them in the instance, and then have the instance execute it's SQL.  It's just a little in transition because of the mix of base-classed cleaning rules and non-base classed cleaning rules.

nishanthpp93: I see the purpose now. I agree this is out of scope and should be tackled in a new ticket. It looks like the bq and non-bq parts of the cleaning rule have to be separated so something like a "local dry-run" is possible. How does the cleaning rule determine if it is being provided mock strings? Is there an explicit parameter declaring so?

lrwb-aou: No.  There is no explicit parameter.  The cleaning rule does not determine if it is mock data or not.  It really doesn't care.  It just takes those parameters and tries to create the SQL.  This works for 26 of 33 cleaning modules.  This is just for creating a report of the sql rules that will be used on a particular dataset.  If notes are needed or an analyst wants to know the sql we are using to clean a dataset, we can get it easily without reading from old logs or providing real dataset names.  6 of the rules that don't conform to this are loading data into BigQuery.  This rule is one that would take a little extra thought of how to achieve both the cleaning rule and the mock sql reporting capability.

nishanthpp93: Makes sense, thanks for clarifying that!

PR #494, opened 2020-04-29
[EDQ-452] Investigate the Heights/Weights Provided by EHR

noahgengel:
Creates a notebook for each of the following:
* each measurement_concept_id / site combination: min, max, mean, stdev, and number of unit_concept_ids used

* each unit_concept_id - the min, max, mean, stdev, and number of sites using.

* also provides some additional organization to the CDR Ops folder (divides into two categories)


noahgengel:
@nishanthpp93 this was a quick analysis requested by Caroline. Figured it could be useful in the CDR Ops folder going forward.

PR #502, opened 2020-05-07
[DC-712] Create Validation Framework for Individual Cleaning Rules

ksdkalluri:

### DIFF
--- a/data_steward/cdr_cleaner/cleaning_rules/base_cleaning_rule.py
+++ b/data_steward/cdr_cleaner/cleaning_rules/base_cleaning_rule.py
@@ -125,7 +125,8 @@ class BaseCleaningRule(AbstractBaseCleaningRule):
                  project_id: str = None,
                  dataset_id: str = None,
                  sandbox_dataset_id: str = None,
-                 depends_on: cleaning_class_list = None):
+                 depends_on: cleaning_class_list = None,
+                 affected_tables: List = None):
         """
         Instantiate a cleaning rule with basic attributes.
 
@@ -168,6 +169,7 @@ class BaseCleaningRule(AbstractBaseCleaningRule):
         self._sandbox_dataset_id = sandbox_dataset_id
         self._issue_urls = issue_urls if issue_urls else []
         self._depends_on_classes = depends_on if depends_on else []
+        self._affected_tables = affected_tables if affected_tables else []
 
         super().__init__()
 
@@ -318,14 +320,19 @@ class BaseCleaningRule(AbstractBaseCleaningRule):
     def get_affected_tables(self):
         """
         Method to get tables that will be modified by the cleaning rule.
+        If getting the tables_affected dynamically logic needs to be
+        implemented in this method.
         """
-        return self.__affected_tables
+        return self._affected_tables
 
     def set_affected_tables(self, affected_tables):
         """
         Set the affected_tables for this class instance
         """
-        self.__affected_tables = affected_tables if affected_tables else []
+        if affected_tables:
+            self._affected_tables = affected_tables
+        else:
+            self._affected_tables = self.get_affected_tables()
 
     affected_tables = property(get_affected_tables, set_affected_tables)
 
--- a/data_steward/cdr_cleaner/cleaning_rules/base_cleaning_rule.py
+++ b/data_steward/cdr_cleaner/cleaning_rules/base_cleaning_rule.py
@@ -331,7 +331,7 @@ class BaseCleaningRule(AbstractBaseCleaningRule):
         Set the affected_tables for this class instance, raises error if affected_tables is not a list
         """
         if affected_tables:
-            if not type(affected_tables) is list:
+            if not isinstance(affected_tables, list):
                 raise TypeError('affected_tables must be of type List')
             else:
                 self._affected_tables = affected_tables
--- a/data_steward/cdr_cleaner/cleaning_rules/base_cleaning_rule.py
+++ b/data_steward/cdr_cleaner/cleaning_rules/base_cleaning_rule.py
@@ -116,7 +116,6 @@ class BaseCleaningRule(AbstractBaseCleaningRule):
     string_list = List[str]
     cleaning_class_list = List[AbstractBaseCleaningRule]
     TABLE_COUNT_QUERY = ''' SELECT COALESCE(COUNT(*), 0) AS row_count FROM `{dataset}.{table}` '''
-    affected_tables_list = List[str]
 
     def __init__(self,
                  issue_numbers: string_list = None,
@@ -126,8 +125,7 @@ class BaseCleaningRule(AbstractBaseCleaningRule):
                  project_id: str = None,
                  dataset_id: str = None,
                  sandbox_dataset_id: str = None,
-                 depends_on: cleaning_class_list = None,
-                 affected_tables: affected_tables_list = None):
+                 depends_on: cleaning_class_list = None):
         """
         Instantiate a cleaning rule with basic attributes.
 
@@ -170,7 +168,6 @@ class BaseCleaningRule(AbstractBaseCleaningRule):
         self._sandbox_dataset_id = sandbox_dataset_id
         self._issue_urls = issue_urls if issue_urls else []
         self._depends_on_classes = depends_on if depends_on else []
-        self._affected_tables = affected_tables if affected_tables else []
 
         super().__init__()
 
@@ -318,20 +315,19 @@ class BaseCleaningRule(AbstractBaseCleaningRule):
         """
         return self._sandbox_dataset_id
 
-    @abstractmethod
-    def get_affected_tables(self, *args, **keyword_args):
+    def get_affected_tables(self):
         """
         Method to get tables that will be modified by the cleaning rule.
+        """
+        return self.__affected_tables
 
-        :return: a list of table names that will be affected by this cleaning rule
+    def set_affected_tables(self, affected_tables):
         """
-        pass
+        Set the affected_tables for this class instance
+        """
+        self.__affected_tables = affected_tables if affected_tables else []
 
-    def set_affected_tables(self, value):
-        """
-        set the affected_tables for this class instance
-        """
-        pass
+    affected_tables = property(get_affected_tables, set_affected_tables)
 
     def get_table_counts(self, client, dataset, tables):
         """
--- a/data_steward/cdr_cleaner/cleaning_rules/base_cleaning_rule.py
+++ b/data_steward/cdr_cleaner/cleaning_rules/base_cleaning_rule.py
@@ -169,7 +169,6 @@ class BaseCleaningRule(AbstractBaseCleaningRule):
         self._sandbox_dataset_id = sandbox_dataset_id
         self._issue_urls = issue_urls if issue_urls else []
         self._depends_on_classes = depends_on if depends_on else []
-        self.set_affected_tables(affected_tables)
 
         super().__init__()
 
@@ -317,7 +316,8 @@ class BaseCleaningRule(AbstractBaseCleaningRule):
         """
         return self._sandbox_dataset_id
 
-    def get_affected_tables(self, *kwargs):
+    @property
+    def affected_tables(self):
         """
         Method to get tables that will be modified by the cleaning rule.
         If getting the tables_affected dynamically logic needs to be
@@ -325,16 +325,15 @@ class BaseCleaningRule(AbstractBaseCleaningRule):
         """
         return self._affected_tables
 
-    def set_affected_tables(self, affected_tables):
+    @affected_tables.setter
+    def affected_tables(self, affected_tables):
         """
         Set the affected_tables for this class instance
         """
         if affected_tables:
             self._affected_tables = affected_tables
         else:
-            self._affected_tables = self.get_affected_tables()
-
-    affected_tables = property(get_affected_tables, set_affected_tables)
+            self._affected_tables = []
 
     def get_table_counts(self, client, dataset, tables):
         """
--- a/data_steward/cdr_cleaner/cleaning_rules/base_cleaning_rule.py
+++ b/data_steward/cdr_cleaner/cleaning_rules/base_cleaning_rule.py
@@ -328,10 +328,13 @@ class BaseCleaningRule(AbstractBaseCleaningRule):
     @affected_tables.setter
     def affected_tables(self, affected_tables):
         """
-        Set the affected_tables for this class instance
+        Set the affected_tables for this class instance, raises error if affected_tables is not a list
         """
         if affected_tables:
-            self._affected_tables = affected_tables
+            if not type(affected_tables) is list:
+                raise TypeError('affected_tables must be of type List')
+            else:
+                self._affected_tables = affected_tables
         else:
             self._affected_tables = []
 
--- a/data_steward/cdr_cleaner/cleaning_rules/base_cleaning_rule.py
+++ b/data_steward/cdr_cleaner/cleaning_rules/base_cleaning_rule.py
@@ -70,15 +70,6 @@ class AbstractBaseCleaningRule(ABC):
         pass
 
     @abstractmethod
-    def get_affected_tables(self, *args, **keyword_args):
-        """
-        Method to get tables that will be modified by the cleaning rule.
-
-        :return: a list of table names that will be affected by this cleaning rule
-        """
-        pass
-
-    @abstractmethod
     def get_query_specs(self, *args, **keyword_args) -> query_spec_list:
         """
         Interface to return a list of query dictionaries.
@@ -125,6 +116,7 @@ class BaseCleaningRule(AbstractBaseCleaningRule):
     string_list = List[str]
     cleaning_class_list = List[AbstractBaseCleaningRule]
     TABLE_COUNT_QUERY = ''' SELECT COALESCE(COUNT(*), 0) AS row_count FROM `{dataset}.{table}` '''
+    affected_tables_list = List[str]
 
     def __init__(self,
                  issue_numbers: string_list = None,
@@ -134,7 +126,8 @@ class BaseCleaningRule(AbstractBaseCleaningRule):
                  project_id: str = None,
                  dataset_id: str = None,
                  sandbox_dataset_id: str = None,
-                 depends_on: cleaning_class_list = None):
+                 depends_on: cleaning_class_list = None,
+                 affected_tables: affected_tables_list = None):
         """
         Instantiate a cleaning rule with basic attributes.
 
@@ -166,6 +159,8 @@ class BaseCleaningRule(AbstractBaseCleaningRule):
         :param depends_on:  a list of rules a cleaning rule depends on when
             running.  Right now, it is used for reporting purposes, but the
             scope may expand in the future.  Default is an empty list.
+        :param affected_tables: a list of tables that are affected by 
+            running this cleaning rule.
         """
         self._issue_numbers = issue_numbers
         self._description = description
@@ -175,6 +170,7 @@ class BaseCleaningRule(AbstractBaseCleaningRule):
         self._sandbox_dataset_id = sandbox_dataset_id
         self._issue_urls = issue_urls if issue_urls else []
         self._depends_on_classes = depends_on if depends_on else []
+        self._affected_tables = affected_tables if affected_tables else []
 
         super().__init__()
 
@@ -321,6 +317,21 @@ class BaseCleaningRule(AbstractBaseCleaningRule):
         Get the sandbox dataset id for this class instance.
         """
         return self._sandbox_dataset_id
+
+    @abstractmethod
+    def get_affected_tables(self, *args, **keyword_args):
+        """
+        Method to get tables that will be modified by the cleaning rule.
+
+        :return: a list of table names that will be affected by this cleaning rule
+        """
+        pass
+
+    def set_affected_tables(self, value):
+        """
+        set the affected_tables for this class instance
+        """
+        pass
 
     def get_table_counts(self, client, dataset, tables):
         """
--- a/data_steward/cdr_cleaner/cleaning_rules/base_cleaning_rule.py
+++ b/data_steward/cdr_cleaner/cleaning_rules/base_cleaning_rule.py
@@ -58,9 +58,14 @@ class AbstractBaseCleaningRule(ABC):
         """
         Run required steps for validation setup
 
-        Method to run .  For example, if your class requires loading a static
-        table, that load operation should be defined here.  It SHOULD NOT BE
-        defined as part of get_query_specs().
+        Method to run to setup validation on cleaning rules that will be updating or deleting the values.
+        For example:
+        if your class updates all the datetime fields you should be implementing the
+        logic to get the initial list of values which adhere to a condition we are looking for.
+
+        if your class deletes a subset of rows in the tables you should be implementing
+        the logic to get the row counts of the tables prior to applying cleaning rule
+
         """
         pass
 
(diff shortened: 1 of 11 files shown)


data_steward/cdr_cleaner/cleaning_rules/base_cleaning_rule.py
```diff
@@ -276,6 +317,53 @@ def get_sandbox_dataset_id(self):
...
+        """
+        counts_dict = dict()
+        for table in tables:
+            query = self.TABLE_COUNT_QUERY.format(dataset=dataset, table=table)
+            count = client.query(query).to_dataframe()
+            counts_dict[table] = count['row_count'][0]
+        return counts_dict
+
+    def validate_delete_rule(self, dataset, sandbox_dataset, sandbox_tables,
+                             tables_affected, initial_counts, client):
```
lrwb-aou, line 338: I would really like to keep the client object out of the class cleaning rules.  That may / may not be possible.  I'm just trying to think through it  right now.

Thoughts?

data_steward/cdr_cleaner/cleaning_rules/base_cleaning_rule.py
```diff
@@ -53,6 +53,26 @@ def setup_rule(self, *args, **keyword_args):
...
+    @abstractmethod
+    def setup_validation(self, *args, **keyword_args):
+        """
+        Run required steps for validation setup
+
+        Method to run .  For example, if your class requires loading a static
+        table, that load operation should be defined here.  It SHOULD NOT BE
+        defined as part of get_query_specs().
+        """
+        pass
```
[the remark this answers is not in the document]
ksdkalluri: This is my mistake, I planned to update the documentation but sidetracked by the bug in the pipeline. I will update the docstring to match the validation.

data_steward/cdr_cleaner/cleaning_rules/base_cleaning_rule.py
```diff
@@ -53,6 +53,26 @@ def setup_rule(self, *args, **keyword_args):
...
+
+    @abstractmethod
+    def get_affected_tables(self, *args, **keyword_args):
+        """
+        Method to get tables that will be modified by the cleaning rule.
+
+        :return: a list of table names that will be affected by this cleaning rule
+        """
+        pass
```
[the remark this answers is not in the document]
ksdkalluri: I had this as a method because some times we need to implement some logic to see which tables will be affected by it. Most of the cleaning rules are dependant on data we get from sites and rdr if the data changes the tables that gets affected might change. SO I thought getting the tables dynamically will give all the tables for that state of data.

data_steward/cdr_cleaner/cleaning_rules/base_cleaning_rule.py
```diff
@@ -129,6 +169,7 @@ def __init__(self,
         self._sandbox_dataset_id = sandbox_dataset_id
         self._issue_urls = issue_urls if issue_urls else []
         self._depends_on_classes = depends_on if depends_on else []
+        self.TABLE_COUNT_QUERY = ''' SELECT COALESCE(COUNT(*), 0) AS row_count FROM `{dataset}.{table}` '''
```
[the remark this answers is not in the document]
ksdkalluri: adding {project} is a good idea. I'll make the change.

I used coalesce because when the tables are empty we get none object as a return value so to make it easy to compute I made it pass 0 as row count when the table is empty.

data_steward/cdr_cleaner/cleaning_rules/base_cleaning_rule.py
ksdkalluri: I thought of creating a client inside the method before adding it. but then I figured we shouldn't create too many clients in the workflow which will increase the running time. so I thought There will be a client instantiated in a cleaning rule so we can just use that.

I am open to using any way that we all can agree on.

data_steward/cdr_cleaner/cleaning_rules/base_cleaning_rule.py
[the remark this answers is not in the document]
lrwb-aou: Nice.  I had not thought about using coalesce for that.  I like it.

data_steward/cdr_cleaner/cleaning_rules/base_cleaning_rule.py
lrwb-aou: This is much better than creating the client in each class.  With this setup, the engine can pass the client it is already using.  

I'm just trying to think of a way to make these functions act like the cleaning rule themselves.  Can they just produce the sql needed and let that be executed by the clean engine?  I'm not sure if it can or can't.  If passing the sql to the engine can't work, then passing the client object from the engine to the validation methods is highly preferable to creating a new client in the validation methods.

data_steward/cdr_cleaner/cleaning_rules/base_cleaning_rule.py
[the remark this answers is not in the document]
lrwb-aou: So, you think there are some cases where the list of returned tables may change?  Is this because data quality has improved and the rows aren't cleaned by a cleaning query?  I'm trying to think of an instance where this would dynamically change.  From my perspective, if a table could be changed by a cleaning rule (meaning it is checked by the rule whether or not an actual change is applied), I would want to know which tables it is checking.  Can you give me an example of the dynamic perspective you are thinking about?

data_steward/cdr_cleaner/cleaning_rules/base_cleaning_rule.py
ksdkalluri: I tried to get it the way you were thinking, but it wouldn't be possible for this scenario because for a cleaning rule we are trying to collect all the queries and run all of them sequentially. the use case here is we are querying for some kind of a result either row_count or state of the data which needs to be saved to make the validation possible. as long as we have something to query and save before running the cleaning rule I couldn't think of a different approach.

data_steward/cdr_cleaner/cleaning_rules/base_cleaning_rule.py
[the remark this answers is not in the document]
ksdkalluri: consider tables like Specimen table we don't actually care about this table right now even if the data is present or not. If we manually add the tables list we might miss this table in the cleaning rule. but if we are getting the table names dynamically using fields or another logic we will not miss this table and we don't need to manually add the table to the list when the all the sites start sending data and we actually need to apply cleaning rules to this table.

data_steward/cdr_cleaner/cleaning_rules/base_cleaning_rule.py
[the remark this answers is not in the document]
lrwb-aou: `TABLE_COUNT_QUERY` feels like a class variable instead of an instance variable.

data_steward/cdr_cleaner/cleaning_rules/base_cleaning_rule.py
lrwb-aou: Ok.  So these methods are validation helper methods.  They are anticipated to be used by the ` setup_validation(self, *args, **keyword_args)` and ` validate_rule(self, *args, **keyword_args)` methods.  This means we expect to pass the `client` object from the engine into these methods (same for the `setup_rule(self, *args, **keyword_args)` method).  I don't expect what is passed from the engine into the class object to change.  What are your thoughts on updating these signatures to specifically include the `client` object, `(client, *args, **keyword_args)`?

data_steward/cdr_cleaner/cleaning_rules/base_cleaning_rule.py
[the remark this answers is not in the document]
lrwb-aou: Maybe this means we need a `setter` method to validate the value before setting the attribute and the `getter` would still return the value.

data_steward/cdr_cleaner/cleaning_rules/base_cleaning_rule.py
ksdkalluri: I think it is a good idea.

data_steward/cdr_cleaner/cleaning_rules/base_cleaning_rule.py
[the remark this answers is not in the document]
lrwb-aou: @ksdkalluri My former Java days are leaking through.  I should have implemented the getters/setters as python properties, similar to this [SO example](https://stackoverflow.com/a/47507855/8132160).  Getting back to why the setter.  Using a setter allows the value for the variable to be validated before it is set instead of when you're getting ready to return the value.  It allows you to raise errors earlier and allows the user to set the value and have that value checked.  I'd be happy to discuss this with you if you would like that.

data_steward/cdr_cleaner/cleaning_rules/base_cleaning_rule.py
```diff
@@ -125,6 +116,7 @@ class BaseCleaningRule(AbstractBaseCleaningRule):
     string_list = List[str]
     cleaning_class_list = List[AbstractBaseCleaningRule]
     TABLE_COUNT_QUERY = ''' SELECT COALESCE(COUNT(*), 0) AS row_count FROM `{dataset}.{table}` '''
+    affected_tables_list = List[str]
```
lrwb-aou, line 119: this should not be a class level variable.  it is still an instance variable.

data_steward/cdr_cleaner/cleaning_rules/base_cleaning_rule.py
```diff
@@ -175,6 +170,7 @@ def __init__(self,
         self._sandbox_dataset_id = sandbox_dataset_id
         self._issue_urls = issue_urls if issue_urls else []
         self._depends_on_classes = depends_on if depends_on else []
+        self._affected_tables = affected_tables if affected_tables else []
```
lrwb-aou, line 173: I think I would set this to something like:
```
self.set_affected_tables(affected_tables)
```
Then the value can be validated by the argument validation methods.
```
def set_affected_tables(self, affected_tables):
    self.__affected_tables = affected_tables if affected_tables else []
```
Then, classes that need to can override this with a more appropriate implementation.  For most classes, this will suffice.

data_steward/cdr_cleaner/cleaning_rules/base_cleaning_rule.py
```diff
@@ -322,6 +318,21 @@ def get_sandbox_dataset_id(self):
         """
         return self._sandbox_dataset_id
 
+    @abstractmethod
```
lrwb-aou, line 321: this getter would not be abstract anymore. it would exist when the extending class is instantiated.

data_steward/cdr_cleaner/cleaning_rules/base_cleaning_rule.py
```diff
@@ -322,6 +318,21 @@ def get_sandbox_dataset_id(self):
...
+        :return: a list of table names that will be affected by this cleaning rule
+        """
+        pass
+
+    def set_affected_tables(self, value):
+        """
+        set the affected_tables for this class instance
+        """
+        pass
+
```
lrwb-aou, line 335: what do you think about adding these as properties methods to the base class?  like
```
def get_affected_tables(self):
    return self.__affected_tables

def set_affected_tables(self, value):
    self.__affected_tables = value if value else []

affected_tables = property(get_affected_tables, set_affected_tables)
```
how do you feel about updating the base class properties the same way?

data_steward/cdr_cleaner/cleaning_rules/base_cleaning_rule.py
ksdkalluri: I made the change but I couldn't figure out how it can be used in __init__  function.

data_steward/cdr_cleaner/cleaning_rules/base_cleaning_rule.py
ksdkalluri: Removing the above class variable is giving making it unable to implement the suggested setter.

data_steward/cdr_cleaner/cleaning_rules/base_cleaning_rule.py
```diff
@@ -276,6 +315,67 @@ def get_sandbox_dataset_id(self):
...
+        """
+        return self.__affected_tables
+
+    def set_affected_tables(self, affected_tables):
+        """
+        Set the affected_tables for this class instance
+        """
+        self.__affected_tables = affected_tables if affected_tables else []
+
+    affected_tables = property(get_affected_tables, set_affected_tables)
```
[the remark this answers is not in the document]
ksdkalluri: ah, now I get it. Thanks for this example @lrwb-aou

ksdkalluri:
> This is potentially problematic. This will allow the `_affected_tables` variable to be set differently under different conditions (constructor setting vs. property setting where a class may override the property). This has the potential of diverging conditions. You should use the setter method in the initialization.
> 
> ```
> self.set_affected_tables(affected_tables)
> ```
> 
> This is an example of potentially diverging behavior, using a setter function and using an assignment statement in the constructor. The constructor allows the variable to be a string but the setter function doesn't not. This is divergent behavior and will be hard to debug.
> 
> ```
> class Example2(object):
>     def __init__(self, var1=None):
>         self._var1 = 'booya'
> 
>     def get_var1(self):
>         return self._var1
> 
>     def set_var1(self, value):
>         if not isinstance(value, int):
>             raise RuntimeError('Wrong type')
> 
>         if value:
>             self._var1 = value
>         else:
>             self._var1 = 3
> 
>     var1 = property(get_var1, set_var1)
> 
> if __name__ == "__main__":
>     print('====================================================')
>     e2 = Example2()
>     print('-----------------------------------------------------')
>     print(f'e2 var1 is:\t{e2.var1}') # will call the method get_var1()  # proves that the variable was initialized with a string value
>     print('-----------------------------------------------------')
> 
>     e2.var1 = 'alsdfjk' # will call the method set_var1().  this will cause a runtime error because the setter doesn't accept string values.
>     print(f'e2 var1 is:\t{e2.var1}') # will call the method get_var1()
> ```

so essentially changing `self._affected_tables = affected_tables if affected_tables else []` to 
`self.set_affected_tables(affected_tables)`  would fix the issue?

lrwb-aou:
> so essentially changing self._affected_tables = affected_tables if affected_tables else [] to
self.set_affected_tables(affected_tables) would fix the issue?

yes, then it would use the same logic to set the variable via the class initialization method or via the variable assignment function.

data_steward/cdr_cleaner/cleaning_rules/clean_mapping.py
```diff
@@ -69,7 +69,8 @@ def __init__(self, project_id, dataset_id, sandbox_dataset_id):
                          affected_datasets=[cdr_consts.RDR],
                          project_id=project_id,
                          dataset_id=dataset_id,
-                         sandbox_dataset_id=sandbox_dataset_id)
+                         sandbox_dataset_id=sandbox_dataset_id,
+                         affected_tables=['_mapping'])
```
ksdkalluri, line 73: @lrwb-aou Clean_mapping uses dynamic logic to get the affected tables, but it has to get two different sets of affected tables. can you suggest some ideas to handle this issue?

lrwb-aou: @ksdkalluri have you considered setting this to a function that returns a list of table names based on the tables in `resource_files/fields` directory?  The validation function for this class may need to implement its own method instead of using the base class method.  This potentially affects all `_mapping` and `_ext` tables.  So I think a list of those should suffice.  You can use a helper method to generate that list.  Does this make sense?  Other suggestions or considerations?

ksdkalluri: yes, I did consider getting tables from fields but the get_tables method has logic that finds the tables in the dataset (mapping or ext) and finds the associated CDM tables. which uses bq to run a query to get all the tables in a dataset. I am starting to think we might also pass the client as the param for the get_affected_tables.

data_steward/cdr_cleaner/cleaning_rules/clean_mapping.py
```diff
@@ -204,6 +212,28 @@ def get_sandbox_tablenames(self):
         """
         raise NotImplementedError("Please fix me")
 
+    def setup_validation(self):
```
lrwb-aou, line 215: should this be `setup_validation(self, client):`?

data_steward/cdr_cleaner/cleaning_rules/clean_mapping.py
```diff
@@ -204,6 +212,28 @@ def get_sandbox_tablenames(self):
...
+        Run required steps for validation setup
+
+        This abstract method was added to the base class after this rule was authored.
+        This rule needs to implement logic to setup validation on cleaning rules that
+        will be updating or deleting the values.
+        Until done no issue exists for this yet.
+        """
+        raise NotImplementedError("Please fix me.")
+
+    def validate_rule(self):
```
lrwb-aou, line 226: should this be `validate_rule(self, client):`?

data_steward/cdr_cleaner/cleaning_rules/ensure_date_datetime_consistency.py
```diff
@@ -155,6 +163,28 @@ def get_sandbox_tablenames(self):
         """
         return []
 
+    def setup_validation(self):
```
lrwb-aou, line 166: should `client` be added to these two method signatures?

data_steward/cdr_cleaner/cleaning_rules/null_concept_ids_for_numeric_ppi.py
```diff
@@ -153,6 +155,28 @@ def setup_rule(self):
     def get_sandbox_tablenames(self):
         return [SAVE_TABLE_NAME]
 
+    def setup_validation(self):
```
lrwb-aou, line 158: same as above, should these signatures contain the `client` parameter?

data_steward/cdr_cleaner/cleaning_rules/rdr_observation_source_concept_id_suppression.py
```diff
@@ -98,6 +99,38 @@ def setup_rule(self):
...
+        """
+        Method to get tables that will be modified by the cleaning rule.
+
+        This abstract method was added to the base class after this rule was authored.This
+        rule needs to implement logic to get the affected tables .
+        Until done no issue exists for this yet.
+        """
+        raise NotImplementedError("Please fix me.")
+
+    def setup_validation(self):
```
lrwb-aou, line 112: same as above, should these signatures contain the `client` parameter?

data_steward/cdr_cleaner/cleaning_rules/ensure_date_datetime_consistency.py
```diff
@@ -10,48 +10,49 @@
...
+# Project imports
+from cdr_cleaner.cleaning_rules.base_cleaning_rule import BaseCleaningRule
+from common import CONDITION_OCCURRENCE, DRUG_EXPOSURE, DEVICE_EXPOSURE, MEASUREMENT, OBSERVATION, \
+    PROCEDURE_OCCURRENCE, SPECIMEN, DEATH, OBSERVATION_PERIOD, VISIT_OCCURRENCE
 
 LOGGER = logging.getLogger(__name__)
 
 TABLE_DATES = {
-    common.CONDITION_OCCURRENCE: {
+    CONDITION_OCCURRENCE: {
```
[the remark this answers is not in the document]
ksdkalluri: sure, I agree.

data_steward/cdr_cleaner/cleaning_rules/base_cleaning_rule.py
```diff
@@ -276,6 +316,72 @@ def get_sandbox_dataset_id(self):
...
+        return self._affected_tables
+
+    @affected_tables.setter
+    def affected_tables(self, affected_tables):
+        """
+        Set the affected_tables for this class instance
+        """
+        if affected_tables:
+            self._affected_tables = affected_tables
+        else:
```
lrwb-aou, line 335: This is defaulting to a list when None.  We should note it in the docstring and enforce it in the value check.  If a value is passed to the setter that is not None and is not a list, an error should be raised.

data_steward/cdr_cleaner/cleaning_rules/clean_mapping.py
```diff
@@ -47,6 +47,13 @@
 ISSUE_NUMBER = 'DC-715'
 
 
+def get_mapping_tables():
+    """
+    Returns list of mapping tables in fields path
+    """
+    return resources.MAPPING_TABLES
```
lrwb-aou, line 54: I think this may need to return a list of `_mapping_` and `_ext` tables for the reporting.  I checked the `generate_ext_tables.py` file.  It basically just creates an extension table for any mapping tables.  You could use a list comprehension to do the same thing here with this `resources.MAPPING_TABLE` variable as your input.

data_steward/cdr_cleaner/cleaning_rules/clean_mapping.py
lrwb-aou: Did you intend to leave `client` out of this signature?

data_steward/cdr_cleaner/cleaning_rules/ensure_date_datetime_consistency.py
lrwb-aou: Did you intend to leave `client` out of this signature?

data_steward/cdr_cleaner/cleaning_rules/null_concept_ids_for_numeric_ppi.py
lrwb-aou: Did you intend to leave `client` out of this signature?

data_steward/cdr_cleaner/cleaning_rules/rdr_observation_source_concept_id_suppression.py
lrwb-aou: Did you intend to leave `client` out of this signature?

data_steward/cdr_cleaner/cleaning_rules/base_cleaning_rule.py
```diff
@@ -276,6 +316,75 @@ def get_sandbox_dataset_id(self):
...
+        """
+        return self._affected_tables
+
+    @affected_tables.setter
+    def affected_tables(self, affected_tables):
+        """
+        Set the affected_tables for this class instance, raises error if affected_tables is not a list
+        """
+        if affected_tables:
+            if not type(affected_tables) is list:
```
lrwb-aou, line 334: https://stackoverflow.com/questions/1549801/what-are-the-differences-between-type-and-isinstance
The python community generally prefers an instance check over a type check.
```
if not isinstance(affect_tables, list):
```
You can see that most argue we shouldn't use either, but in this case I think it is acceptable.  Because we want to set the variable to a good value that can be trusted and used later.

tests/unit_tests/data_steward/cdr_cleaner/cleaning_rules/null_concept_ids_for_numeric_ppi_test.py
```diff
@@ -13,13 +13,13 @@
...
-
+from cdr_cleaner.cleaning_rules.null_concept_ids_for_numeric_ppi import NullConceptIDForNumericPPI, \
+    SAVE_TABLE_NAME, SANDBOX_QUERY, CLEAN_NUMERIC_PPI_QUERY
 # Project imports
 from constants.bq_utils import WRITE_TRUNCATE
 from constants.cdr_cleaner import clean_cdr as clean_consts
-from cdr_cleaner.cleaning_rules.null_concept_ids_for_numeric_ppi import NullConceptIDForNumericPPI, \
-    SAVE_TABLE_NAME, SANDBOX_QUERY, CLEAN_NUMERIC_PPI_QUERY
+
+# Third party imports
```
lrwb-aou, line 22: nit:  just an odd comment and import ordering.

PR #504, opened 2020-05-07
[DC-798] Fix failing ehr_union and site validation caused by person table fields

ksdkalluri:
[DC-798] Fix failing ehr_union and site validation caused by person table fields

[DC-798]: https://precisionmedicineinitiative.atlassian.net/browse/DC-798

### DIFF
--- a/data_steward/resources/fields/post_deid_person.json
+++ b/data_steward/resources/fields/post_deid_person.json
@@ -106,5 +106,23 @@
         "name": "ethnicity_source_concept_id", 
         "mode": "nullable", 
         "description": "A foreign key to the ethnicity concept that refers to the code used in the source."
+    }, 
+    {
+        "type": "integer", 
+        "name": "sex_at_birth_concept_id", 
+        "mode": "nullable", 
+        "description": "[All of Us OMOP extension] A foreign key to the biological sex at birth concept."
+    }, 
+    {
+        "type": "integer", 
+        "name": "sex_at_birth_source_concept_id", 
+        "mode": "nullable", 
+        "description": "[All of Us OMOP extension] A foreign key to the biological sex at birth source concept."
+    }, 
+    {
+        "type": "string", 
+        "name": "sex_at_birth_source_value", 
+        "mode": "nullable", 
+        "description": "[All of Us OMOP extension] The source code for the biological sex at birth."
     }
 ]
--- a/data_steward/tools/snapshot_by_query.py
+++ b/data_steward/tools/snapshot_by_query.py
@@ -147,7 +147,7 @@ if __name__ == '__main__':
                         dest='snapshot_dataset_id',
                         help='Name of the new dataset that needs to be created',
                         required=True)
-    parser.add_argument('-ds',
+    parser.add_argument('-s',
                         '--data_stage',
                         action='store',
                         dest='data_stage',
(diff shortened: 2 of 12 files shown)


data_steward/resources/fields/post_deid_person.json
```diff
@@ -0,0 +1,128 @@
+[
```
nishanthpp93, line 1: This file can cause issues in `resources.cdm_schemas` if it is not excluded within that function.

data_steward/tools/snapshot_by_query.py
```diff
@@ -125,7 +147,14 @@ def create_snapshot_dataset(project_id, dataset_id, snapshot_dataset_id):
                         dest='snapshot_dataset_id',
                         help='Name of the new dataset that needs to be created',
                         required=True)
+    parser.add_argument('-ds',
+                        '--data_stage',
```
nishanthpp93, line 151: You cannot have multiple characters in short option format, argparse will throw an error. It needs to be a single character. This is because sometimes multiple options can be specified that way (here, `-ds` would mean both `-d` and `-s` options are set)

data_steward/tools/snapshot_by_query.py
```diff
@@ -125,7 +147,14 @@ def create_snapshot_dataset(project_id, dataset_id, snapshot_dataset_id):
...
+                        action='store',
+                        dest='data_stage',
+                        help='Stage of the cdr generation',
+                        choices=['pre_deid', 'post_deid'],
+                        required=True)
```
nishanthpp93, line 156: Instead of passing a parameter like this, can we make it check if the supplied input dataset has a person table with the extra columns and if it does, use the new person schema, else stick to the old one?

data_steward/tools/snapshot_by_query.py
ksdkalluri: i'll change it to a single character

data_steward/tools/snapshot_by_query.py
ksdkalluri: I actually thought about getting from the table, but this is going to go away in the near future so I didn't spend much time to change it.

nishanthpp93: It's a simpler change than what you are currently doing. All it needs is to query the dataset with `SELECT * from {proj}.{dataset}.INFORMATION_SCHEMA.COLUMNS` and look for `sex_at_birth_concept_id ` in the `result_df.['column_name']` column. You wouldn't have to modify any of the bash scripts and it is easy to miss that parameter while doing it manually.

data_steward/resources/fields/post_deid_person.json
ksdkalluri: updated

data_steward/tools/snapshot_by_query.py
ksdkalluri: @nishanthpp93 Updated the data_stage issue

data_steward/tools/snapshot_by_query.py
```diff
@@ -20,13 +27,33 @@ def create_empty_dataset(project_id, dataset_id, snapshot_dataset_id):
...
+        if table == PERSON and has_at_birth_column(dataset_id):
+            table_id = table
+            table_name = 'post_deid_person'
+        else:
+            table_id = table
+            table_name = table
+        create_standard_table(table_name,
+                              table_id,
+                              drop_existing=True,
+                              dataset_id=snapshot_dataset_id)
```
nishanthpp93, line 56: nit: This might be simpler
```suggestion
    for table in resources.CDM_TABLES:
        table_id = table
        table_name = table
        if table == PERSON and has_at_birth_column(dataset_id):
            table_name = 'post_deid_person'
        create_standard_table(table_name,
                              table_id,
                              drop_existing=True,
                              dataset_id=snapshot_dataset_id)
```

PR #505, opened 2020-05-07
[DC-703] Add cleaning rule from DC-537 to the pipeline

dcampbell-vumc:
Created `null_concept_ids_for_nmeric_ppi.py` module. Created `null_concept_ids_for_nmeric_ppi_test.py` unit and integration tests.

### DIFF
--- a/data_steward/cdr_cleaner/cleaning_rules/null_concept_ids_for_numeric_ppi.py
+++ b/data_steward/cdr_cleaner/cleaning_rules/null_concept_ids_for_numeric_ppi.py
@@ -1,50 +1,54 @@
 """
+Nullify concept ids for numeric PPIs from the RDR observation dataset
+
 Original Issues: DC-537, DC-703
+
+The intent is to null concept ids (value_source_concept_id, value_as_concept_id, value_source_value,
+value_as_string) for numeric PPIs from the RDR observation dataset. The changed records should be
+archived in the dataset sandbox.
 """
 
 # Python imports
 import logging
 
 # Project imports
+import constants.cdr_cleaner.clean_cdr as cdr_consts
 from cdr_cleaner.cleaning_rules.base_cleaning_rule import BaseCleaningRule
 from constants.bq_utils import WRITE_TRUNCATE
-import constants.cdr_cleaner.clean_cdr as cdr_consts
+from sandbox import check_and_create_sandbox_dataset
 
 LOGGER = logging.getLogger(__name__)
-
 
 SAVE_TABLE_NAME = "dc_703_obs_changed_rows_saved"
 
 # Query to create tables in sandbox with the rows that will be removed per cleaning rule
 SANDBOX_QUERY = """
 CREATE OR REPLACE TABLE 
-  `{project}.{sandbox_dataset}.{intermediary_table}` AS (
+    `{project}.{sandbox_dataset}.{intermediary_table}` AS (
 SELECT *
 FROM 
-  `{project}.{dataset}.observation`
+    `{project}.{dataset}.observation`
 WHERE
-  questionnaire_response_id IS NOT NULL
+    questionnaire_response_id IS NOT NULL
 AND 
-  value_as_number IS NOT NULL
+    value_as_number IS NOT NULL
 AND 
-  (value_source_concept_id IS NOT NULL OR value_as_concept_id IS NOT NULL))
+    (value_source_concept_id IS NOT NULL OR value_as_concept_id IS NOT NULL))
 """
 
-# Query to modify all selected rows
 CLEAN_NUMERIC_PPI_QUERY = """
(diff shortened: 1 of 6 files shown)


data_steward/cdr_cleaner/cleaning_rules/null_concept_ids_for_numeric_ppi.py
```diff
@@ -0,0 +1,134 @@
...
+SELECT *
+FROM 
+    `{project}.{dataset}.observation`
+WHERE
+    questionnaire_response_id IS NOT NULL
+AND 
+    value_as_number IS NOT NULL
+AND 
+    (value_source_concept_id IS NOT NULL OR value_as_concept_id IS NOT NULL))
+"""
```
amcgrenera-vumc, line 36: Please update queries using Jinja, per the discussion in the eng sync. Here is an example Lauren sent me:

https://github.com/all-of-us/curation/blob/develop/tests/integration_tests/data_steward/cdr_cleaner/cleaning_rules/bigquery_tests_base.py#L44

dcampbell-vumc: Should I go ahead and set up my Jinja environment in my module? I know we discussed putting it in a utils file.

data_steward/cdr_cleaner/cleaning_rules/null_concept_ids_for_numeric_ppi.py
```diff
@@ -0,0 +1,134 @@
...
+    NULL AS value_as_string)
+FROM `{project}.{dataset}.observation` AS obs 
+WHERE 
+   questionnaire_response_id IS NOT NULL
+AND
+   value_as_number IS NOT NULL
+AND
+   (value_source_concept_id IS NOT NULL OR value_as_concept_id IS NOT NULL)
+"""
+
```
amcgrenera-vumc, line 52: When running this query I realized this is only selecting the 1m+ rows that are effected by the where clause. I asked in the eng channel about select all or select each column individually and Krishna suggested selecting each column individually, so I think the query below should work better:

SELECT
  observation_id,
  person_id,
  observation_concept_id,
  observation_date,
  observation_datetime,
  observation_type_concept_id,
  value_as_number,
  CASE
    WHEN value_as_number IS NOT NULL AND value_source_concept_id IS NOT NULL AND (value_source_concept_id IS NOT NULL OR value_as_concept_id IS NOT NULL) THEN NULL
END
  AS value_as_string,
  CASE
    WHEN value_as_number IS NOT NULL AND value_source_concept_id IS NOT NULL AND (value_source_concept_id IS NOT NULL OR value_as_concept_id IS NOT NULL) THEN NULL
END
  AS value_as_concept_id,
  qualifier_concept_id,
  unit_concept_id,
  provider_id,
  visit_occurrence_id,
  observation_source_value,
  observation_source_concept_id,
  unit_source_value,
  qualifier_source_value,
  CASE
    WHEN value_as_number IS NOT NULL AND value_source_concept_id IS NOT NULL AND (value_source_concept_id IS NOT NULL OR value_as_concept_id IS NOT NULL) THEN NULL
END
  AS value_source_concept_id,
  CASE
    WHEN value_as_number IS NOT NULL AND value_source_concept_id IS NOT NULL AND (value_source_concept_id IS NOT NULL OR value_as_concept_id IS NOT NULL) THEN NULL
END
  AS value_source_value,
  questionnaire_response_id
FROM
  `{project}.{dataset}.observation`

data_steward/cdr_cleaner/cleaning_rules/null_concept_ids_for_numeric_ppi.py
amcgrenera-vumc: Just checked with Lauren, she suggests the following, so I would go ahead and set up the Jinja env in your own module for now: 

> good question.  if you want to set up the environment for yourself in your own module, then you could go ahead and write your queries with a template the way you expect.  when we implement the utility, you could then just import the environment and remove the environment you created from your module without needing to change your sql and go through another round of SQL reviews.  so I would probably do that for now.

amcgrenera-vumc:
Hey Danyel, everything looks good to me with this PR once you update the queries using Jinja and update `clean_numeric_ppi_query`. I compared this cleaning rule to other cleaning rules that are utilizing the base cleaning rule as I was unfamiliar, this helped me better understand the base cleaning rule, so thank you!

data_steward/cdr_cleaner/cleaning_rules/null_concept_ids_for_numeric_ppi.py
dcampbell-vumc: Okay, I will set it up globally for now.

PR #509, opened 2020-05-14
[EDQ-462] Small Fixes in Data Quality Metrics Processes

noahgengel:
* Ensured consistency in the means of calculating error rates (reporting 'error rate' rather than a mix of 'error rate' and 'success rate')

* Implemented a temporary workaround for a bug within the duplicates notebook (regarding datetime casting)

* Uses 0 values instead of nan values in the metrics_over_time module

* Fixed the dictionaries and lists with respect to the constants file.

* Changed the means of generating the full site names in the combined/deid datasets

* Changed setup functions of the duplicates metric (relic of old setups)


noahgengel:
@nishanthpp93 small commit to fix some bugs

PR #512, opened 2020-05-18
[DC-654] Adjust De-id generalizations to account for new gender ident…

amcgrenera-vumc:
…ity response option


amcgrenera-vumc:
@dcampbell-vumc doh, I was querying based off the `value_source_concept_id` I am running into the same issue. Looking into this.

PR #513, opened 2020-05-18
[DC-866] Fix union_ehr failing to save achilles_heel_results

ksdkalluri:
Add unioned_ehr to whitelist, where it can be prepended to the Achilles tables.


ksdkalluri:
Yes, I tested  it.

PR #514, opened 2020-05-18
[DC-505] Fix drug class query division by zero error.

ksdkalluri:


data_steward/constants/validation/main.py
```diff
@@ -43,37 +43,39 @@
...
-        concept_id
+SELECT
+  init.*,
+  CASE
+    WHEN ( SELECT COUNT(*) FROM `{project_id}.{dataset_id}.{table_id}`) > 0
+    THEN CONCAT(CAST(ROUND(init.count/( SELECT COUNT(*) FROM `{project_id}.{dataset_id}.{table_id}`)*100, 2) AS STRING), '%')
+  ELSE
+  '0'
+END
+  AS percentage
```
nishanthpp93, line 54: nit: the `END` is for the `CASE`, so I would format them as follows
```suggestion
  CASE
    WHEN ( SELECT COUNT(*) FROM `{project_id}.{dataset_id}.{table_id}`) > 0
    THEN CONCAT(CAST(ROUND(init.count/( SELECT COUNT(*) FROM `{project_id}.{dataset_id}.{table_id}`)*100, 2) AS STRING), '%')
    ELSE '0'
  END AS percentage
```

ksdkalluri: done updated!

PR #516, opened 2020-05-20
[DC-867] Validation on sites failing due to unfinished jobs

nishanthpp93:
* Extend wait time for jobs to complete, add test


nishanthpp93:
Yes, this is because we use exponential backoff. Earlier, the total time was 1+2+4+8+...+256 seconds. Now the extra retry will make it 1+2+4+8+...+256+512 seconds. It should be enough, if not there's something else going wrong.

PR #517, opened 2020-05-21
[EDQ-458] Create Additional Analyses for Height/Weight

noahgengel:
* Additions for h/w analyses that can be used to better understand the justifications for exclusion criteria
* Minor improvements in the documentation for metrics_over_time modules


noahgengel:
@nishanthpp93 this is related to it; I'm documenting ways that we could potentially improve the height/weight cleaning rule and this more or less supports my suggestions.

PR #521, opened 2020-05-26
[DC-792] Sandbox Dataset Creation Bug

amcgrenera-vumc:


data_steward/utils/bq.py
```diff
@@ -286,3 +286,64 @@ def get_latest_validation_dataset_id(project_id):
...
+    if dataset_id is None:
+        raise RuntimeError("Cannot create a dataset without a name")
+
+    if description is None:
+        raise RuntimeError("Will not create a dataset without a description")
+
+    if project_id is None:
+        LOGGER.info(f"You should specify project_id for a reliable experience."
+                    f"Defaulting to {os.environ.get(PROJECT_ID)}.")
+        project_id = os.environ.get(PROJECT_ID)
```
nishanthpp93, line 315: It looks like a lot of these checks are already in the `define_dataset` function. It might be better to call that function within this to be consistent and reuse code.

data_steward/utils/bq.py
```diff
@@ -286,3 +286,64 @@ def get_latest_validation_dataset_id(project_id):
...
+
+    if overwrite_existing is None:
+        overwrite_existing = True
+    elif not overwrite_existing:
+        overwrite_existing = False
+    else:
+        overwrite_existing = True
```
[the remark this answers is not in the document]
amcgrenera-vumc: The `None` is what was in the `bq_utils.create_dataset`. It defaults to None, if it is not provided, and overwrites; per the description `:param overwrite_existing: determine if dataset should be overwritten if already exists, defaults to true/overwrite` Let me know if this needs to be changed. I just didn't want to alter want was already in place.

data_steward/utils/bq.py
amcgrenera-vumc: This is a good idea, I will make that change!

amcgrenera-vumc: Quick question-- `define_dataset` assumes the dataset does not exist and raises a conflict if it already exists. `create_dataset` was initially written in `bq_utils` to default to overwrite and does not throw a conflict if the dataset exists (which is what I maintained in the function in `bq`. Do you think this could cause issues or should it default the same and throw a conflict if the dataset exists?

amcgrenera-vumc: One more thing- `define_dataset` requires a lable_or_tag, which `create_dataset` previously did not. And neither does `create_sandbox`, let me know if they should all require, which will need to adjust the usage as well.

data_steward/utils/bq.py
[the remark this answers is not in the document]
nishanthpp93: I think it makes sense to change it from `None` to `True`. We shouldn't have to stick to legacy conventions in new code, but thanks for the reasoning on why it was added. When we make the change from `bq_utils` to `bq`, we need to make sure all usages conform to the new convention.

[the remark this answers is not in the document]
amcgrenera-vumc: Changing it to `True` will then default to `False` if the overwrite_existing parameter is not given. The currently logic I think is just wonky to being with. Both `None` and `not overwrite_existing` should both default to `True` if we want to follow what it says in the notes. I'm thinking it would make the most sense to just change the logic to the following:

```
if overwrite_existing is False:
   overwrite_existing = False
else:
   overwrite_existing = True
```

that way both `True` and None will default to `True`

data_steward/utils/bq.py
nishanthpp93: I think it would be good to have a check if it exists before starting to create the dataset. If `overwrite_existing` is set to `False` and it exists, an error should be raised stating that the dataset already exists. Is this the same error raised by `define_dataset`? 
It's good to add the `label_or_tag` parameter as well since we want to enforce this practice (sandbox should have the `sandbox` tag I think). Try making the changes incl. usages, but if the scope blows up a lot, add it as a new ticket and a different PR.

data_steward/utils/bq.py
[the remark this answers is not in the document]
nishanthpp93: This snippet of code is not really changing much. I think an error should be raised if `overwrite_existing` is neither `True` nor `False` (the value in the function definition should be `True` as well). If it is boolean, there are no issues and the `exists_ok=overwrite_existing` parameter setting should handle those cases.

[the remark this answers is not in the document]
amcgrenera-vumc: yeah that makes sense to me, and with `sandbox.create_sandbox_dataset` it is always declaring that parameter as `False`. So in this case do I just remove that if clause entirely?

data_steward/utils/bq.py
amcgrenera-vumc: They are the same errors raised if `overwrite_existing` is set to False and it already exists in `define_dataset`, my only concern is if someone wants to set `overwrite_existing` to `True` (not sure if that will even be wanted). Then the function `create_dataset` would not work with `define_dataset`. But for now, all `sandbox.create_sandbox_dataset` sets that parameter to `False` so that won't be an issue with any current usages.

There is no sandbox tag, but I can define that in `create_sandbox_dataset`. Is there anything specific I should set these to? For now I am leaving as blank strings. 

Take a look when you get the chance and let me know if my code makes more sense now using `define_dataset`.

data_steward/utils/bq.py
```diff
@@ -286,3 +286,46 @@ def get_latest_validation_dataset_id(project_id):
...
                           reverse=True)[0][1]
     return None
+
+
+def create_dataset(project_id,
+                   dataset_id,
+                   description=None,
+                   friendly_name=None,
+                   label_or_tag=None,
+                   overwrite_existing=None):
```
nishanthpp93, line 296: I think `overwrite_existing=False` could work here. Is there any specific reason we need to use `None`?

data_steward/utils/bq.py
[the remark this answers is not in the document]
nishanthpp93: Yeah removing is best.

data_steward/utils/bq.py
amcgrenera-vumc: yep you're totally right, updating now!

data_steward/utils/bq.py
nishanthpp93: I think there can be a check for existence before that step. `define_dataset` should not need to be called if the dataset already exists.
I think you can leave the labels/tags blank for now, at least until we come up with a convention.

tests/integration_tests/data_steward/sandbox_test.py
```diff
@@ -0,0 +1,35 @@
...
+class RetractDataBqTest(unittest.TestCase):
+
+    @classmethod
+    def setUpClass(cls):
+        print('**************************************************************')
+        print(cls.__name__)
+        print('**************************************************************')
+
+    def setUp(self):
+        self.project_id = 'aou-res-curation-test'
```
nishanthpp93, line 18: This needs to be updated to use `app_identity.get_application_id`

data_steward/utils/bq.py
amcgrenera-vumc: I've added in a check!

tests/integration_tests/data_steward/sandbox_test.py
```diff
@@ -0,0 +1,36 @@
+#Python imports
+import unittest
+
+# Project Imports
+from data_steward import sandbox
```
nishanthpp93, line 5: This import is failing in the integration test. `data_steward` is the home directory in the python path so we don't need to specify it.
```suggestion
import sandbox
```

data_steward/utils/bq.py
```diff
@@ -286,3 +287,51 @@ def get_latest_validation_dataset_id(project_id):
...
+                          tags for the dataset = Dict[str, '']
+    :param overwrite_existing: determine if dataset should be overwritten if already exists, defaults to true/overwrite
+    :return: a new dataset returned from the API
+    """
+    client = get_client(project_id)
+
+    # Check to see if dataset already exists if overwrite_existing is False
+    all_datasets = [d.dataset_id for d in list_datasets(project_id)]
+    if overwrite_existing is False and dataset_id in all_datasets:
+        raise RuntimeError("Dataset already exists")
```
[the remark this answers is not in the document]
amcgrenera-vumc: good suggestion!

amcgrenera-vumc:
hey Nishanth-- I'm confused about the integration test for `remove_operational_pii_fields_test.py`.  Currently `sandbox.create_sandbox_dataset` does not take the `overwrite_existing` parameter, it sets it to `False` when it calls `bq.create_dataset`. If i added in that parameter, would I need to add that in all usages?

nishanthpp93:
That's a good point. Maybe the test needs to use `get_or_create_dataset` for the sandbox since it already exists instead of creating it.

amcgrenera-vumc:
Good call, I've made that change! My linting wasn't passing so I will make sure all checks are passing, but I've adjusted the code per all our comments.

amcgrenera-vumc:
@nishanthpp93 thanks for adding these tests! The circleci checks are passing, so I will go ahead and merge.

PR #525, opened 2020-06-01
[DC-702] Add cleaning rule from DC-520 to the pipeline

nishanthpp93:
* Add new obs source concept
* Use Jinja Template instead of format

### DIFF
--- a/data_steward/cdr_cleaner/cleaning_rules/rdr_observation_source_concept_id_suppression.py
+++ b/data_steward/cdr_cleaner/cleaning_rules/rdr_observation_source_concept_id_suppression.py
@@ -16,6 +16,7 @@ Include the observation_source_concept_id 903079 to the list
 import logging
 
 # Third party imports
+from jinja2 import Template
 
 # Project imports
 from common import OBSERVATION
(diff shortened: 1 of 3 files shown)


data_steward/cdr_cleaner/cleaning_rules/rdr_observation_source_concept_id_suppression.py
```diff
@@ -6,36 +6,54 @@
...
+from common import OBSERVATION
 from cdr_cleaner.cleaning_rules.base_cleaning_rule import BaseCleaningRule
 from constants.bq_utils import WRITE_TRUNCATE
 from constants.cdr_cleaner import clean_cdr as cdr_consts
 
 LOGGER = logging.getLogger(__name__)
 
+OBS_SRC_CONCEPTS = '43530490,43528818,43530333,903079'
+
 SAVE_TABLE_NAME = 'dc_529_obs_rows_dropped'
```
amcgrenera-vumc, line 31: Should this be changed to `dc_702_obs_rows_dropped`? Since DC-529 only included the first three observation concept ids? or a combination of the two issue numbers `dc_529_702_obs_rows_dropped`? Not sure how to handle sandbox names for issues with more than one ticket, may be a conversation for eng sync meeting.

data_steward/cdr_cleaner/cleaning_rules/rdr_observation_source_concept_id_suppression.py
```diff
@@ -6,36 +6,54 @@
...
 
+ISSUE_NUMBERS = ['DC-529', 'DC-702']
```
amcgrenera-vumc, line 33: Should 'DC-520' also be added to this?

amcgrenera-vumc:
@nishanthpp93 this looks good to me and I have all tests passing when I run locally. Just a couple comments on the ticket numbers.

data_steward/cdr_cleaner/cleaning_rules/rdr_observation_source_concept_id_suppression.py
nishanthpp93: I think it should.

data_steward/cdr_cleaner/cleaning_rules/rdr_observation_source_concept_id_suppression.py
nishanthpp93: Good point! I think the table name could get longer with 502. Would you mind bringing this up at the meeting? We should set a naming convention for sandboxed tables. If we change the name it may make it harder to find previous sandboxed tables with the old name, but not sure if they're important.

amcgrenera-vumc: I agree, don't like when table names are too long. For now I think we can just leave it as is. But I will add this to the agenda for the eng sync tomorrow!

PR #526, opened 2020-06-02
[DC-899] Update rdr dataset name in import_rdr_omop.sh

ksdkalluri:


data_steward/tools/import_rdr_omop.sh
```diff
@@ -58,7 +58,7 @@ export GOOGLE_CLOUD_PROJECT="${app_id}"
 export BIGQUERY_DATASET_ID="${OUTPUT_DATASET}"
 
 today=$(date '+%Y%m%d')
-RDR_DATASET="${today}_rdr"
+RDR_DATASET="rdr${today}"
```
nishanthpp93, line 61: Can the `RDR_DATASET` be changed to a user parameter? I think we want the date to be based on when the RDR export happened rather than the current date, and it's best to let the user decide the name. The parameter could be named `rdr_dataset` or `rdr_dataset_id`.

data_steward/tools/import_rdr_omop.sh
```diff
@@ -57,12 +62,12 @@ export GOOGLE_APPLICATION_CREDENTIALS="${KEY_FILE}"
...
 
-today=$(date '+%Y%m%d')
-RDR_DATASET="${today}_rdr"
+
+RDR_DATASET="rdr${RDR_UPLOAD_DATE}"
 
 source "${TOOLS_DIR}/set_path.sh"
 
-bq mk -f --description "RDR DUMP loaded from ${RDR_DIRECTORY} on ${today}" "${GOOGLE_CLOUD_PROJECT}:${RDR_DATASET}"
+bq mk -f --description "RDR DUMP loaded from ${RDR_DIRECTORY} on ${RDR_UPLOAD_DATE}" "${GOOGLE_CLOUD_PROJECT}:${RDR_DATASET}"
```
nishanthpp93, line 70: nit: Change to `dated` instead of `on` since it can be confusing.
```suggestion
bq mk -f --description "RDR DUMP loaded from ${RDR_DIRECTORY} dated ${RDR_UPLOAD_DATE}" "${GOOGLE_CLOUD_PROJECT}:${RDR_DATASET}"
```

PR #532, opened 2020-06-09
[DC-859] Adjust de-id generalizations to account for multiple selecti…

amcgrenera-vumc:
…ons for sexual orientation

### DIFF
--- a/data_steward/deid/config/ids/config.json
+++ b/data_steward/deid/config/ids/config.json
@@ -62,19 +62,14 @@
         "SEXUAL-ORIENTATION":[
             {
                 "comment": [
-                    "aggregate multi-sexual orientation answers before generalizing ",
-                    "single sexual orientation answers."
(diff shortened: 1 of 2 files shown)


amcgrenera-vumc:
You're right, I was basing this off the `RACE` example. I added in `True` for duplicates in the single generalization, which didn't seem to help. I also tried the following copying from `GENDER` but still getting the same results....

```{
                "comment": [
                    "multi sex orientation generalization rule, the values the count is ",
                    "limited to is provided by the 'on' key.  it is best to ",
                    "perform aggregate functions before other generalizations"
                ],
                "apply": "COUNT",
                "into": 2000000003,
                "qualifier": "> 1",
                "values": [903096, 903079, 1585901, 1585902, 1585903, 1585904],
                "drop_duplicates": "True"
            }
```

Another strange thing, it does seem to drop all other duplicates, except for my test data..

PR #535, opened 2020-06-16
[EDQ-522] Increase versatility of DQ issue panel reporter

noahgengel:
* DQ issue panel can now accommodate sites that appear in 'newer' analytics reports that do not exist in 'reference' analytics reports 
* Adds saou_uabhs to the dictionaries and lists
* Newer dates in applicable modules
* Docstring formatting in the DQ issue panel

### DIFF
--- a/data_steward/analytics/table_metrics/metrics_over_time/dictionaries_and_lists.py
+++ b/data_steward/analytics/table_metrics/metrics_over_time/dictionaries_and_lists.py
@@ -275,7 +275,8 @@ full_names = {
     "cpmc_ucd": "UC Davis",
     "ipmc_rush": "Rush University",
     "va": "United States Department of Veterans Affairs - Boston",
-    "saou_umc": "University Medical Center (UA Tuscaloosa)"
+    "saou_umc": "University Medical Center (UA Tuscaloosa)",
+    "saou_usahs": "University of South Alabama"
 }
 
 row_count_col_names = [
--- a/data_steward/analytics/tools/dq_issue_tracker/dictionaries_and_lists.py
+++ b/data_steward/analytics/tools/dq_issue_tracker/dictionaries_and_lists.py
@@ -78,7 +78,7 @@ relevant_links = {
     "drug_success":
     "https://sites.google.com/view/ehrupload/"
     "data-quality-metrics/drug-ingredient-integration-rate?authuser=0",
-    
+
     "erroneous_dates":
     "https://sites.google.com/view/ehrupload/"
     "data-quality-metrics/erroneous-dates?authuser=0",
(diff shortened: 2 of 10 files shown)


data_steward/analytics/table_metrics/metrics_over_time/dictionaries_and_lists.py
```diff
@@ -275,7 +275,8 @@
     "cpmc_ucd": "UC Davis",
     "ipmc_rush": "Rush University",
     "va": "United States Department of Veterans Affairs - Boston",
-    "saou_umc": "University Medical Center (UA Tuscaloosa)"
+    "saou_umc": "University Medical Center (UA Tuscaloosa)",
+    "saou_usahs": "University of South Alabama"
```
nishanthpp93, line 279: All hpo-related items need to use the lookup table. We shouldn't need to update a new site in multiple places

data_steward/analytics/tools/dq_issue_tracker/dictionaries_and_lists.py
```diff
@@ -152,7 +152,8 @@
     "cpmc_ucd": "UC Davis",
     "ipmc_rush": "Rush University",
     "va": "United States Department of Veterans Affairs - Boston",
-    "saou_umc": "University Medical Center (UA Tuscaloosa)"
+    "saou_umc": "University Medical Center (UA Tuscaloosa)",
+    "saou_usahs": "University of South Alabama"
```
nishanthpp93, line 156: Same note here

data_steward/analytics/table_metrics/metrics_over_time/dictionaries_and_lists.py
noahgengel: @nishanthpp93 
This script actually does not use BigQuery at all (it is intended to be used on the reports that I run every week) so we cannot use a lookup table unless we integrate this into the pipeline.

nishanthpp93: I see. Can this still use BigQuery without being integrated into the pipeline?

noahgengel: @nishanthpp93 Potentially but that would be a big change - something that would warrant its own ticket and need to be tested thoroughly. I can make a ticket in EDQ to investigate whether or not that is feasible.

nishanthpp93: We shouldn't be manually updating sites in different places because it's easy to miss them. Keep me in the loop on the ticket as well. I'll merge this for now.

noahgengel: @nishanthpp93 I documented this in EDQ-534. It is low priority at the moment (see the 'NOTE') so it will fall under the responsibilities of my successors.

PR #536, opened 2020-06-16
[DC-700] Add cleaning rule from DC-414 to the pipeline

ksdkalluri:
[DC-700] Update unit normalization cleaning rule to extend BaseCleaningRule.
Add upload_csv_data_to_bq_table() function in bq.
Add _unit_mapping.csv and _unit_mapping.json(schema)

[DC-700]: https://precisionmedicineinitiative.atlassian.net/browse/DC-700


data_steward/cdr_cleaner/cleaning_rules/unit_normalization.py
```diff
@@ -0,0 +1,302 @@
...
+    import cdr_cleaner.args_parser as parser
+    import cdr_cleaner.clean_cdr_engine as clean_engine
+
+    ARGS = parser.parse_args()
+    clean_engine.add_console_logging(ARGS.console_log)
+    unit_normalization = UnitNormalization(ARGS.project_id, ARGS.dataset_id,
+                                           ARGS.sandbox_dataset_id)
+    client = bq.get_client(ARGS.project_id)
+    UnitNormalization.setup_rule(unit_normalization, client=client)
+    query_list = UnitNormalization.get_query_specs(unit_normalization)
```
lrwb-aou, line 298: You would call these methods from the `unit_normalization` instance you created.  Like:
```
# save the result in case you want to use it.  Otherwise, you can just call the method without assigning the return value to a variable
result = unit_normalization.setup_rule(client)
query_list = unit_normalization.get_query_specs()
```

data_steward/cdr_cleaner/cleaning_rules/unit_normalization.py
```diff
@@ -0,0 +1,302 @@
...
+        desc = 'Units for labs/measurements will be normalized using unit_mapping.csv.'
+        super().__init__(issue_numbers=['DC414', 'DC700'],
+                         description=desc,
+                         affected_datasets=[cdr_consts.DEID_CLEAN],
+                         affected_tables=['measurement'],
+                         project_id=project_id,
+                         dataset_id=dataset_id,
+                         sandbox_dataset_id=sandbox_dataset_id)
+
+    def get_mapping_list(self, unit_mappings_list):
```
lrwb-aou, line 198: Since this is a draft PR, I don't know if you're finished with this or not.  But this isn't being used anywhere right now.

data_steward/cdr_cleaner/cleaning_rules/unit_normalization.py
```diff
@@ -0,0 +1,302 @@
...
+
+        # creating _unit_mapping table
+        unit_mapping_table = f'{self.get_project_id()}.{self.get_dataset_id()}.{UNIT_MAPPING_TABLE}'
+        try:
+            bq.create_tables(client, self.get_project_id(),
+                             [unit_mapping_table])
+            # Uploading data to _unit_mapping table
+            unit_mappings_csv_path = os.path.join(resources.resource_files_path,
+                                                  UNIT_MAPPING_FILE)
+            result = bq.upload_csv_data_to_bq_table(
```
lrwb-aou, line 234: This result should check for an error property on the result.  The errors are not automatically raised.  This is the [LoadJob](https://googleapis.dev/python/bigquery/latest/generated/google.cloud.bigquery.job.LoadJob.html) documentation.  This is a [stack overflow answer](https://stackoverflow.com/questions/57039771/how-to-get-detailed-big-query-error-by-using-python) showing how it is used.  We should either do this here or possibly in the utility you have created.

data_steward/cdr_cleaner/cleaning_rules/unit_normalization.py
ksdkalluri: my bad this will be removed.

data_steward/cdr_cleaner/cleaning_rules/unit_normalization.py
```diff
@@ -0,0 +1,283 @@
...
+            measurement_table=MEASUREMENT)
+
+        update_query = dict()
+        update_query[cdr_consts.QUERY] = UNIT_NORMALIZATION_QUERY.render(
+            project_id=self.get_project_id(),
+            dataset_id=self.get_dataset_id(),
+            unit_table_name=UNIT_MAPPING_TABLE,
+            measurement_table=MEASUREMENT)
+        update_query[cdr_consts.DESTINATION_TABLE] = MEASUREMENT
+        update_query[cdr_consts.DISPOSITION] = bq_consts.WRITE_TRUNCATE
```
mark-velez, line 246: `WRITE_TRUNCATE` would wipe out the schema, right? Have you ever looked into instead using `CREATE OR REPLACE TABLE ... AS query_statement` ([source](https://cloud.google.com/bigquery/docs/reference/standard-sql/data-definition-language#creating_a_clustered_table_from_the_result_of_a_query))?

```sql 
CREATE OR REPLACE TABLE source_dataset.measurement(
  measurement_id INT64 NOT NULL OPTIONS(description='...')
  /* ... */
) AS 
SELECT m.* 
FROM source_dataset.measurement m
WHERE /* ... */
```

I'm not suggesting we do this here, but it is something I'm trying out somewhere else and I wanted to know if you'd looked into it already.

data_steward/resource_files/fields/_unit_mapping.json
```diff
@@ -0,0 +1,22 @@
+[
```
mark-velez, line 1: Good idea to prefix this with '_': it will be recognized as internal by `resources.cdm_schemas()`. :)

tests/integration_tests/data_steward/cdr_cleaner/cleaning_rules/unit_normalization_test.py
```diff
@@ -0,0 +1,84 @@
...
+        self.assertRaises(NotFound, table_exists(client, intermediary_table))
+
+        # run setup_rule and see if the table is created
+        self.query_class.setup_rule(client)
+
+        try:
+            client.get_table(intermediary_table)
+            result = True
+        except NotFound:
+            result = False
```
mark-velez, line 75:
```suggestion
        actual_table = client.get_table(intermediary_table)
        self.assertIsNotNone(actual_table.created)
```

tests/integration_tests/data_steward/cdr_cleaner/cleaning_rules/unit_normalization_test.py
```diff
@@ -0,0 +1,84 @@
...
+
+        client = bq.get_client(self.project_id)
+
+        def table_exists(bq_client, table):
+            try:
+                bq_client.get_table(table)
+            except NotFound:
+                return
+
+        self.assertRaises(NotFound, table_exists(client, intermediary_table))
```
mark-velez, line 66: I contest the value of this part of the test- whether or not the lookup exists in the test environment is up to how we have set it up. We should be sure it doesn't exist.

tests/unit_tests/data_steward/cdr_cleaner/cleaning_rules/unit_normalization_test.py
```diff
@@ -0,0 +1,110 @@
...
+
+    def test_setup_rule(self):
+        # Test
+        def run_setup(client):
+            try:
+                self.query_class.setup_rule(client)
+            except RuntimeError:
+                return
+
+        self.assertRaises(RuntimeError, run_setup(self.client))
```
mark-velez, line 43: Sorry, not sure the idea behind `run_setup(client)` above- here's my suggestion.
```suggestion
        with self.assertRaises(RuntimeError) as c:
             self.query_class.setup_rule(None)
        self.assertEqual(str(c.exception), 'Specify BigQuery client object')
```

tests/integration_tests/data_steward/cdr_cleaner/cleaning_rules/unit_normalization_test.py
ksdkalluri: I agree.

data_steward/cdr_cleaner/cleaning_rules/unit_normalization.py
ksdkalluri: I looked into this solution and thought this would change a lot of our cleaning process and wanted to discuss in the ENG-Sync before actually making that decision.

mark-velez: Ok, just checking. Closing this thread.

data_steward/cdr_cleaner/cleaning_rules/unit_normalization.py
```diff
@@ -0,0 +1,278 @@
...
+        """
+        Validates the cleaning rule which deletes or updates the data from the tables
+        """
+        pass
+
+    def get_sandbox_tablenames(self):
+        sandbox_table_names = list()
+        sandbox_table_names.append(self._issue_numbers[0].lower() + '_' +
+                                   self._affected_tables[0])
+        return sandbox_table_names
```
mark-velez, line 261: This function `get_sandbox_tablenames()` might be more appropriate for rule documentation/logging than the one usage in `get_query_specs()` where it looks like we want a helper to fetch the name of the one sandbox table this rule happens to have. If this rule should need more than one sandbox table then `get_sandbox_tablenames()` won't be fit for this purpose (which list position is the one I'm looking for?). So I suggest a helper function like below. If this rule were eventually to evolve to need more than one sandbox table, then it could be refactored to accept a `base_name` parameter or something.

```suggestion
    def get_sandbox_table_name(self):
        return f'{self._issue_numbers[0].lower()}_measurement'

    def get_sandbox_tablenames(self):
        return [get_sandbox_table_name()]
```

tests/integration_tests/data_steward/cdr_cleaner/cleaning_rules/unit_normalization_test.py
```diff
@@ -0,0 +1,69 @@
...
+        self.dataset_id = os.environ.get('COMBINED_DATASET_ID')
+        self.sandbox_id = self.dataset_id + '_sandbox'
+
+        self.query_class = UnitNormalization(self.project_id, self.dataset_id,
+                                             self.sandbox_id)
+
+        self.assertEqual(self.query_class.get_project_id(), self.project_id)
+        self.assertEqual(self.query_class.get_dataset_id(), self.dataset_id)
+        self.assertEqual(self.query_class.get_sandbox_dataset_id(),
+                         self.sandbox_id)
```
mark-velez, line 50: Code duplication: this is already in the unit test

tests/integration_tests/data_steward/cdr_cleaner/cleaning_rules/unit_normalization_test.py
```diff
@@ -0,0 +1,69 @@
...
+        client = bq.get_client(self.project_id)
+        # run setup_rule and see if the table is created
+        self.query_class.setup_rule(client)
+
+        actual_table = client.get_table(intermediary_table)
+        self.assertIsNotNone(actual_table.created)
+
+        query = test_query.render(intermediary_table=intermediary_table)
+        result = bq.query(query, self.project_id)
+        self.assertEqual(result.empty, False)
```
mark-velez, line 66: How about another test here to see that `setup_rule()` behaves as expected when the table exists already?

data_steward/utils/bq.py
```diff
@@ -65,6 +65,39 @@ def get_table_schema(table_name, fields=None):
...
+
+    LOGGER.info(f"Uploading {fq_file_path} data to {dataset_id}.{table_name}")
+    with open(fq_file_path, "rb") as source_file:
+        job = client.load_table_from_file(source_file,
+                                          table_ref,
+                                          job_config=job_config)
+    try:
+        result = job.result()  # Waits for table load to complete.
+    except (BadRequest, OSError, AttributeError, TypeError, ValueError):
+        raise RuntimeError(f"Unable to load data to table {table_name}")
```
mark-velez, line 96: Good job throwing the error back. I suggest logging the exception as well so we can troubleshoot it.
```suggestion
        message = f"Unable to load data to table {table_name}"
        LOGGER.exception(message)
        raise RuntimeError(message)
```

tests/integration_tests/data_steward/cdr_cleaner/cleaning_rules/unit_normalization_test.py
```diff
@@ -0,0 +1,73 @@
...
+        actual_table = client.get_table(intermediary_table)
+        self.assertIsNotNone(actual_table.created)
+
+        # test if exception is raised if table already exists
+        with self.assertRaises(RuntimeError) as c:
+            self.query_class.setup_rule(client)
+
+        self.assertEqual(
+            str(c.exception),
+            "Unable to create tables: ['aou-res-curation-test.krishna_combined._unit_mapping']"
```
mark-velez, line 65: This test does not pass when I run it locally. Use var `intermediary_table` to get the table name.

tests/integration_tests/data_steward/cdr_cleaner/cleaning_rules/unit_normalization_test.py
```diff
@@ -0,0 +1,73 @@
...
+        )
+
+        query = test_query.render(intermediary_table=intermediary_table)
+        result = bq.query(query, self.project_id)
```
mark-velez, line 69: I think we agreed we will change `bq.query()` so that it returned `QueryJob` rather than assuming a dataframe is always wanted- please use `client.query(query)` in the meantime.

tests/unit_tests/data_steward/cdr_cleaner/cleaning_rules/unit_normalization_test.py
```diff
@@ -0,0 +1,107 @@
...
+    def setUpClass(cls):
+        print('**************************************************************')
+        print(cls.__name__)
+        print('**************************************************************')
+
+    def setUp(self):
+        self.project_id = 'foo_project'
+        self.dataset_id = 'foo_dataset'
+        self.sandbox_id = 'foo_sandbox'
+        self.client = None
```
mark-velez, line 25: I think this is not used and can be removed.

ksdkalluri: These are being used in setup_rule tests.

PR #537, opened 2020-06-17
[DC-729] Create an email notification system for sites when they make a new submission

nishanthpp93:
* [DC-729] Add mandrill package and dependencies
* [DC-729] Fetch contact list, create email notification script
* [DC-729] Add email notification test, verify sheet data pulled
* [DC-729] Add stubs, refactor process_hpo to facilitate email
* [DC-729] Add template for email, add reporting function in main
* [DC-729] Add message contents and constants
* [DC-729] Add timezone, logo, tests
* [DC-729] Fix failing tests

[DC-729]: https://precisionmedicineinitiative.atlassian.net/browse/DC-729
[DC-729]: https://precisionmedicineinitiative.atlassian.net/browse/DC-729
[DC-729]: https://precisionmedicineinitiative.atlassian.net/browse/DC-729
[DC-729]: https://precisionmedicineinitiative.atlassian.net/browse/DC-729
[DC-729]: https://precisionmedicineinitiative.atlassian.net/browse/DC-729
[DC-729]: https://precisionmedicineinitiative.atlassian.net/browse/DC-729
[DC-729]: https://precisionmedicineinitiative.atlassian.net/browse/DC-729
[DC-729]: https://precisionmedicineinitiative.atlassian.net/browse/DC-729


nishanthpp93:
Thanks for reviewing @ksdkalluri.
I didn't want to use `five_person_results.html`, but it is the only results file that we have which is updated. I had to generate it after running validation on our five person test data through the integration test. I removed `person_only_results.html` since it uses the old html format which we no longer use.
Other than that the only html files we have are `bootstrap_style.html` which is not a html document and only contains styling information, and `hpo_report.html` which is actually a Jinja template file and not a html document that can be used (it contains { % variable } which can cause issues while processing it as a html document).

I also want to mention that this html document will not be used in any test right now, since I have disabled the integration test that sends the email.

PR #539, opened 2020-06-23
[EDQ-355] Notebook for Incomplete Measurement Mappings

noahgengel:
* Notebook contains information with respect to measurement_source_concept_ids/measurement_source_values that failed to map to the appropriate measurement_concept_ids/value_as_concept_ids.
* Notebook can be expanded in the future with additional concept_class_ids and/or concept relationships.
* As of right now, the findings indicate that 'incomplete' mappings are due to ETL issues confined to a small number of sites. Consequently, further explorations on this front can be deprioritzed in favor of system-wide issues.


noahgengel:
@nishanthpp93 could you perhaps take a look at this when you get the chance? This is referring to a finding we have had for quite some time. I want to ensure it gets populated into the GH before next week if possible.

PR #540, opened 2020-06-25
[EDQ-513] Integrate the 'Visit Date Disparity' Metric into the Current Schema

noahgengel:
* Reworked the metrics_over_time module by implementing f-strings, leveraging a more robust 'constants' file, and adding visit_date_disparity metrics where applicable
* Reworked the dq_issue_tracker module by implementing f-strings, implementing a 'constants' file, and adding visit_date_disparity metrics where applicable
* Reworked the email_generator module by implementing f-strings, implementing a 'constants' file, and adding visit_date_disparity metrics where applicable
* Changed the column names for the two visit_date_disparity notebooks to ensure they more closely resemble the names of similar notebooks
* Added a visit_date_disparity visualization notebook.

### DIFF
--- a/data_steward/analytics/tools/dq_issue_tracker/general_functions.py
+++ b/data_steward/analytics/tools/dq_issue_tracker/general_functions.py
@@ -11,7 +11,7 @@ import os
 import pandas as pd
 import sys
 import datetime
-from class_definitions import HPO
+import constants
 
 
 def load_files(sheet_name, file_name):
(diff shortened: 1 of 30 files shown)


noahgengel:
@nishanthpp93 Could you merge this when you get the chance. I want to ensure this is approved before my final day if possible.

Also - as an aside - I recognize that there is a great deal of redundant code and many classes/constants that could be localized into a single file. I figured, however, that I should make this PR before I leave the DRC to ensure no work is lost. The potential means of improvement (namely localization) are documented in EDQ-540.

data_steward/analytics/tools/dq_issue_tracker/general_functions.py
```diff
@@ -45,24 +45,24 @@ def load_files(sheet_name, file_name):
...
 
         if sheet.empty:
-            print("WARNING: No data found in the {} sheet "
-                  "in dataframe {}".format(
-                   sheet_name, file_name))
+            print(f"""WARNING: No data found in the {sheet_name} sheet
+                  in dataframe {file_name}""")
 
     except Exception as ex:  # sheet not in specified excel file
         if type(ex).__name__ == "FileNotFoundError":
```
nishanthpp93, line 52: You could use this instead
```suggestion
    except FileNotFoundError as ex:  # sheet not in specified excel file
```
This will run into other errors and fail, but the behavior will be the same as using `sys.exit(0)`, without the print statement.
Using `Exception` is not typically recommended since it is a blanket term and will not allow the script to fail when it should, potentially causing issues downstream. I'm guessing that's why the script is set to exit with `sys.exit(0)`.

Also, this is a job for Python's `logging` module. You can set a logger to do this, and call `logger.exception(msg)` where the message is your print statement. For other Exceptions, the program should be allowed to fail.

All of the print messages should ideally be handled by a logging module with different levels, like `info`, `warning`, `error` etc., but this can be a future improvement. You might want to add this to the ticket.

PR #543, opened 2020-06-29
[EDQ-542] Fix Logical Inconsistencies within the VDD Metric

noahgengel:
* Aligns the VDD metric with the specifications outlined in EDQ-514
* Ensures the presence of all 8 logic points, looks only at before visit start times/after visit end times, always use the start date/datetime fields, etc.


noahgengel:
forgot to also update the `combined_deid` version of this script. closing and creating a new PR.

PR #551, opened 2020-07-08
[DC-704] Add cleaning rule from DC-539 to the pipeline

ksdkalluri:


data_steward/cdr_cleaner/clean_cdr.py
```diff
@@ -157,6 +158,7 @@
     (remove_aian_participants.get_queries,),
     (validate_missing_participants.delete_records_for_non_matching_participants,
     ),
+    (DropDuplicatePpiQuestionsAndAnswers,),
```
amcgrenera-vumc, line 161: I noticed in the ticket it said this should be applied to the `RDR export`-- should this be moved up to `RDR_CLEANING_CLASSES`?

data_steward/cdr_cleaner/cleaning_rules/drop_duplicate_ppi_questions_and_answers.py
```diff
@@ -0,0 +1,227 @@
...
+) on (trim(lower(observation_source_value))=lower(concept_code))
+join `{{project}}.{{dataset}}.concept_relationship` on (concept_id_1=concept_id)
+where vocabulary_id='PPI' and c.invalid_reason is null
+and relationship_id = 'Maps to')
+""")
+
+DELETE_DUPLICATE_ANSWERS = jinja_env.from_string("""
+select * 
+from `{{project}}.{{dataset}}.{{clinical_table_name}}` o
+where value_source_concept_id in (select value_source_concept_id
```
amcgrenera-vumc, line 84: Since this isn't a DML statement, and you are truncating the query return to overwrite the table, shouldn't this be `NOT IN`?

data_steward/cdr_cleaner/clean_cdr.py
ksdkalluri: Oh my bad! I'll change it. Thanks for finding it!

data_steward/cdr_cleaner/cleaning_rules/drop_duplicate_ppi_questions_and_answers.py
ksdkalluri: This should be, I am updating it

amcgrenera-vumc:
@ksdkalluri thanks for the changes, this looks good to me!

PR #552, opened 2020-07-09
[DC-964] Access issue preventing retrieval of HPO contact list via query

nishanthpp93:
* Add logging for email client and stacktrace
* Use REST API to access spreadsheet linked table

### DIFF
--- a/data_steward/utils/auth.py
+++ b/data_steward/utils/auth.py
@@ -17,7 +17,8 @@ def get_access_token(scopes):
     :param scopes: List of Google scopes as strings
     :return: access token as string
     """
-    url = f'{METADATA_URL}instance/service-accounts/{SERVICE_ACCOUNT}/token?scopes={scopes}'
+    scopes_str = ','.join(scopes)
+    url = f'{METADATA_URL}instance/service-accounts/{SERVICE_ACCOUNT}/token?scopes={scopes_str}'
     # Request an access token from the metadata server.
     r = requests.get(url, headers=METADATA_HEADERS)
     r.raise_for_status()
--- a/data_steward/utils/auth.py
+++ b/data_steward/utils/auth.py
@@ -1,0 +1,26 @@
+# Python imports
+import requests
+
+# Third party imports
+
+# Project imports
+
+METADATA_URL = 'http://metadata.google.internal/computeMetadata/v1/'
+METADATA_HEADERS = {'Metadata-Flavor': 'Google'}
+SERVICE_ACCOUNT = 'default'
+
+
+def get_access_token(scopes):
+    """
+    Retrieves an access_token in App Engine for the default service account
+
+    :param scopes: List of Google scopes as strings
+    :return: access token as string
+    """
+    url = f'{METADATA_URL}instance/service-accounts/{SERVICE_ACCOUNT}/token?scopes={scopes}'
+    # Request an access token from the metadata server.
+    r = requests.get(url, headers=METADATA_HEADERS)
+    r.raise_for_status()
+    # Extract the access token from the response.
+    access_token = r.json()['access_token']
+    return access_token
--- a/data_steward/utils/auth.py
+++ b/data_steward/utils/auth.py
@@ -1,5 +1,5 @@
 # Python imports
-import requests
+import requests as req
 
 # Third party imports
 from google.auth import iam
@@ -23,7 +23,7 @@ def get_access_token(scopes):
     scopes_str = ','.join(scopes)
     url = f'{METADATA_URL}instance/service-accounts/{SERVICE_ACCOUNT}/token?scopes={scopes_str}'
     # Request an access token from the metadata server.
-    r = requests.get(url, headers=METADATA_HEADERS)
+    r = req.get(url, headers=METADATA_HEADERS)
     r.raise_for_status()
     # Extract the access token from the response.
     access_token = r.json()['access_token']
--- a/data_steward/utils/auth.py
+++ b/data_steward/utils/auth.py
@@ -31,6 +31,14 @@ def get_access_token(scopes):
 
 
 def delegated_credentials(credentials, subject, scopes):
+    """
+    Generate credentials with scopes. This needs the 'SA token creator' role for the SA
+
+    :param credentials: Credentials object to add scopes/subject to
+    :param subject: Subject to add to credentials
+    :param scopes: Scopes to add to credentials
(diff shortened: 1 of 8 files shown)


data_steward/validation/main.py
```diff
@@ -615,7 +640,9 @@ def extract_date_from_rdr_dataset_id(rdr_dataset_id):
         # TODO remove dependence on date string in RDR dataset id
         rdr_date = rdr_date[:4] + '-' + rdr_date[4:6] + '-' + rdr_date[6:]
         return rdr_date
-    raise ValueError(f"{rdr_dataset_id} is not a valid rdr_dataset_id")
+    else:
```
mark-velez, line 643: This is another higher level issue I'm glad you bring up: how to handle unset environment variables. Please put this in a separate branch and create a new ticket as we discussed.

data_steward/validation/email_notification.py
```diff
@@ -11,17 +11,63 @@
...
+
+
+def get_access_token():
+    url = f'{METADATA_URL}instance/service-accounts/{SERVICE_ACCOUNT}/token?scopes={SCOPES}'
+    # Request an access token from the metadata server.
+    r = requests.get(url, headers=METADATA_HEADERS)
+    r.raise_for_status()
+    # Extract the access token from the response.
+    access_token = r.json()['access_token']
+    return access_token
```
mark-velez, line 41: Glad you figured this out! I recommend relocating this to a different package `utils.auth` and refactoring `get_access_token()` so that it
1. receives the requested scope(s) as a parameter `scope: List[str]` and 
1. has a docstring 😄

data_steward/validation/email_notification.py
```diff
@@ -11,17 +11,63 @@
...
+
+
+def post_query(project_id):
+    url = f'https://www.googleapis.com/bigquery/v2/projects/{project_id}/queries'
```
mark-velez, line 45: It's possible someone else may query either the current or some other drive-backed table and run into the same scope issue. We should aim to prevent their having to reinvent this solution.

At minimum I recommend refactoring this by 
1. relocating it to `utils.bq`
1. having it accept the parameters `query: str` and `external_data_scopes:List[str]`
1. adding a docstring 😄 

Please give due diligence (<1 hour) to possibly injecting the token into the client somehow. I'll let you know if I find anything.

tests/integration_tests/data_steward/validation/email_notification_test.py
```diff
@@ -32,7 +31,11 @@ def setUp(self):
             'submission_error': False
         }
 
-    def test_hpo_contact_list(self):
+    @mock.patch(
+        'validation.email_notification.query_sheet_linked_bq_table_app_engine')
+    def test_hpo_contact_list(self, query_rest_api):
+        query_rest_api.return_value = en.query_sheet_linked_bq_table_compute_engine(
```
mark-velez, line 37: I recommended creating the drive-backed table within `EmailNotificationTest.setUp()`- but even overlooking the GDrive aspect and just loading a regular table (using `fake_dict` below) would help better encapsulate the test resources. If neither are feasible then indicate in a comment any special requirements of this test (e.g. account access to a particular spreadsheet,  existence of a table backed by the spreadsheet).

data_steward/validation/email_notification.py
nishanthpp93: Done! Unfortunately, I couldn't find any info on using access tokens for the client library. They really want everything done under the hood and require a credentials object instead to determine the access tokens.

tests/integration_tests/data_steward/validation/email_notification_test.py
nishanthpp93: I think `query_sheet_linked_bq_table_app_engine` and other sheet related functions should be in `utils/bq.py` instead of the email script. What do you think @mark-velez?

nishanthpp93: I also found out that it cannot create a spreadsheet. A spreadsheet needs to be created separately and granted permissions to the service account. It can, however, link the spreadsheet to a table, and I have added that in the test.

mark-velez: I think your inclination that reusable bq-related functions should be relocated to `utils.bq` is right. I recommend moving `post_query` and somehow making it explicit that the function is specialized to handle the scope issue (e.g. refactoring it to be `query_with_scopes(scopes: List[str])`, including a detailed doc string, etc.). I think we should avoid putting anything in `utils.bq` that regards specific tables as it is less likely to be more generally useful.

nishanthpp93: I agree, this was my thought as well. I've made them all of them available for general usage.

data_steward/utils/auth.py
```diff
@@ -0,0 +1,70 @@
...
+        headers = {'Authorization': f'Bearer {access_token}'}
+    """
+    scopes_str = ','.join(scopes)
+    url = f'{METADATA_URL}instance/service-accounts/{SERVICE_ACCOUNT}/token?scopes={scopes_str}'
+    # Request an access token from the metadata server.
+    r = req.get(url, headers=METADATA_HEADERS)
+    r.raise_for_status()
+    # Extract the access token from the response.
+    access_token = r.json()['access_token']
+    return access_token
```
mark-velez, line 32: nit: this is no longer needed, can be removed

data_steward/utils/auth.py
```diff
@@ -0,0 +1,70 @@
...
+
+
+def delegated_credentials(credentials, subject, scopes):
```
mark-velez, line 35: nit: subject can be an optional param
```suggestion
def delegated_credentials(credentials, scopes, subject=None):
```

data_steward/utils/bq.py
```diff
@@ -5,28 +5,37 @@
 import logging
 import os
 
-from google.api_core.exceptions import GoogleAPIError, BadRequest
 # Third-party imports
+from google.api_core.exceptions import GoogleAPIError, BadRequest
 from google.cloud import bigquery
+import google.auth
```
mark-velez, line 11: nit: unused import

nishanthpp93: The `google.auth.default()` in `get_client()` is using this. I'll make this `from google.auth import default`

mark-velez: Sorry, my mistake- looks like a false alarm from the IDE

PR #558, opened 2020-07-13
[DC-971] Develop utility and call data in line with DC-954

dcampbell-vumc:
Created `deactivated_participants.py` to fetch participants with `suspensionStatus = NO_CONTACT`


data_steward/utils/deactivated_participants.py
```diff
@@ -0,0 +1,98 @@
...
+    not_done = True
+    data = []
+
+    while not_done:
+        resp = requests.get(url, headers=headers)
+        if not resp or resp.status_code != 200:
+            print(f'Error: API request failed because {resp}')
+        else:
+            r_json = resp.json()
+            data += r_json['entry']
```
lrwb-aou, line 75: This is a dangerous statement.  If the `entry` key isn't in the dictionary, this will raise a KeyError.  Instead, you should do:  `data += r_json.get('entry', {})`.  This will append an empty dictionary if `entry` is not present.

data_steward/utils/deactivated_participants.py
```diff
@@ -0,0 +1,98 @@
...
+            if 'link' in r_json:
+                url = r_json['link'][0]['url']
+            else:
+                not_done = False
```
lrwb-aou, line 79: nit:  Honestly, this is hard for me to easily conceptualize.  I have an easier time with naming the flag as `done`.
```suggestion
    done = False
    data = []

    while not done:
        resp = requests.get(url, headers=headers)
        
        if not resp or resp.status_code != 200:
            print(f'Error: API request failed because {resp}')
        else:
            r_json = resp.json()
            data += r_json.get('entry', {})
            
            if 'link' in r_json:
                link_obj = r_json.get('link')
                url = link_obj[0].get('url')
            else:
                done = True
```

data_steward/utils/deactivated_participants.py
```diff
@@ -0,0 +1,98 @@
+"""
```
lrwb-aou, line 1: I suggest renaming this file.  The purpose of building the utility is to make something reusable.  This file name is extremely specific.  We will probably need to visit the ParticipantSummary API more for other status types in the future, i.e. `deceased`, `withdrawn`, etc.  

I suggest naming it something like `participant_summary_requests.py`.  This would allow a natural place to add that type of information.

I've made some specific comments in here but have a larger structural suggestion.  I would refactor some of the functionality out of `get_deactivated_participants`.  Specifically, the loop where you are getting your data.  You could move that loop into a helper function that could be use by `get_deactivated_participants` now and by some other participant summary api reader function in the future.  If you would like to, we can have a call and discuss it.

dcampbell-vumc: I understand moving the loop fetching the data out and into it's own function. I wasn't aware that this utility was going to do more than just fetch deactivated participants. It makes sense to rename the file along with moving the fetching data loop as it could be used in various functions that would get other types of data like you mentioned.

data_steward/utils/deactivated_participants.py
```diff
@@ -0,0 +1,98 @@
...
+Utility to fetch and store deactivated participants information.
+
+Original Issues: DC-797, DC-971 (sub-task), DC-972 (sub-task)
+
+The intent of this module is to call and store deactivated participants information by leveraging the RDR Participant
+    Summary API. Deactivated participant information stored is `participantID`, `suspensionStatus`, and `suspensionTime`.
+"""
+
+# Python imports
+import pandas
```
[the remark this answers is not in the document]
dcampbell-vumc: Moved to third party import section

data_steward/utils/deactivated_participants.py
```diff
@@ -0,0 +1,98 @@
+"""
+Utility to fetch and store deactivated participants information.
+
+Original Issues: DC-797, DC-971 (sub-task), DC-972 (sub-task)
```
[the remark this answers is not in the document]
dcampbell-vumc: Removed :p

data_steward/utils/participant_summary_requests.py
```diff
@@ -0,0 +1,114 @@
...
+    participant_data = get_participant_data(url, headers)
+
+    deactivated_participants_cols = columns
+
+    deactivated_participants = []
+    # loop over participant summary records, insert participant data in same order as good_cols.
+    for entry in participant_data:
+        item = []
+        for col in deactivated_participants_cols:
+            for key, val in entry['resource'].items():
```
lrwb-aou, line 106: This is another potentially unsafe dictionary usage.  Suggestion:

` entry.get('resource', {}).items()`.

And I still owe you a KeyError example script.

data_steward/utils/participant_summary_requests.py
```diff
@@ -0,0 +1,114 @@
...
+    }
+
+    # Make request to get API version. This is the current RDR version for reference
+    # See https://github.com/all-of-us/raw-data-repository/blob/master/opsdataAPI.md for documentation of this api.
+    url = 'https://{0}.appspot.com/rdr/v1/ParticipantSummary?_sort=lastModified&suspensionStatus={1}'.format(
+        project_id, 'NO_CONTACT')
+
+    participant_data = get_participant_data(url, headers)
+
+    deactivated_participants_cols = columns
```
lrwb-aou, line 99: Clarification:  Are you planning to pass columns?

tests/unit_tests/data_steward/utils/participant_summary_requests_test.py
```diff
@@ -0,0 +1,91 @@
...
+                    'suspensionTime': '2018-12-07T08:21:14'
+                }
+            }]
+        }
+
+    @mock.patch('utils.participant_summary_requests.get_access_token',
+                return_value='ya29.12345')
+    def test_get_access_token(self, mock_get_access_token):
+
+        self.assertEqual(mock_get_access_token(), self.access_token)
```
lrwb-aou, line 71: I don't think this test works as you intend it to work.  You are mocking the entire function you are trying to unit test.  This will always pass.  It looks like you are asserting that the mock you create (which doesn't run any code, but replaces all of the get_access_token function) equals the access_token value you set in `setUp(self)`.  To test `get_access_token`, you should mock lines 37 and/or 38 in participant_summary_requests.py.  You will have to set a "good enough" mock return value for the function to finish executing.  Then you can `assertEqual(value returned from get_access_token, expected value of access token)`.

tests/unit_tests/data_steward/utils/participant_summary_requests_test.py
```diff
@@ -0,0 +1,91 @@
...
+
+    @mock.patch('utils.participant_summary_requests.get_access_token',
+                return_value='ya29.12345')
+    @mock.patch(
+        'utils.participant_summary_requests.get_deactivated_participants')
+    @mock.patch('utils.participant_summary_requests.requests.get')
+    def test_get_deactivated_participants(self, mock_get,
+                                          mock_get_deactivated_participants,
+                                          mock_access_token):
```
lrwb-aou, line 80: This test should be broken into two unit tests now.  One for `get_deactivated_participants` and one for `get_participant_data`.  The one for `get_participant_data` will make sure the response is iterated correctly by comparing the function return value to the expected return value.  The one for `get_deactivated_participants` will make sure the response data is correctly iterated to make the expected dataframe.  If you've got questions about this, I can talk through it.

data_steward/utils/participant_summary_requests.py
```diff
@@ -0,0 +1,114 @@
...
+    :param sa_key: path that points to your cdr-ops SA key
+
+    :return: returns the access_token
+    """
+
+    scopes = [
+        'https://www.googleapis.com/auth/cloud-platform', 'email', 'profile'
+    ]
+
+    service_account_file = sa_key
```
lrwb-aou, line 32: What happens if you pass `None` in to the `sa_key`?

data_steward/utils/participant_summary_requests.py
dcampbell-vumc: Yes, I thought it would be best not to hardcode the columns as the ticket doesn't state what exact columns to include.

tests/unit_tests/data_steward/utils/participant_summary_requests_test.py
dcampbell-vumc: I was working on changing this test function because you're right, this doesn't really test anything. I was having issues mocking those lines and will try to recreate the issue and discuss if I on a call later.

data_steward/utils/participant_summary_requests.py
dcampbell-vumc: If there isn't an `sa_key` or if it's `None` then an access token can't be generated. I will add a check here to make sure that one is passed.

lrwb-aou: I thought that might be the case.  I was wondering how we should handle that.  We can explicitly check and raise an error if it isn't given or we can let the software raise the error for us when the `service _account_file` variable is used.  But I don't know what error that would raise.

dcampbell-vumc: I'll see what error it raises and post it here.

dcampbell-vumc: Here's the error:

`TypeError: expected str, bytes or os.PathLike object, not NoneType`

dcampbell-vumc:
@lrwb-aou after talking to @nishanthpp93 about the circleCI unit test check error: ``google.auth.exceptions.DefaultCredentialsError: Could not automatically determine credentials. Please set GOOGLE_APPLICATION_CREDENTIALS or explicitly create credentials and re-run the application``
it was determined that I cannot unit test the `get_access_token` due to it explicitly calling external services. I will need to add this to the integration test for [DC-972](https://precisionmedicineinitiative.atlassian.net/browse/DC-972). I noted this change in the ticket. The unit test is now passing.

lrwb-aou:
@dcampbell-vumc Actually, mocking explicitly allows you to mock calls to external products and set expected return values to allow unit testing.  So, unit testing could be done.  But in this instance, the function isn't doing anything else, so mocking won't provide the same type of validation.  You would instead be interested in making sure the mock was called with the expected values.  Yes, an integration test will be nice.  

> @lrwb-aou after talking to @nishanthpp93 about the circleCI unit test check error: `google.auth.exceptions.DefaultCredentialsError: Could not automatically determine credentials. Please set GOOGLE_APPLICATION_CREDENTIALS or explicitly create credentials and re-run the application`
> it was determined that I cannot unit test the `get_access_token` function due to it explicitly calling external services. I will need to add this to the integration test for [DC-972](https://precisionmedicineinitiative.atlassian.net/browse/DC-972). I noted this change in the ticket. The unit test is now passing.

lrwb-aou:
[unit_test_patch.txt](https://github.com/all-of-us/curation/files/5008058/unit_test_patch.txt)
Here is a patch file for you to test.  You can use it by downloading, changing to your `curation` directory, and applying with `git apply <path to downloaded patch>`.  I also ran this command from `curation`.  `pylint data_steward/utils/participant_summary_requests.py `.  This give the following info.  You can ignore the first line and look at the other 4 suggestions.

```
data_steward/utils/participant_summary_requests.py:68:0: W0612: Unused variable 'get_deactivated_participants' (unused-variable)  <<-- this comment does not need to be addressed.
data_steward/utils/participant_summary_requests.py:12:0: C0411: third party import "from google.auth import default" should be placed before "from utils import auth" (wrong-import-order)
data_steward/utils/participant_summary_requests.py:13:0: C0411: third party import "import google.auth.transport.requests as req" should be placed before "from utils import auth" (wrong-import-order)
data_steward/utils/participant_summary_requests.py:14:0: C0411: third party import "import requests" should be placed before "from utils import auth" (wrong-import-order)
data_steward/utils/participant_summary_requests.py:15:0: C0411: third party import "import pandas" should be placed before "from utils import auth" (wrong-import-order)
```

lrwb-aou:
Thank you for the changes.  I'll get out of your way now.  :)

PR #563, opened 2020-07-15
[DC-940] Creates Person_Ext Schema

ksdkalluri:


ksdkalluri:
> One change: the `src_id` field that originally existed in the ext table needs to be added to store mapping information.

We use person_id as is pre-deid and post de-id it will be replaced by research_id. Do we want to have src_id field, since person_ext is populated post de-id.

nishanthpp93:
This has nothing to do with research_id. `src_id` is the source hpo_id which is being populated by `generate_ext_tables.py`. We will also need a ticket to refactor that script if it's not already created.

data_steward/resource_files/fields/person_ext.json
```diff
@@ -0,0 +1,44 @@
...
+        "type": "integer", 
+        "name": "person_id", 
+        "mode": "required", 
+        "description": "A unique identifier for each person."
+    },
+    {
+        "type": "integer",
+        "name": "src_id",
+        "mode": "nullable",
+        "description": "A foreign key to the participant's source EHR site."
```
nishanthpp93, line 12: This should be "string" since it can be like `EHR Site 123`
```suggestion
        "type": "string",
        "name": "src_id",
        "mode": "nullable",
        "description": "A foreign key to the participant's source EHR site."
```

data_steward/resource_files/fields/person_ext.json
```diff
@@ -0,0 +1,44 @@
...
+        "type": "integer", 
+        "name": "state_of_residence_concept_id", 
+        "mode": "nullable", 
+        "description": "[All of Us OMOP extension] A foreign key to the participant's state of residence concept."
+    }, 
+    {
+        "type": "string", 
+        "name": "state_of_residence_source_value", 
+        "mode": "nullable", 
+        "description": "[All of Us OMOP extension] The source code for the participant's state of the residence."
```
nishanthpp93, line 24: nit: typo
```suggestion
        "description": "[All of Us OMOP extension] The source code for the participant's state of residence."
```

PR #564, opened 2020-07-16
[DC-891] Suppress COVID EHR Structured Concepts

amcgrenera-vumc:


data_steward/utils/concept_ids_suppression.py
```diff
@@ -0,0 +1,67 @@
+"""
```
ksdkalluri, line 1: This script is more similar to the scripts that we have in `data_steward/tools` which are mostly one-off or the scripts we need to run our cdr generation pipeline. `data_steward/utils` is being used to store the scripts that wrap functions in a third-party library to our project needs. let's maintain the same organization.

amcgrenera-vumc:
@ksdkalluri perfect, good catch, thanks!

PR #567, opened 2020-07-29
[DC-986] Fix Vocabulary dataset id in ci/setup.sh

ksdkalluri:


lrwb-aou:
I think we also need to update `VOCABULARY_DATASET` in `curation/data_steward/common.py`.  This should ideally be set by a config file or environment variable, but is hard coded here and is incorrect.

PR #568, opened 2020-07-31
[DC-699] clean implausible measurements. automate scripts from DC-481

lrwb-aou:
cleans implausible measurements
updates how schema files are used to create tables.

### DIFF
--- a/data_steward/resources.py
+++ b/data_steward/resources.py
@@ -120,6 +120,9 @@ def fields_for(table, sub_path=None):
     Uses os.walk to traverse subdirectories
 
     :param table: The table to get a schema for
+    :param sub_path: A string identifying a sub-directory in resource_files/fields.
+        If provided, this directory will be searched.
+    :returns: a json object representing the fields for the named table
     """
     path = os.path.join(fields_path, sub_path if sub_path else '')
 
@@ -140,6 +143,9 @@ def fields_for(table, sub_path=None):
         raise RuntimeError(
             f"Unable to read schema file because multiple schemas exist for:\t"
             f"{table} in path {path}")
+    elif unique_count == 0:
+        raise RuntimeError(
+            f"Unable to find schema file for {table} in path {path}")
 
     with open(json_path, 'r') as fp:
         fields = json.load(fp)
@@ -210,6 +216,7 @@ def cdm_schemas(include_achilles=False, include_vocabulary=False):
     :return:
     """
     result = dict()
+    # TODO:  update this code as part of DC-1015 and remove this comment
     for dir_path, _, files in os.walk(fields_path):
         for f in files:
             file_path = os.path.join(dir_path, f)
--- a/data_steward/resources.py
+++ b/data_steward/resources.py
@@ -2,12 +2,15 @@ import csv
 import hashlib
 import inspect
 import json
+import logging
 import os
 from io import open
 
 import cachetools
 
 from common import ACHILLES_TABLES, ACHILLES_HEEL_TABLES, VOCABULARY_TABLES, PROCESSED_TXT, RESULTS_HTML
+
+LOGGER = logging.getLogger(__name__)
 
 base_path = os.path.dirname(
     os.path.abspath(inspect.getfile(inspect.currentframe())))
@@ -111,9 +114,22 @@ def achilles_index_files():
 
 
 def fields_for(table):
-    json_path = os.path.join(fields_path, table + '.json')
+    """
+    Return the json schema for any table identified in the fields directory.
+
+    Uses os.walk to traverse subdirectories
+
+    :param table: The table to get a schema for
+    """
+    for dirpath, _, files in os.walk(fields_path):
+        for filename in files:
+            if filename[:-5] == table:
+                json_path = os.path.join(dirpath, filename)
+                break
+
     with open(json_path, 'r') as fp:
         fields = json.load(fp)
+
     return fields
 
 
@@ -126,6 +142,19 @@ def is_internal_table(table_id):
     :return: True if specified table is an internal table, False otherwise
     """
     return table_id.startswith('_')
+
+
+def is_extension_table(table_id):
+    """
+    Return True if specified table is an OMOP extension table.
+
+    Extension tables provide additional detail about an OMOP records taht does
+    not inherently fit in with the OMOP common data model.
+
+    :param table_id: identifies the table
+    :return: True if specified table is an internal table, False otherwise
+    """
+    return table_id.endswith('_ext')
 
 
 def is_mapping_table(table_id):
@@ -167,27 +196,31 @@ def cdm_schemas(include_achilles=False, include_vocabulary=False):
     :return:
     """
     result = dict()
-    for f in os.listdir(fields_path):
-        file_path = os.path.join(fields_path, f)
-        with open(file_path, 'r') as fp:
-            file_name = os.path.basename(f)
-            table_name = file_name.split('.')[0]
-            schema = json.load(fp)
-            include_table = True
-            if table_name in VOCABULARY_TABLES and not include_vocabulary:
-                include_table = False
-            elif table_name in ACHILLES_TABLES + ACHILLES_HEEL_TABLES and not include_achilles:
-                include_table = False
-            elif is_internal_table(table_name):
-                include_table = False
-            elif is_pii_table(table_name):
-                include_table = False
-            elif is_id_match(table_name):
-                include_table = False
-            elif table_name == 'post_deid_person':
-                include_table = False
-            if include_table:
-                result[table_name] = schema
+    for dir_path, _, files in os.walk(fields_path):
+        for f in files:
+            file_path = os.path.join(dir_path, f)
+            with open(file_path, 'r') as fp:
+                file_name = os.path.basename(f)
+                table_name = file_name.split('.')[0]
+                schema = json.load(fp)
+                include_table = True
+                if table_name in VOCABULARY_TABLES and not include_vocabulary:
+                    include_table = False
+                elif table_name in ACHILLES_TABLES + ACHILLES_HEEL_TABLES and not include_achilles:
+                    include_table = False
+                elif is_internal_table(table_name):
+                    include_table = False
+                elif is_pii_table(table_name):
+                    include_table = False
+                elif is_id_match(table_name):
+                    include_table = False
+                elif is_extension_table(table_name):
+                    include_table = False
+                elif table_name == 'post_deid_person':
+                    include_table = False
+                if include_table:
+                    result[table_name] = schema
+
     return result
 
 
--- a/data_steward/resources.py
+++ b/data_steward/resources.py
@@ -113,7 +113,7 @@ def achilles_index_files():
     return achilles_index_files
 
 
-def fields_for(table):
+def fields_for(table, sub_path=None):
     """
     Return the json schema for any table identified in the fields directory.
 
@@ -121,11 +121,25 @@ def fields_for(table):
 
     :param table: The table to get a schema for
     """
-    for dirpath, _, files in os.walk(fields_path):
+    path = os.path.join(fields_path, sub_path if sub_path else '')
+
+    # default setting
+    json_path = os.path.join(path, table + '.json')
+
+    unique_count = 0
+    for dirpath, _, files in os.walk(path):
+        if sub_path and os.path.basename(sub_path) != os.path.basename(dirpath):
+            continue
+
         for filename in files:
             if filename[:-5] == table:
                 json_path = os.path.join(dirpath, filename)
-                break
+                unique_count = unique_count + 1
+
+    if unique_count > 1:
+        raise RuntimeError(
+            f"Unable to read schema file because multiple schemas exist for:\t"
+            f"{table} in path {path}")
 
     with open(json_path, 'r') as fp:
         fields = json.load(fp)
--- a/tests/integration_tests/data_steward/cdr_cleaner/cleaning_rules/bigquery_tests_base.py
+++ b/tests/integration_tests/data_steward/cdr_cleaner/cleaning_rules/bigquery_tests_base.py
@@ -251,6 +251,12 @@ class BaseTest:
             # The query class that is being executed.
             cls.query_class = None
 
+        def setUp(self):
+            """
+            Add data to the tables for the rule to run on.
+            """
+            super().setUp()
+
         def default_test(self, tables_and_test_values):
             """
             Test passing the query specifications to the clean engine module.
--- a/tests/integration_tests/data_steward/cdr_cleaner/cleaning_rules/bigquery_tests_base.py
+++ b/tests/integration_tests/data_steward/cdr_cleaner/cleaning_rules/bigquery_tests_base.py
@@ -105,7 +105,6 @@ class BaseTest:
             """
             for table in cls.fq_table_names + cls.fq_sandbox_table_names:
                 cls.client.delete_table(table, not_found_ok=True)
-                pass
 
         def drop_rows(self, fq_table_name):
             """
@@ -132,7 +131,6 @@ class BaseTest:
             for table in self.fq_table_names + self.fq_sandbox_table_names:
                 self.drop_rows(table)
                 self.client.delete_table(table)
-                pass
 
         def load_test_data(self, sql_statements=None):
             """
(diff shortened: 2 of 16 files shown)


data_steward/resources.py
```diff
@@ -96,10 +99,37 @@ def achilles_index_files():
...
+    :param table: The table to get a schema for
+    """
+    path = os.path.join(fields_path, sub_path if sub_path else '')
+
+    # default setting
+    json_path = os.path.join(path, table + '.json')
+
+    unique_count = 0
+    for dirpath, _, files in os.walk(path):
+        if sub_path and os.path.basename(sub_path) != os.path.basename(dirpath):
```
mark-velez, line 117: I'm not sure what the intent is here- is this to ensure that `sub_path` leads directly to the desired file (by ceasing traversal)? It seems that this would continue walking, but only when directories with the same name are nested e.g. `fields_for('person', 'foo/bar')` could be used to retrieve `foo/bar/bar/bar/person.json`.

I'm slightly conflicted as to whether this should `os.walk` even when `sub_path` is provided or if it should just assume `sub_path` points directly to the file; both approaches might potentially be useful. The docstring could indicate more explicitly the behavior when `sub_path` is provided. Theoretically both strategies could be supported [in the future].

data_steward/resources.py
```diff
@@ -153,27 +196,31 @@ def cdm_schemas(include_achilles=False, include_vocabulary=False):
...
-                include_table = False
-            elif is_pii_table(table_name):
-                include_table = False
-            elif is_id_match(table_name):
-                include_table = False
-            elif table_name == 'post_deid_person':
-                include_table = False
-            if include_table:
-                result[table_name] = schema
+    for dir_path, _, files in os.walk(fields_path):
```
mark-velez, line 199: This works with the current state of `resource_files` but doesn't do the same name collision check as `fields_for` does; If I remember correctly and `os.walk` does do a depth-first search then a theoretical file `resource_files/post_deid/person.json` would thereby get returned rather than `resource_files/person.json`. Please let me know if I'm mistaken.

Even still we might want to account for possible collisions. Could we and would it help to put the standard schemas (clinical + vocabulary + achilles + heel) in `resource_files/omop`? If it's scope creepy, then maybe we could just raise an error on name collision here until such time that we can reorg all schemas into folders as we'd like.

data_steward/resources.py
lrwb-aou: @mark-velez Yes.  This is exactly why I wanted to talk to you.  I wasn't sure if it made sense to let the function act one way with one set of parameters and another way with a different set of parameters.  

I don't like the idea of the nested directories, `foo/bar/bar/bar/person.json`.  My intention was to avoid finding `fields/person.json` if `table='person' and sub_path='foo'`.  In this case, more than one `person.json` file exists but we are specifically interested in finding the person schema at `fields/foo/person.json`.

With that thought, I suppose the `sub_path` parameter makes it explicit where we want to find the schema.  I'll move in the direction of:  if a `sub_path` is specified, that is the only directory that is traversed.

data_steward/resources.py
```diff
@@ -96,10 +99,37 @@ def achilles_index_files():
...
+    for dirpath, _, files in os.walk(path):
+        if sub_path and os.path.basename(sub_path) != os.path.basename(dirpath):
+            continue
+
+        for filename in files:
+            if filename[:-5] == table:
+                json_path = os.path.join(dirpath, filename)
+                unique_count = unique_count + 1
+
+    if unique_count > 1:
```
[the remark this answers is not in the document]
lrwb-aou: Information is good.  Is this something to raise an error for or something we log and move on?

data_steward/resources.py
lrwb-aou: Yes, this will need to be adjusted, but that does feel like scope creep.  I think this is an engineering improvement and another ticket.

And os.walk does use a depth first search.  But is defaults to `topdown=True`, meaning it will find files higher in the directory system first.

tests/unit_tests/data_steward/resources_test.py
```diff
@@ -65,3 +68,77 @@ def test_cdm_schemas(self):
...
+        Testing that fields for works as expected with sub-directory structures.
+        """
+        # preconditions
+
+        # test
+        actual_fields = resources.fields_for('person')
+
+        # post conditions
+        person_path = os.path.join('data_steward', 'resource_files', 'fields',
+                                   'person.json')
```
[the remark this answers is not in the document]
lrwb-aou: I'll fix this.

data_steward/resources.py
lrwb-aou: I built a throw away directory structure like:
```
temp/
temp/person.json
temp/alpha/person.json
temp/test/person.json
```
I then ran the following in an interpreter and got the following results.
```
>>> import os
>>> for dir_path, _, files in os.walk('temp'):
...   print(f"{dir_path}\t\t\t{files}")
... 
temp			['person.json']
temp/test			['person.json']
temp/alpha			['person.json']
>>> for dir_path, _, files in os.walk('temp', topdown=False):
...   print(f"{dir_path}\t\t\t{files}")
... 
temp/test			['person.json']
temp/alpha			['person.json']
temp			['person.json']
>>> for dir_path, _, files in os.walk('temp', topdown=True):
...   print(f"{dir_path}\t\t\t{files}")
... 
temp			['person.json']
temp/test			['person.json']
temp/alpha			['person.json']
```
It is looping through the files in a topdown directory order by default.

mark-velez: Got it @lrwb-aou. It makes sense to include this as part of a bigger folder re-org ticket. And I think this is ok since walking would resolve to the existing schemas.

data_steward/resources.py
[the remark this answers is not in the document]
mark-velez: An error would be raised anyway a couple lines down in `open(json_path)`, but an error message like `f"Unable to locate a schema file for {table} in {path}"` might be more helpful.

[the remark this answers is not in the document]
lrwb-aou: Good point!  Added a `RuntimeError`.

data_steward/resources.py
lrwb-aou: I definitely want to do a folder re-org.  It would ease multiple areas.  But I just think that is an engineering improvement this ticket doesn't especially need.

data_steward/resources.py
mark-velez: _It could_ make sense to choose a search strategy based on params, but my preference would be to add that level of complexity only after a clear need is demonstrated. I think your choice should work well with existing usages. 👍

data_steward/resources.py
lrwb-aou: Created DC-1015 to address this.

tests/integration_tests/data_steward/cdr_cleaner/cleaning_rules/bigquery_tests_base.py
```diff
@@ -247,6 +251,12 @@ def initialize_class_vars(cls):
             # The query class that is being executed.
             cls.query_class = None
 
+        def setUp(self):
+            """
+            Add data to the tables for the rule to run on.
+            """
+            super().setUp()
```
dcampbell-vumc, line 258: Can you explain the reason why this was added again?

lrwb-aou: Cleaning classes that extend this class may define their own `setUp` function.  If the extending class calls `super().setUp()`, it will be routed to this function.  If this function does not exist, it will instead be routed to `BigQueryTestBase.setUp()`.  Both setups are functional.  Without this function, the hierarchy looks like:
```
class BigQueryTestBase():
    def setUp():

class CleaningRulesTestBase(BigQueryTestBase):

class ExtendingCleaningRuleTest(CleaningRulesTestBase):
    def setUp():
        # calls the setup function in BigQueryTestBase
        super().setUp()  
```

With it, the hierarchy is:
```
class BigQueryTestBase():
    def setUp():

class CleaningRulesTestBase(BigQueryTestBase):
    def setUp():
        # calls the setup function in BigQueryTestBase
        super().setUp()  

class ExtendingCleaningRuleTest(CleaningRulesTestBase):
    def setUp():
        # calls the setup function in CleaningRulesTestBase
        super().setUp()  
```

I added it to the Cleaning Rules Test Base, to give a visual clue to future developers of where to add cleaning rule specific setUp functions.  There may come a time when we want to add something for cleaning rules but not necessarily for all big query setup functions.

Does this  make sense?  Or would a call help?

dcampbell-vumc: Yes, that makes sense! Thank you!

tests/unit_tests/data_steward/resources_test.py
```diff
@@ -65,3 +68,76 @@ def test_cdm_schemas(self):
...
+        # post conditions
+        person_path = os.path.join(resources.base_path, 'resource_files',
+                                   'fields', 'person.json')
+        with open(person_path, 'r') as fp:
+            expected_fields = json.load(fp)
+
+        self.assertEqual(actual_fields, expected_fields)
+
+    @mock.patch('resources.os.walk')
+    def test_fields_for_duplicate_files(self, mock_walk):
```
dcampbell-vumc, line 90: I'm going to attempt to explain this test function so I know what's going on --

The walk results are:
`foo/bar/duplicate.json`
`foo/bar/unique1.json`
`foo/bar/baz/duplicate.json`
`foo/bar/baz/unique.json`

For the first test this isn't good because we have two json files `duplicate.json` which would throw a RuntimeError.

For the second test (I'm less sure about), it's ensuring that `baz/duplicate.json` returns the json with the right fields inside, right?

tests/unit_tests/data_steward/resources_test.py
```diff
@@ -65,3 +68,76 @@ def test_cdm_schemas(self):
...
+            os.path.join('foo', 'bar', 'baz'), '',
+            ['duplicate.json', 'unique2.json']
+        ]]
+
+        mock_walk.return_value = walk_results
+
+        data = '[{"id": "fake id desc", "type": "fake type"}]'
+        json_data = json.loads(data)
+
+        # test
```
dcampbell-vumc, line 134: How is this test different than the last test? 
`fake_path = 'baz'` vs. `fake_path = os.path.join('foo', 'bar', 'baz')`
Is `fake_path = 'baz'` not considered a defined path?

tests/unit_tests/data_steward/resources_test.py
```diff
@@ -65,3 +68,76 @@ def test_cdm_schemas(self):
...
             % result_vocab_tables)
+
+    def test_fields_for(self):
+        """
+        Testing that fields for works as expected with sub-directory structures.
+        """
+        # preconditions
+
+        # test
+        actual_fields = resources.fields_for('person')
```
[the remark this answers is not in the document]
lrwb-aou: This is the original function signature.  I wanted to make sure the new implementation didn't break existing function calls (there are quite a few throughout the code base).  I used the person table because it is a prominent omop table and is used extensively. Also, it's `post_deid_person.json` counterpart can cause trouble.  So I want to make sure the `fields_for` function is finding the correct file.
Yes, I explicitly open the person file I expect `fields_for` to open.  This is a file that exists as part of our repository, so I can expect it to be present.  If that file goes missing, it would mean something has changed structurally within our code and the failing test would alert us that whatever change was made is a breaking change.  I compare that json object to the json object I get from `fields_for` to make sure it's opening the correct file.

tests/unit_tests/data_steward/resources_test.py
lrwb-aou: good questions.  i'll add some comments to this test for future reference.  I'll address your questions here and we can talk via slack if you want.

Python's `os.walk` [returns a list of tuples](https://docs.python.org/3/library/os.html#os.walk).  Each tuple is of the form `(directory_string, list of contained directories, list of files)`.  So your directory structure is correct.  

The `fields_for` implementation doesn't care about the contained directories and pointedly ignores them using `_`.  Because they are ignored, that is represented as a blank list in `walk_results`.  If this were an accurate depiction of `os.walk` based on the given structure, then walk results should be:
```suggestion
walk_results = [
    (os.path.join('foo', 'bar'), ['baz'], ['duplicate.json', 'unique1.json']),
    (os.path.join('foo', 'bar', 'baz'), [], ['duplicate.json', 'unique2.json'])
]
```
I'll probably make this change also.

The first test highlights a concern @mark-velez had.  If we have two files with the same name in different sub-directories and don't indicate which one we expect, then `fields_for` doesn't know which one is correct.  It could make an assumption and return either the first or last it finds.  But instead of building in assumptions, Mark's suggestion of throwing an error is better.  So, with this `walk_results` and the existing function signature, `duplicate.json` would be found multiple times.  This test verifies the error is raised (which is what we want), so the test passes only when the error is raised.

The second test uses the `sub_path` parameter to dictate which sub-directory to walk.  `fields_for` will create the path `<path_to_curation_repo/curation/data_steward/resource_files/fields/<sub_path_parameter>`.  This directory will be walked.  When this directory is walked, `duplicate.json` is found and the upper level duplicate file is not.  So, the file is unique and a `RuntimeError` isn't raised.  The file is then opened and the json response object is returned.  The patch provides mock data for the file open function to read and return.  This just lets the test wrap up nicely by checking the equality of the expected json object with what it actually returns.

tests/integration_tests/data_steward/cdr_cleaner/cleaning_rules/rdr_observation_source_concept_id_suppression_test.py
```diff
@@ -79,17 +79,17 @@ def setUp(self):
...
         for tmpl in self.insert_fake_participants_tmpls:
             query = tmpl.render(fq_table_name=self.fq_table_names[0],
                                 age_prediabetes=age_prediabetes,
                                 meds_prediabetes=meds_prediabetes,
                                 now_prediabetes=now_prediabetes,
                                 prefer_not_to_ans=prefer_not_to_ans)
-            load_statements.append(query)
+            self.load_statements.append(query)
 
-        self.load_test_data(load_statements)
```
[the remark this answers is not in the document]
lrwb-aou: Close.  Because the tables are created in the individual test setup function now, it doesn't exist at this point.  The table isn't created until after `super().setUp()` is called.  So I can't load data into the table because it doesn't exist at this point.  I could have called
```
super().setUp()
self.load_test_data(load_statements)
```
I instead moved the load from the test setup to the first part of the test definition.  This is the same sequence as the code segment listed here, but achieved slightly differently.

tests/integration_tests/data_steward/cdr_cleaner/cleaning_rules/rdr_observation_source_concept_id_suppression_test.py
```diff
@@ -98,6 +98,8 @@ def test(self):
         Validates pre-conditions, test execution and post conditions based on
         the load statements and the tables_and_counts variable.
         """
+        self.load_test_data(self.load_statements)
```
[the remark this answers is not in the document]
lrwb-aou: Yes.  The table has to exist before we can load data into it.

tests/unit_tests/data_steward/resources_test.py
lrwb-aou: I think you're right.  This test isn't providing useful data.

lrwb-aou: This was the original test for testing the sub-directory path, but I copied to contents to the previous unit test because it helped illustrate how we could have multiple files with duplicate names if we specify the sub-directory filepath.  I'll remove this one now.

tests/integration_tests/data_steward/cdr_cleaner/cleaning_rules/rdr_observation_source_concept_id_suppression_test.py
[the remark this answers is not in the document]
dcampbell-vumc: Okay, gotcha!

tests/unit_tests/data_steward/resources_test.py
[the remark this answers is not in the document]
dcampbell-vumc: This makes sense.

tests/unit_tests/data_steward/resources_test.py
dcampbell-vumc: Yeah adding some comments would be helpful here :D also, that change makes sense to me.

lrwb-aou:
@dcampbell-vumc Thank you for the thorough review!

PR #569, opened 2020-08-04
[DC-984] Update clean engine to accommodate old-style and new-style cleaning rules

nishanthpp93:
* Add setup_rule, remove unitnorm specific code


data_steward/cdr_cleaner/clean_cdr_engine.py
```diff
@@ -140,6 +150,222 @@ def clean_dataset(project=None, statements=None, data_stage=stage.UNSPECIFIED):
...
+
+        query_list = instance.get_query_specs()
+    else:
+        query_function = clazz
+        try:
+            query_list = clazz(project_id, dataset_id, sandbox_dataset_id)
+        except TypeError:
+            # raised when called with the 3 parameters and only 2 are needed
+            query_list = clazz(project_id, dataset_id)
+    return query_list, query_function
```
mark-velez, line 278: I think the expectation is that this [well-named] function `infer_rule()` would be side-effect free. I recommend removing the client parameter and deferring invocation of `setup_rule()` to whoever is calling this func. Right now It's a bit awkward that this calls the query list func while also returning it for someone else to use- of course these are remnants of the past. Another consequence is that `setup_rule()` calls are not set to occur in the sequence we discussed (e.g. `rule_1.setup(), rule_1.query1(), rule_1.query2(), rule_2.setup() ...`).

What do you think of moving the call to `get_module_info()` here? It would seem appropriate to me at the moment as the purpose of this function seems to be to assemble some universal representation of old and new style rules. Also right now the "clean_dataset" use case is getting the module info (just for logging) but the "get_rules_documentation" use case (where this would shine brightest) does not currently get module info. 

To solve this, I'm wondering if we could figure out a way for this function to return a data structure similar to this:
```
(query_spec_func, setup_rule_func, module_info)
```

For new-style rules it be relatively straightforward:
```
return instance.get_query_specs, instance.setup_rule, get_module_info(instance.get_query_specs)
```
Old-style rules might need a little creativity. Something like this may work:
```
def old_style_query_specs():
  try:
    query_list = clazz(project_id, dataset_id, sandbox_dataset_id)
  except TypeError:
    # raised when called with the 3 parameters and only 2 are needed
    query_list = clazz(project_id, dataset_id)
  return query_list
def empty_setup_rule(client):
  pass

return old_style_query_specs, empty_setup_rule, get_module_info(instance.get_query_specs)
```
and `project_id`, `dataset_id`, `sandbox_dataset_id` should evaluate fine if I understand Python closures correctly. But even if not there's different ways to encapsulate them in another closure. This comment is heavy, let's discuss it soon.

mark-velez: Possible test harness on closure:
```
project_id = 'fake_project'
dataset_id = 'fake_dataset'
sandbox_dataset_id = 'fake_sandbox'
query_func, setup_rule, module_info = infer_rule(clazz, project_id, dataset_id, sandbox_dataset_id)
project_id = 'another_project_id'
query_func()
```

data_steward/cdr_cleaner/clean_cdr_engine.py
```diff
@@ -140,6 +150,222 @@ def clean_dataset(project=None, statements=None, data_stage=stage.UNSPECIFIED):
...
+            # wait for job to complete
+            query_job.result()
+        except (GoogleCloudError, TOError) as exp:
+            LOGGER.exception(
+                ce_consts.FAILURE_MESSAGE_TEMPLATE.render(client.project,
+                                                          **module_info_dict,
+                                                          **query_dict,
+                                                          exception=exp))
+            failures.append(query_dict[cdr_consts.QUERY])
+            continue
```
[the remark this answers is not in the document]
nishanthpp93: This change was due to migration from `bq_utils` to `utils/bq`. I referred to https://googleapis.dev/python/bigquery/latest/generated/google.cloud.bigquery.job.QueryJob.html#google.cloud.bigquery.job.QueryJob.result. Let me know if the earlier errors will suffice to catch these. I can also re-raise the error if you think it might be prudent.

data_steward/cdr_cleaner/clean_cdr_engine.py
nishanthpp93: Thanks for the detailed explanation, Mark! I have created the closure, please let me know if anything looks off. I will create the test soon.

data_steward/cdr_cleaner/clean_cdr_engine.py
```diff
@@ -140,6 +149,226 @@ def clean_dataset(project=None, statements=None, data_stage=stage.UNSPECIFIED):
...
+    :param sandbox_dataset_id: identifies the dataset to store sandbox tables
+    :return: rule_info: dictionary of information about the rule
+        keys:
+            query_function: function that generates query_list
+            setup_function: function that sets up the tables for the rule
+            function_name: name of the query function
+            module_name: name of the module containing the function
+            line_no: points to the source line where query_function is
+    """
+    if inspect.isclass(clazz):
```
mark-velez, line 246: This is good, better to do `inspect.isclass()` than try catching. It seems prudent to also do `issubclass(clazz, BaseCleaningRule)`, is there a reason to exclude that check? I recommend also raising an error if `inspect.isclass(clazz) and not issubclass(clazz, BaseCleaningRule)`.

data_steward/cdr_cleaner/clean_cdr_engine.py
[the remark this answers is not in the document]
mark-velez: As discussed please re-raise

data_steward/cdr_cleaner/clean_cdr_engine.py
```diff
@@ -140,6 +149,226 @@ def clean_dataset(project=None, statements=None, data_stage=stage.UNSPECIFIED):
...
+                    index, query_count, **rule_info, **query_dict))
+            job_config = generate_job_config(client.project, query_dict)
+
+            module_name = rule_info[cdr_consts.MODULE_NAME]
+            query_job = client.query(query=query_dict.get(cdr_consts.QUERY),
+                                     job_config=job_config,
+                                     job_id_prefix=f'clean_{module_name}_')
+
+            # wait for job to complete
+            query_job.result()
```
mark-velez, line 216: A job may complete but have job.errors. Please raise an exception here and ensure we log it and the rule_info, query_dic, etc. below.

data_steward/cdr_cleaner/clean_cdr_engine.py
```diff
@@ -140,6 +149,226 @@ def clean_dataset(project=None, statements=None, data_stage=stage.UNSPECIFIED):
...
+        except (GoogleCloudError, TOError) as exp:
+            LOGGER.exception(
```
mark-velez, line 218: Please raise exception as discussed.

data_steward/cdr_cleaner/clean_cdr_engine.py
```diff
@@ -140,6 +149,226 @@ def clean_dataset(project=None, statements=None, data_stage=stage.UNSPECIFIED):
...
+    job_config.destination = destination_table
+    job_config.use_legacy_sql = query_dict.get(cdr_consts.LEGACY_SQL, False)
+    # allow_large_results can only be used if use_legacy_sql=True
+    job_config.allow_large_results = job_config.use_legacy_sql
+    job_config.write_disposition = query_dict.get(cdr_consts.DISPOSITION,
+                                                  bq_consts.WRITE_EMPTY)
+    return job_config
+
+
+def run_queries(client, query_list, rule_info):
```
mark-velez, line 192: Change return type to `List[QueryJob]`

data_steward/cdr_cleaner/clean_cdr_engine.py
mark-velez: Ooh also include the `job_id`!

data_steward/cdr_cleaner/clean_cdr_engine.py
```diff
@@ -140,6 +149,226 @@ def clean_dataset(project=None, statements=None, data_stage=stage.UNSPECIFIED):
...
+    sandbox_dataset_id = sandbox.get_sandbox_dataset_id(dataset_id)
+
+    all_queries_list = []
+    for rule in rules:
+        clazz = rule[0]
+        query_function, _, rule_info = infer_rule(clazz, project_id, dataset_id,
+                                                  sandbox_dataset_id)
+        # added to test closure
+        project_id = 'incorrect_proj'
+        dataset_id = 'incorrect_ds'
```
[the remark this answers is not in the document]
nishanthpp93: Great catch, @mark-velez! I missed this completely.

tests/integration_tests/data_steward/cdr_cleaner/clean_cdr_engine_test.py
```diff
@@ -0,0 +1,74 @@
...
+    def tearDown(self) -> None:
+        for table_path in test_util.NYC_FIVE_PERSONS_FILES:
+            table_file = os.path.basename(table_path)
+            table = table_file.split('.')[0]
+            fq_table = f'{self.project_id}.{self.dataset_id}.{table}'
+            self.client.delete_table(fq_table, not_found_ok=True)
+        sandbox_dataset_id = sandbox.get_sandbox_dataset_id(self.dataset_id)
+        self.client.delete_dataset(sandbox_dataset_id,
+                                   delete_contents=True,
+                                   not_found_ok=True)
```
mark-velez, line 74: I appreciate how thorough the test is, but I contest the dependency on any specific cleaning rules here. If either should change or have a bug it can add work to be done here which could otherwise have been avoided. I think we could avoid these issues, make the tests more explicit and improve the running time by defining fake rules inline. Please let me know what you think:

```suggestion
        def setUp(self):
        self.project_id = get_application_id()
        self.dataset_id = os.environ.get('UNIONED_DATASET_ID')
        self.client = bq.get_client(self.project_id)
        self.delete_sandbox()

    def test_clean_dataset(self):
        fake_rule_class_query = 'SELECT "FakeRuleClass"'
        fake_rule_func_query = 'SELECT "fake_rule_func"'

        class FakeRuleClass(BaseCleaningRule):

            def __init__(self, project_id, dataset_id, sandbox_dataset_id):
                super().__init__(issue_numbers=[''],
                                 description='',
                                 affected_datasets=[cdr_consts.UNIONED],
                                 affected_tables=[],
                                 project_id=project_id,
                                 dataset_id=dataset_id,
                                 sandbox_dataset_id=sandbox_dataset_id)

            def get_sandbox_tablenames(self):
                pass

            def setup_rule(self, client, *args, **keyword_args):
                pass

            def setup_validation(self, client, *args, **keyword_args):
                pass

            def get_query_specs(self, *args, **keyword_args):
                return [{cdr_consts.QUERY: fake_rule_class_query}]

            def validate_rule(self, client, *args, **keyword_args):
                pass

        def fake_rule_func(project_id, dataset_id):
            return [{cdr_consts.QUERY: fake_rule_func_query}]

        # get_query_list returns query dicts for class- and func-based rules
        # (theoretically this could be a unit test)
        expected_queries = {fake_rule_class_query, fake_rule_func_query}
        query_dicts = ce.get_query_list(project_id=self.project_id,
                                        dataset_id=self.dataset_id,
                                        rules=[(FakeRuleClass,),
                                               (fake_rule_func,)])
        actual_queries = set(
            query_dict[cdr_consts.QUERY] for query_dict in query_dicts)
        self.assertSetEqual(expected_queries, actual_queries)

        # clean_dataset returns jobs associated with all rules' queries
        jobs = ce.clean_dataset_v1(project_id=self.project_id,
                                   dataset_id=self.dataset_id,
                                   rules=[(FakeRuleClass,), (fake_rule_func,)])
        actual_job_queries = set(job.query for job in jobs)
        self.assertEqual(expected_queries, actual_job_queries)
        # the jobs are completed
        self.assertTrue(all(job.state == 'DONE' for job in jobs))
        self.assertTrue(all(job.ended for job in jobs))
        self.delete_sandbox()

        # specific errors associated with failed queries are raised
        fake_rule_func_err_query = f'SELECT 1 FROM {self.dataset_id}.test_not_found'

        def fake_rule_func_err(project_id, dataset_id):
            return [{cdr_consts.QUERY: fake_rule_func_err_query}]

        with self.assertRaises(NotFound) as c:
            ce.clean_dataset_v1(project_id=self.project_id,
                                dataset_id=self.dataset_id,
                                rules=[(FakeRuleClass,), (fake_rule_func_err,)])
        self.assertEqual(c.exception.query_job.query, fake_rule_func_err_query)

    def delete_sandbox(self):
        sandbox_dataset_id = sandbox.get_sandbox_dataset_id(self.dataset_id)
        self.client.delete_dataset(sandbox_dataset_id,
                                   delete_contents=True,
                                   not_found_ok=True)

    def tearDown(self) -> None:
        self.delete_sandbox()
```

data_steward/cdr_cleaner/clean_cdr_engine.py
```diff
@@ -140,6 +149,199 @@ def clean_dataset(project=None, statements=None, data_stage=stage.UNSPECIFIED):
...
+    :param rules: a list of cleaning rule objects/functions as tuples
+    :param data_stage: an enum to indicate what stage of the cleaning this is
+    """
+    if project_id is None or project_id == '' or project_id.isspace():
+        project_id = app_identity.get_application_id()
+        LOGGER.info(f"project_id not provided, using default {project_id}")
+
+    # Set up
+    client = bq.get_client(project_id=project_id)
+    sandbox_dataset_id = sandbox.create_sandbox_dataset(project_id=project_id,
```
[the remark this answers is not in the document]
nishanthpp93: I'd go with 2 as well, thanks for pointing this out @mark-velez!

data_steward/cdr_cleaner/clean_cdr_engine.py
```diff
@@ -140,6 +149,199 @@ def clean_dataset(project=None, statements=None, data_stage=stage.UNSPECIFIED):
...
+        cdr_consts.MODULE_NAME: module_name,
+        cdr_consts.LINE_NO: line_no,
+    }
+    return query_function, setup_function, rule_info
+
+
+def get_query_list(project_id=None,
+                   dataset_id=None,
+                   rules=None,
+                   data_stage=cdr_consts.DataStage.UNSPECIFIED):
```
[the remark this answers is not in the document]
nishanthpp93: Yes, I had forgotten this. Thanks for reminding me @mark-velez, I was unsure so I left it unused.

nishanthpp93:
Thanks @mark-velez, it is delightful to delete large blocks of code and I am enjoying every moment of it. For the `CleaningRulesTestBase`, I added a `query_rule` parameter in addition to the existing `query_class` so I don't have to make too many changes to all the integration tests and still run the engine. I also added a `combined_dataset_id` parameter because `generate_ext_tables` required it. Let me know if these steps are okay.

nishanthpp93:
Closing this and opening #605

PR #571, opened 2020-08-05
[DC-989] Update DEID to Reference Existing PID-RID Mapping Table

ksdkalluri:
Adds copying _deid_map table from rdr step to combine_ehr_rdr.py


data_steward/tools/combine_ehr_rdr.py
```diff
@@ -46,10 +46,30 @@
...
+    trim_blocks=True,
+    lstrip_blocks=True,
+    # syntax highlighting should be better
+    # with these comment delimiters
+    comment_start_string='--',
+    comment_end_string=' --')
+
+COPY_PID_RID_QUERY = jinja_env.from_string("""
+select * from `{{project}}.{{dataset}}.{{pid_rid_table}}`
+""")
```
lrwb-aou, line 72: I'm hesitant about creating the `_deid_map` table in the combined dataset code.  The combined dataset has no reason to know about a mapping table created for and used later in the pipeline (this is a criticism of the mapping table's historical placement).  It would be better (not part of your PR) to store this mapping table in a sandbox dataset in the future.  So, I'm concerned that putting this table copy code in this module would make that transition harder in the future.  It would make more sense to either create this mapping table from `curation/data_steward/tools/run_deid.py` where it can run once, or it can replace the `create_participant_mapping_table()` function in `curation/data_steward/deid/aou.py`.  This has the disadvantage of running once for every table that is deidentified.

data_steward/tools/combine_ehr_rdr.py
```diff
@@ -361,10 +381,58 @@ def load_mapped_person():
...
+
+    # This else condition is a stop gap fix until RDR starts sending the _pid_rid_mapping table
+    elif bq_utils.table_exists(
+            combine_consts.PID_RID_MAPPING_TABLE,
+            dataset_id=combine_consts.PIPELINE_TABLES_DATASET):
+        copy_deid_map_table(deid_map_mapping_table, project_id,
+                            combine_consts.PIPELINE_TABLES_DATASET, client)
+        logging.info(
+            f"{project_id}.{combine_consts.PIPELINE_TABLES_DATASET}.{combine_consts.PID_RID_MAPPING_TABLE} \
+                    copied to the table {deid_map_mapping_table}")
```
lrwb-aou, line 422: I'm not sure what advantage this stop gap solution is creating.  We have already copied the `_deid_map` table to a `pipeline_tables.pid_rid_mapping`.  The deid code will not be updating that table (some other code will be developed for that), but deid can use it as if it exists because it has already been moved.  Is there a design decision I have overlooked?

data_steward/tools/combine_ehr_rdr.py
```diff
@@ -46,10 +46,30 @@
...
+    # will not cause extra white space
+    trim_blocks=True,
+    lstrip_blocks=True,
+    # syntax highlighting should be better
+    # with these comment delimiters
+    comment_start_string='--',
+    comment_end_string=' --')
+
+COPY_PID_RID_QUERY = jinja_env.from_string("""
+select * from `{{project}}.{{dataset}}.{{pid_rid_table}}`
```
lrwb-aou, line 71: This query doesn't take into account the participant's age.  It copies the pid/rid table in its entirety.  Ideally, the copy will only copy data for those participants whose age is less than the maximum age.

data_steward/tools/combine_ehr_rdr.py
```diff
@@ -361,10 +381,58 @@ def load_mapped_person():
...
 
+def copy_deid_map_table(pid_rid_mapping_table, project_id, dataset_id, client):
+    job_config = bq.bigquery.QueryJobConfig(destination=pid_rid_mapping_table)
+    q = COPY_PID_RID_QUERY.render(
+        project=project_id,
+        dataset=dataset_id,
+        pid_rid_table=combine_consts.PID_RID_MAPPING_TABLE)
+
+    query_job = client.query(q, job_config=job_config)
+    query_job.result()
```
lrwb-aou, line 392: This call does block to make sure the query finishes, but you haven't checked to make sure there are no exceptions.
```suggestion
    query_job.result()
    if query_job.exception():
        logging.error(f"The deid_map table did not generate successfully")
```

data_steward/tools/run_deid.py
```diff
@@ -211,9 +248,69 @@ def parse_args(raw_args=None):
...
+    # Create _pid_rid_mapping table
+    project_id = app_identity.get_application_id()
+    client = bq.get_client(project_id)
+    deid_map_mapping_table = f'{project_id}.{deid_map_dataset_name}.{DEID_MAP_TABLE}'
+    bq.create_tables(client,
+                     project_id, [deid_map_mapping_table],
+                     exists_ok=False)
+    # Copy pid_rid_mapping records to _deid_map table
+    if bq_utils.table_exists(PID_RID_MAPPING_TABLE,
+                             dataset_id=PIPELINE_TABLES_DATASET):
```
lrwb-aou, line 301: Did you consider using a DDL statement to create the table?  What are the advantages/disadvantages the explicitly creating the table and checking for its existence instead of using a DDL statement?  It seems like a `Create Table If Not Exists` [DDL statement](https://cloud.google.com/bigquery/docs/reference/standard-sql/data-definition-language#creating_a_table_only_if_the_table_doesnt_exist) would achieve the same results with fewer API calls.  If you did use a DDL statement, I think you would want to add a `try/except/else` to catch and log when the [table already exists.](https://cloud.google.com/bigquery/docs/reference/standard-sql/data-definition-language#creating_a_new_table)

I think I may be missing a design decision.  I'm definitely interested in your thoughts.

ksdkalluri: You are right I should have used DDL in the first place. It didn't come into my mind while I was adding an API call.

PR #572, opened 2020-08-06
[DC-972] Store data pulled by utility

dcampbell-vumc:
Created utility to store deactivated participant information


dcampbell-vumc:
Because I had to do mocking for the integration test, it looks very similar to the unit test....

dcampbell-vumc:
@lrwb-aou, I forgot to add a `tearDown` function to the integration test so `_deactivated_participants` is still in the `pipeline_logging` dataset. I can't use `delete_dataset` because there are other tables in `pipeline_logging`. Our `bq.py` module doesn't have a delete tables function, however `bq_utils.py` has one `bq_utils.delete_table()`. Should I use that? We should probably make a ticket to eventually create a delete tables function in `bq.py` right?

tests/integration_tests/data_steward/utils/participant_summary_requests_test.py
```diff
@@ -0,0 +1,185 @@
...
+
+    @classmethod
+    def setUpClass(cls):
+        print('**************************************************************')
+        print(cls.__name__)
+        print('**************************************************************')
+
+    def setUp(self):
+        self.project_id = app_identity.get_application_id()
+        self.dataset_id = 'pipeline_tables'
```
lrwb-aou, line 41: @dcampbell-vumc You've parameterized `dataset_id` in your code.  You could make this any dataset you choose.  Unless I've missed a spot.  Then you could delete the entirety of the dataset without worrying about removing someone else's test code.  Actually, you should change this hard coded name.  This will cause collisions between branches.

dcampbell-vumc: I normally would have used `fake_dataset` for the dataset_id however, I thought having it named that would cause collisions since several integration tests use `fake_dataset`. I probably am misremembering though...

lrwb-aou: The problem exists either way, both are hard coded names.  You need to use a parameterized variable for this.  The cleaning rules set the dataset id by doing something like this:
```suggestion
        dataset_id = os.environ.get('COMBINED_DATASET_ID')
```
This uses the available environment variables.  These are set in the circleci config as a combination of your user name and branch name.  This means they will be unique.

dcampbell-vumc: Ah, I understand now!

lrwb-aou:
Please do a rebase.  I think there are some recent changes that this might need.

data_steward/utils/participant_summary_requests.py
```diff
@@ -65,16 +66,34 @@ def get_participant_data(url, headers):
...
+    if not dataset_id or not isinstance(dataset_id, str):
+        raise RuntimeError(f'Please provide a dataset_id')
+
+    if not tablename or not isinstance(tablename, str):
+        raise RuntimeError(
+            f'Please provide a tablename to house deactivated participant data')
+
+    if not columns or not isinstance(columns, list):
+        raise RuntimeError(
+            'Please provide a list of columns to be pushed to BigQuery table')
```
lrwb-aou, line 95: The `isinstance` check can accommodate the presence of `None`.  So, these could all be simplified to:

```
if not isinstance(variable, type):
```

dcampbell-vumc: I simplified all the parameter check statements

data_steward/utils/participant_summary_requests.py
```diff
@@ -104,4 +123,35 @@ def get_deactivated_participants(project_id, columns):
...
+
+    :return: returns a dataset with the participant data
+    """
+
+    # Parameter check
+    if not project_id or not isinstance(project_id, str):
+        raise RuntimeError(
+            f'Please specify the project in which to create the tables')
+
+    stored_dataset = pandas_gbq.to_gbq(df,
```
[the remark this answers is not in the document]
dcampbell-vumc: Is pylint is actually hitting the API and returning the deactivated participants dataframe generated in `get_deactivated_participants`?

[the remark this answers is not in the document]
lrwb-aou: No.  Pylint is a static analysis tool, (meaning it doesn't run your software).  This is the `pandas_gbq` [documentation](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.DataFrame.to_gbq.html).  If you follow the source link, you can see it doesn't return a value.  Pylint marked it as an error because you are trying to assign a value from a function that doesn't return a value.

data_steward/resource_files/fields/_deactivated_participants.json
```diff
@@ -0,0 +1,20 @@
+[
```
lrwb-aou, line 1: I would suggest adding this definition to a sub-directory in `fields`.  Possibly something like `static_tables`, `lookup_tables`, or even `pipeline_tables`.

data_steward/utils/participant_summary_requests.py
```diff
@@ -101,7 +120,64 @@ def get_deactivated_participants(project_id, columns):
...
+    df = pandas.DataFrame(updated_deactivated_participants,
                           columns=deactivated_participants_cols)
 
-    return df
+    # To store dataframe in a BQ dataset table
+    destination_table = dataset_id + '.' + tablename
+
+    dataset = store_participant_data(df, project_id, destination_table)
+
+    return dataset
```
lrwb-aou, line 134: Based on how you decide to / not to return values from `store_participant_data`, the `dataset` variable may be `None`.  If it is `None`, you **may** want to do something like:
```suggestion
    store_participant_data(df, project_id, destination_table)

    return '.'.join([project_id, destination_table])
```

tests/integration_tests/data_steward/utils/participant_summary_requests_test.py
```diff
@@ -0,0 +1,197 @@
...
+                          self.project_id, None, self.tablename, self.columns)
+        self.assertRaises(RuntimeError, psr.get_deactivated_participants,
+                          self.project_id, self.dataset_id, None, self.columns)
+        self.assertRaises(RuntimeError, psr.get_deactivated_participants,
+                          self.project_id, self.dataset_id, self.tablename,
+                          None)
+
+    @mock.patch('utils.participant_summary_requests.store_participant_data')
+    @mock.patch(
+        'utils.participant_summary_requests.get_deactivated_participants')
```
lrwb-aou, line 152: This test can't be testing `get_deactivated_participants` or `store_participant_data` because both functions are mocked.  As part of the integration test, you want your tested code to actually build the table in BigQuery to make sure your test is verifying your code functions correctly.  The only part of the test you want to mock is the response from the RDR API, `@mock.patch('utils.participant_summary_requests.requests.get')`.  

I've been looking at how to do this and have a partial patch file.  We can talk about this via a call if you would like.

data_steward/utils/participant_summary_requests.py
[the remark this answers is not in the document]
dcampbell-vumc: Gotcha, I removed the value assignment so `store_participant_data` is returning `pandas_gbq.to_gbq(...)`. Ran pylint and no longer getting that error.

data_steward/utils/participant_summary_requests.py
dcampbell-vumc: Yep, `dataset` was returning `None`. I took your suggestion and changed the return statement of `get_deactivated_participants`

data_steward/resource_files/fields/_deactivated_participants.json
dcampbell-vumc: Created a subdirectory called `lookup_tables` and put `_deactivated_participants` in it.

data_steward/resource_files/fields/_deactivated_participants.json
```diff
@@ -0,0 +1,20 @@
...
+    {
+        "type": "string",
```
lrwb-aou, line 3: It might be better if this parameter is an `integer` instead of a `string`.  Especially since `person_id` is an `integer` in [omop tables](https://www.ohdsi.org/web/wiki/doku.php?id=documentation:cdm:person).

data_steward/utils/participant_summary_requests.py
```diff
@@ -101,7 +120,64 @@ def get_deactivated_participants(project_id, columns):
...
+    # determines the index of participantId in columns list
+    participant_id = columns.index('participantId')
+
+    updated_participant_list = []
+
+    for participant in participant_list:
+        participant[participant_id] = int(participant[participant_id][1:])
+        updated_participant_list.append(participant)
+
+    return updated_participant_list
```
lrwb-aou, line 157: I've been toying with pandas some more and think you could make this a pandas helper function that can just be applied to an existing column.  You can create custom functions to apply to columns like in this [stackoverflow post](https://stackoverflow.com/questions/50947447/convert-values-of-a-column-in-pandas).

To test, you can run this in an interpreter:
```
import pandas as pd

# create the dataframe
df = pd.DataFrame([['P111'],['P222'],['P333']], columns=['pid'])

# view dataframe
df

# view dataframe info
df.info(verbose=True)

# define custom function
def participant_id_to_int(pid):
    return int(pid[1:])

# apply the custom function
df['pid'] = df['pid'].apply(participant_id_to_int)

# view dataframe
df

# view dataframe info
df.info(verbose=True)
```
Then this could just be another transformation you apply to your dataframe without you needing to identify the column index of the `participant_id`.

dcampbell-vumc: Oh, I like how concise this is!

lrwb-aou: This definition is assuming the first character is always an alphabetic character.  You could add checks in.  It might not be necessary because if the format of this string changes, it will have to be communicated across multiple teams, as a lot of systems could break.  But for an example, you could use:

```
import string

def participant_id_to_int(pid):
    return int(pid.strip(string.ascii_letters))
```
This would perform a case-insensitive strip of all leading and trailing alphabetic characters.  You can find documentation in the [string API](https://docs.python.org/3.7/library/stdtypes.html?highlight=strip#string-methods) by searching for strip.

tests/integration_tests/data_steward/utils/participant_summary_requests_test.py
```diff
@@ -0,0 +1,189 @@
...
+        # Test
+        psr.store_participant_data(self.fake_dataframe, self.project_id,
+                                   self.destination_table)
+
+        # Post conditions
+        values = [(111, 'NO_CONTACT', parser.parse('2018-12-07T08:21:14 UTC')),
+                  (222, 'NO_CONTACT', parser.parse('2018-12-07T08:21:14 UTC'))]
+        self.assertTableValuesMatch(
+            '.'.join([self.project_id, self.destination_table]), self.columns,
+            values)
```
lrwb-aou, line 189: If we go with `person_id` in the big query table, this will have to be changed also.
```suggestion
        bq_columns = ['person_id] + self.columns[1:]
        self.assertTableValuesMatch(
            '.'.join([self.project_id, self.destination_table]), bq_columns,
            values)
```

tests/integration_tests/data_steward/utils/participant_summary_requests_test.py
```diff
@@ -0,0 +1,189 @@
...
+        # Tests
+        psr.get_deactivated_participants(self.project_id, self.dataset_id,
+                                         self.tablename, self.columns)
+
+        # Post conditions
+        values = [(111, 'NO_CONTACT', parser.parse('2018-12-07T08:21:14 UTC')),
+                  (222, 'NO_CONTACT', parser.parse('2018-12-07T08:21:14 UTC'))]
+        self.assertTableValuesMatch(
+            '.'.join([self.project_id, self.destination_table]), self.columns,
+            values)
```
lrwb-aou, line 173: If we go with `person_id` in the big query table, this will have to be changed.
```suggestion
        bq_columns = ['person_id] + self.columns[1:]
        self.assertTableValuesMatch(
            '.'.join([self.project_id, self.destination_table]), bq_columns,
            values)
```

data_steward/utils/participant_summary_requests.py
```diff
@@ -104,4 +124,56 @@ def get_deactivated_participants(project_id, columns):
...
 
-    return df
+    # Converts column `suspensionTime` from string to timestamp
+    for column in deactivated_participants_cols:
+        if column == 'suspensionTime':
+            df[column] = pandas.to_datetime(df[column])
+
+    # Transforms participantId to an integer string
+    df['participantId'] = df['participantId'].apply(participant_id_to_int)
+
```
lrwb-aou, line 134: You could use this to do the renaming:
```suggestion
    # rename participantId column to person_id
    df = df.rename(columns={'participantId': 'person_id'})
    
```

data_steward/resource_files/fields/lookup_tables/_deactivated_participants.json
```diff
@@ -0,0 +1,20 @@
+[
+    {
+        "type": "INTEGER",
+        "name": "participantId",
```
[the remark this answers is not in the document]
dcampbell-vumc: Yes, I can change it `person_id` it's consistent.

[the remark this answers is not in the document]
dcampbell-vumc: I changed the description too. I ripped the old one straight off RDR's API documentation :p

data_steward/utils/participant_summary_requests.py
```diff
@@ -104,4 +124,56 @@ def get_deactivated_participants(project_id, columns):
     df = pandas.DataFrame(deactivated_participants,
                           columns=deactivated_participants_cols)
 
-    return df
+    # Converts column `suspensionTime` from string to timestamp
+    for column in deactivated_participants_cols:
+        if column == 'suspensionTime':
```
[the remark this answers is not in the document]
dcampbell-vumc: Changed 👍🏾

tests/integration_tests/data_steward/utils/participant_summary_requests_test.py
dcampbell-vumc: Changed 👍🏾

tests/integration_tests/data_steward/utils/participant_summary_requests_test.py
dcampbell-vumc: Changed 👍🏾

data_steward/utils/participant_summary_requests.py
dcampbell-vumc: Added 👍🏾

dcampbell-vumc:
I'm going to push my changes and see if the integration test check fails again, I looked at the Circle history for this branch and this was the first time it failed because of the integration test.

lrwb-aou:
> I'm going to push my changes and see if the integration test check fails again, I looked at the Circle history for this branch and this was the first time it failed because of the integration test.

Yea.  It's weird.  It intermittently fails when I run it locally.  I haven't been able to consistently either make it pass or fail yet.

data_steward/utils/participant_summary_requests.py
```diff
@@ -104,4 +124,57 @@ def get_deactivated_participants(project_id, columns):
...
 
-    return df
+    # Converts column `suspensionTime` from string to timestamp
+    if 'suspensionTime' in deactivated_participants_cols:
+        df['suspensionTime'] = pandas.to_datetime(df['suspensionTime'])
+
+    # Transforms participantId to an integer string
+    # Rename participantId column to person_id
+    df['participantId'] = df['participantId'].apply(participant_id_to_int)
+    df = df.rename(columns={'participantId': 'person_id'})
```
lrwb-aou, line 134: If we are not passing column names and we use underscores as delimiters instead of camel casing, this should be updated to:

```suggestion
    bq_columns = {'participantId': 'person_id', 'suspensionTime': 'suspension_time', 'suspensionStatus': 'suspension_status'}
    df = df.rename(columns=bq_columns)
```

If we do still pass the column names, and want to use camel casing, we can use a regular expression to create a list of strings that are formatted correctly.  Then we can zip the contents together to create a dictionary.  What do you think about this?

tests/integration_tests/data_steward/utils/participant_summary_requests_test.py
```diff
@@ -0,0 +1,190 @@
...
+        cls.destination_table = cls.dataset_id + '.' + cls.tablename
+
+        cls.fq_table_names = [
+            f"{cls.project_id}.{cls.dataset_id}.{cls.tablename}"
+        ]
+        super().setUpClass()
+
+    def setUp(self):
+        self.columns = ['participantId', 'suspensionStatus', 'suspensionTime']
+        self.bq_columns = ['person_id'] + self.columns[1:]
```
lrwb-aou, line 56: This would need to be updated if you choose underscore delimiters in favor of camel casing.

data_steward/resource_files/fields/lookup_tables/_deactivated_participants.json
```diff
@@ -0,0 +1,20 @@
+[
+    {
+        "type": "INTEGER",
+        "name": "person_id",
+        "mode": "required",
+        "description": "The person_id created by removing the first character form the RDR participantId"
+    },
+    {
+        "type": "STRING",
+        "name": "suspensionStatus",
```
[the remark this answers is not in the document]
dcampbell-vumc: Changed 👍🏾

data_steward/resource_files/fields/lookup_tables/_deactivated_participants.json
```diff
@@ -0,0 +1,20 @@
...
+        "mode": "required",
+        "description": "The status of suspension for a participant. Enumerated values: NOT_SUSPENDED and NO_CONTACT"
+    },
+    {
+        "type": "TIMESTAMP",
+        "name": "suspensionTime",
```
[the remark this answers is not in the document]
dcampbell-vumc: Changed 👍🏾

data_steward/utils/participant_summary_requests.py
```diff
@@ -65,16 +67,34 @@ def get_participant_data(url, headers):
     return participant_data
 
 
-def get_deactivated_participants(project_id, columns):
+def get_deactivated_participants(project_id, dataset_id, tablename, columns):
```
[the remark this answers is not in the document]
dcampbell-vumc: Yep, I believe we do. These are the RDR columns (I think) that are used to create the DataFrame and in loop through the participant summary records before we do any transformations to the columns.

data_steward/utils/participant_summary_requests.py
dcampbell-vumc: yes, this would be best, I wanted columns to be passed in because they could possibly change in the future.

tests/integration_tests/data_steward/utils/participant_summary_requests_test.py
dcampbell-vumc: Updated 👍🏾

dcampbell-vumc:
Woohoo! 🥳🥳 I'll merge it now and set the ticket to done. Thanks so much for the thorough review!

PR #574, opened 2020-08-11
[DC-912] Fix table identification logic in completeness.py

nishanthpp93:
* Fix is_hpo_col logic and update test
* Remove re-raising error due to failed metric

### DIFF
--- a/data_steward/validation/metrics/completeness.py
+++ b/data_steward/validation/metrics/completeness.py
@@ -6,7 +6,6 @@ from io import open
 # Project imports
 import bq_utils
 import resources
-from common import CDM_TABLES, PII_TABLES
 from constants.validation.metrics import completeness as consts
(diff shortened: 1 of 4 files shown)


data_steward/validation/metrics/completeness.py
```diff
@@ -1,7 +1,13 @@
+# Python imports
+from io import open
+
+# Third party imports
+
+# Project imports
 import bq_utils
 import resources
+from common import CDM_TABLES, PII_TABLES
```
amcgrenera-vumc, line 9: I'm not seeing `PII_TABLES` being utilized. Is that import necessary?

nishanthpp93: You're right, I was using this earlier and forgot to remove it.

PR #576, opened 2020-08-12
[DC-990] Remove DEID Max Age Methodology

lrwb-aou:
* ensures the logic of building the mapping table is removed
* performs a double check to make sure the mapping table only has age valid participants
* passes the max age parameter to aou.py for the check.
* updates the aou signature in the unit test check


data_steward/deid/aou.py
```diff
@@ -277,16 +276,12 @@ def initialize(self, **args):
         Press.initialize(self, **args)
         LOGGER.info(f"BEGINNING de-identification on table:\t{self.tablename}")
 
-        age_limit = args['age_limit']
+        age_limit = args.get('age_limit', 89)
```
nishanthpp93, line 279: I think it would be good to define a constant `MAX_AGE=89` and reuse it so we won't have to change it in multiple places if we need to.

data_steward/deid/aou.py
```diff
@@ -298,24 +293,36 @@ def initialize(self, **args):
...
-                }
-            }
-            map_table = self.get_dataframe(sql=map_sql, query_config=job_config)
-            if not map_table.shape[0] > 0:
-                LOGGER.error(
-                    "Unable to initialize Deid. _deid_map has no participants to load."
-                )
+        # ensure mapping table only contains participants within age limits
+        sql = (f"SELECT DISTINCT p.person_id, "
+               f"EXTRACT(YEAR FROM CURRENT_DATE()) - year_of_birth AS age "
```
nishanthpp93, line 298: How strict do we need to be for the age cutoff? I'm wondering if `birth_datetime` is a more accurate cutoff. If it is null, we could resort to using `year_of_birth` (possibly along with `day_of_birth` and `month_of_birth` unless they are null as well). What are your thoughts on this?

data_steward/deid/aou.py
```diff
@@ -298,24 +293,36 @@ def initialize(self, **args):
...
+            LOGGER.error(
+                f"Unable to initialize Deid. {map_tablename} table cannot be "
+                f"joined to {self.idataset}.person table to verify age requirements."
+            )
+
+        # ensure no age ineligible participants are available in the mapping table
+        ineligible_person_table = person_table[person_table.age >= age_limit]
+        if ineligible_person_table.shape[0] > 0:
+            LOGGER.error(f"Some age ineligible participants are available in "
+                         f"{map_tablename}.  Deid is bailing out!!")
```
nishanthpp93, line 319: Haha I love this message! The two statements could probably be combined by replacing `Some` with the actual count below.

lrwb-aou: Added this.

data_steward/deid/aou.py
lrwb-aou: It's a good idea.  I wasn't sure if it was scope creep.  I will make this change.

data_steward/deid/aou.py
lrwb-aou: I think I want to leave this one alone.  This is how Steve wrote Deid before it got handed off to us.  I like this approach because `year_of_birth` is a required field while `birth_datetime` is nullable.

tests/unit_tests/data_steward/deid/parser_test.py
```diff
@@ -63,7 +63,7 @@ def test_parse_args(self):
 
         # setting correct_parameter_dict values not set in setUp function
         correct_parameter_dict['cluster'] = False
-        correct_parameter_dict['age-limit'] = 89
+        correct_parameter_dict['age_limit'] = 89
```
nishanthpp93, line 66: Would this need to use the constant as well?

data_steward/deid/aou.py
```diff
@@ -298,24 +294,37 @@ def initialize(self, **args):
...
+
+        # ensure no age ineligible participants are available in the mapping table
+        ineligible_person_table = person_table[person_table.age >= age_limit]
+        if ineligible_person_table.shape[0] > 0:
+            LOGGER.error(f"{ineligible_person_table.shape[0]} age ineligible "
+                         f"participants are available in "
+                         f"{map_tablename}.  Deid is bailing out!!")
+
+        LOGGER.info(f"map table contains {eligible_person_table.shape[0]} "
+                    f"records.")
```
nishanthpp93, line 324: nit: This can be removed since it is the same as above

tests/unit_tests/data_steward/deid/parser_test.py
lrwb-aou: It should.  I updated this.  Thanks for the catch.

data_steward/deid/aou.py
lrwb-aou: Which line is it duplicating?

lrwb-aou:
> Added a couple of nits. Were you able to run this on the test dataset without any issues?
> 
> Side note: It looks like a couple of integration tests are failing but they seem unrelated to these changes. Do you know why they are getting affected?

I was able to run it in the test environment without any issues.  There should be steps I followed in the ticket.  
I also noticed the failing integration tests and don't know why it's happening.  It's something with the retraction module, so I don't know why these changes are causing it.  I'll look into now that I have my other PR open.

lrwb-aou:
@nishanthpp93 I am going to delete this branch, `lb/remove_deid_builtin_max_age`, and rename it, `lb/remove_de_id_builtin_max_age`.  This should fix the integration test issue.  I may have to close this PR and issue a new one.

PR #578, opened 2020-08-14
[DC-1011] Add person_ext schema

mark-velez:


tests/unit_tests/data_steward/resources_test.py
```diff
@@ -86,6 +86,16 @@ def test_fields_for(self):
 
         self.assertEqual(actual_fields, expected_fields)
 
+        # test
+        actual_fields = resources.fields_for('person_ext')
+        person_ext_path = os.path.join(resources.base_path, 'resource_files',
+                                       'fields', 'extension_tables',
+                                       'person_ext.json')
```
nishanthpp93, line 93: Do you think this path (or paths of the folders) should be added to resources for reuse? I notice that we are using `os.walk` which could render them unnecessary anyway.

I also wanted to mention that even though this is picked up by `resource.cdm_schemas` via `os.walk` without having to specify the full path, it is being excluded as a CDM file using `is_extension_table()`. Is that correct?

mark-velez: Yes, I think it would probably be useful to add these search paths to `resources` but I chose not to do it here since there is no concrete use case for it yet. Perhaps it will happen in upcoming tickets that are using `person_ext`. I might prefer to defer adding paths until they are _strictly needed_ (in order to disambiguate files having the same name) because the way we organize these files might change by then.

You are right, this file is scanned by `resources.cdm_schemas` but it is excluded because of `is_extension_table()`. In the future I hope we can eliminate that sequence of conditionals by reorganizing schemas into folders

nishanthpp93: That makes sense, thanks @mark-velez

PR #588, opened 2020-08-24
[DC-1042] Update ParticipantSummary API utility for correct pagination

amcgrenera-vumc:


data_steward/utils/participant_summary_requests.py
```diff
@@ -51,17 +51,21 @@ def get_participant_data(url, headers):
...
         if not resp or resp.status_code != 200:
             print(f'Error: API request failed because {resp}')
         else:
             r_json = resp.json()
             participant_data += r_json.get('entry', {})
             if 'link' in r_json:
                 link_obj = r_json.get('link')
-                url = link_obj[0].get('url')
+                link_url = link_obj[0].get('url')
+                next_url = original_url + '&' + link_url[link_url.find('_token'
```
dcampbell-vumc, line 67: Can you explain what this line is doing 😅

dcampbell-vumc: Explanation of line 67 provided by @amcgrenera-vumc:
 
Instead of using the link the API provides in the returned JSON (to go to the next page of results) we are creating the link to use to go to the next page. The link provided by the API is incorrect, it drops off some of the parameters we initially provide. So we are grabbing the token from the link the API provides and appending that to our original URL to keep and make sure the initial parameters are still being sent.

The drop off of parameters is an issue on the RDR side that he said they will eventually fix their code, but this is a workaround until they are able to do so.

PR #591, opened 2020-08-26
[DC-1013] Create person table with state of residence columns using person_ext

nishanthpp93:
* Create tool to update person table with new cols

### DIFF
--- a/data_steward/tools/recreate_person.py
+++ b/data_steward/tools/recreate_person.py
@@ -1,0 +1,79 @@
+import logging
+
+from utils import bq
+
+LOGGER = logging.getLogger(__name__)
+
+NEW_PERSON_QUERY = """
+SELECT
+    p.* EXCEPT (state_of_residence_concept_id, state_of_residence_source_value),
+    ext.state_of_residence_concept_id,
+    ext.state_of_residence_source_value
+FROM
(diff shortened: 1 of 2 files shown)


data_steward/tools/recreate_person.py
```diff
@@ -0,0 +1,79 @@
...
+    parser.add_argument('-d',
+                        '--dataset_id',
+                        action='store',
+                        dest='dataset_id',
+                        help='Dataset to modify person table',
+                        required=True)
+
+    ARGS = parser.parse_args()
+    bq_client = bq.get_client(ARGS.project_id)
+    update_person(bq_client, ARGS.project_id, ARGS.dataset_id)
```
ksdkalluri, line 79: This script looks good and the integration tests work fine. But, 2 things we need to get rid of the materialized table out from the scope. The second is it's good to call this script somewhere (maybe in a shell script) where we copy the cdr form prod to output-prod.

nishanthpp93: Added notes in the ticket and CDR playbook

data_steward/tools/recreate_person.py
```diff
@@ -0,0 +1,79 @@
+import logging
```
ksdkalluri, line 1: Looks good! please merge once the integration tests pass.

ksdkalluri (approved):

PR #592, opened 2020-08-26
[EDQ-552] Add description of check local file check if upload fails in results.html

ksdkalluri:


data_steward/constants/validation/email_notification.py
```diff
@@ -31,6 +31,9 @@
 {% endif %}
 Please review the <code>results.html</code> submission report attached to this email{% if submission_error %}<br>
 and resolve the errors before making a new submission{% endif %}.<br>
+If Your files have not been successfully uploaded to the DRC.
```
dcampbell-vumc, line 34:
```suggestion
Your files have not been successfully uploaded to the DRC.
```

data_steward/resource_files/templates/hpo_report.html
```diff
@@ -233,7 +233,11 @@ <h1>
     </p>
 {% if submission_error is defined %}
     <div class="alert alert-danger">
-        Validation failed for the following reason: {{ submission_error }}.
+        Validation failed for the following reason: {{ submission_error }}.<br>
+        
+        If Your files have not been successfully uploaded to the DRC.
```
dcampbell-vumc, line 238:
```suggestion
        Your files have not been successfully uploaded to the DRC.
```

ksdkalluri: I added this because right now, we don't just check all the files are loaded and report that it's failed. so I added an "IF" so that if that's the case they can try running the local_validation.

dcampbell-vumc: Gotcha. That makes sense. This is good to go then🥳🥳

PR #593, opened 2020-08-27
[DC-1009] Truncate Data from RDR Export after a cutoff date

ksdkalluri:


data_steward/cdr_cleaner/cleaning_rules/truncate_rdr_using_date.py
```diff
@@ -0,0 +1,174 @@
...
+WHERE
+ {{field_name}} > '{{cutoff_date}}'
+""")
+
+
+class TruncateRdrData(BaseCleaningRule):
+    """
+    Apply value ranges to ensure that values are reasonable and to minimize the likelihood
+    of sensitive information (like phone numbers) within the free text fields.
+    """
```
dcampbell-vumc, line 74: This needs to have the updated description for `TruncateRdrData`

data_steward/cdr_cleaner/cleaning_rules/truncate_rdr_using_date.py
```diff
@@ -0,0 +1,174 @@
...
+        """
+        Run required steps for validation setup
+        """
+        pass
+
+    def validate_rule(self, client):
+        """
+        Validates the cleaning rule which deletes or updates the data from the tables
+        """
+        pass
```
dcampbell-vumc, line 151: Should we add `raise NotImplementedError("Please fix me.")` to these so that when we do need to implement them they are easily found?

nishanthpp93:
Thanks for requesting my input @dcampbell-vumc. You've already covered everything for the review so I don't have much to add. The queries are generated correctly. Your point about the `NonImplementedError` is accurate. I think this makes sense for `setup_validation` and `validate_rule`, but not for `setup_rule`. If there is nothing to set up, the rule should `pass`. This is called by the engine and it can break if it runs into the `NonImplementedError`. We use this error primarily for something that needs to be implemented for running a rule but isn't yet, causing unexpected behavior. It helps prevent the rule from running.

On a final note, I wanted to confirm that this rule is inclusive of the `8/1` date (and shouldn't cut-off at `7/31`).

ksdkalluri:
@nishanthpp93 @dcampbell-vumc I made the necessary changes. and I confirmed that the date should be 08/01 before starting to work on this ticket.

PR #594, opened 2020-08-27
[DC-686] Remove EHR data past date of deactivation for deactivated participants

dcampbell-vumc:
-  Created `remove_ehr_data_past_deactivation_date.py` module to utilize `retract_deactivated_pids.py` and `participant_summary_requests.py` modules to remove EHR data of deactivated participants past the date of deactivation

- Created unit and integration test modules to properly test `remove_ehr_data_past_deactivation_date`

- Modified `participant_summary_request.py` module to convert the `suspensionTime` (`deactivation_date`) field to be of type `DATE` instead of `TIMESTAMP`

### DIFF
--- a/data_steward/cdr_cleaner/clean_cdr.py
+++ b/data_steward/cdr_cleaner/clean_cdr.py
@@ -94,8 +94,8 @@ UNIONED_EHR_CLEANING_CLASSES = [
     (remove_records_with_wrong_date.get_remove_records_with_wrong_date_queries,
     ),
     (invalid_procedure_source.get_remove_invalid_procedure_source_queries,),
-    (CleanMappingExtTables,),
     (remove_ehr_data.remove_ehr_data_queries,),
+    (CleanMappingExtTables,),
 ]
 
 RDR_CLEANING_CLASSES = [
--- a/data_steward/cdr_cleaner/clean_cdr.py
+++ b/data_steward/cdr_cleaner/clean_cdr.py
@@ -95,7 +95,7 @@ UNIONED_EHR_CLEANING_CLASSES = [
     ),
     (invalid_procedure_source.get_remove_invalid_procedure_source_queries,),
     (CleanMappingExtTables,),
-    (remove_ehr_data,),
+    (remove_ehr_data.remove_ehr_data,),
 ]
 
 RDR_CLEANING_CLASSES = [
--- a/data_steward/cdr_cleaner/clean_cdr.py
+++ b/data_steward/cdr_cleaner/clean_cdr.py
@@ -95,7 +95,7 @@ UNIONED_EHR_CLEANING_CLASSES = [
     ),
     (invalid_procedure_source.get_remove_invalid_procedure_source_queries,),
     (CleanMappingExtTables,),
-    (remove_ehr_data.remove_ehr_data,),
+    (remove_ehr_data.remove_ehr_data_queries,),
 ]
 
 RDR_CLEANING_CLASSES = [
--- a/data_steward/cdr_cleaner/clean_cdr.py
+++ b/data_steward/cdr_cleaner/clean_cdr.py
@@ -39,6 +39,7 @@ import cdr_cleaner.cleaning_rules.round_ppi_values_to_nearest_integer as round_p
 import cdr_cleaner.cleaning_rules.temporal_consistency as bad_end_dates
 import cdr_cleaner.cleaning_rules.update_family_history_qa_codes as update_family_history
 import cdr_cleaner.cleaning_rules.valid_death_dates as valid_death_dates
+import cdr_cleaner.cleaning_rules.remove_ehr_data_past_deactivation_date as remove_ehr_data
 import cdr_cleaner.manual_cleaning_rules.clean_smoking_ppi as smoking
 import cdr_cleaner.manual_cleaning_rules.negative_ppi as negative_ppi
 import cdr_cleaner.manual_cleaning_rules.ppi_drop_duplicate_responses as ppi_drop_duplicates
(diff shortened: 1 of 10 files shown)


data_steward/cdr_cleaner/clean_cdr.py
```diff
@@ -85,7 +86,8 @@
     (remove_records_with_wrong_date.get_remove_records_with_wrong_date_queries,
     ),
     (invalid_procedure_source.get_remove_invalid_procedure_source_queries,),
-    (CleanMappingExtTables,)
+    (CleanMappingExtTables,),
+    (remove_ehr_data.remove_ehr_data_queries,),
```
amcgrenera-vumc, line 90: I think this needs to be above `CleanMappingExtTables`, I had recently added a cleaning rule in DEID_BASE and lauren had the following comment:

(CleanMappingExtTables,), needs to be the last rule run in this list. It's purpose is to drop any mapping records that no longer have a record in the OMOP tables.

dcampbell-vumc: I changed the location!

amcgrenera-vumc:
@dcampbell-vumc when running the integration test, I'm getting a query_list of zero. The test data entered with the `/curation/tests/test_data/nyc_five_person' files have specific dates, so you will need to make sure the test data you are entering for `self.deactivated_participants` deactivated dates will counter against the nyc test data, so some rows will need to be retracted. Unless you wrote the test to return zero?

data_steward/cdr_cleaner/cleaning_rules/remove_ehr_data_past_deactivation_date.py
```diff
@@ -0,0 +1,68 @@
...
+    their deactivation date
+
+    :param project_id: BQ name of the project
+    :param ticket_number: Jira ticket number to identify and title sandbox tables
+    :param pids_project_id: deactivated participants PIDs table in BQ's project_id
+    :param pids_dataset_id: deactivated participants PIDs table in BQ's dataset_id
+    :param tablename: The name of the table to house the deactivated participant data
+    """
+
+    ehr_union_dataset = bq_utils.get_unioned_dataset_id()
```
amcgrenera-vumc, line 42: Is there a reason you are only removing from the unioned dataset?

dcampbell-vumc: Yes, the ticket says:

Include a check for the `ehr_union` dataset that drops all rows of EHR data dated after the date of deactivation for participants who have deactivated from the Program.

I took that as this rule is only looking at the `ehr_union` dataset which is why I passed it as the dataset to check. Did I interpret that right?

dcampbell-vumc:
> @dcampbell-vumc when running the integration test, I'm getting a query_list of zero. The test data entered with the `/curation/tests/test_data/nyc_five_person' files have specific dates, so you will need to make sure the test data you are entering for `self.deactivated_participants` deactivated dates will counter against the nyc test data, so some rows will need to be retracted. Unless you wrote the test to return zero?

It definitely wasn't supposed to be zero. I'll look at the test data.

tests/unit_tests/data_steward/cdr_cleaner/cleaning_rules/remove_ehr_data_past_deactivation_date_test.py
```diff
@@ -0,0 +1,199 @@
...
+
+        mock_pid_exist.return_value = 1
+        sandbox = mock_check_sandbox.return_value = 'foo_pid_dataset_id_sandbox'
+
+        # test
+        returned_queries = red.remove_ehr_data_queries(self.project_id,
+                                                       self.ticket_number,
+                                                       self.pids_project_id,
+                                                       self.pids_dataset_id,
+                                                       self.tablename)
```
amcgrenera-vumc, line 112: Sorry one more thing-- should this be `self.pids_table`?

data_steward/cdr_cleaner/cleaning_rules/remove_ehr_data_past_deactivation_date.py
```diff
@@ -0,0 +1,68 @@
...
+
+if __name__ == '__main__':
+    ARGS = rdp.parse_args()
+    rdp.add_console_logging(ARGS.console_log)
+
+    remove_ehr_data_queries = remove_ehr_data_queries(ARGS.project_id,
+                                                      ARGS.ticket_number,
+                                                      ARGS.pids_project_id,
+                                                      ARGS.pids_dataset_id,
+                                                      ARGS.tablename)
```
amcgrenera-vumc, line 64: here too-- I think this needs to be `self.pids_table` to match the arguments in `retract_deactivated_pids.py`

tests/unit_tests/data_steward/cdr_cleaner/cleaning_rules/remove_ehr_data_past_deactivation_date_test.py
dcampbell-vumc: should `self.tablename` be `self.pids_table`? If that's the question then it's the same thing just with a different name.

data_steward/cdr_cleaner/cleaning_rules/remove_ehr_data_past_deactivation_date.py
dcampbell-vumc: Changed!

data_steward/cdr_cleaner/cleaning_rules/remove_ehr_data_past_deactivation_date.py
amcgrenera-vumc: Caroline confirmed this (I just wanted to double check!)

PR #595, opened 2020-08-28
[DC-1012] Create CR for person_ext

amcgrenera-vumc:


data_steward/cdr_cleaner/clean_cdr.py
```diff
@@ -179,7 +180,8 @@
     (CleanHeightAndWeight,),  # dependent on MeasurementRecordsSuppression
     (UnitNormalization,),  # dependent on CleanHeightAndWeight
     (DropZeroConceptIDs,),
-    (CleanMappingExtTables,)
+    (CleanMappingExtTables,),
+    (CreatePersonExtTable,)
```
lrwb-aou, line 184: `(CleanMappingExtTables,),` needs to be the last rule run in this list.  It's purpose is to drop any mapping records that no longer have a record in the OMOP tables.

data_steward/cdr_cleaner/cleaning_rules/create_person_ext_table.py
```diff
@@ -0,0 +1,130 @@
...
+import constants.cdr_cleaner.clean_cdr as cdr_consts
+from cdr_cleaner.cleaning_rules.base_cleaning_rule import BaseCleaningRule
+from constants.bq_utils import WRITE_TRUNCATE
+from utils.bq import JINJA_ENV
+
+# Query to create person_ext table
+PERSON_EXT_TABLE_QUERY = JINJA_ENV.from_string("""
+SELECT p.person_id, e.src_id,
+o.value_source_concept_id AS state_of_residence_concept_id,
+c.concept_id AS state_of_residence_source_concept_id,
```
lrwb-aou, line 28: This wasn't asked for in the ticket.

data_steward/cdr_cleaner/clean_cdr.py
amcgrenera-vumc: will adjust the order!

data_steward/cdr_cleaner/cleaning_rules/create_person_ext_table.py
amcgrenera-vumc: I have no idea where I thought I saw that... Removing now!

data_steward/cdr_cleaner/clean_cdr.py
```diff
@@ -179,6 +180,7 @@
     (CleanHeightAndWeight,),  # dependent on MeasurementRecordsSuppression
     (UnitNormalization,),  # dependent on CleanHeightAndWeight
     (DropZeroConceptIDs,),
+    (CreatePersonExtTable,),
```
lrwb-aou, line 183: This should actually be in `DEID_BASE_CLEANING_CLASSES` so it can exist for both output datasets.

data_steward/cdr_cleaner/cleaning_rules/create_person_ext_table.py
```diff
@@ -0,0 +1,129 @@
...
+        Validates the cleaning rule which deletes or updates the data from the tables
+        """
+        raise NotImplementedError("Please fix me.")
+
+    def get_sandbox_tablenames(self):
+        sandbox_table_names = list()
+        for i in range(0, len(self._affected_tables)):
+            sandbox_table_names.append(self._issue_numbers[0].lower() + '_' +
+                                       self._affected_tables[i])
+        return sandbox_table_names
```
lrwb-aou, line 112: This isn't actually accurate.  This rule doesn't make use of sandbox tables.  This should actually return an empty list.

data_steward/cdr_cleaner/cleaning_rules/create_person_ext_table.py
```diff
@@ -0,0 +1,129 @@
...
+            query_list.append({
+                cdr_consts.QUERY:
+                    PERSON_EXT_TABLE_QUERY.render(project=self.project_id,
+                                                  dataset=self.dataset_id),
+                cdr_consts.DESTINATION_TABLE:
+                    table,
+                cdr_consts.DESTINATION_DATASET:
+                    self.dataset_id,
+                cdr_consts.DISPOSITION:
+                    WRITE_TRUNCATE
```
lrwb-aou, line 84: If you were to change the query to a DDL statement, (like `CREATE OR REPLACE TABLE...`), you could drop these parameters entirely.

data_steward/cdr_cleaner/clean_cdr.py
amcgrenera-vumc: moving

data_steward/cdr_cleaner/cleaning_rules/create_person_ext_table.py
amcgrenera-vumc: I had a feeling, I tried taking this entire function out and realized it broke the class, so I left as is, but I will have this return an empty list instead, thank you!

data_steward/cdr_cleaner/cleaning_rules/create_person_ext_table.py
amcgrenera-vumc: I thought we weren't supposed to use DDL statements? Is this not the case anymore?

lrwb-aou: We still aren't using DML statements frequently (`update`, `delete`), but DDL statements are ok (create a table from a result set).  This is an optional suggestion and is not a required change.

amcgrenera-vumc: ahh was confusing DML vs DDL!

data_steward/cdr_cleaner/cleaning_rules/create_person_ext_table.py
```diff
@@ -0,0 +1,126 @@
...
+        Initialize the class with proper information.
+
+        Set the issue numbers, description and affected datasets. As other tickets may affect
+        this SQL, append them to the list of Jira Issues.
+        DO NOT REMOVE ORIGINAL JIRA ISSUE NUMBERS!
+        """
+        desc = ('Create person_ext table')
+        super().__init__(issue_numbers=['DC1012'],
+                         description=desc,
+                         affected_datasets=[cdr_consts.DEID_CLEAN],
```
lrwb-aou, line 58: This should be `DEID_BASE` because it is specifically applied with those rules.
```suggestion
                         affected_datasets=[cdr_consts.DEID_BASE],
```

amcgrenera-vumc: good catch, thank you!

PR #597, opened 2020-09-01
[DC-1017] Update custom concepts.

ksdkalluri:


ksdkalluri:
> Apologies for missing the tabs at the end of the last two lines and thanks for catching this. Everything looks good.
> There is another reference of 3000000001 that needs to be replaced
> at
> 
> https://github.com/all-of-us/curation/blob/2a0c7a7510ed9d94d682becc3e5fafdf93139cf6/data_steward/cdr_cleaner/cleaning_rules/repopulate_person_post_deid.py#L140

Fixed it @ChaoPang

PR #599, opened 2020-09-04
[DC-1051] Update "PPI Drop Duplicates" Rule to exclude COPE responses

ksdkalluri:


data_steward/cdr_cleaner/cleaning_rules/drop_ppi_duplicate_responses.py
```diff
@@ -0,0 +1,214 @@
...
+SELECT
+  *
+FROM
+  `{{project}}.{{dataset}}.observation`
+WHERE
+  observation_id IN 
+""")
+
+REMOVE_DUPLICATE_TEMPLATE = jinja_env.from_string("""
+DELETE
```
amcgrenera-vumc, line 59: Should this be updated to not use a DML statement? Replace with `SELECT` and `WRITE TRUNCATE`?

amcgrenera-vumc:
@ksdkalluri This looks good to me, unit and integration tests are passing locally for me as well. I have one comment on the DML statement, but otherwise this looks good to me.

data_steward/cdr_cleaner/cleaning_rules/drop_ppi_duplicate_responses.py
ksdkalluri: @amcgrenera-vumc Because it is affecting a large amount of data and either of them is performing similarly, so I left it as is. I can update it if it is needed.

amcgrenera-vumc: No if it makes more sense to leave as `DELETE` that's fine!

amcgrenera-vumc (approved):

PR #601, opened 2020-09-07
[DC-1061] Modify Numeric PPI response cleaning rule to include "11+" generalization

dcampbell-vumc:
(1) Added new queries `HOUSEHOLD_SIZE_SANDBOX_QUERY` and `CLEAN_HOUSEHOLD_SIZE_QUERY` to generalize participants with more than 11 members in their household
(2) Modified unit and integration tests `clean_ppi_numeric_fields_using_parameters_test` to properly test changes to `CleanPPINumericFieldsUsingParameters` module


data_steward/cdr_cleaner/cleaning_rules/clean_ppi_numeric_fields_using_parameters.py
```diff
@@ -82,6 +86,51 @@
...
+""")
+
+CLEAN_HOUSEHOLD_SIZE_QUERY = jinja_env.from_string("""SELECT
+    observation_id,
+    person_id,
+    observation_concept_id,
+    observation_date,
+    observation_datetime,
+    observation_type_concept_id,
+CASE
```
ksdkalluri, line 126: I will suggest adding the conditions in the existing query, so we can run only 2 queries instead of 4 queries. 
You can add the new ticket number at the end. you can still use the 827 as the main ticket number.

data_steward/cdr_cleaner/cleaning_rules/clean_ppi_numeric_fields_using_parameters.py
```diff
@@ -82,6 +86,51 @@
...
+    WHEN observation_source_concept_id IN (1333015, 1585889) AND (value_as_number < 0 OR value_as_number > 20) THEN NULL
+  ELSE value_as_number
+END AS
+    value_as_number,
+    value_as_string,
+CASE
+    WHEN observation_source_concept_id IN (1333015, 1585889) AND (value_as_number < 0 OR value_as_number > 20) THEN 2000000013
```
ksdkalluri, line 133: By this condition, `2000000013` concept_id is being updated to households less than 0 and greater than 20. you should be only updating the concept for all the values that are greater than 11.

Please take a look at this condition and update it accordingly.

ksdkalluri (changes requested):

data_steward/cdr_cleaner/cleaning_rules/clean_ppi_numeric_fields_using_parameters.py
```diff
@@ -166,19 +163,16 @@ def setup_validation(self, client):
...
         """
-        pass
+        raise NotImplementedError("Please fix me.")
 
     def get_sandbox_tablenames(self):
-        sandbox_table_names = list()
-        sandbox_table_names.append(self._issue_numbers[0] +
-                                   self._affected_tables[0])
-        return sandbox_table_names
+        return [INVALID_VALUES_SANDBOX]
```
ksdkalluri, line 175:
```suggestion
        return f'{self._issue_numbers[0].lower()}_{self._affected_tables[0]}'
```

You can use this method to return the sandbox table name, like earlier. Since you have all the tickets listed in the class, it would be easier to find the original ticket. that shouldn't be a problem.

ksdkalluri (changes requested):

data_steward/cdr_cleaner/cleaning_rules/clean_ppi_numeric_fields_using_parameters.py
```diff
@@ -71,17 +63,19 @@
     observation_datetime,
     observation_type_concept_id,
 CASE
-    WHEN observation_concept_id IN (1585889, 1585890) AND (value_as_number < 0 OR value_as_number > 20) THEN NULL
+    WHEN observation_concept_id IN (1585889, 1585890) AND (value_as_number < 0) THEN NULL
     WHEN observation_concept_id IN (1585795, 1585802, 1585864, 1585870, 1585873, 1586159, 1586162) AND (value_as_number < 0 OR value_as_number > 99) THEN NULL
     WHEN observation_concept_id = 1585820 AND (value_as_number < 0 OR value_as_number > 255) THEN NULL
+    WHEN observation_source_concept_id IN (1333015, 1585889) AND (value_as_number < 0) THEN NULL
```
ksdkalluri, line 69: My bad Danyel, you are right value_as_number should be set to NULL for < 0 and > 11 values. I just re-read the scope. and only for the values > 11 value_as_concept_id should be set to 2000000013.

ksdkalluri (changes requested):

data_steward/cdr_cleaner/cleaning_rules/clean_ppi_numeric_fields_using_parameters.py
```diff
@@ -71,17 +59,19 @@
     observation_datetime,
     observation_type_concept_id,
 CASE
-    WHEN observation_concept_id IN (1585889, 1585890) AND (value_as_number < 0 OR value_as_number > 20) THEN NULL
+    WHEN observation_concept_id IN (1585889, 1585890) AND (value_as_number < 0) THEN NULL
```
ksdkalluri, line 62: You still need to leave the value_as_number  > 20 here, so the values for the 1585890 will still be set to NULL.

ksdkalluri (changes requested):

ksdkalluri (changes requested):
Please use the suggested when conditions to fix the issue.

ksdkalluri (approved):

PR #604, opened 2020-09-08
[DC-1053] Cleaning Rule for fixing the unmapped survey answers

ChaoPang:
- completed the cleaning rule fix_unmapped_survey_answers
- added the unit test for fixed_unmapped_survey_answers
- added the integration test for fix_unmapped_survey_answers

### DIFF
--- a/data_steward/cdr_cleaner/cleaning_rules/fix_unmapped_survey_answers.py
+++ b/data_steward/cdr_cleaner/cleaning_rules/fix_unmapped_survey_answers.py
@@ -48,9 +48,6 @@ Actually we don't need to do anything for this case because value_as_concept_id 
 
 import logging
 
-# Third party imports
-from jinja2 import Environment
-
 # Project imports
 import constants.cdr_cleaner.clean_cdr as cdr_consts
 from cdr_cleaner.cleaning_rules.base_cleaning_rule import BaseCleaningRule, query_spec_list
@@ -147,7 +144,7 @@ class FixUnmappedSurveyAnswers(BaseCleaningRule):
                 SANDBOX_FIX_UNMAPPED_SURVEY_ANSWERS_QUERY.render(
                     project=self.project_id,
                     sandbox_dataset=self.sandbox_dataset_id,
-                    sandbox_table=self.get_sandbox_tablenames()[0],
+                    sandbox_table=self.sandbox_table_for(OBSERVATION),
                     dataset=self.dataset_id)
         }
 
@@ -156,7 +153,7 @@ class FixUnmappedSurveyAnswers(BaseCleaningRule):
                 UPDATE_FIX_UNMAPPED_SURVEY_ANSWERS_QUERY.render(
                     project=self.project_id,
                     sandbox_dataset=self.sandbox_dataset_id,
-                    sandbox_table=self.get_sandbox_tablenames()[0],
+                    sandbox_table=self.sandbox_table_for(OBSERVATION),
                     dataset=self.dataset_id),
             cdr_consts.DESTINATION_TABLE:
                 OBSERVATION,
@@ -184,11 +181,22 @@ class FixUnmappedSurveyAnswers(BaseCleaningRule):
         raise NotImplementedError("Please fix me.")
 
     def get_sandbox_tablenames(self):
-        sandbox_table_names = list()
-        for i in range(0, len(self._affected_tables)):
-            sandbox_table_names.append('_'.join(self._issue_numbers).lower() +
-                                       '_' + self._affected_tables[i])
-        return sandbox_table_names
+        return [
+            self.sandbox_table_for(affected_table)
+            for affected_table in self._affected_tables
+        ]
+
+    def sandbox_table_for(self, affected_table):
+        """
+        A helper function to retrieve the sandbox table name for the affected_table
+        :param affected_table: 
+        :return: 
+        """
+        if affected_table not in self._affected_tables:
+            raise LookupError(
+                f'{affected_table} is not define as an affected table in {self._affected_tables}'
+            )
+        return f'{"_".join(self._issue_numbers).lower()}_{affected_table}'
 
 
 if __name__ == '__main__':
--- a/data_steward/cdr_cleaner/cleaning_rules/fix_unmapped_survey_answers.py
+++ b/data_steward/cdr_cleaner/cleaning_rules/fix_unmapped_survey_answers.py
@@ -1,0 +1,203 @@
+"""Original Issues: DC-1043, DC-1053 
+
+PPI answers are mapped to standard answer concepts in concept_relationship through 'Maps to value' (it has been this 
+case historically and it is still the case now), however, there are a bunch of PPI answer concepts (619 in total in 
+rdr20200807) missing such relationships in concept_relationship. Interestingly, the corresponding standard PPI 
+answer concepts could be found through 'Maps to'.  This might have been a bug in the vocabulary and they 
+probably should’ve used 'Maps to value' for mapping Answer concepts. 
+
+There are 619 unique answers (value_source_concept_ids) in observation that are mapped to a '0' value_as_concept_id, 
+out of which 368 are standard concepts, 202 are non-standard concepts that could be mapped to a standard concept 
+through Maps to in concept_relationship, and 49 are deprecated concepts that do not map to anything.
+
+For the 368 standard answers, we could just use it as-is for populating value_as_concept_id. However, among the 202 
+non-standard concepts, not all of the mapped standard concepts are classified as 'Answer' and they could belong to 
+other concept_classes. Below is a breakdown of the concept_class_id of the mapped concepts: 
+
+95 Context-dependent
+89 Answer
+8 Clinical Finding
+6 Question
+3 Unit
+1 Module
+
+Question or Module concept classes don't make sense so will get excluded. In conclusion, 
+
+1. For 368 standard concepts --> Set value_as_concept_id to the source_value_concept_id.
+2. For 49 deprecated concepts --> Set value_as_concept_id to 0. Actually we don't need to do anything for this case 
+because visit_as_concept_id is already 0.
+3, For 202 non-standard concepts --> Set value_as_concept_id to standard concept ids mapped through 'Maps to' 
+for the concept classes ('Answer', 'Context-dependent', 'Clinical Finding', 'Unit') only. 
+"""
+
+import logging
+
+# Third party imports
+from jinja2 import Environment
+
+# Project imports
+import constants.cdr_cleaner.clean_cdr as cdr_consts
+from cdr_cleaner.cleaning_rules.base_cleaning_rule import BaseCleaningRule, query_spec_list
+from constants.bq_utils import WRITE_TRUNCATE
+from common import OBSERVATION
+
+LOGGER = logging.getLogger(__name__)
+
+jinja_env = Environment(
+    # help protect against cross-site scripting vulnerabilities
+    autoescape=True,
+    # block tags on their own lines
+    # will not cause extra white space
+    trim_blocks=True,
+    lstrip_blocks=True,
+    # syntax highlighting should be better
+    # with these comment delimiters
+    comment_start_string='--',
+    comment_end_string=' --')
+
+JIRA_ISSUE_NUMBERS = ['DC1043', 'DC1053']
+
+# Query to create tables in sandbox with the rows that will be updated per cleaning rule
+SANDBOX_FIX_UNMAPPED_SURVEY_ANSWERS_QUERY = jinja_env.from_string("""
+CREATE OR REPLACE TABLE `{{project}}.{{sandbox_dataset}}.{{sandbox_table}}` AS
+SELECT
+  o.*,
+  COALESCE(cr.concept_id_2, 0) AS new_value_as_concept_id
+FROM
+  `{{project}}.{{dataset}}.observation` AS o
+JOIN
+  `{{project}}.{{dataset}}.concept_relationship` AS cr
+ON
+  o.value_source_concept_id = cr.concept_id_1 AND cr.relationship_id = 'Maps to'
+JOIN 
+  `{{project}}.{{dataset}}.concept` AS c
+ON
+  c.concept_id = cr.concept_id_2 AND c.concept_class_id IN ('Answer', 'Context-dependent', 'Clinical Finding', 'Unit')
+WHERE
+    o.observation_type_concept_id = 45905771 --Observation Recorded from a Survey --
+        AND o.value_source_concept_id <> 0 AND o.value_as_concept_id = 0
+""")
+
+UPDATE_FIX_UNMAPPED_SURVEY_ANSWERS_QUERY = jinja_env.from_string("""
+SELECT
+  o.observation_id,
+  o.person_id,
+  o.observation_concept_id,
+  o.observation_date,
+  o.observation_datetime,
+  o.observation_type_concept_id,
+  o.value_as_number,
+  o.value_as_string,
+  COALESCE(o2.new_value_as_concept_id, o.value_as_concept_id, 0) AS value_as_concept_id,
+  o.qualifier_concept_id,
+  o.unit_concept_id,
+  o.provider_id,
+  o.visit_occurrence_id,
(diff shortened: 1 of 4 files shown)


data_steward/cdr_cleaner/cleaning_rules/fix_unmapped_survey_answers.py
```diff
@@ -0,0 +1,203 @@
...
+case historically and it is still the case now), however, there are a bunch of PPI answer concepts (619 in total in 
+rdr20200807) missing such relationships in concept_relationship. Interestingly, the corresponding standard PPI 
+answer concepts could be found through 'Maps to'.  This might have been a bug in the vocabulary and they 
+probably should’ve used 'Maps to value' for mapping Answer concepts. 
+
+There are 619 unique answers (value_source_concept_ids) in observation that are mapped to a '0' value_as_concept_id, 
+out of which 368 are standard concepts, 202 are non-standard concepts that could be mapped to a standard concept 
+through Maps to in concept_relationship, and 49 are deprecated concepts that do not map to anything.
+
+For the 368 standard answers, we could just use it as-is for populating value_as_concept_id. However, among the 202
```
mark-velez, line 13: I might be getting thrown off by "however" on this line- to be sure, please let me know if this is correct. Here's a diagram to illustrate my understanding of how `value_as_concept_id` is derived for these rows. 
image
We intend to replace the value in `value_as_concept_id` with that in `value_source_concept_id` whenever the latter is a standard concept. For rows whose `value_source_concept_id` refers to a non-standard concept, we populate `value_as_concept_id` with a standard concept obtained via `Maps to` (`maps_to_concept` in the diagram) which belongs to one of the appropriate classes (Context-dependent, Answer, Clinical Finding, Unit). No special treatment is needed for deprecated concepts.

data_steward/cdr_cleaner/cleaning_rules/fix_unmapped_survey_answers.py
```diff
@@ -0,0 +1,203 @@
+"""Original Issues: DC-1043, DC-1053 
+
+PPI answers are mapped to standard answer concepts in concept_relationship through 'Maps to value' (it has been this
```
mark-velez, line 3: I realize pylint passes, but should [max line length be 100](https://github.com/all-of-us/curation/blob/develop/.pylintrc#L371)? I configured my editor so it complains about this. Let me know if I'm wrong- I'm happy to start using lines that are a little longer. :)

data_steward/cdr_cleaner/cleaning_rules/fix_unmapped_survey_answers.py
```diff
@@ -0,0 +1,203 @@
...
+3 Unit
+1 Module
+
+Question or Module concept classes don't make sense so will get excluded. In conclusion, 
+
+1. For 368 standard concepts --> Set value_as_concept_id to the source_value_concept_id.
+2. For 49 deprecated concepts --> Set value_as_concept_id to 0. Actually we don't need to do anything for this case 
+because visit_as_concept_id is already 0.
+3, For 202 non-standard concepts --> Set value_as_concept_id to standard concept ids mapped through 'Maps to' 
+for the concept classes ('Answer', 'Context-dependent', 'Clinical Finding', 'Unit') only.
```
mark-velez, line 30: I recommend excluding information particular to the current RDR or vocabulary datasets (e.g. number of concepts) as this quickly becomes outdated (it already has).

tests/unit_tests/data_steward/cdr_cleaner/cleaning_rules/fix_unmapped_survey_answers_test.py
```diff
@@ -0,0 +1,107 @@
...
+3 Unit
+1 Module
+
+Question or Module concept classes don't make sense so will get excluded. In conclusion, 
+
+1. For 368 standard concepts --> Set value_as_concept_id to the source_value_concept_id.
+2. For 49 deprecated concepts --> Set value_as_concept_id to 0. Actually we don't need to do anything for this case 
+because visit_as_concept_id is already 0.
+3, For 202 non-standard concepts --> Set value_as_concept_id to standard concept ids mapped through 'Maps to' 
+for the concept classes ('Answer', 'Context-dependent', 'Clinical Finding', 'Unit') only.
```
mark-velez, line 30: Do you feel strongly about repeating this comment? It will be easy for the two copies to get out of sync.

mark-velez (commented):
@ChaoPang I **very much** appreciate the docstring you provided here! Before I dive into reviewing the code, I want to make sure that I correctly understand the approach as it is described. Could you please see my comments and let me know?

data_steward/cdr_cleaner/cleaning_rules/fix_unmapped_survey_answers.py
```diff
@@ -0,0 +1,203 @@
...
+    # help protect against cross-site scripting vulnerabilities
+    autoescape=True,
+    # block tags on their own lines
+    # will not cause extra white space
+    trim_blocks=True,
+    lstrip_blocks=True,
+    # syntax highlighting should be better
+    # with these comment delimiters
+    comment_start_string='--',
+    comment_end_string=' --')
```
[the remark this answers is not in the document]
ChaoPang: @mark-velez thanks for this, I didn't know there was an environment in `utils.bq`. Will update this

data_steward/cdr_cleaner/cleaning_rules/fix_unmapped_survey_answers.py
ChaoPang: @mark-velez very nice diagram and this clearly illustrates the logic used in the query, this is way better than words (especially my words). 
This is a 100% accurate conceptual representation of the logic. I just wanted to mention that there might be a slight discrepancy in the query though, specifically for cases where `value_source_concept_id` is standard, the query replaces `value_as_concept_id` with the mapped concept from `concept_relationship`, which is the same as `value_source_concept_id` because the standard concepts map to themselves as well. This simplifies the query

data_steward/cdr_cleaner/cleaning_rules/fix_unmapped_survey_answers.py
ChaoPang: Agreed

tests/unit_tests/data_steward/cdr_cleaner/cleaning_rules/fix_unmapped_survey_answers_test.py
ChaoPang: I have no feeling for this comment, I will delete it.

data_steward/cdr_cleaner/cleaning_rules/fix_unmapped_survey_answers.py
ChaoPang: Will do

data_steward/cdr_cleaner/cleaning_rules/fix_unmapped_survey_answers.py
```diff
@@ -0,0 +1,198 @@
...
+Actually we don't need to do anything for this case because visit_as_concept_id is already 0. 
+
+3, For non-standard concepts --> Set value_as_concept_id to standard concept ids mapped through 
+'Maps to' for the concept classes ( 'Answer', 'Context-dependent', 'Clinical Finding', 
+'Unit') only. """
+
+import logging
+
+# Third party imports
+from jinja2 import Environment
```
amcgrenera-vumc, line 41: this can be removed now since you are using the `jinja_env` from `utils.bq`

data_steward/cdr_cleaner/cleaning_rules/fix_unmapped_survey_answers.py
```diff
@@ -0,0 +1,198 @@
...
+SELECT
+  o.*,
+  COALESCE(cr.concept_id_2, 0) AS new_value_as_concept_id
+FROM
+  `{{project}}.{{dataset}}.observation` AS o
+JOIN
+  `{{project}}.{{dataset}}.concept_relationship` AS cr
+ON
+  o.value_source_concept_id = cr.concept_id_1 AND cr.relationship_id = 'Maps to'
+JOIN
```
amcgrenera-vumc, line 66: I didn't realize you could do an `AND` in the `ON` statement, this is cool and good to know!

data_steward/cdr_cleaner/cleaning_rules/fix_unmapped_survey_answers.py
amcgrenera-vumc: Thank you both, the diagram and description are super helpful in understanding this query!

tests/integration_tests/data_steward/cdr_cleaner/cleaning_rules/fix_unmapped_survey_answers_test.py
```diff
@@ -0,0 +1,184 @@
...
+                (1585609,'Asian Specific: Asian Specific Indian','Observation','PPI','Answer',null,'AsianSpecific_AsianSpecificIndian','2017-05-22','2099-12-31',null),
+                (1384515,'Kidney Conditions: Kidney Stones','Observation','PPI','Answer',null,'KidneyConditions_KidneyStones','2018-10-09','2099-12-31',null),
+                (4179221,'History of calculus of kidney','Observation','SNOMED','Context-dependent','S','429025008','2008-01-31','2099-12-31',null),
+                (4088548,'Seen by general practitioner','Observation','SNOMED','Clinical Finding','S','185278000','1970-01-01','2099-12-31',null),
+                (1586151,'Asian: Asian Specific','Observation','PPI','Question','S','Asian_AsianSpecific','2017-05-17','2099-12-31',null),
+                (1333079,'Moderately unhappy','Observation','PPI','Answer','S','mhwb_a_40','2020-05-07','2099-12-31',null),
+                (9580,'month','Unit','UCUM','Unit','S','mo','1970-01-01','2099-12-31',null),
+                (1332724,'Months','Observation','PPI','Answer',null,'cope_a_198','2020-05-07','2099-12-31',null)
+        """)
```
amcgrenera-vumc, line 120: To piggy back on Mark's comment for max line length of 100, I also configured my editor so it complains about this too.

amcgrenera-vumc:
@ChaoPang just a couple comments above. But this looks good and makes sense to me if @mark-velez is good with it! 

The unit and integration tests are passing for me locally, so you may just need to re-run the integration_test in CircleCI, it seems to be finicky lately. 

I also wanted to ask if this cleaning rule needs to be added to the pipeline? If so, `clean_cdr.py` will need to be updated with this class and added to the `RDR_CLEANING_CLASSES`

data_steward/cdr_cleaner/cleaning_rules/fix_unmapped_survey_answers.py
```diff
@@ -0,0 +1,198 @@
+"""Original Issues: DC-1043, DC-1053 
+
```
mark-velez, line 2: I suggest a TL;DR of the problem and how this rule addresses it like:
```suggestion
"""Original Issues: DC-1043, DC-1053 
PPI records may have value_as_concept_id set to 0 by the RDR ETL as a result of changes to the 
vocabulary. This rule sets value_as_concept_id to an appropriate value for these records. Before 
they are updated rows are stored in a sandbox table with the same columns as observation and 
an additional column new_value_as_concept_id.

The RDR ETL populates value_as_concept_id with a concept in the concept_relationship table 
having the 'Maps to value' relationship with value_source_concept_id, however concept_relationship
may not have such a mapping for all PPI answer concepts and in these cases the associated 
value_as_concept_id will get a default value of 0. For these rows this rule uses concepts related 
to value_source_concept_id by 'Maps to' instead. In order to prevent usage of erroneous mappings 
which may be present in the vocabulary, the target concept is also constrained to a limited set 
of concept classes ('Answer', 'Context-dependent', 'Clinical Finding', 'Unit').
```
Arguably some of the comment is TMI as the characterization pertains specifically to the current vocabulary/RDR data. For example I think the numbered bullets are observations - see what I did there? :) - about how concepts end up being mapped currently but these may not always hold true; the update query does not distinguish between standard vs. non-standard or deprecated concepts.

data_steward/cdr_cleaner/cleaning_rules/fix_unmapped_survey_answers.py
```diff
@@ -0,0 +1,198 @@
...
+Context-dependent
+Answer
+Clinical Finding
+Question
+Unit
+Module
+
+Question or Module concept classes don't make sense so will get excluded. In conclusion, 
+
+1. For standard concepts --> Set value_as_concept_id to the source_value_concept_id.
```
mark-velez, line 29:
```suggestion
1. For standard concepts --> Set value_as_concept_id to the value_source_concept_id. 
```

data_steward/cdr_cleaner/cleaning_rules/fix_unmapped_survey_answers.py
```diff
@@ -0,0 +1,198 @@
...
+Question
+Unit
+Module
+
+Question or Module concept classes don't make sense so will get excluded. In conclusion, 
+
+1. For standard concepts --> Set value_as_concept_id to the source_value_concept_id. 
+
+2. For deprecated concepts --> Set value_as_concept_id to 0. 
+Actually we don't need to do anything for this case because visit_as_concept_id is already 0.
```
mark-velez, line 32:
```suggestion
Actually we don't need to do anything for this case because value_as_concept_id is already 0. 
```

tests/integration_tests/data_steward/cdr_cleaner/cleaning_rules/fix_unmapped_survey_answers_test.py
```diff
@@ -0,0 +1,184 @@
...
+
+    def setUp(self):
+        """
+        Create empty tables for the rule to run on
+        """
+        # Create the observation, concept, and concept_relationship tables required for the test
+        super().setUp()
+
+    def test_fix_unmapped_survey_answers(self):
+        """
```
mark-velez, line 62: OMG this test is so good :1st_place_medal:

tests/integration_tests/data_steward/cdr_cleaner/cleaning_rules/fix_unmapped_survey_answers_test.py
```diff
@@ -0,0 +1,184 @@
...
+                (8, 888888, 0, '2015-07-15', 45905771, 0, 1585536)
+                """)
+        insert_observation_query = observation_tmpl.render(
+            fq_dataset_name=f'{self.project_id}.{self.dataset_id}')
+
+        # Load test data
+        self.load_test_data([
+            insert_observation_query, insert_concept_query,
+            insert_concept_relationship_query
+        ])
```
mark-velez, line 161: Executing a single bigquery script would be a lot more efficient:
```suggestion
        self.load_test_data([
            f'''{insert_observation_query};
                {insert_concept_query};
                {insert_concept_relationship_query}'''
        ])
```

mark-velez (approved):
This is great. Please take or leave my suggestions as you wish.

tests/integration_tests/data_steward/cdr_cleaner/cleaning_rules/fix_unmapped_survey_answers_test.py
mark-velez: Yeah my editor is complaining but in this case I don't think I'd want the line length <= 100. I guess I'll try to make my editor not complain on test scripts. :man_shrugging:

data_steward/cdr_cleaner/cleaning_rules/fix_unmapped_survey_answers.py
ChaoPang: Thanks for catching this! Will do

data_steward/cdr_cleaner/cleaning_rules/fix_unmapped_survey_answers.py
ChaoPang: Writing has always been my weakness. Your summary is way better than mine, I will accept it blindly.

amcgrenera-vumc (approved):

PR #605, opened 2020-09-08
[DC-984] Update clean engine to accommodate old-style and new-style cleaning rules

nishanthpp93:
Rebased version of #569


data_steward/cdr_cleaner/clean_cdr_engine.py
```diff
@@ -37,129 +41,225 @@ def add_console_logging(add_handler):
...
-                                    cdr_consts.MODULE_NAME_DEFAULT_VALUE)
-        function_name = statement.get(cdr_consts.FUNCTION_NAME,
-                                      cdr_consts.FUNCTION_NAME_DEFAULT_VALUE)
+    if project_id is None or project_id == '' or project_id.isspace():
+        project_id = app_identity.get_application_id()
+        LOGGER.info(f"project_id not provided, using default {project_id}")
+
+    # Set up
+    client = bq.get_client(project_id=project_id)
+    sandbox_dataset_id = create_sandbox(project_id, dataset_id, client)
```
mark-velez, line 71: Data hiding the sandbox like this seems to add complexity which _I think_ would otherwise be avoided if it were a `clean_dataset()` argument.

The behavior of `clean_dataset()` when the sandbox dataset already exists is ambiguous. If the sandbox is implicitly determined by `clean_dataset()` then the function will either A) fail or B) make the decision to proceed and overwrite (or whatever- some decision in which the caller is given no say). If `sandbox_dataset_id` is an argument, perhaps the same assumptions made about other args (`project_id`, `dataset_id`) would apply- the resource exists and this process has appropriate access.

Another angle is we can say the signature of `clean_dataset()` consists of three types of args:
1. the cleaning rules to apply (`rules`)
2. the params required by all cleaning rules (`project_id`, `dataset_id`,)
3. the params which may or may not be used by some of the cleaning rules (`kwargs`)

Looking at it this way it would seem that `sandbox_dataset_id` is missing from 2. Now some of the legacy rules do not accept `sandbox_dataset_id` but I understand this is just a technicality resulting from their being authored prior to our sandboxing practices. They could (if not "should") be refactored to use a sandbox.

I think making `sandbox_dataset_id` a parameter would also increase testability. I recall there was a test we looked at last week that needed implicit knowledge of what the `sandbox_dataset_id` was.

data_steward/cdr_cleaner/clean_cdr_engine.py
```diff
@@ -37,129 +41,225 @@ def add_console_logging(add_handler):
...
+    labels = {'type': 'sandbox', 'dataset_id': dataset_id}
+    sandbox_dataset = bq.define_dataset(project_id,
+                                        sandbox_dataset_id,
+                                        description=f'Sandbox for {dataset_id}',
+                                        label_or_tag=labels)
+    sandbox_dataset = client.create_dataset(sandbox_dataset, exists_ok=True)
+    return sandbox_dataset.dataset_id
+
+
+def clean_dataset(project_id=None, dataset_id=None, rules=None, **kwargs):
```
mark-velez, line 55: I think we should like `project_id`, `dataset_id`, `rules` to be required rather than optional arguments.
1. Currently if `rules` is not provided this function will raise an error when it gets to the for loop. This is avoided if the argument is required.
1. The caller may simply forget to set one of these args and inadvertently edit resources (e.g. env var for project is set to prod instead of test).
1. The default values make calls to this function dependent on external configuration (env variables) which can be ambiguous and difficult to manage. If it is a required arg then all the configs are explicit and in one place.
1. There do not appear to be any calls to `clean_dataset()` where the defaults are used. This could be because there has not been a need. I would advise adding defaults to parameters like these only when there is a very important use case.

data_steward/cdr_cleaner/clean_cdr_engine.py
nishanthpp93: Sure Mark, I was debating that as well. I thought of including it in `kwargs` earlier but a separate parameter is better. Do you think the sandbox should still be created by the engine?

data_steward/cdr_cleaner/clean_cdr_engine.py
```diff
@@ -37,129 +41,225 @@ def add_console_logging(add_handler):
...
+    """
+    params = inspect.signature(clazz).parameters
+    mandatory_params = ['project_id', 'dataset_id']
+    rule_params = [param for param in params if param not in mandatory_params]
+    # filter kwargs based on required params
+    kwargs = {k: v for k, v in kwargs.items() if k in rule_params}
+    if any(param not in kwargs.keys() for param in rule_params):
+        raise ValueError(
+            f'Params {[param for param in rule_params if param not in kwargs.keys()]} '
+            f'not provided for cleaning rule {clazz.__name__}')
```
mark-velez, line 180: Yes! This should allow inversion of control such that the rules being applied dictate to the runner the required args. Very cool!

Based on the docs I put together this will need to handle parameters with defaults- perhaps something like this:
```suggestion
    rule_params = {k: v for k, v in params.items() if k not in mandatory_params}
    # filter kwargs based on required params
    kwargs = {k: v for k, v in kwargs.items() if k in rule_params.keys()}
    missing = [k for k, v in rule_params.items() 
               if k not in kwargs.keys() and v is inspect.Parameter.empty]
    if any(missing):
        raise ValueError(
            f'Params {missing} '
            f'not provided for cleaning rule {clazz.__name__}')
```

For testability I recommend extracting out a `get_all_params` function:
```python
def get_all_params(clazz) -> List[str]:
"""
Get all parameters required by the specified cleaning rule
"""
``` 
Naming. Maybe call it `get_custom_kwargs` rather than `validate_params` :man_shrugging:? Naming.

data_steward/cdr_cleaner/clean_cdr_engine.py
```diff
@@ -37,129 +41,225 @@ def add_console_logging(add_handler):
...
+        setup_function: function that sets up the tables for the rule
+        rule_info: dictionary of information about the rule
+            keys:
+                query_function: function that generates query_list
+                setup_function: function that sets up the tables for the rule
+                function_name: name of the query function
+                module_name: name of the module containing the function
+                line_no: points to the source line where query_function is
+    """
+    kwargs = validate_params(clazz, **kwargs)
```
mark-velez, line 203: Is it possible for this call to move up a level? As it stands a missing parameter may only be caught **after** the engine has already visited (e.g. executed) one or more rules. Alternatively, `clean_dataset` and `get_query_list` could collect the full set of parameters required by all rules and proceed **iff** the minimum set are provided. This would save a whole lot of pain.

mark-velez: I think it would be appropriate for a function which collects all the params required by a run and compares them with the supplied args to have the word "validate" in its name.

data_steward/cdr_cleaner/cleaning_rules/deid/fitbit_dateshift.py
```diff
@@ -48,8 +48,8 @@ def __init__(self, project_id, dataset_id, sandbox_dataset_id,
                          dataset_id=dataset_id,
                          sandbox_dataset_id=sandbox_dataset_id,
                          affected_tables=self.tables,
-                         map_dataset=map_dataset_id,
-                         map_table=map_tablename,
+                         mapping_dataset_id=mapping_dataset_id,
+                         mapping_table_id=DEID_MAP,
```
mark-velez, line 52: @lrwb-aou is it safe to assume [at least for Erwin release] that any instance of `FitbitDateShiftRule` should have `mapping_table_id = "_deid_map"` or do we need to parameterize `mapping_table_id`?

tests/integration_tests/data_steward/cdr_cleaner/cleaning_rules/bigquery_tests_base.py
```diff
@@ -58,6 +58,9 @@ def initialize_class_vars(cls):
             # this should always be for a test project.  a check is implemented to
             # make sure this is true
             cls.project_id = ''
+            cls.dataset_id = ''
+            cls.sandbox_id = ''
+            cls.kwargs = {}
```
mark-velez, line 63: Is this in order to call `clean_dataset()` in `default_test()`? If so, it seems like kwargs should live in `CleaningRulesTestBase`. Please add a comment describing what this is.

data_steward/cdr_cleaner/clean_cdr_engine.py
```diff
@@ -37,129 +41,225 @@ def add_console_logging(add_handler):
...
-    else:
-        LOGGER.warning(
-            f"No clean rules successfully applied to {project}.{data_stage}")
+    :param clazz: Clean class or clean function to check the signature of
+    :param kwargs: optional arguments provided by the user
+    :return: filtered dictionary of kwargs
+    :raises ValueError: if a required param for 'clazz' is missing from kwargs
+    """
+    params = inspect.signature(clazz).parameters
+    mandatory_params = ['project_id', 'dataset_id']
```
mark-velez, line 173: I recommend putting this in a consts file and renaming to something like `CLEAN_ENGINE_REQUIRED_PARAMS`

data_steward/cdr_cleaner/clean_cdr.py
```diff
@@ -200,498 +186,105 @@
...
+        dest='data_stage',
+        action='store',
+        type=DataStage,
+        choices=list([s for s in DataStage if s is not DataStage.UNSPECIFIED]),
+        help='Specify the dataset')
+    return engine_parser
 
 
-def clean_combined_dataset(project_id=None, dataset_id=None):
+def get_dynamic_parser(optional_args):
```
mark-velez, line 220: I'm not sure that/why a parser is needed after the arguments have already been consumed- let's discuss this soon. I think a dynamic parser might be viable in the long-run but for now just turning input into kwargs should be sufficient.

data_steward/cdr_cleaner/clean_cdr.py
```diff
@@ -200,498 +186,105 @@
...
-                "Please run `python cdr_cleaner/clean_cdr.py -h` to list help "
-                "commands.  One or more of project-id, dataset-id, "
-                "mapping-dataset, or mapping-table are missing.  These are "
-                "required to clean fitbit data.")
+    args, kwargs = fetch_args_kwargs()
+
+    rules = DATA_STAGE_RULES_MAPPING[args.data_stage]
+    for rule in rules:
+        clazz = rule[0]
+        kwargs.update({'sandbox_dataset_id': None})
```
mark-velez, line 274: I think this code might smell better if `sandbox_dataset_id` were more like `project_id` and `dataset_id`.

tests/integration_tests/data_steward/cdr_cleaner/cleaning_rules/bigquery_tests_base.py
```diff
@@ -304,16 +307,18 @@ def default_test(self, tables_and_test_values):
...
                 self.assertRowIDsMatch(fq_table_name, fields, values)
 
-            if self.query_class:
-                # test:  get the queries
-                query_list = self.query_class.get_query_specs()
-
+            if self.rule_instance:
                 # test: run the queries
-                engine.clean_dataset(self.project_id, query_list)
+                with mock.patch('cdr_cleaner.clean_cdr_engine.create_sandbox'
```
mark-velez, line 312: This mock wouldn't be needed if sandbox_dataset_id were an argument of clean_dataset

tests/integration_tests/data_steward/cdr_cleaner/cleaning_rules/bigquery_tests_base.py
lrwb-aou: Agree with @mark-velez on this.  All three of these are added to the base class but are only used by the inheriting class.  Do they need to be added here?  Can they be parameterized?

data_steward/cdr_cleaner/clean_cdr_engine.py
nishanthpp93: I've moved it up a level to `clean_cdr.py` as `validate_all_params`

data_steward/cdr_cleaner/clean_cdr.py
nishanthpp93: Yes, the `dynamic_parser` is one way to achieve that. It dynamically creates parser args based on the input.

mark-velez (approved):

PR #613, opened 2020-09-14
[DC-1069] Update Generate Combined Dataset

dcampbell-vumc:
Modified bash script, added labels to each created dataset

### DIFF
--- a/data_steward/tools/generate_combined_dataset.sh
+++ b/data_steward/tools/generate_combined_dataset.sh
@@ -141,7 +141,7 @@ bq rm -r -d "${combined_staging}"
 combined_release="${combined}_release"
 
 # Create a dataset for data browser team
-bq mk --dataset --description "${version} Release version of combined dataset with ${rdr_dataset} + ${unioned_ehr_dataset}" --label "phase:clean" --label "release_tag:${dataset_release_tag}" --label "de_identified:false" ${app_id}:${combined_release}
+bq mk --dataset --description "${version} Release version of combined dataset with ${rdr_dataset} + ${unioned_ehr_dataset}" --label "phase:release" --label "release_tag:${dataset_release_tag}" --label "de_identified:false" ${app_id}:${combined_release}
 
 "${TOOLS_DIR}/table_copy.sh" --source_app_id ${app_id} --target_app_id ${app_id} --source_dataset ${combined} --target_dataset ${combined_release}
 
--- a/data_steward/tools/generate_combined_dataset.sh
+++ b/data_steward/tools/generate_combined_dataset.sh
@@ -95,7 +95,7 @@ export UNIONED_DATASET_ID="${unioned_ehr_dataset}"
 export COMBINED_DATASET_ID="${combined_backup}"
 export BIGQUERY_DATASET_ID="${unioned_ehr_dataset}"
 
-bq mk --dataset --description "${version} combined raw version of  ${rdr_dataset} + ${unioned_ehr_dataset}" --label "type:combined_raw" --label "release_tag:${dataset_release_tag}" --label "de_identified:false" ${app_id}:${combined_backup}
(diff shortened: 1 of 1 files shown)


data_steward/tools/generate_combined_dataset.sh
```diff
@@ -95,7 +95,7 @@ export UNIONED_DATASET_ID="${unioned_ehr_dataset}"
 export COMBINED_DATASET_ID="${combined_backup}"
 export BIGQUERY_DATASET_ID="${unioned_ehr_dataset}"
 
-bq mk --dataset --description "${version} combined raw version of  ${rdr_dataset} + ${unioned_ehr_dataset}" ${app_id}:${combined_backup}
+bq mk --dataset --description "${version} combined raw version of  ${rdr_dataset} + ${unioned_ehr_dataset}" --label "type:combined_raw" --label "release_tag:${dataset_release_tag}" --label "de_identified:false" ${app_id}:${combined_backup}
```
lrwb-aou, line 98: This might be something to discuss in the engineering channel at large.  But I think this type could just be `backup` and the `type` on line 18 could just be `staging`.  That way we can easily get list all backups for a release_tag with a query like:
```
SELECT
 * 
FROM
 INFORMATION_SCHEMA.SCHEMATA_OPTIONS
 where option_name = 'labels' and 
option_value like '%("type", "backup")%'
-- instead of this
-- option_value like '%("type", "combined_raw")%' or option_value like '%("type", "backup")%'
-- I tested with this.
-- option_value like '%("release_tag", "dcampbell_dc1069_test")%'
```

data_steward/tools/generate_combined_dataset.sh
```diff
@@ -127,20 +127,21 @@ data_stage='combined'
 python "${CLEANER_DIR}/clean_cdr.py" --data_stage ${data_stage} -s 2>&1 | tee combined_cleaning_log_"${combined}".txt
 
 # Create a snapshot dataset with the result
-python "${TOOLS_DIR}/snapshot_by_query.py" -p "${app_id}" -d "${combined_staging}" -n "${combined}"
+python "${TOOLS_DIR}/snapshot_by_query.py" --project_id "${app_id}" --dataset_id "${combined_staging}" --snapshot_dataset_id "${combined}"
 
-bq update --description "${version} combined clean version of ${rdr_dataset} + ${unioned_ehr_dataset}" ${app_id}:${combined}
+bq update --description "${version} combined clean version of ${rdr_dataset} + ${unioned_ehr_dataset}" --set_label "type:combined_clean" --set_label "release_tag:${dataset_release_tag}" --set_label "de_identified:false" ${app_id}:${combined}
```
lrwb-aou, line 132: I would suggest making this type `clean`.  Again, I think this should be discussed with the engineers before making a final decision.

data_steward/tools/generate_combined_dataset.sh
```diff
@@ -127,20 +127,21 @@ data_stage='combined'
...
 
+# Remove intermediary datasets
 bq rm -r -d "${combined_staging}_sandbox"
 bq rm -r -d "${combined_staging}"
 
 combined_release="${combined}_release"
 
 # Create a dataset for data browser team
-bq mk --dataset --description "${version} Release version of combined dataset with ${rdr_dataset} + ${unioned_ehr_dataset}" ${app_id}:${combined_release}
+bq mk --dataset --description "${version} Release version of combined dataset with ${rdr_dataset} + ${unioned_ehr_dataset}" --label "type:combined_clean_release" --label "release_tag:${dataset_release_tag}" --label "de_identified:false" ${app_id}:${combined_release}
```
lrwb-aou, line 144: I think this `type` would just be a `release`.

lrwb-aou (commented):
I would like to take the options for `type` back to the engineers.  Just so we can get some feedback and a little consistency between scripts.  I'll happily update my type labels as needed.  What are your thoughts?

lrwb-aou (approved):
This looks good!  Thank you!

PR #617, opened 2020-09-16
[DC-1087] fix broken integration test

lrwb-aou:
* integration test needed to be updated after the clean engine update changes


ksdkalluri (approved):
Looks good to me!

PR #620, opened 2020-09-17
[DC-1083] Add remove deactivated participants rule changes to run_curation_pipeline.sh

ksdkalluri:


data_steward/tools/run_curation_pipeline.sh
```diff
@@ -14,6 +14,10 @@ Usage: run_curation_pipeline.sh
   --dataset_release_tag <release tag for the CDR>
   --ehr_cutoff <ehr_cut_off date format yyyy-mm-dd>
   --rdr_export_date <date RDR export is run format yyyy-mm-dd>
+  --ticket_number <Ticket number to append to sandbox table names>
```
lrwb-aou, line 17: you may want to mention this is to remove deactivated participants.

lrwb-aou (approved):
I added an optional comment.  It might help to specify what the ticket_number parameter is for.  It is technically sound.

data_steward/tools/run_curation_pipeline.sh
ksdkalluri: I'll make this change, and push it in a new PR.

PR #615, opened 2020-09-16
[DC-1083] / [DC-1086] Move remove_ehr_data_post_deactivation cleaning rule from clean_cdr to generate_unioned_ehr_dataset.sh script

ksdkalluri:
Add apply cleaning rule step to generate_unioned_ehr_dataset.sh

PR #622, opened 2020-09-17
[DC-1088] Bash script sandbox update

nishanthpp93:
* Updated bash scripts to create sandbox with the appropriate labels

### DIFF
--- a/data_steward/tools/deid_runner.sh
+++ b/data_steward/tools/deid_runner.sh
@@ -150,7 +150,7 @@ python "${TOOLS_DIR}/add_cdr_metadata.py" --component "insert" --project_id ${AP
 bq update --description "Sandbox created for storing records affected by the cleaning rules applied to ${cdr_deid_base}" --set_label "phase:sandbox" --set_label "release_tag:${dataset_release_tag}" --set_label "de_identified:true" "${APP_ID}":"${cdr_deid_base_sandbox}"
 
 # remove intermediary datasets
-bq rm -r -d "${cdr_deid_base_staging}_sandbox"
+bq rm -r -d "${cdr_deid_base_staging_sandbox}"
 bq rm -r -d "${cdr_deid_base_staging}"
 
 # create empty de-id_clean dataset to apply cleaning rules
@@ -181,7 +181,7 @@ bq update --description "${version} De-identified Clean version of ${cdr_deid_ba
 bq update --description "Sandbox created for storing records affected by the cleaning rules applied to ${cdr_deid_clean}" --set_label "phase:sandbox" --set_label "release_tag:${dataset_release_tag}" --set_label "de_identified:true" "${APP_ID}":"${cdr_deid_clean_sandbox}"
 
 # remove intermediary datasets
-bq rm -r -d "${cdr_deid_clean_staging}_sandbox"
+bq rm -r -d "${cdr_deid_clean_staging_sandbox}"
 bq rm -r -d "${cdr_deid_clean_staging}"
 
 unset PYTHONPATH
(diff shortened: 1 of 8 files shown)


ChaoPang (commented):
A general comment, it seems that we create `dataset_staging_sandbox` as well as `dataset_sandbox`, we first create all the sandboxed tables in `dataset_staging_sandbox` when running `clean_cdr.py`, then copy the content from `dataset_staging_sandbox` to `dataset_sandbox`. I think that's because previously `sandbox_dataset` used to be coupled with the corresponding `dataset` because it's automatically inferred by `check_and_create_sandbox_dataset`. Now that we have decoupled the two datasets, I don't think we need to create `dataset_staging_sandbox` anymore. When calling `clean_cdr.py`, we could just use `dataset_sandbox` directly.

nishanthpp93:
Thanks for looking at it @ChaoPang. This would be a significant change from our current process. I am just following precedence that was already agreed upon in https://github.com/all-of-us/curation/pull/615/commits/5031eb1e75b0c021ed5b953ad7bcc745ee1511fb that was reviewed and merged recently after discussion on the eng sync channel. If we need to change this, we need to start a new discussion.

data_steward/tools/deid_runner.sh
```diff
@@ -133,7 +144,10 @@ bq update --description "${version} De-identified Base version of ${cdr_id}" --s
...
 
 #copy sandbox dataset
-"${TOOLS_DIR}/table_copy.sh" --source_app_id ${APP_ID} --target_app_id ${APP_ID} --source_dataset "${cdr_deid_base_staging}_sandbox" --target_dataset "${cdr_deid_base}_sandbox"
+"${TOOLS_DIR}/table_copy.sh" --source_app_id ${APP_ID} --target_app_id ${APP_ID} --source_dataset "${cdr_deid_base_staging_sandbox}" --target_dataset "${cdr_deid_base_sandbox}"
+
+# Update sandbox description
+bq update --description "Sandbox created for storing records affected by the cleaning rules applied to ${cdr_deid_base}" --set_label "phase:sandbox" --set_label "release_tag:${dataset_release_tag}" --set_label "de_identified:true" "${APP_ID}":"${cdr_deid_base_sandbox}"
 
 # remove intermediary datasets
 bq rm -r -d "${cdr_deid_base_staging}_sandbox"
```
ChaoPang, line 153: Use the environment `${cdr_deid_base_staging_sandbox}`

data_steward/tools/deid_runner.sh
```diff
@@ -142,23 +156,29 @@ bq rm -r -d "${cdr_deid_base_staging}"
...
 
 #copy sandbox dataset
-"${TOOLS_DIR}/table_copy.sh" --source_app_id ${APP_ID} --target_app_id ${APP_ID} --source_dataset "${cdr_deid_clean_staging}_sandbox" --target_dataset "${cdr_deid_clean}_sandbox"
+"${TOOLS_DIR}/table_copy.sh" --source_app_id ${APP_ID} --target_app_id ${APP_ID} --source_dataset "${cdr_deid_clean_staging_sandbox}" --target_dataset "${cdr_deid_clean_sandbox}"
+
+# Update sandbox description
+bq update --description "Sandbox created for storing records affected by the cleaning rules applied to ${cdr_deid_clean}" --set_label "phase:sandbox" --set_label "release_tag:${dataset_release_tag}" --set_label "de_identified:true" "${APP_ID}":"${cdr_deid_clean_sandbox}"
 
 # remove intermediary datasets
 bq rm -r -d "${cdr_deid_clean_staging}_sandbox"
```
ChaoPang, line 184: Use the environment variable `${cdr_deid_clean_staging_sandbox}`

ChaoPang (changes requested):
Everything looks good except two minor comments.

ChaoPang (approved):
Looks good now

PR #623, opened 2020-09-18
[DC-1098] Add ELSE clauses to null_concept_ids_for_numeric_ppi

lrwb-aou:
adds else clauses
updates integration test


dcampbell-vumc (approved):
The changes look good and make sense to me. Ran the unit and integration tests and both passed fine.

PR #627, opened 2020-09-22
[DC-1106] Add missing backtick in delete query for Operational PII field CR

nishanthpp93:
* Add backtick in DELETE query


ksdkalluri (approved):
Please update the commit message to a different message before merging. If not it will add 2 commits with the same message.

nishanthpp93:
Yeah, I'll create a new ticket for this

PR #628, opened 2020-09-22
[DC-1028] [DC-1155] Add script to remove select sites' EHR data

mark-velez:


tests/unit_tests/data_steward/tools/purge_hpo_data_test.py
```diff
@@ -0,0 +1,98 @@
...
+        results = _filter_hpo_tables(self.hpo_cdm_tables, 'fake_hpo3')
+        self.assertListEqual([], results)
+
+    def _mock_client_with(self, tables):
+        """
+        Get a mock BQ client object
+        :param tables: names of tables the client should list
+        :return: the mock client object
+        """
+        mock_client = mock.MagicMock(wraps=Client(project=self.project_id))
```
lrwb-aou, line 78: There is something wrong with this mock client instantiation.  It took me a little bit to reproduce the error, but you can reproduce it locally if you run:
```
unset GOOGLE_APPLICATION_CREDENTIALS
cd curation
./tests/run_tests.sh -s unit -r purge_hpo*
```

I tried creating a mock credentials object, but that didn't work well with the wrapped client object.  The only thing I was able to do locally after unsetting the `GOOGLE_APPLICATION_CREDENTIALS` variable was
```suggestion
        mock_client = mock.MagicMock()
```
I tried creating a credentials object with bad data, but it ran into errors in the pyasn1 module and I didn't want to look deeper than that.

data_steward/tools/purge_hpo_data.py
```diff
@@ -0,0 +1,100 @@
+"""
+Empty all CDM tables associated with one or more HPO sites from an EHR dataset
+
+This can be used to exclude sites' data from the EHR dataset before 
+generating the unioned dataset which will be used for the CDR.
```
lrwb-aou, line 5: I think I'm right.  You're using this doc string as the command line description for the module, correct?  I like the idea.  It might be a good idea to explicitly state that this module does not delete the empty tables.  I know the first line states you are emptying the CDM tables, but since I was just skimming the comment I didn't pay **close** attention.  That's why it **might** help to explicitly state the tables are not deleted.  Because that was my first question.  Why not just delete the tables?  It's probably something in the generation process that I'm just not familiar with.

data_steward/tools/purge_hpo_data.py
```diff
@@ -0,0 +1,100 @@
+"""
```
lrwb-aou, line 1: you mentioned not loving the module name.  I'll list some options. 
- purge_site_data
- exclude_site_submission
- exclude_site_submission_data

 I _think_ I like `exclude_site_submission` because it seems like we are only interested in dropping the info a site submitted (instead of non-CDM tables).   This is purely optional.  I just wanted to offer some ideas.

lrwb-aou (changes requested):
Would appreciate your feedback on some of these comments.  We will need to fix the unit test before merging.  The other feedback could be addressed now or in a new PR, whichever makes the most sense.

data_steward/tools/purge_hpo_data.py
```diff
@@ -0,0 +1,100 @@
...
+    :param hpo_ids: Identifies the HPO sites whose data should be purged
+    :return: Query job associated with removing all the records
+    :raises RuntimeError if CDM tables associated with a site are not found in the dataset
+    """
+    LOGGER.debug(
+        f'purge_hpo_data called with dataset={dataset.dataset_id} and hpo_ids={hpo_ids}'
+    )
+    all_tables = list(
+        client.list_tables(dataset=dataset,
+                           max_results=LIST_TABLES_MAX_RESULTS))
```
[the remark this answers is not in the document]
mark-velez: I can no longer find the doc page that said the default `max_results`- but can you believe it's a measly 50! Anyway I will follow your first suggestion and leave a comment in there for now.

I'd also like to avoid a hard coded value, but I'm conflicted about how to go about it. The caller's intent is always to purge HPO data from "all tables where they are found." A user-supplied `max_results` parameter would be far removed from the intent and it can be difficult to predict what would happen when the value provided is less than the number of tables in the dataset.

Here are some other alternatives I'd like to discuss with you (**after** this CDR gen).
1. Determine `max_results` based on metadata query
```python
  # This query should be free of cost as compared to INFORMATION_SCHEMA.TABLES
  client.query(f"SELECT COUNT(1) max_results FROM `{project_id}.{dataset_id}.__TABLES__`")
```
2. Forgo `max_results` altogether and instead generate and execute dynamic query in a script.
A. Get the tables based on `hpo_id` prefix.
    ```sql
    DECLARE qs DEFAULT(
    ARRAY(
        -- assuming my_dataset and my_site are filled e.g. using jinja
        SELECT FORMAT('DELETE FROM %s.%s.%s', table_catalog, table_schema, table_name) 
        FROM `my_dataset.INFORMATION_SCHEMA.COLUMNS`
        WHERE column_name='person_id' AND table_name LIKE 'my_site_%'
    )
    );
    DECLARE n DEFAULT(ARRAY_LENGTH(qs));
    DECLARE i DEFAULT(0);
    DECLARE q STRING;
    WHILE i < n DO
    SET q = qs[ARRAY_OFFSET(i)];
    EXECUTE IMMEDIATE q;
    SET i = i + 1;
    END WHILE;
    ```
    B. Same idea as A but get the tables based on labels we apply e.g. `{"hpo_id": "some_hpo"}` and avoid the weirdness when one `hpo_id` is a substring of another (e.g. `site_a` &#8800; `site_a_1` &#8800; `site_a_1_b`).

tests/unit_tests/data_steward/tools/purge_hpo_data_test.py
mark-velez: As discussed, I changed it to a plain ol' `MagicMock` minus the `wraps` bit

data_steward/tools/purge_hpo_data.py
```diff
@@ -0,0 +1,100 @@
...
+                f'No tables found for {hpo_id} in dataset {dataset.dataset_id}. '
+                f'Ensure the specified arguments are correct.')
+        tables_to_purge.extend(hpo_tables)
+    script = DELETE_QUERY_TPL.render(tables_to_purge=tables_to_purge)
+    LOGGER.debug(f'purge_hpo_data about to start script:\n {script}')
+    return client.query(script)
+
+
+def main(credentials, project_id, dataset_id, hpo_ids: List[str]):
+    client = bigquery.Client.from_service_account_json(credentials)
```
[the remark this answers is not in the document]
mark-velez: I agree- this script is simple and it'd be a good place to start using impersonation. I'm not sure of a few parts and I would prefer to discuss these with you beforehand. Could we look at impersonation separate from this PR?

For example, I'm not sure whether/how to get stored user credentials to use as source credentials. My impression is that we'd want the end-user (e.g. mark.velez) credentials to be used to acquire impersonated credentials for the service account (i.e. cdr-ops); theoretically this would enable us to more easily track down the individual associated with API activity. Currently we'd have to peruse BigQuery log events associated with the activity in question to find the SA key and then dig into the service account key logs to find out who created that key.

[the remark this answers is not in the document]
lrwb-aou: @mark-velez I used impersonation in the copy_prototype.py script I created and gave to Nishanth to run.  He stripped the service account impersonation part out of that script, but you can still see the original script on DC-1030.  My example used my personal account service account key, because the script requires that level of credentials.  That key file had to be passed to access the impersonated service account key file.  So, I would still need service account credentials, but they would be my credentials, not the credentials of the service account running the code.  This is an example of my log file via [Google Logging](https://console.cloud.google.com/logs/viewer?project=aou-res-curation-test&minLogLevel=0&expandAll=false&timestamp=2020-10-05T17%3A20%3A44.366000000Z&customFacets&limitCustomFacetWidth=true&dateRangeStart=2020-09-21T13%3A20%3A44.000Z&dateRangeEnd=2020-10-05T17%3A20%3A44.618Z&interval=CUSTOM&resource=bigquery_resource&scrollTimestamp=2020-09-29T22%3A07%3A39.310000000Z&advancedFilter=resource.type%3D%22bigquery_resource%22%0Aresource.labels.project_id%3D%22aou-res-curation-test%22%0Atimestamp%3D%222020-09-29T22%3A07%3A36.790000000Z%22%0AinsertId%3D%22i1dg89d15rl%22).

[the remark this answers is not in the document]
lrwb-aou: I think we can continue exploring service account impersonation separately from this PR.

data_steward/tools/purge_hpo_data.py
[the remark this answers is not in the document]
lrwb-aou: I think we both agree option 1 is better than a hard coded value and possibly more maintainable than option 2.

data_steward/tools/purge_hpo_data.py
```diff
@@ -0,0 +1,100 @@
...
+                        help='Identifies the project')
+    parser.add_argument('-d',
+                        '--dataset_id',
+                        required=True,
+                        help='Identifies the dataset to purge data from')
+    parser.add_argument(
+        'hpo_ids',
+        nargs='+',
+        help='Identifies the HPO sites whose data should be purged')
+    ARGS = parser.parse_args()
```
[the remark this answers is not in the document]
lrwb-aou: If you use the pipeline_logging.py module, there will definitely be a few changes required in that file.

lrwb-aou (changes requested):
There is at least one change needed to calculating `max_results`.

data_steward/tools/purge_hpo_data.py
mark-velez: Just renamed it to exclude_site_submission

data_steward/tools/purge_hpo_data.py
mark-velez: Thanks- the docstring is now more explicit.

data_steward/tools/purge_hpo_data.py
```diff
@@ -0,0 +1,100 @@
...
+import bq_utils
+import resources
+
+LOGGER = logging.getLogger(__name__)
+LIST_TABLES_MAX_RESULTS = 5000
+DELETE_QUERY_TPL = bq.JINJA_ENV.from_string("""
+{%- for table in tables_to_purge -%}
+DELETE FROM `{{table.project}}.{{table.dataset_id}}.{{table.table_id}}` WHERE 1=1;
+{%- endfor -%}
+""")
```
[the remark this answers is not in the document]
mark-velez: Thanks! Of course I look at this a little more critically [now that BigQuery supports TRUNCATE](https://cloud.google.com/bigquery/docs/reference/standard-sql/dml-syntax#truncate_table_statement), but I haven't tested it yet and I think we should probably just get this PR out of the way and do that separately. :)

data_steward/tools/purge_hpo_data.py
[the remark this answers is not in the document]
mark-velez: @lrwb-aou in updating this to use `pipeline_logging` I first tried to use its `setup_logger`, but it did not behave as I'd expected and I could not locate specific requirements. (@dcampbell-vumc I woulda pinged ya but you were doing heroic feats on two wheels at the time. :biking_woman: Please feel free to chime in on this.)

Anyway I [added a function](https://github.com/all-of-us/curation/pull/628/files#diff-5aaa0476cdf524ff7dccac764598bf2667022c2f9a0ef2d3a280b0115c5d24d5R137) `get_logger` wherein I tried to consolidate the logic contained in the other functions. Please let me know if it is sufficient for our logging standardization needs and we could cut out extra code.

data_steward/utils/pipeline_logging.py
```diff
@@ -111,3 +121,58 @@ def setup_logger(log_filepath_list, console_logging=True):
...
+    >>> LOGGER.addHandler(logging.FileHandler('my_custom.log'))
+    >>> def func(p1, p2):
+    >>>     LOGGER.debug(f"func called with {p1}, {p2}")
+    """
+    logger = logging.getLogger(logger_name)
+    # explicitly set logger level to DEBUG otherwise
+    # the level would be that of the closest ancestor
+    logger.setLevel(logging.DEBUG)
+    # prevent inheriting handlers from ancestors
+    logger.propagate = False
```
lrwb-aou, line 157: This logging setup makes for some difficult log reading.  With this line and lines 166 and 176 as written, the console looks like you would expect it to, all log messages being displayed as they are received.  The log files are hard to read.  I set up a test with the following in a file
```
import logging
import tools.exclude_site_submission as ess
from utils import pipeline_logging

# assuming a pipeline setup like yours was used in this fake module too
logger = pipeline_logging.get_logger(__name__)

logger.info("Random logging message")

if __name__ == '__main__':
    logger.info("STARTINGGGGGGGGG")
    # calling your code like referencing it from another module
    ess.main('/Users/lbiggers/.keys/aou-res-curation-test-me.json', 'aou-res-curation-test', 'lrwb_aou_ehr_ops_test', ['ca', 'fl'])
    logger.info("DONEEEEEE")
```
When I run this with the code, I get two separate log files, one titled `logs/__main__.py` and one titled `logs/tools.exclude_site_submission.log`.  Both files have log lines, but even though `logs/__main__.py` called the other module, neither module has logs from the other file.  So it doesn't look like the console with a "stream of conciousness" type logging.  To get that I would have to mix the two log files.

To get a single log file with the interwoven log statements, I had to remove this restriction and change the lines below from `logger.addHandler` to `logging.getLogger('').addHandler`.  Similar to this portion of the [logging cookbook](https://docs.python.org/3/howto/logging-cookbook.html#logging-to-multiple-destinations).

What issues were you addressing with this setup?

tests/bq_test_helpers.py
```diff
@@ -0,0 +1,96 @@
...
+from google.cloud import bigquery
+from google.cloud.bigquery import TableReference
+from google.cloud.bigquery.table import TableListItem
+
+_TYPE_TO_FIELD_TYPE = {
+    str(int): 'INT64',
+    str(str): 'STRING',
+    str(float): 'FLOAT64',
+    str(datetime.date): 'DATE',
+    str(datetime.datetime): 'DATETIME'
```
lrwb-aou, line 19: Should add TIMESTAMP to this list.

tests/bq_test_helpers.py
```diff
@@ -0,0 +1,96 @@
...
+    :return: a mock RowIterator
+    """
+    mock_row_iter = mock.MagicMock(spec=bigquery.table.RowIterator)
+    mock_row_iter.total_rows = len(rows)
+    row0 = rows[0]
+    if isinstance(row0, OrderedDict):
+        _rows = rows
+    else:
+        warnings.warn('Provide rows as `OrderedDict` if type conversion '
+                      'results in unexpected behavior.')
```
lrwb-aou, line 72: So any time I use this without providing an ordered dictionary, it's going to fuss at me?

tests/unit_tests/data_steward/utils/pipeline_logging_test.py
```diff
@@ -69,6 +96,33 @@ def test_create_logger(self, mock_get_logger, mock_file_handler,
...
+                logger.debug('debug message')
+                # debug messages are logged to file
+                self.assertEqual(file_emit.messages, ['debug message'])
+                # debug messages are NOT logged to stderr
+                self.assertEqual(0, stderr_emit.call_count)
+
+                logger.info('info message')
+                self.assertEqual(file_emit.messages,
+                                 ['debug message', 'info message'])
+                self.assertListEqual(stderr_emit.messages, ['info message'])
```
lrwb-aou, line 123: What does this give us that using the [standard context manager](https://docs.python.org/3.7/library/unittest.html#unittest.TestCase.assertLogs) does not give us?

tests/unit_tests/data_steward/utils/pipeline_logging_test.py
```diff
@@ -14,11 +14,38 @@
 import unittest
 import logging
 import mock
-import os
 from datetime import datetime
 
 # Project imports
-from utils.pipeline_logging import generate_paths, create_logger, setup_logger
+from utils.pipeline_logging import generate_paths, create_logger, setup_logger, get_logger
```
lrwb-aou, line 20: at this point, i'd just import this as an alias and update the places these are called.

```suggestion
import utils.pipeline_logging as pl
```

data_steward/bq_utils.py
```diff
@@ -536,6 +537,12 @@ def list_tables(dataset_id=None,
       for table in result['tables']:
           print table['id']
     """
+    warnings.warn(
```
lrwb-aou, line 540: Maybe add a TODO here to remind us to remove this stuff in something like 6 months?

lrwb-aou (commented):
So, the exclude script is good.  I think we covered that in the previous review.  I'd appreciate talking about the logging changes and your design considerations for them.

mark-velez:
Thanks @lrwb-aou. Let's discuss logging when I get back on Wed. There's some good feedback in here which I may be able to address by then.

data_steward/tools/exclude_site_submission.py
```diff
@@ -0,0 +1,100 @@
...
+
+from utils import bq
+import bq_utils
+import resources
+
+LOGGER = pipeline_logging.get_logger(__name__)
+
+DELETE_QUERY_TPL = bq.JINJA_ENV.from_string("""
+{%- for table in tables_to_empty -%}
+DELETE FROM `{{table.project}}.{{table.dataset_id}}.{{table.table_id}}` WHERE 1=1;
```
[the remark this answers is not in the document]
mark-velez: Haha I'm _so close_ to being able to hit the backtick key without looking at my keyboard now!

data_steward/tools/exclude_site_submission.py
```diff
@@ -0,0 +1,100 @@
...
+                f'Ensure the specified arguments are correct.')
+        tables_to_empty.extend(hpo_tables)
+    script = DELETE_QUERY_TPL.render(tables_to_empty=tables_to_empty)
+    LOGGER.debug(f'exclude_site_submission about to start script:\n {script}')
+    return client.query(script)
+
+
+def main(credentials, project_id, dataset_id, hpo_ids: List[str]):
+    client = bigquery.Client.from_service_account_json(credentials)
+    dataset = bigquery.DatasetReference(project_id, dataset_id)
```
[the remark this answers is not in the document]
mark-velez: Yes, we discussed this previously- it's my practice to resolve conversations so I know what remains for me to do, but they do  get harder to find as a result. Ah [here it is](https://github.com/all-of-us/curation/pull/628#discussion_r493678071).

data_steward/tools/exclude_site_submission.py
```diff
@@ -0,0 +1,100 @@
...
+                        required=True,
+                        help='Identifies the project')
+    parser.add_argument('-d',
+                        '--dataset_id',
+                        required=True,
+                        help='Identifies the dataset to exclude site data from')
+    parser.add_argument(
+        'hpo_ids',
+        nargs='+',
+        help='Identifies the HPO sites whose submissions should be excluded')
```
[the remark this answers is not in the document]
mark-velez: I already got questions about this when we ran it. :) How about we make it a required option like this?
```suggestion
        '-i',
        '--hpo_ids',
        required=True,
        nargs='+',
        help='Identifies the HPO sites whose submissions should be excluded')
```

data_steward/utils/bq.py
```diff
@@ -19,6 +19,10 @@
 from constants.utils import bq as consts
 from resources import fields_for
 
+_MAX_RESULTS_PADDING = 100
+"""
+Constant added to table count in order to list all table results 
+"""
```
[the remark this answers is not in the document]
mark-velez: My tools don't seem to show hashed comments. I discovered this "variable docstrings" practice in google's code and found that PyCharm and VS Code both show these types of comments in various contexts. Tools like epydoc and sphinx would also render these docstring-like comments but seem to ignore the hashed comments.

tests/bq_test_helpers.py
mark-velez: Hmm yeah I think I meant to use `TIMESTAMP` rather than `DATETIME` here. (A quick scan on a recent combined dataset shows 0 `DATETIME` fields vs 32 `TIMESTAMP` fields in our tables.) Let me know if you had something else in mind.

tests/bq_test_helpers.py
mark-velez: Yeah, I suspect there may be edge cases where passing a list of plain dicts yields inconsistent row structure. If this is true it would likely be difficult to troubleshoot the issue. I thought adding this to the function docstring might be TL;DR, but we can do that if you think it's a better option.

tests/unit_tests/data_steward/utils/pipeline_logging_test.py
mark-velez: The context manager has a lot of the same, but I couldn't get it to tell me which handlers log records get routed to. I wanted to ensure that debugs go to the file handler as it seemed this could be perturbed by e.g. prior logging initialization. Please let me know if you have recommendations- I googled the heck out of this and would definitely prefer tried-and-true approaches over this one.

data_steward/bq_utils.py
mark-velez: I think I can do better- I created a ticket DC-1157 and listed the funcs to remove. Too many TODOs don't get done. :) I found other companies following this warning practice prior to removing library code.

mark-velez:
@lrwb-aou I wanted to get as much done on this as I can since I will be out for a few days. I changed the PR title because the build check was failing. I pushed commits I thought were not too controversial and closed the associated conversations; please unresolve if needed. I plan to cover the remaining items when I get back.

dcampbell-vumc:
@mark-velez @lrwb-aou I merged the changes to fix the intermittent unit test failure for `pipeline_logging_test.py` so y'all may want to do a pull if you haven't already. 👍🏾

data_steward/tools/exclude_site_submission.py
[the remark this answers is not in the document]
mark-velez: @mark-velez remove the short option `-i`

tests/bq_test_helpers.py
```diff
@@ -0,0 +1,96 @@
...
+        >>> d = {'item': 'book', 'qty': 2, 'price': 1.99}
+        >>> bq_test_helpers.fields_from_dict(d)
+            [SchemaField('item', 'STRING', 'NULLABLE', None, ()),
+             SchemaField('qty', 'INT64', 'NULLABLE', None, ()),
+             SchemaField('price', 'FLOAT64', 'NULLABLE', None, ())]
+    """
+    return [_field_from_key_value(key, value) for key, value in row.items()]
+
+
+def mock_query_result(rows: List[dict_types]):
```
mark-velez, line 57: @mark-velez require second arg List[str] if first arg is List[dict]

tests/unit_tests/data_steward/tools/exclude_site_submission_test.py
```diff
@@ -0,0 +1,93 @@
...
+        full_table_id = self._full_table_id(table_id)
+        return list_item_from_table_id(full_table_id)
+
+    def _table_list_item(self, hpo_id, cdm_table) -> TableListItem:
+        table_id = bq_utils.get_table_id(hpo_id, cdm_table)
+        return self._table_id_to_list_item(table_id)
+
+    def test_filter_hpo_tables(self):
+        # clinical tables for specified hpo are returned
+        results = _filter_hpo_tables(self.hpo_cdm_tables, 'fake_hpo1')
```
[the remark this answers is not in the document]
mark-velez: Will remove since next test essentially tests this part out anyway

data_steward/utils/pipeline_logging.py
```diff
@@ -111,3 +121,58 @@ def setup_logger(log_filepath_list, console_logging=True):
...
+
+    :param logger_name: name of the logger (usually set to __name__)
+    :return: the logger
+    :example:
+    >>> LOGGER = get_logger(__name__)
+    >>> # Additional handlers can be added if needed
+    >>> LOGGER.addHandler(logging.FileHandler('my_custom.log'))
+    >>> def func(p1, p2):
+    >>>     LOGGER.debug(f"func called with {p1}, {p2}")
+    """
```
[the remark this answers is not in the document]
mark-velez: I will refactor and then close this

data_steward/utils/pipeline_logging.py
```diff
@@ -111,3 +121,58 @@ def setup_logger(log_filepath_list, console_logging=True):
...
+    logger = logging.getLogger(logger_name)
+    # explicitly set logger level to DEBUG otherwise
+    # the level would be that of the closest ancestor
+    logger.setLevel(logging.DEBUG)
```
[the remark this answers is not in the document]
mark-velez: I will do this with parameter name file_log_level

data_steward/utils/pipeline_logging.py
```diff
@@ -111,3 +159,80 @@ def setup_logger(log_filepath_list, console_logging=True):
...
+    """
+    Get a dictionary which describes the logging configuration
+    
+    :param filename: Create the FileHandler using the specified filename.
+    :param level: Set the root logger level to the specified level.
+    :return: the default configuration dict
+    """
+    # copy _DEFAULT_CONFIG to avoid modifying it
+    default_config = dict(_DEFAULT_CONFIG)
+    default_config['root']['level'] = level
```
lrwb-aou, line 191: I assume level is one of the `logging.[INFO, DEBUG, WARNING, ERROR]` levels.  You should make this clear.  I don't expect the others to pick up on this on their own.  It wouldn't hurt to mention the possible enumerated values in the function doc string either.  Also, what about the case where someone didn't provide a level?  Is there a default value or default behavior?

tests/bq_test_helpers.py
mark-velez: We must determine if GCE is 3.7; if so then dict would work just fine and we can avoid the warning altogether.

mark-velez: It's still up in the air so I will implement as we discussed above

data_steward/utils/pipeline_logging.py
```diff
@@ -111,3 +159,80 @@ def setup_logger(log_filepath_list, console_logging=True):
...
+    Log exception info to root logger. Used as a hook for uncaught exceptions prior 
+    to system exit.
+    
+    :param exc_type: type of the exception
+    :param exc_value: the exception
+    :param exc_traceback: the traceback associated with the exception
+    """
+    root_logger = logging.getLogger()
+    root_logger.critical('Uncaught exception',
+                         exc_info=(exc_type, exc_value, exc_traceback))
```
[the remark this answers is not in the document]
mark-velez: @lrwb-aou Take a look and let me know if this is ok as a last resort logging of uncaught exceptions

data_steward/utils/pipeline_logging.py
mark-velez: I'll add more detailed doc and also make everything except configure private

mark-velez:
Hey @lrwb-aou please let me know if there's anything left to address before I rebase and remove obsolete code.

data_steward/utils/pipeline_logging.py
```diff
@@ -9,9 +9,57 @@
...
+            'formatter': 'default',
+        },
+        _FILE_HANDLER: {
+            'class': 'logging.FileHandler',
+            'mode': 'a',
+            'formatter': 'default'
+        }
+    },
+    'root': {
+        'handlers': [_CONSOLE_HANDLER, _FILE_HANDLER]
```
lrwb-aou, line 54: @mark-velez Actually, I do have another comment.  This is adding a Stream Handler, no matter what.  Stream handlers are fine / great / wanted when a user invokes the code from the command line.  They are not wanted when the script is called by a cron job or some other automated action.  Adding the stream handler should only be done if it is specifically invoked from the command line.  That's why a flag is passed from the command line to the logging setup in the clean cdr engine module.

data_steward/utils/pipeline_logging.py
```diff
@@ -111,3 +159,86 @@ def setup_logger(log_filepath_list, console_logging=True):
...
+    
+    :param filename: Create the FileHandler using the specified filename.
+    :param level: Set the root logger level to the specified level 
+                  (i.e. logging.{DEBUG,INFO,WARNING,ERROR}).
+    :return: the default configuration dict
+    """
+    # copy _DEFAULT_CONFIG to avoid modifying it
+    default_config = dict(_DEFAULT_CONFIG)
+    default_config['root']['level'] = level
+    file_log_handler_dict = default_config['handlers'][_FILE_HANDLER]
```
lrwb-aou, line 197: To be safe and to further establish a convention, I would change this line to:

```suggestion
    file_log_handler_dict = default_config.get('handlers', {}).get(_FILE_HANDLER, {})
```

data_steward/utils/pipeline_logging.py
```diff
@@ -111,3 +159,86 @@ def setup_logger(log_filepath_list, console_logging=True):
...
+    :param exc_type: type of the exception
+    :param exc_value: the exception
+    :param exc_traceback: the traceback associated with the exception
+    """
+    root_logger = logging.getLogger()
+    root_logger.critical('Uncaught exception',
+                         exc_info=(exc_type, exc_value, exc_traceback))
+
+
+def configure(level=logging.INFO):
```
lrwb-aou: This assumes the user always wants a stream handler.  We shouldn't make that assumption.  Processes that are automatically started won't want to add a stream handler, just a file handler.  We should allow passing a flag into to the configure function to determine whether or not to add a stream handler.

```suggestion
def configure(level=logging.INFO, console_log=False):
```

data_steward/utils/pipeline_logging.py
```diff
@@ -111,3 +159,86 @@ def setup_logger(log_filepath_list, console_logging=True):
...
+    >>> def func(p1):
+    >>>     LOGGER.debug(f"func called with p1={p1}")
+    >>>     LOGGER.info("func called")
+    >>>
+    >>> if __name__ == '__main__':
+    >>>     pipeline_logging.configure()
+    >>>     func(1, 2)
+    """
+    filename = _get_log_file_path()
+    default_config = _get_config(filename, level)
```
lrwb-aou: These lines could be altered.  The `_get_config(level, console_logging)` function should contain the call to `_get_log_file_path()` because we don't let users alter this value.  There's no reason to call it here and pass it as a parameter to another internal function.  We could just call it where needed in the internal function.

```suggestion
    default_config = _get_config(level, console_log)
```

data_steward/utils/pipeline_logging.py
```diff
@@ -111,3 +159,86 @@ def setup_logger(log_filepath_list, console_logging=True):
...
+    The location is DEFAULT_LOG_DIR and the file name is the current
+    date formatted using FILENAME_FMT.
+
+    :return: absolute path to the log file
+    """
+    date_str = _get_date_str()
+    return os.path.join(DEFAULT_LOG_DIR, f'{date_str}.log')
+
+
+def _get_config(filename, level):
```
lrwb-aou: This function signature may change to:
```suggestion
def _get_config(level, console_log):
```

data_steward/utils/pipeline_logging.py
```diff
@@ -111,3 +159,86 @@ def setup_logger(log_filepath_list, console_logging=True):
...
+    :param filename: Create the FileHandler using the specified filename.
+    :param level: Set the root logger level to the specified level 
+                  (i.e. logging.{DEBUG,INFO,WARNING,ERROR}).
+    :return: the default configuration dict
+    """
+    # copy _DEFAULT_CONFIG to avoid modifying it
+    default_config = dict(_DEFAULT_CONFIG)
+    default_config['root']['level'] = level
+    file_log_handler_dict = default_config['handlers'][_FILE_HANDLER]
+    file_log_handler_dict['filename'] = filename
```
lrwb-aou, line 198: To be safe and to further establish a convention, I would change these lines to:

```suggestion
    file_log_handler_dict = default_config.get('handlers', {}).get(_FILE_HANDLER, {})
    file_log_handler_dict['filename'] = _get_log_file_path()
```

data_steward/utils/pipeline_logging.py
```diff
@@ -111,3 +159,86 @@ def setup_logger(log_filepath_list, console_logging=True):
...
+    return default_config
```
lrwb-aou, line 203: You could add the console handler with these commands:
```suggestion
    if console_log:
        root_handlers = default_config.get('root').get('handlers')
        root_handlers.append(_CONSOLE_HANDLER)
        default_config['root']['handlers'] = root_handlers 

    return default_config
```

data_steward/utils/pipeline_logging.py
mark-velez: Thanks @lrwb-aou , I added `add_console_handler` to `configure()` which addresses this.

data_steward/utils/pipeline_logging.py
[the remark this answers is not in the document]
lrwb-aou: @mark-velez This works as described.

lrwb-aou (commented):
I've got some thoughts.  Let me know what you think.

data_steward/utils/pipeline_logging.py
mark-velez: Should we get rid of `_DEFAULT_CONFIG` and instead construct the dict config using procedural logic then? If in the suggestion above `file_log_handler_dict` is empty it will need more attributes added (i.e. `class`, `mode`, `formatter`).

data_steward/utils/pipeline_logging.py
mark-velez: Thanks, I've incorporated your suggestion.

tests/bq_test_helpers.py
```diff
@@ -0,0 +1,114 @@
...
+        >>> d = {'item': 'book', 'qty': 2, 'price': 1.99}
+        >>> bq_test_helpers.fields_from_dict(d)
+            [SchemaField('item', 'STRING', 'NULLABLE', None, ()),
+             SchemaField('qty', 'INT64', 'NULLABLE', None, ()),
+             SchemaField('price', 'FLOAT64', 'NULLABLE', None, ())]
+    """
+    return [_field_from_key_value(key, value) for key, value in row.items()]
+
+
+def _to_ordered_dict(d: dict, key_order: Iterable[Any] = None):
```
lrwb-aou, line 57: Same as below comment.  Not sure the `Iterable[Any]` is actually what you want.  Maybe this would be a time to create a NamedType?

lrwb-aou (commented):
Very few comments.  The three comments are about conventions.

tests/bq_test_helpers.py
```diff
@@ -0,0 +1,114 @@
...
+        return OrderedDict(d)
+    else:
+        if key_order is None:
+            raise ValueError(
+                'Parameter key_order is required in order to convert'
+                ' a dict with multiple items to OrderedDict')
+    return OrderedDict((key, d[key]) for key in key_order)
+
+
+def mock_query_result(rows: List[dict_types], key_order: Iterable[Any] = None):
```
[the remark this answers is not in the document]
mark-velez: Theoretically passing a string _could_ work, but wouldn't be advisable..? 
```
scalar_st = dict(a=1, b=2, c=3)
mock_iter = mock_query_result([scalar_st], key_order='abc')
```
In cases where we don't care about the order I wanted to be able to do `scalar_st.keys()` which returns the mystical (to me) [view object](https://docs.python.org/3/library/stdtypes.html#dict-views). I have to admit I'm no expert on the type hierarchy here- could you provide a preferable subsuming type/expression? Whatever the case I will also note the deprecation like you mention.

data_steward/utils/pipeline_logging.py
```diff
@@ -111,3 +129,116 @@ def setup_logger(log_filepath_list, console_logging=True):
...
+    :param exc_type: type of the exception
+    :param exc_value: the exception
+    :param exc_traceback: the traceback associated with the exception
+    """
+    root_logger = logging.getLogger()
+    root_logger.critical('Uncaught exception',
+                         exc_info=(exc_type, exc_value, exc_traceback))
+
+
+def configure(level=logging.INFO, add_console_handler=True):
```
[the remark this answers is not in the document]
mark-velez: I updated it to default to `False` and updated the unit tests accordingly.

tests/bq_test_helpers.py
```diff
@@ -0,0 +1,114 @@
...
+from google.cloud.bigquery.table import TableListItem
+
+_TYPE_TO_FIELD_TYPE = {
+    str(int): 'INT64',
+    str(str): 'STRING',
+    str(float): 'FLOAT64',
+    str(datetime.date): 'DATE',
+    str(datetime.datetime): 'TIMESTAMP'
+}
+dict_types = Union[Dict, OrderedDict]
```
lrwb-aou, line 21: I didn't go through creating a NamedTuple because I like this setup you've got here.  So I created this very small patch file.  What do you think?  I know we could let someone use strings, but if these helper functions were used by an integration test, a string object iterable would blow this stuff up.  Because I find it unlikely that we'll have single character column names in an integration test.  On that subject, do you intend for these functions to be used by integration and unit tests alike?

[non_string_iterables.txt](https://github.com/all-of-us/curation/files/5497459/non_string_iterables.txt)

mark-velez: The patch would permit `key_order` to be `List[tuple]` or `List[List]` whereas it is most likely to be passed as a `List[str]` ([example](https://github.com/all-of-us/curation/blob/mv/purge_hpo/tests/unit_tests/data_steward/utils/bq_test.py#L188-L189)) as it should represent the keys within each row (type=dict) in the order that the result schema. A value belonging to any immutable (hashable?) type is permitted to be used as a key in a dictionary.

I thought maybe the patch had a typo and `non_string_iterable_types` should be the container type (?), but my IDE is complaining this is invalid.
image

lrwb-aou: There is a lot of debate about how to use [Iterable without using a string](https://github.com/python/typing/issues/256).  I don't think there's been a good answer for how to proceed forward.  I'm dropping this.

tests/bq_test_helpers.py
lrwb-aou: Dropping this as even the python community can't come to an agreement on it.

lrwb-aou (approved):
This is good.  Just need to remove the dead code.

PR #632, opened 2020-09-24
[DC-1028] Fixes for retraction of deactivated participant data

mark-velez:


data_steward/cdr_cleaner/cleaning_rules/remove_ehr_data_past_deactivation_date.py
```diff
@@ -41,13 +41,16 @@ def remove_ehr_data_queries(project_id, ticket_number, pids_project_id,
     ehr_union_dataset = bq_utils.get_unioned_dataset_id()
```
lrwb-aou, line 41: The only suggestion I have is to update with some comments.  We were both a little confused about `project_id` vs `pids_project_id` and what should be passed to which function.  The docstring doesn't distinguish between the two well enough to allow someone to read it and just know what is supposed to happen.  I can't select that area of the file to comment on, ugh.

data_steward/cdr_cleaner/cleaning_rules/remove_ehr_data_past_deactivation_date.py
```diff
@@ -41,13 +41,16 @@ def remove_ehr_data_queries(project_id, ticket_number, pids_project_id,
...
+    df = psr.get_deactivated_participants(pids_project_id, pids_dataset_id,
+                                          tablename,
+                                          DEACTIVATED_PARTICIPANTS_COLUMNS)
+    # To store dataframe in a BQ dataset table
+    destination_table = pids_dataset_id + '.' + tablename
+    psr.store_participant_data(df, project_id, destination_table)
     # creates sandbox and truncate queries to run for deactivated participant data drops
-    queries = rdp.create_queries(project_id, ticket_number, pids_project_id,
-                                 pids_dataset_id, tablename, ehr_union_dataset)
+    queries = rdp.create_queries(project_id, ticket_number, project_id,
```
lrwb-aou, line 51: I would probably add a comment here about why we are using `project_id` twice.  Something like,
```
    # the deactivated participants table was stored in the project, project_id, so we are using the same value as
    # the third parameter to rdp to represent the project containing the deactivated_pids table.
```

lrwb-aou (commented):
The logic looks good.  It would be good to add/update comments about the things we found confusing.

lrwb-aou (changes requested):
Along with adding some comments to make it easier to understand what this code is doing, the suggested changes cause unit test failures.  We need to mock the storage component since it is no longer covered by the existing mock.

data_steward/cdr_cleaner/cleaning_rules/remove_ehr_data_past_deactivation_date.py
mark-velez: I added comments and updated the docstring like you suggested

mark-velez:
@lrwb-aou thanks for your review. Now I'm just waiting/hoping all checks will pass. :crossed_fingers:

lrwb-aou (approved):
This looks good.

PR #634, opened 2020-09-28
[DC-1110] Clean Engine does not raise error if query fails in some cases

nishanthpp93:
* [DC-1110] Test to verify if clean engine raises error when query fails

[DC-1110]: https://precisionmedicineinitiative.atlassian.net/browse/DC-1110


nishanthpp93:
This was found to not be the case. A cleaning rule was creating the jobs which didn't have the necessary checks to verify if they failed. This will be addressed when the cleaning rule is refactored to use the base CR class.

PR #635, opened 2020-09-29
[DC-1112] Export VOCABULARY_DATASET in unioned script

mark-velez:
* Needed by populate_route_ids CR


lrwb-aou (approved):
Would it be worth adding a comment either here or  before calling the clean engine that there is a cleaning rule that requires this environment variable be set?

lrwb-aou (approved):
Thanks!

PR #637, opened 2020-10-02
[DC-1031] Update sandbox_dataset name in deid_runner.sh

ksdkalluri:


data_steward/tools/deid_runner.sh
```diff
@@ -70,8 +70,8 @@ gcloud auth activate-service-account --key-file="${key_file}"
 gcloud config set project "${APP_ID}"
 
 cdr_deid="${dataset_release_tag}_deid"
```
mark-velez, line 72: Is it true that the variable `cdr_deid` is also not needed and the usages below (from lines 82 and 102) are incorrect?

```
export BIGQUERY_DATASET_ID="${cdr_deid}"

bq mk --dataset --description "${version} sandbox dataset to apply cleaning rules on ${cdr_deid}" --label "phase:sandbox" --label "release_tag:${dataset_release_tag}" --label "de_identified:true" "${APP_ID}":"${registered_cdr_deid_sandbox}"
```

mark-velez (changes requested):
Please see if we can remove variable `cdr_deid` and its usages.

mark-velez (approved):

PR #640, opened 2020-10-05
[DC-1102] Create validation dataset with identity match tables

nishanthpp93:
* [DC-1102] Fix table schemas
* [DC-1102] Add tests


ksdkalluri (approved):
Looks good to me.

PR #641, opened 2020-10-05
[DC-1117] add backticks to achilles scripts

lrwb-aou:
* add needed backticks to achilles file scripts


mark-velez:
@lrwb-aou I did a regex search and it looks like you got all of them! :+1:

mark-velez (approved):

PR #642, opened 2020-10-06
[DC-1121] add backticks to combined script SQL statements

lrwb-aou:
* adds backticks to the combined script SQL statements
* updates the associated unit test


dcampbell-vumc (approved):
Quick and easy review! Verified that backticks were added at all necessary places! 👍🏾 Can merge once integration test passes!

dcampbell-vumc (approved):
The integration test looks good! Verified that backticks were applied at the proper places!

PR #644, opened 2020-10-07
[DC-1063] Fix Broken Link in results.html

dcampbell-vumc:
Updated the link in `hpo_report.html` file


nishanthpp93 (approved):
Just tested this and it works!

PR #646, opened 2020-10-08
[DC-1099] Update Vocabulary reference and remove extra tabs in concept.csv

ksdkalluri:


mark-velez (approved):
This looks good, thanks!

PR #648, opened 2020-10-09
[DC-916] required_labs.py assumes concept tables exists in target ehr dataset

dcampbell-vumc:
Added condition to `required_labs.py` to check if `MEASUREMENT_CONCEPT_SETS_DECENDANTS_TABLE` and `MEASUREMENT_CONCEPT_SETS_TABLE` exists which will be created if they do not.

### DIFF
--- a/data_steward/validation/metrics/required_labs.py
+++ b/data_steward/validation/metrics/required_labs.py
@@ -1,13 +1,15 @@
+# Python imports
 import logging
 
+# Third party imports
 import googleapiclient
 import oauth2client
-import os
+from google.cloud.exceptions import NotFound
 
+# Project imports
 import app_identity
 import bq_utils
 import common
-import resources
 from constants import bq_utils as bq_consts
 from utils import bq
 from validation.metrics.required_labs_sql import (IDENTIFY_LABS_QUERY,
@@ -29,10 +31,16 @@ def load_measurement_concept_sets_table(project_id, dataset_id):
     :return: None
     """
 
-    client = bq.get_client(project_id)
     table_name = f'{project_id}.{dataset_id}.{MEASUREMENT_CONCEPT_SETS_TABLE}'
 
-    if not bq_utils.table_exists(MEASUREMENT_CONCEPT_SETS_TABLE, dataset_id):
+    client = bq.get_client(project_id)
+    dataset = client.dataset(dataset_id)
+    table_ref = dataset.table(MEASUREMENT_CONCEPT_SETS_TABLE)
+
+    # will check to see if MEASUREMENT_CONCEPT_SETS_TABLE exists, table will be created if it is not found
+    try:
+        client.get_table(table_ref)
+    except NotFound:
         bq.create_tables(client=client,
                          project_id=project_id,
                          fq_table_names=[table_name],
@@ -67,14 +75,32 @@ def load_measurement_concept_sets_descendants_table(project_id, dataset_id):
     :return: None
     """
 
-    client = bq.get_client(project_id)
-    table_name = f'{project_id}.{dataset_id}.{MEASUREMENT_CONCEPT_SETS_DESCENDANTS_TABLE}'
+    descendants_table_name = f'{project_id}.{dataset_id}.{MEASUREMENT_CONCEPT_SETS_DESCENDANTS_TABLE}'
+    vocab_table_name = f'{project_id}.{dataset_id}.{common.VOCABULARY}'
 
-    if not bq_utils.table_exists(MEASUREMENT_CONCEPT_SETS_DESCENDANTS_TABLE,
-                                 dataset_id):
+    client = bq.get_client(project_id)
+    dataset = client.dataset(dataset_id)
+    descendants_table_ref = dataset.table(
+        MEASUREMENT_CONCEPT_SETS_DESCENDANTS_TABLE)
+    vocab_table_ref = dataset.table(common.VOCABULARY)
+
+    # will check to see if MEASUREMENT_CONCEPT_SETS_DESCENDANTS_TABLE exists, will be created if table is not found
+    try:
+        client.get_table(descendants_table_ref)
+    except NotFound:
         bq.create_tables(client=client,
                          project_id=project_id,
-                         fq_table_names=[table_name],
+                         fq_table_names=[descendants_table_name],
+                         exists_ok=False,
+                         fields=None)
+
+    # will check to see if VOCABULARY table exists, will be created if table is not found
+    try:
+        client.get_table(vocab_table_ref)
+    except NotFound:
+        bq.create_tables(client=client,
+                         project_id=project_id,
+                         fq_table_names=[vocab_table_name],
                          exists_ok=False,
                          fields=None)
 
--- a/data_steward/validation/metrics/required_labs.py
+++ b/data_steward/validation/metrics/required_labs.py
@@ -3,7 +3,6 @@ import logging
 import googleapiclient
 import oauth2client
 import os
-from google.cloud.exceptions import NotFound
 
 import app_identity
 import bq_utils
@@ -17,8 +16,6 @@ from validation.metrics.required_labs_sql import (IDENTIFY_LABS_QUERY,
 LOGGER = logging.getLogger(__name__)
 
 MEASUREMENT_CONCEPT_SETS_TABLE = '_measurement_concept_sets'
-MEASUREMENT_CONCEPT_SETS_TABLE_DISPOSITION = bq.bigquery.job.WriteDisposition.WRITE_TRUNCATE
-
 MEASUREMENT_CONCEPT_SETS_DESCENDANTS_TABLE = '_measurement_concept_sets_descendants'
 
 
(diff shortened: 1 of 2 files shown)


data_steward/validation/metrics/required_labs.py
```diff
@@ -54,6 +67,17 @@ def load_measurement_concept_sets_descendants_table(project_id, dataset_id):
     :return: None
     """
 
+    client = bq.get_client(project_id)
+    table_name = f'{project_id}.{dataset_id}.{MEASUREMENT_CONCEPT_SETS_DESCENDANTS_TABLE}'
+
+    if not bq_utils.table_exists(MEASUREMENT_CONCEPT_SETS_DESCENDANTS_TABLE,
```
nishanthpp93, line 73: Same comment here about `bq`/`client`

data_steward/validation/metrics/required_labs.py
```diff
@@ -54,6 +67,17 @@ def load_measurement_concept_sets_descendants_table(project_id, dataset_id):
...
+                                 dataset_id):
+        bq.create_tables(client=client,
+                         project_id=project_id,
+                         fq_table_names=[table_name],
+                         exists_ok=False,
+                         fields=None)
+
```
nishanthpp93, line 80: The ticket also mentions copying the vocabulary files if missing. This can be added here, before the query `IDENTIFY_LABS_QUERY` is called.

nishanthpp93 (changes requested):
Suggested some changes based on the ticket. The vocabulary dataset can be found via `common.py`

data_steward/validation/metrics/required_labs.py
```diff
@@ -54,6 +75,35 @@ def load_measurement_concept_sets_descendants_table(project_id, dataset_id):
     :return: None
     """
 
+    descendants_table_name = f'{project_id}.{dataset_id}.{MEASUREMENT_CONCEPT_SETS_DESCENDANTS_TABLE}'
+    vocab_table_name = f'{project_id}.{dataset_id}.{common.VOCABULARY}'
```
nishanthpp93, line 79: The `concept` table and the `concept_ancestor` table need to be verified to exist in `dataset_id`. If they don't, they need to be copied from the dataset `common.VOCABULARY` into `dataset_id`.

nishanthpp93 (commented):
Thanks for updating! Added a suggestion for the vocabulary tables/dataset

data_steward/validation/metrics/required_labs.py
dcampbell-vumc: Okay, I was off on that one, I understand now and will make the changes.

data_steward/validation/metrics/required_labs.py
```diff
@@ -15,6 +21,10 @@
 MEASUREMENT_CONCEPT_SETS_TABLE = '_measurement_concept_sets'
 MEASUREMENT_CONCEPT_SETS_DESCENDANTS_TABLE = '_measurement_concept_sets_descendants'
 
+ROW_COUNT_QUERY = JINJA_ENV.from_string("""SELECT table_id, row_count
+FROM `{{project_id}}.{{dataset_id}}.__TABLES__`
+WHERE table_id IN ('concept', 'concept_ancestor')""")
```
nishanthpp93, line 26: You don't have to add a qualifier for the concept/concept_ancestor here. This will help in determining whether the `MEASUREMENT_CONCEPT_SETS_DESCENDANTS_TABLE` table exists

data_steward/validation/metrics/required_labs.py
```diff
@@ -54,6 +80,67 @@ def load_measurement_concept_sets_descendants_table(project_id, dataset_id):
...
+    descendants_table_ref = dataset.table(
+        MEASUREMENT_CONCEPT_SETS_DESCENDANTS_TABLE)
+    concept_table_ref = dataset.table(common.CONCEPT)
+    concept_ancestor_table_ref = dataset.table(common.CONCEPT_ANCESTOR)
+
+    # query will check the row counts of each table in the specified dataset
+    row_count_query = ROW_COUNT_QUERY.render(project_id=project_id,
+                                             dataset_id=dataset_id)
+
+    job = client.query(row_count_query)
```
nishanthpp93, line 106: This part can be in a new function, like `check_and_copy_tables` or so. This will help make it unit-testable.

data_steward/validation/metrics/required_labs.py
```diff
@@ -54,6 +80,67 @@ def load_measurement_concept_sets_descendants_table(project_id, dataset_id):
...
+
+    # will check to see if MEASUREMENT_CONCEPT_SETS_DESCENDANTS_TABLE exists, will be created if table is not found
+    try:
+        client.get_table(descendants_table_ref)
+    except NotFound:
+        bq.create_tables(client=client,
+                         project_id=project_id,
+                         fq_table_names=[descendants_table_name],
+                         exists_ok=False,
+                         fields=None)
```
nishanthpp93, line 142: All of these can be removed, since you are already getting a data frame above, called `results`. This can be used for these tasks

data_steward/validation/metrics/required_labs.py
```diff
@@ -54,6 +80,67 @@ def load_measurement_concept_sets_descendants_table(project_id, dataset_id):
...
+
+    # will check if either the CONCEPT and CONCEPT_ANCESTOR tables is empty or not
+    # If so, the CONCEPT and CONCEPT_ANCESTOR tables will be copied from the common.VOCABULARY
+    # table to the destination dataset
+    if results.total_rows == 0:
+        client.copy_table(concept_source_table, concept_dest_table)
+        client.copy_table(concept_ancestor_source_table,
+                          concept_ancestor_dest_table)
+    else:
+        pass
```
nishanthpp93, line 117: You might want to use/create the `results` as a dataframe and search for the existence of the tables `concept`, `concept_ancestor` and `MEASUREMENT_CONCEPT_SETS_DESCENDANTS_TABLE`, under the `table_id` column. If they exists, you can look for the `row_count` field and check if it is `0`. Only then would you have to initiate the copy for the tables which are `0`

nishanthpp93 (commented):
Added some comments on using the results from the `__TABLES__` query.

data_steward/validation/metrics/required_labs.py
dcampbell-vumc: This makes sense to me. I will work on this today.

data_steward/validation/metrics/required_labs.py
```diff
@@ -15,6 +21,75 @@
...
+                          concept_ancestor_dest_table)
+
+    # checks if any tables in the results_dataframe are empty, if CONCEPT and CONCEPT_ANCESTOR tables are empty,
+    # they are copied from the CONCEPT and CONCEPT_ANCESTOR tables in common.VOCABULARY
+    if (results_dataframe['row_count'] == 0).any():
+        if common.CONCEPT in (results_dataframe['table_id']).any():
+            client.copy_table(concept_source_table, concept_dest_table)
+        if common.CONCEPT_ANCESTOR in (results_dataframe['table_id']).any():
+            client.copy_table(concept_ancestor_source_table,
+                              concept_ancestor_dest_table)
```
nishanthpp93, line 75: This may not be doing what is intended. It is checking if any entries in `row_count` are `0`, and if so, replace the `concept` and `concept_ancestor` tables regardless of their row counts. This should ideally create a new dataframe where a filter is applied such that row_count is 0, and then check if the tables exist in this filtered dataframe.

data_steward/validation/metrics/required_labs.py
```diff
@@ -15,6 +21,75 @@
...
+    row_count_query = ROW_COUNT_QUERY.render(project_id=project_id,
+                                             dataset_id=dataset_id)
+
+    results_dataframe = client.query(row_count_query).to_dataframe()
+
+    # checks if CONCEPT and CONCEPT_ANCESTOR tables exist, if they don't, they are copied from the
+    # CONCEPT and CONCEPT_ANCESTOR tables in common.VOCABULARY
+    if common.CONCEPT not in (results_dataframe['table_id']).any():
+        client.copy_table(concept_source_table, concept_dest_table)
+    if common.CONCEPT_ANCESTOR not in (results_dataframe['table_id']).any():
```
nishanthpp93, line 64: nit: You can use `.values` to treat it like a Python List. E.g
```suggestion
    if common.CONCEPT_ANCESTOR not in results_dataframe['table_id']).values:
```

nishanthpp93 (commented):
Thank you for adding the test, it makes it so much easier to go through! Added a suggestion and a nit

data_steward/validation/metrics/required_labs.py
dcampbell-vumc: Ah, that makes sense. I'll make the changes.

data_steward/validation/metrics/required_labs.py
```diff
@@ -15,6 +21,76 @@
...
+    # checks if MEASUREMENT_CONCEPT_SETS_TABLE and MEASUREMENT_CONCEPT_SETS_DESCENDANTS_TABLE exist, if they
+    # do not exist, they will be created
+    if MEASUREMENT_CONCEPT_SETS_TABLE not in results_dataframe['table_id']:
+        bq.create_tables(client=client,
+                         project_id=project_id,
+                         fq_table_names=[concept_sets_table_name],
+                         exists_ok=True,
+                         fields=None)
+    if MEASUREMENT_CONCEPT_SETS_DESCENDANTS_TABLE not in results_dataframe[
+            'table_id']:
```
nishanthpp93, line 87: These two statements need to use `results_dataframe['table_id'].values`

nishanthpp93 (commented):
One last change and we should be good

nishanthpp93 (approved):
Looks good! Thank you for making all the changes @dcampbell-vumc

PR #651, opened 2020-10-13
[DC-1133] Add clean_height_weight missing sandbox dataset parameter

mark-velez:


ChaoPang (changes requested):
`dataset_id` is not a parameter in `INSERT_NEW_ROWS_QUERY`, please remove it from another occurrence at https://github.com/all-of-us/curation/blob/1aac63ee5fe99bc8ba2a4c504246a4593734ba7b/data_steward/cdr_cleaner/cleaning_rules/clean_height_weight.py#L730

The corresponding unit test is broken and please fix, an example at [gist](https://gist.github.com/ChaoPang/0ed9efba4722c55eb14a76cb4c16dcee)

mark-velez:
Thanks @ChaoPang. Next time you might want to try GitHub's [suggestion feature](https://docs.github.com/en/free-pro-team@latest/github/collaborating-with-issues-and-pull-requests/commenting-on-a-pull-request#adding-line-comments-to-a-pull-request). The bug was not caught in part because the unit tests are circular; I will create an integration test which would catch it. Not sure if I should remove the whole test case or mark the tests as skip for now. I will consult the original author on that point.

dcampbell-vumc (approved):

ChaoPang:
@dcampbell-vumc thanks for looking at this and approving it in a timely fashion, it would've taken me a while to do so. 

@mark-velez I agree, the suggestions are definitely better! Do you know how to make suggestions or comments on the unchanged code that is not in the PR? I actually tried to add suggestions but Github didn't seem to allow me to do it on those unchanged lines.

ChaoPang (approved):
Approve!

mark-velez:
@ChaoPang I see why you gisted now- I've also not found a convenient way in a GitHub PR to reference unchanged code. Please do share if you should make any discoveries in that area!

PR #652, opened 2020-10-13
[DC-1134] Fix return type and add backticks in drop_zero_concept_ids

mark-velez:


ksdkalluri (approved):
Looks good to me, But you might need to update the unit test as well.

PR #653, opened 2020-10-14
[DC-1136] Create a notebook for visualizing height and weight distributions

ChaoPang:
* moved the common functions for creating histgrams to measurement_hist_utils
* dded height_weight_distribution.py and refactored unit_standardization.py


dcampbell-vumc (approved):
Looks great! Was able to get the notebook working and see the height and weight distributions. I am confident this will help when we refactor the height and weight cleaning rule!

PR #658, opened 2020-10-23
[DC-1016] Pipeline Logging unit test intermittently fails

dcampbell-vumc:
* added datetime mocking to ensure slight changes in time won't result
in test failure in `test_logger_setup` and `test_generate_path` modules
* changed `logging_path` in `setUp` to ensure date matches the mocked date

### DIFF
--- a/tests/unit_tests/data_steward/utils/pipeline_logging_test.py
+++ b/tests/unit_tests/data_steward/utils/pipeline_logging_test.py
@@ -49,7 +49,7 @@ class PipelineLoggingTest(unittest.TestCase):
         # tests if the lists are equal regardless of order
         self.assertCountEqual(results, self.log_path)
 
-        # tests if the lists are
+        # tests if the lists are equal
         self.assertEqual(results, self.log_path)
         self.assertListEqual(results, self.log_path)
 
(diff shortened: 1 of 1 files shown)


tests/unit_tests/data_steward/utils/pipeline_logging_test.py
```diff
@@ -32,16 +31,26 @@ def setUpClass(cls):
...
+
         # checks that log_path is generated properly
         results = generate_paths(self.log_file_list)
-        self.assertEquals(results, self.log_path)
+
+        # tests if the lists are equal regardless of order
+        self.assertCountEqual(results, self.log_path)
+
+        # tests if the lists are equal
+        self.assertEqual(results, self.log_path)
```
nishanthpp93, line 53: nit: This is possibly redundant since there is a check for `self.assertListEqual` in the next line

nishanthpp93 (approved):
Looks good! Added a small nit but it's not necessary

tests/unit_tests/data_steward/utils/pipeline_logging_test.py
dcampbell-vumc: I can get rid of `self.assertEqual(results, self.log_path)` 👍🏾

PR #659, opened 2020-10-26
[DC-1154] Update Snapshot Description

dcampbell-vumc:
* Added command line argument for date in form of yyyy-mm-dd
* Updated the ehr dataset description to utilize the date

### DIFF
--- a/data_steward/tools/create_ehr_and_rdr_snapshot.sh
+++ b/data_steward/tools/create_ehr_and_rdr_snapshot.sh
@@ -9,7 +9,6 @@ Usage: create_ehr_snapshot.sh
   --rdr_dataset <RDR dataset ID>
   --validation_dataset <Validation dataset ID>
   --dataset_release_tag <release tag for the CDR>
-  --snapshot_date <date snapshot is generated format yyyy-mm-dd>
 "
 
 while true; do
@@ -34,10 +33,6 @@ while true; do
     dataset_release_tag=$2
     shift 2
     ;;
-  --snapshot_date)
-    snapshot_date=$2
-    shift 2
-    ;;
   --)
     shift
     break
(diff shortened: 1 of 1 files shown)


data_steward/tools/create_ehr_and_rdr_snapshot.sh
```diff
@@ -9,6 +9,7 @@ Usage: create_ehr_snapshot.sh
   --rdr_dataset <RDR dataset ID>
   --validation_dataset <Validation dataset ID>
   --dataset_release_tag <release tag for the CDR>
+  --snapshot_date <date snapshot is generated format yyyy-mm-dd>
```
ChaoPang, line 12: I don't think this was specified in the ticket. A general question though, should we provide the `snapshot_date` as a parameter or should we get the date using `snapshot_date=$(date +'%Y-%m-%d')` in the bash script directly?

ChaoPang (commented):
@dcampbell-vumc I just left a question here, I am not so sure which approach we want to take for getting `snapshot_date`? I don't think this was specified in the ticket, should we provide the snapshot_date as a parameter or should we get the date using `snapshot_date=$(date +'%Y-%m-%d')` in the bash script directly?

data_steward/tools/create_ehr_and_rdr_snapshot.sh
dcampbell-vumc: Oh, to be honest, I didn't know there was another way to do it 😅 I think it would be better if we didn't need to supply it as a parameter.

lrwb-aou:
> @dcampbell-vumc I just left a question here, I am not so sure which approach we want to take for getting `snapshot_date`? I don't think this was specified in the ticket, should we provide the snapshot_date as a parameter or should we get the date using `snapshot_date=$(date +'%Y-%m-%d')` in the bash script directly?

I prefer just getting it directly from the system.

dcampbell-vumc:
> > @dcampbell-vumc I just left a question here, I am not so sure which approach we want to take for getting `snapshot_date`? I don't think this was specified in the ticket, should we provide the snapshot_date as a parameter or should we get the date using `snapshot_date=$(date +'%Y-%m-%d')` in the bash script directly?
> 
> I prefer just getting it directly from the system.

Yep, that's what I just changed it to and made sure everything is working properly!

ChaoPang (approved):
Looks good now

PR #660, opened 2020-10-27
[DC-1109] Tables fail to copy in snapshot_by_query

nishanthpp93:
* [DC-1109] Add src_table_id field to mapping tables
* [DC-1109] Add _mapping_person and fix tests
* [DC-1109] Update combined to accommodate new mapping
* [DC-1109] Use f-strings

[DC-1109]: https://precisionmedicineinitiative.atlassian.net/browse/DC-1109
[DC-1109]: https://precisionmedicineinitiative.atlassian.net/browse/DC-1109
[DC-1109]: https://precisionmedicineinitiative.atlassian.net/browse/DC-1109
[DC-1109]: https://precisionmedicineinitiative.atlassian.net/browse/DC-1109

### DIFF
--- a/data_steward/constants/tools/combine_ehr_rdr.py
+++ b/data_steward/constants/tools/combine_ehr_rdr.py
@@ -14,180 +14,181 @@ EHR_TABLES_TO_COPY = ['death']
 DOMAIN_TABLES = list(
     set(cdm.tables_to_map()) - set(RDR_TABLES_TO_COPY + EHR_TABLES_TO_COPY))
 TABLES_TO_PROCESS = RDR_TABLES_TO_COPY + EHR_TABLES_TO_COPY + DOMAIN_TABLES
-LEFT_JOIN = (
-    ' LEFT JOIN'
-    ' ('
-    ' SELECT *'
-    ' FROM ('
-    ' SELECT'
-    ' *,'
-    ' row_number() OVER (PARTITION BY {prefix}.{field}, {prefix}.src_hpo_id ) '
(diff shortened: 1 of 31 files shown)


dcampbell-vumc (approved):
Thanks for updating `combine_eehr_rdr.py` to f-strings, it made it much more readable for me! 

I ran the unit test and verified that the `mapping_*` tables had nulls in `src_dataset_id` column as necessary!

nishanthpp93:
Thanks for the feedback @dcampbell-vumc. I realized I didn't update the format strings to f-strings in `ehr_union.py` and have made that change. I also filled the `NULL` columns to be more informative with the correct `src_dataset_id` and `src_table_id` because while using `WRITE_TRUNCATE` on `NULL` columns with `STRING` type, BQ changes the type to `INTEGER` for some reason.

dcampbell-vumc:
> Thanks for the feedback @dcampbell-vumc. I realized I didn't update the format strings to f-strings in `ehr_union.py` and have made that change. I also filled the `NULL` columns to be more informative with the correct `src_dataset_id` and `src_table_id` because while using `WRITE_TRUNCATE` on `NULL` columns with `STRING` type, BQ changes the type to `INTEGER` for some reason.

Ah, I should have asked that as a clarifying question whether `ehr_union.py` should have the strings converted to f-strings 😅 thanks for making that change though! Also, yes, having the columns be more informative is a good call as well. Strange that BQ changes the type to `INTEGER` 🤔 

Once the integration test check passes this will be good to go!

PR #661, opened 2020-10-29
[DC-1144] recreate_person.py fails due to partitioning mismatch on person table via DDL

nishanthpp93:
* [DC-1144] Add ALTER TABLE DDL and use UPDATE to preserve schema

### DIFF
--- a/data_steward/tools/recreate_person.py
+++ b/data_steward/tools/recreate_person.py
@@ -7,20 +7,31 @@ from common import JINJA_ENV
 
 LOGGER = logging.getLogger(__name__)
 
-NEW_PERSON_QUERY = """
-SELECT
-    p.* EXCEPT (state_of_residence_concept_id, state_of_residence_source_value),
-    ext.state_of_residence_concept_id,
-    ext.state_of_residence_source_value
-FROM
-    {{person.project}}.{{person.dataset_id}}.{{person.table_id}} p
-LEFT JOIN
-    {{person_ext.project}}.{{person_ext.dataset_id}}.{{person_ext.table_id}} ext
-USING
-    (person_id)
-"""
+ADD_COLUMNS_QUERY = JINJA_ENV.from_string("""
+ALTER TABLE `{{person.project}}.{{person.dataset_id}}.{{person.table_id}}`
+ADD COLUMN IF NOT EXISTS state_of_residence_concept_id INT64 
+    OPTIONS(description="[All of Us OMOP extension] A foreign key for the state of residence"),
+ADD COLUMN IF NOT EXISTS state_of_residence_source_value STRING 
+    OPTIONS(description="[All of Us OMOP extension] The source code for the state of residence")
+""")
 
-NEW_PERSON_QUERY_TMPL = JINJA_ENV.from_string(NEW_PERSON_QUERY)
+UPDATE_PERSON_QUERY = JINJA_ENV.from_string("""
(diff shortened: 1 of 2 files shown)


tests/integration_tests/data_steward/tools/recreate_person_test.py
```diff
@@ -52,22 +50,32 @@ def setUp(self):
...
-        self.person = Table.from_string(self.fq_person)
-        self.person_ext = Table.from_string(self.fq_person_ext)
+        self.person = gbq.Table.from_string(self.fq_person)
+        self.person_ext = gbq.Table.from_string(self.fq_person_ext)
 
         self.client = bq.get_client(self.project_id)
-        for table in self.table_ids:
-            self.client.delete_table(table, not_found_ok=True)
-        bq.create_tables(self.client, self.project_id, self.table_ids)
+        self.tearDown()
```
dcampbell-vumc, line 59: Can you explain why this `tearDown()` function is here?

dcampbell-vumc (commented):
The new `recreate_person.py` module makes much more sense to me now! Just had a couple general questions before approval!

data_steward/tools/recreate_person.py
```diff
@@ -7,20 +7,31 @@
...
+ADD_COLUMNS_QUERY = JINJA_ENV.from_string("""
+ALTER TABLE `{{person.project}}.{{person.dataset_id}}.{{person.table_id}}`
+ADD COLUMN IF NOT EXISTS state_of_residence_concept_id INT64 
+    OPTIONS(description="[All of Us OMOP extension] A foreign key for the state of residence"),
+ADD COLUMN IF NOT EXISTS state_of_residence_source_value STRING 
+    OPTIONS(description="[All of Us OMOP extension] The source code for the state of residence")
+""")
+
+UPDATE_PERSON_QUERY = JINJA_ENV.from_string("""
+UPDATE `{{person.project}}.{{person.dataset_id}}.{{person.table_id}}` p
```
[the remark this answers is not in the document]
nishanthpp93: Yes, our primary motivation for using DDL was the time it takes (also depends on the query, but for most cleaning rules it works well). However, because we use a `WRITE_TRUNCATE` we lose the schema. DML is advantageous in preserving the schema and since the query affects only 2 columns and BQ is architecture columnar, this performs relatively quickly. I tried to `WRITE_TRUNCATE` to a specific partition as we did earlier to preserve the schema but it looks like the Python BQ client doesn't have that capability yet.

tests/integration_tests/data_steward/tools/recreate_person_test.py
nishanthpp93: I used it to remove any existing person/person_ext tables that could interfere with the tests. Removing them beforehand ensures the tables can be created and populated as expected for the test.

dcampbell-vumc: That makes sense, thank you!

data_steward/tools/recreate_person.py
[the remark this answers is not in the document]
dcampbell-vumc: That's right, that was our main concern was the schema being wiped out. Thanks for explaining this!

dcampbell-vumc (approved):
Thanks for clarifying all my queries! Everything looks good and this can be merged!

PR #662, opened 2020-11-03
[DC-1167] generate_ext_tables Formatting Fix

dcampbell-vumc:
* Updated strings to f-strings
* Updated SQL to use JINJA_ENV

### DIFF
--- a/data_steward/tools/generate_ext_tables.py
+++ b/data_steward/tools/generate_ext_tables.py
@@ -7,7 +7,7 @@ import constants.cdr_cleaner.clean_cdr as cdr_consts
 from resources import fields_for, MAPPING_TABLES
 from common import JINJA_ENV
 
-LOGGER = logging.getLogger("__name__")
+LOGGER = logging.getLogger(__name__)
 
 EXT_FIELD_TEMPLATE = [{
     "type": "integer",
@@ -131,8 +131,7 @@ def convert_to_bq_string(mapping_list):
     """
     bq_insert_list = []
     for hpo_rdr_item in mapping_list:
-        bq_insert_list.append("(\"{hpo_rdr_id}\", \"{src_id}\")".format(
-            hpo_rdr_id=hpo_rdr_item[0], src_id=hpo_rdr_item[1]))
+        bq_insert_list.append(f"(\"{hpo_rdr_item[0]}\", \"{hpo_rdr_item[1]}\")")
     bq_insert_string = ', '.join(bq_insert_list)
     return bq_insert_string
 
--- a/data_steward/tools/generate_ext_tables.py
+++ b/data_steward/tools/generate_ext_tables.py
@@ -131,7 +131,7 @@ def convert_to_bq_string(mapping_list):
     """
     bq_insert_list = []
     for hpo_rdr_item in mapping_list:
-        bq_insert_list.append(f"(\"{hpo_rdr_item[0]}\", \"{hpo_rdr_item[1]}\")")
+        bq_insert_list.append(f'("{hpo_rdr_item[0]}", "{hpo_rdr_item[1]}")')
     bq_insert_string = ', '.join(bq_insert_list)
     return bq_insert_string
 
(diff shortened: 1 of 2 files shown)


data_steward/tools/generate_ext_tables.py
```diff
@@ -5,11 +5,10 @@
 import constants.bq_utils as bq_consts
 import constants.cdr_cleaner.clean_cdr as cdr_consts
 from resources import fields_for, MAPPING_TABLES
+from common import JINJA_ENV
 
 LOGGER = logging.getLogger("__name__")
 
-LOGGER = logging.getLogger(__name__)
-
```
lrwb-aou, line 12: This should be left in.  The `__name__` variable is a special python variable used to name each module.  This is setting the name of the python logger to the module name.  So you see logs like this.
```
2020-11-03 12:18:03,439 - INFO - __main__ - using dynamic extension table schema for table: procedure_occurrence_ext
```

data_steward/tools/generate_ext_tables.py
```diff
@@ -5,11 +5,10 @@
 import constants.bq_utils as bq_consts
 import constants.cdr_cleaner.clean_cdr as cdr_consts
 from resources import fields_for, MAPPING_TABLES
+from common import JINJA_ENV
 
 LOGGER = logging.getLogger("__name__")
```
lrwb-aou, line 10: This needs to be removed.  This is setting the logger name to the string __name__.  So you would see logs like this.
```
2020-11-03 12:18:03,439 - INFO - __name__ - using dynamic extension table schema for table: procedure_occurrence_ext
```

data_steward/tools/generate_ext_tables.py
```diff
@@ -160,7 +159,7 @@ def create_and_populate_source_mapping_table(project_id, dataset_id):
                                    SITE_MAPPING_FIELDS,
```
lrwb-aou, line 159: I would suggest changing lines 134 and 135 to using f-strings too.  I can't comment on that line directly because it hasn't been changed in this PR.

```
        bq_insert_list.append(f'("{hpo_rdr_item[0]}", "{hpo_rdr_item[1]}")')

File: tests/unit_tests/data_steward/tools/generate_ext_tables_test.py [hunk 4]
```diff
@@ -159,7 +159,7 @@ def test_generate_ext_table_queries(self, mock_mapping_tables,
                     common.PERSON, common.DEATH, common.FACT_RELATIONSHIP
```
lrwb-aou, line 159: Line 63 in this test could be changed to:
```
        ext_table = f'foo{gen_ext.EXT_TABLE_SUFFIX}'
```

Line 101 could be changed to:
```
        mapping_observation = f'{gen_ext.MAPPING_PREFIX}{observation_table_id}'
```

Line 140-141 could become:
```
        hpo_bq_list.append(f'("{gen_ext.RDR}", "{gen_ext.PPI_PM}")')
```

lrwb-aou (changes requested):
The logger name has to be changed.  The others are suggestions about more places we could be using f-strings.

data_steward/tools/generate_ext_tables.py
```diff
@@ -132,8 +131,7 @@ def convert_to_bq_string(mapping_list):
     """
     bq_insert_list = []
     for hpo_rdr_item in mapping_list:
-        bq_insert_list.append("(\"{hpo_rdr_id}\", \"{src_id}\")".format(
-            hpo_rdr_id=hpo_rdr_item[0], src_id=hpo_rdr_item[1]))
+        bq_insert_list.append(f"(\"{hpo_rdr_item[0]}\", \"{hpo_rdr_item[1]}\")")
```
lrwb-aou, line 134: This is optional. I think this is more readable.  Changing that first double quote to a single quote means we don't need to need to escape the double quotes in the string.
```suggestion
        bq_insert_list.append(f'("{hpo_rdr_item[0]}", "{hpo_rdr_item[1]}")')
```

lrwb-aou (approved):
This works well.  Nice changes.  I have one other suggestion.  I think it might help with maintainability, but it does not make any functional changes.  So I will leave it to your discretion.  :)

lrwb-aou (approved):
Looks good!

PR #663, opened 2020-11-04
[DC-1159] Send slack alert when unable to write files to a bucket

dcampbell-vumc:
* Added bucket write access error handler to `app_errors.py`
* Modified `app_errors_test.py` to test new bucket write access error handler functionality


ChaoPang (changes requested):
It looks like the exceptions (e.g. HttpError and BucketDoesNotExistError) are caught in `main.py` but they are not raised again after being logged to stackdriver, so as a consequence those errors are silently discarded and error handlers would not catch those. 
https://github.com/all-of-us/curation/blob/a45ccc4a1568be9885aea94ab549a4e2ba5f9ab6/data_steward/validation/main.py#L532

Actually, we probably want to go through `main.py` to find all try-catch blocks and add `raise error` where it is missing so that the error messages are being propagated to error handlers.

data_steward/curation_logging/test_main.py
```diff
@@ -0,0 +1,22 @@
+"""
+Notes: -- if dev is using macOS and gets error:
```
ChaoPang, line 2: `test_main.py` was something I created for easy testing, however, this script may not belong to the repository and this is like a precursor of the unit test.  I do love your notes here. I am wondering if you could move the notes to `slack_logging_handler.py` or `slack_logging_handler_test.py`

dcampbell-vumc: I'm not sure it would make sense to put the notes in those places as the logging messages aren't being sent to the slack channel. Would it be beneficial to create another test function to send them to the test slack channel? We would need to give everyone access to that test channel, will thee same slack token work for everyone?

tests/unit_tests/data_steward/curation_logging/slack_logging_handler_test.py
```diff
@@ -0,0 +1,47 @@
+"""
```
ChaoPang, line 1:
```suggestion
"""
Unit test for slack_logging_handler module

Original Issue: DC-1159

Ensures that the slack logging messages are properly captured and sent to the curation slack alert channel
"""

# Python imports
import os
import logging
import mock
import unittest

# Project imports
from curation_logging.slack_logging_handler import initialize_slack_logging
from utils.slack_alerts import SLACK_TOKEN, SLACK_CHANNEL

TEST_SLACK_TOKEN = 'test_slack_token'
TEST_CHANNEL_NAME = 'channel_name'

INFO_MESSAGE = 'Do not send this'
WARNING_MESSAGE = 'logging.warning slack message sent by logging handler!'
CRITICAL_MESSAGE = 'logging.critical slack message sent by logging handler!'
ERROR_MESSAGE = 'logging.error slack message sent by logging handler!'


class SlackLoggingHandlerTest(unittest.TestCase):

    @classmethod
    def setUpClass(cls):
        print('**************************************************************')
        print(cls.__name__)
        print('**************************************************************')

    @mock.patch.dict(os.environ, {
        SLACK_TOKEN: TEST_SLACK_TOKEN,
        SLACK_CHANNEL: TEST_CHANNEL_NAME
    })
    @mock.patch('curation_logging.slack_logging_handler.post_message')
    def test_initialize_slack_logging(self, mock_post_message):
        initialize_slack_logging()

        logging.info(INFO_MESSAGE)
        logging.debug(INFO_MESSAGE)

        logging.warning(WARNING_MESSAGE)
        logging.critical(CRITICAL_MESSAGE)
        logging.error(ERROR_MESSAGE)

        self.assertEqual(mock_post_message.call_count, 3)

        mock_post_message.assert_any_call(WARNING_MESSAGE)
        mock_post_message.assert_any_call(CRITICAL_MESSAGE)
        mock_post_message.assert_any_call(ERROR_MESSAGE)
```

data_steward/utils/slack_alerts.py
```diff
@@ -52,6 +71,23 @@ def _get_slack_channel_name():
...
+    client = _get_slack_client()
+    channel_name = _get_slack_channel_name()
+    response = client.conversations_list(limit=2000)
+    if response.status_code == 200:
+        for channel in response.data['channels']:
+            if 'test' in channel['name']:
+                print(channel['name'])
+            if channel['name'] == channel_name:
+                return True
+    return False
```
ChaoPang, line 88: We might need to use try... except to catch `SlackApiError` if `client.conversations_list` fails to run. We don't want to raise `SlackApiError` in this case because the unit tests that import `validation.main.py` will fail because they don't have valid slack_token and slack_channel set up. 

```suggestion
def _is_channel_available():
    """
    Test if the Slack channel is available
    :return:
    """
    client = _get_slack_client()
    channel_name = _get_slack_channel_name()
    try:
        response = client.conversations_list(limit=sys.maxsize) -- use the max size allowed to avoid a hardcoded int
        if response.status_code == 200:
            for channel in response.data['channels']:
                if channel['name'] == channel_name:
                    return True
    except SlackApiError as e:
        logging.error(e)
    return False
```

data_steward/curation_logging/slack_logging_handler.py
```diff
@@ -0,0 +1,44 @@
...
+def _is_raised_from_itself(record):
+    return record.module in __name__
+
+
+def initialize_slack_logging():
+    """
+    Setup Slack logging
+    """
+    if SLACK_TOKEN in os.environ and SLACK_CHANNEL in os.environ:
+        if check_channel_and_token:
```
ChaoPang, line 37: I think we can just check the channel directly since it internally checks whether the token is authorized. 

```suggestion
        if is_channel_available():
```

data_steward/curation_logging/slack_logging_handler.py
```diff
@@ -0,0 +1,44 @@
...
+            if not _is_raised_from_itself(record):
+                post_message(record.msg % record.args)
+        except SlackConfigurationError:
+            logging.exception(
+                'Slack is not configured for posting messages, refer to playbook.'
+            )
+            raise
+
+def _is_raised_from_itself(record):
+    return record.module in __name__
```
ChaoPang, line 29: nit: this is just a personal preference, I would define `_is_raised_from_itself` as a class method in `SlackLoggingHandler`. `self.__module__` seems more understandable than `__name__` although they refer to the same thing. 

```suggestion
    def _is_raised_from_itself(self, record):
        return record.module in self.__module__
```

data_steward/utils/slack_alerts.py
```diff
@@ -39,6 +42,22 @@ def _get_slack_token():
...
+    slack_token = _get_slack_token()
+    response = client.auth_test(token=slack_token)
+    if response.status_code == 200:
+        if 'ok: True' in response:
+            return True
+        else:
+            return False
+    return False
+
+
```
ChaoPang, line 60: Since we can just call `is_channel_available()` directly, I am not sure if we still need to keep this function around?What do you think?

data_steward/utils/slack_alerts.py
```diff
@@ -39,6 +42,22 @@ def _get_slack_token():
...
+    """
+    client = _get_slack_client()
+    slack_token = _get_slack_token()
+    response = client.auth_test(token=slack_token)
+    if response.status_code == 200:
+        if 'ok: True' in response:
+            return True
+        else:
+            return False
+    return False
```
ChaoPang, line 58: We might need to add try....except for this because `client.auth_test` could throw SlackApiError. We don't want to raise SlackApiError in this case because the unit tests that import validation.main.py will fail because they don't have valid slack_token and slack_channel set up.

```suggestion
    try:
        response = client.auth_test(token=slack_token)
        return response.status_code == 200 and response['ok']
    except SlackApiError as e:
        logging.error(e)
    return False
```

data_steward/utils/slack_alerts.py
```diff
@@ -62,6 +98,13 @@ def _get_slack_client():
     return slack.WebClient(slack_token)
 
 
+def check_channel_and_token():
+    if _is_channel_available and _is_token_valid:
+        return True
+    else:
+        return False
+
+
```
ChaoPang, line 107: This function could potentially go.

data_steward/curation_logging/slack_logging_handler.py
```diff
@@ -0,0 +1,44 @@
...
+     Logging handler to send messages to a Slack Channel.
+    """
+
+    def __init__(self):
+        super().__init__(level=logging.WARNING)
+
+    def emit(self, record):
+
+        try:
+            if not _is_raised_from_itself(record):
```
ChaoPang, line 20: Just a comment to explain why we add this check is here. 

```suggestion
            # this is added for preventing the infinite loop from happening
            if not _is_raised_from_itself(record):
```

data_steward/utils/slack_alerts.py
dcampbell-vumc: Gotcha! Before if there wasn't 200 response code returned from `conversations_list` it would return false raising all sorts of errors in unit test modules that depended on having the `SLACK_TOKEN` and `CHANNEL_NAME`.

data_steward/curation_logging/slack_logging_handler.py
dcampbell-vumc: I got an error running the unit test when I defined this as a class method because I had it as `return record.module in self.__name__` NOT `return record.module in self.__module__`. I changed it and now it's working as expected!

data_steward/utils/slack_alerts.py
dcampbell-vumc: I agree that we don't need this anymore since `is_channel_available` is using the token, if the token doesn't exist, the channel won't be found right? I will remove the `is_token_valid` function.

data_steward/utils/slack_alerts.py
dcampbell-vumc: I removed it because I removed the `is_token_valid` function and we will be calling `is_channel_available` directly in the `slack_logging_handler` module.

data_steward/curation_logging/slack_logging_handler.py
dcampbell-vumc: Yep, because I will definitely forget :p

ChaoPang (changes requested):
Please review the comments and let me know how to proceed

data_steward/utils/slack_alerts.py
```diff
@@ -22,7 +25,7 @@ class SlackConfigurationError(RuntimeError):
     """
 
     def __init__(self, msg):
-        super(SlackConfigurationError, self).__init__()
+        super(SlackConfigurationError, self).__init__(msg)
```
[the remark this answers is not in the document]
dcampbell-vumc: I added this because it wasn't printing what the actual error was. Before when it failed it would just say:  
```raise SlackConfigurationError(UNSET_SLACK_TOKEN_MSG)```
I had to look through the actual code to see what that error was. When I added `msg`, it would say: 
```raise SlackConfigurationError(UNSET_SLACK_TOKEN_MSG)
utils.slack_alerts.SlackConfigurationError: Slack token not set in environment variable SLACK_TOKEN```

Is there a way to show this without adding `msg`?

ChaoPang (approved):

PR #666, opened 2020-11-11
[DC-1111] Update _mapping_person.person_id semantics

mark-velez:


nishanthpp93 (approved):
Looks good! Verified that the mapping person table contains the person_id in both columns.

mark-velez:
Thanks @nishanthpp93 ! I "fixed" the integration test -it was CaSiNg and  white   space  differences in the SQL. I went ahead and moved the test to unit test and also removed a redundant integration test that was already covered by a unit test. That should reduce the build time at least a little. Just waiting for the build to pass now before merging. :crossed_fingers:

PR #668, opened 2020-11-12
[DC-1177] Logging error causes union jobs to fail

nishanthpp93:
* [DC-1177] Remove long queries from log messages and use job_id

[DC-1177]: https://precisionmedicineinitiative.atlassian.net/browse/DC-1177


ChaoPang (approved):
Looks good

PR #671, opened 2020-11-13
[DC-807] Update Invalid Foreign Keys CR to New Base Class

dcampbell-vumc:
- Updated null_invalid_foreign_keys.py to inherit from BaseCleaningRule
- Added NullInvalidForeignKeys to COMBINED_CLEANING_CLASSES list in clean_cdr.py
- Created unit test null_invalid_foreign_keys_test.py to ensure python is functioning (this will not address SQL functionality)


dcampbell-vumc:
Had trouble updating branch so created a copy, made changes to the copied branch and pushed. Thought it would be easier to create new PR instead of trying to merge the two local branches. Previous PR comments [here](https://github.com/all-of-us/curation/pull/669)

data_steward/cdr_cleaner/cleaning_rules/null_invalid_foreign_keys.py
```diff
@@ -21,84 +29,215 @@
...
+                    table,
+                cdr_consts.DESTINATION_DATASET:
+                    self.dataset_id,
+                cdr_consts.DISPOSITION:
+                    bq_consts.WRITE_TRUNCATE
+            }
+
+            queries_list.append(invalid_foreign_key_query)
+
+            sandbox_query = {
```
ChaoPang, line 196: `sandbox_query` should be created before `invalid_foreign_key_query` because we want to sandbox records before `invalid_foreign_key_query` updates the content of the table.

data_steward/cdr_cleaner/cleaning_rules/null_invalid_foreign_keys.py
```diff
@@ -21,84 +29,215 @@
...
+LEFT_JOIN = JINJA_ENV.from_string("""
+    LEFT JOIN `{{dataset_id}}.{{table}}` {{prefix}} 
+    ON t.{{field}} = {{prefix}}.{{field}}""")
 
+SANDBOX_QUERY = JINJA_ENV.from_string("""
+    CREATE OR REPLACE TABLE
+    `{{project_id}}.{{sandbox_dataset_id}}.{{intermediary_table}}` AS (
+    SELECT {{cols}}
+    FROM `{{project_id}}.{{dataset_id}}.{{table_name}}` t
+    {{join_exp}})""")
```
ChaoPang, line 46: `SANDBOX_QUERY` is doing the same thing as `INVALID_FOREIGN_KEY_QUERY`. I think this cleaning rule is designed to update the records whose foreign keys are invalid, so we need a slightly different query to identify such records for sandboxing.

data_steward/cdr_cleaner/cleaning_rules/null_invalid_foreign_keys.py
```diff
@@ -21,84 +29,215 @@
...
+        no changes to the column name.
+
+        :param table: single table in the list of affected tables
+        :return: all the formatted column names for the table
+        """
+        col_exprs = []
+        foreign_keys = self.get_foreign_keys(table)
+        for field in self.get_field_names(table):
+            if field in foreign_keys:
+                col_expr = '{x}.'.format(x=field[:3]) + field
```
ChaoPang, line 136:
```suggestion
                col_expr = f'{field[:3]}.field'
```

data_steward/cdr_cleaner/cleaning_rules/null_invalid_foreign_keys.py
```diff
@@ -21,84 +29,215 @@
...
+        """
+        This method generates the LEFT_JOIN query. Only columns that are foreign keys will be
+        used in the query generation.
+
+        :param table: single table in the list of affected tables
+        :return: LEFT_JOIN query expression
+        """
+        join_expression = []
+        for key in FOREIGN_KEYS_FIELDS:
+            if key in self.get_foreign_keys(table):
```
ChaoPang, line 152: It might be more efficient to iterate through `self.get_foreign_keys(table)` and check if the key is in `FOREIGN_KEYS_FIELDS`, the CDM table doesn't usually contain all key fields defined in `FOREIGN_KEYS_FIELDS`, so the number of iterations would be smaller.

data_steward/cdr_cleaner/cleaning_rules/null_invalid_foreign_keys.py
```diff
@@ -21,84 +29,215 @@
...
-                dataset_id=dataset_id,
-                project=project_id,
-                join_expr=full_join_expression)
-            query[cdr_consts.DESTINATION_TABLE] = table
-            query[cdr_consts.DISPOSITION] = bq_consts.WRITE_TRUNCATE
-            query[cdr_consts.DESTINATION_DATASET] = dataset_id
-            queries_list.append(query)
-    return queries_list
+                    table_alias = self.get_mapping_table(
+                        '{x}'.format(x=key)[:-3])
```
ChaoPang, line 157: I don't really know why `mapping_table` is used here for validating the foreign keys, I think there might be tacit knowledge I don't know about, @ksdkalluri perhaps you could better comment on this?

ChaoPang (changes requested):
I like the new structure of the code and it's so much easier to read and test now! Thanks for making the effort to refactor the content. I left some new comments, I looked at the content of the code a bit more and have some questions there, I also mentioned @ksdkalluri in one of the questions I had regarding how the query is constructed.

data_steward/cdr_cleaner/cleaning_rules/null_invalid_foreign_keys.py
```diff
@@ -21,84 +29,215 @@
...
+                        join_expr=join_expression),
+                cdr_consts.DESTINATION_DATASET:
+                    self.dataset_id,
+                cdr_consts.DESTINATION_TABLE:
+                    table,
+                cdr_consts.DISPOSITION:
+                    bq_consts.WRITE_TRUNCATE
+            }
+
+            sandbox_queries_list.append(sandbox_query)
```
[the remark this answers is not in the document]
dcampbell-vumc: Ah, I see the issue here!

data_steward/cdr_cleaner/cleaning_rules/null_invalid_foreign_keys.py
dcampbell-vumc: @ChaoPang - yeah, this didn't seem right to me but it would be hard for me to test because this ticket doesn't include creating the integration test. Should I try to fix this now or do so in the integration test? @ksdkalluri what do you think I should do here? Fix now or fix during the creation of the integration test?

data_steward/cdr_cleaner/cleaning_rules/null_invalid_foreign_keys.py
dcampbell-vumc: This isn't working as listed, changed to this, `col_expr = f'{field[:3]}.{field}'`, so that `field` is a variable, not printed string. This was resulting in the `col_expr` to print columns as 'loc.field' instead of 'loc.location_id'

data_steward/cdr_cleaner/cleaning_rules/null_invalid_foreign_keys.py
dcampbell-vumc: I'm not sure why it's there either, I don't have the historical knowledge of it so I'm hoping @ksdkalluri will know why it's here.

ksdkalluri: we use mapping tables to join on to the domain tables since we only need is the domain_ids(primary_keys) for each table which are available in _mapping_tables and its less expensive to join onto the mapping tables than to join onto the domain tables.

data_steward/cdr_cleaner/cleaning_rules/null_invalid_foreign_keys.py
ksdkalluri: Sandboxing query should be updated to get only the rows that are null when joined using a foreign key.

for example, we can use a similar query to this.

```
select m.* from measurement as m 
left join _mapping_visit_occurrence as mv using(visit_occurrence_id) 
where mv.visit_occurrence_id is NULL

```
to sandbox invalid visit_occurrence foreign keys in the measurement table.

ksdkalluri (changes requested):
please let me know if you have any other questions.

data_steward/cdr_cleaner/cleaning_rules/null_invalid_foreign_keys.py
```diff
@@ -39,11 +39,11 @@
...
 SANDBOX_QUERY = JINJA_ENV.from_string("""
-    CREATE OR REPLACE TABLE
+    CREATE OR REPLACE TABLE 
     `{{project_id}}.{{sandbox_dataset_id}}.{{intermediary_table}}` AS (
-    SELECT {{cols}}
-    FROM `{{project_id}}.{{dataset_id}}.{{table_name}}` t
-    {{join_exp}})""")
+        SELECT t.* FROM `{{project_id}}.{{dataset_id}}.{{table_name}}` AS t 
+        LEFT JOIN {{dataset_id}}.{{table}} AS tc USING({{cols}}) 
+        WHERE tc.{{cols}} IS NULL)""")
```
ksdkalluri, line 46: @dcampbell-vumc this query is will not work since some tables have multiple foreign keys that need to be joined to different tables. so for each table, the join expression will be different. and {{cols}} is a list of all the columns for any given table.  So the join won't work. Rather you should be updating the join expression that is being generated for each table to filter the invalid keys.

ksdkalluri (changes requested):

data_steward/cdr_cleaner/cleaning_rules/null_invalid_foreign_keys.py
```diff
@@ -21,84 +29,257 @@
...
+                    cdr_consts.DISPOSITION:
+                        bq_consts.WRITE_TRUNCATE
+                }
+
+                sandbox_queries_list.append(sandbox_query)
+
+                invalid_foreign_key_query = {
+                    cdr_consts.QUERY:
+                        INVALID_FOREIGN_KEY_QUERY.render(
+                            cols=cols,
```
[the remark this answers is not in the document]
dcampbell-vumc: I'm a bit confused at this. On line 205, `cols` currently is defined as `cols = self.get_col_expression(table)` and the `get_cols_expression` returns `return ', '.join(col_expres)`. Isn't that the same thing you're suggesting or is this different?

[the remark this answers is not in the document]
ksdkalluri: You are right. The get_cols_expression is returning a string. The issue is in the test.
in `test_get_query_specs()` these utility `mock_get_col_expression` is being mocked for which the return value is list. If you just update the `self.person_col_expression` to the string expression that should fix this.

ksdkalluri (approved):
This looks good! I would suggest running the cleaning rule to log all the queries that will be generated by the CR and validate them in the BQ console before merging.

ChaoPang (approved):
I approve!

PR #674, opened 2020-11-18
[DC-1187] Add RDR evaluation queries to curation notebook

mark-velez:


data_steward/analytics/cdr_ops/rdr_export_qc.py
```diff
@@ -0,0 +1,316 @@
...
+v_dict['june'] = set(version[version['cope_month'] == 'june']
+                     ['questionnaire_response_id'].unique())
+v_dict['july'] = set(version[version['cope_month'] == 'july']
+                     ['questionnaire_response_id'].unique())
+
+# +
+from itertools import combinations
+
+for d1, d2 in combinations(v_dict.keys(), 2):
+    print(f"{d1}-{d2}: {len(v_dict[d1].intersection(v_dict[d2]))}")
```
mark-velez, line 302:
```suggestion
query = f"""
SELECT
    questionnaire_response_id,
    COUNT(*) n
FROM `{cope_version_map}` 
GROUP BY questionnaire_response_id
HAVING n > 1
"""
pd.read_gbq(query, project_id=project_id, dialect='standard')
```
@rfrancis1 would it suffice to check for duplicate `questionnaire_response_id`? The only way there would be overlap between versions is if it repeated and as far as I know multiple `cope_version_map` rows should never have the same `questionnaire_response_id`.

data_steward/analytics/cdr_ops/rdr_export_qc.py
```diff
@@ -0,0 +1,325 @@
...
+    Extract params from the query string of a URL
+    
+    :returns: dict where the keys are parameter names
+    """
+    split_url = urllib.parse.urlsplit(url)
+    params = urllib.parse.parse_qs(split_url.query)
+    for key, value in params.items():
+        if len(value) == 1:
+            params[key] = value[0]
+    return params
```
lrwb-aou, line 70: Investigate using a `parameter` mark in the cell to define parameters.

lrwb-aou (commented):
We discussed most of this in our call earlier.  Just wondering about your thoughts on adding the `project_id` to the query string vs providing it as an optional argument to `read_gbq`.

data_steward/analytics/cdr_ops/rdr_export_qc.py
rfrancis1: That might work, Mark.

data_steward/analytics/cdr_ops/rdr_export_qc.py
```diff
@@ -0,0 +1,325 @@
...
+
+# -
+
+params = get_url_params(NOTEBOOK_URL)
+validate_params(params, REQUIRED_PARAMS)
+
+project_id = params['project_id']
+old_rdr = params['old_rdr']
+new_rdr = params['new_rdr']
+cope_version_map = params['cope_version_map']
```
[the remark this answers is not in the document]
mark-velez: Thanks @lrwb-aou I just created this cell for the convenience of having individual variables and I'm not thrilled with it. But keying off the dict in all the queries like below looks icky.
```python
query = f'SELECT ... {params["old_rdr"]}'
```
Another way might be for get_url_params() to return a (project_id, old_rdr, new_rdr, version_map) tuple. What do you think? I'm open to other suggestions too.

data_steward/analytics/cdr_ops/rdr_export_qc.py
```diff
@@ -0,0 +1,325 @@
...
+v_dict = {}
+v_dict['may'] = set(version[version['cope_month'] == 'may']
+                    ['questionnaire_response_id'].unique())
+v_dict['june'] = set(version[version['cope_month'] == 'june']
+                     ['questionnaire_response_id'].unique())
+v_dict['july'] = set(version[version['cope_month'] == 'july']
+                     ['questionnaire_response_id'].unique())
+
+# +
+from itertools import combinations
```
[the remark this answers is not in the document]
mark-velez: All the imports should be at the top now that the cell is removed.

data_steward/analytics/cdr_ops/rdr_export_qc.py
```diff
@@ -0,0 +1,325 @@
...
+WHERE prev_code.value IS NULL OR curr_code.value IS NULL
+;
+''');
+DECLARE query STRING DEFAULT(REPLACE(query_template, '${CURR}', @new_rdr));
+SET query = REPLACE(query, '${PREV}', @old_rdr);
+EXECUTE IMMEDIATE query;
+
+# # Check if observation_source_value vs concept ids
+
+query = """
```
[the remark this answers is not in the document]
mark-velez: I have updated the notebook to use f-strings where possible

data_steward/analytics/cdr_ops/rdr_export_qc.py
```diff
@@ -0,0 +1,325 @@
...
+SELECT 
+    SUM(CASE WHEN observation_source_concept_id IS NULL THEN 1 ELSE 0 END) as n_null_source_concept_id
+    , SUM(CASE WHEN observation_source_concept_id=0 THEN 1 ELSE 0 END) as n_zero_source_concept_id
+FROM `{dataset}.observation`
+WHERE observation_source_value IS NOT NULL
+and observation_source_value != ''
+"""
+pd.read_gbq(query.format(dataset=new_rdr),
+            project_id=project_id,
+            dialect='standard')
```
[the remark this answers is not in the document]
mark-velez: @lrwb-aou I agree- it is safer to use fully qualified table names in queries wherever possible. I will make that change. Do you think we should keep the `project_id` arg in calls calls to `read_gbq()`?

data_steward/analytics/cdr_ops/rdr_export_qc.py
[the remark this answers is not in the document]
lrwb-aou: For the purposes of this PR, I withdraw my comment about putting this in a function.  I personally like this, even if it feels a bit wordy.
```
query = f'SELECT ... {params.get("old_rdr")}'
```
Then I'm more likely to know where that value came from.  I prefer to use the `get()` function for convention sake.  Just to make sure the dictionary value is safely accessed.

Or see the code suggestion above about setting these values in the validation function.

data_steward/analytics/cdr_ops/rdr_export_qc.py
```diff
@@ -0,0 +1,307 @@
...
+
+def validate_params(params, required_params):
+    """
+    Raise an error if a required parameter is missing from the params dict
+    """
+    for required_param in required_params:
+        if required_param not in params:
+            raise ValueError(f'Missing query parameter `{required_param}`')
+    print(f'All required parameters are specified.\n{params}')
+
```
lrwb-aou, line 81: You could add the following to this to set the global values after validating them.
```suggest
    global project_id, old_rdr, new_rdr, cope_version_map
    project_id = params.get('project_id')
    old_rdr = params.get('old_rdr')
    new_rdr = params.get('new_rdr')
    cope_version_map = params.get('cope_version_map')
```

data_steward/analytics/cdr_ops/rdr_export_qc.py
```diff
@@ -0,0 +1,307 @@
...
+ curr.row_count AS {new_rdr}, 
+ prev.row_count AS {old_rdr}, 
+ (curr.row_count - prev.row_count) row_diff 
+FROM {new_rdr}.__TABLES__ curr 
+FULL OUTER JOIN {old_rdr}.__TABLES__ prev
+  USING (table_id)
+WHERE curr.table_id IS NULL OR prev.table_id IS NULL
+'''
+pd.read_gbq(query, project_id=project_id, dialect='standard')
+
```
lrwb-aou, line 107: Adding comments to tell me what I should expect as the result of running these queries would be helpful.  Right now, I wouldn't necessarily know how to interpret the results of any of these queries.

lrwb-aou (commented):
Would like to discuss a few comments before checking this in tomorrow.

data_steward/analytics/cdr_ops/rdr_export_qc.py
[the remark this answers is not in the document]
mark-velez: The `project_id` parameter confuses me too- the pandas-gbq docs could be more descriptive. I assume that in this case it is just used to resolve object names which are not fully qualified within queries, and, if the parameter is not supplied then it is obtained from "default credentials" as described by google. If this is correct, I think I would prefer that all objects be fully qualified in queries and rely on neither of these fallbacks (i.e. throw an error for any objects that are not fully qualified). One reason is reliance on the fallback can have ambiguous/unintended consequences if queries reference objects in multiple projects. Yes, let's discuss in the morning.

data_steward/analytics/cdr_ops/rdr_export_qc.py
```diff
@@ -0,0 +1,308 @@
...
+
+def validate_params(params, required_params):
+    """
+    Raise an error if a required parameter is missing from the params dict
+    """
+    for required_param in required_params:
+        if required_param not in params:
+            raise ValueError(f'Missing query parameter `{required_param}`')
+    global project_id, old_rdr, new_rdr, cope_version_map
```
lrwb-aou, line 77: For some reason, `cope_version_map` is still here.

data_steward/analytics/cdr_ops/rdr_export_qc.py
```diff
@@ -0,0 +1,308 @@
...
+# # Missing tables
+
+query = f'''
+SELECT 
+ COALESCE(curr.table_id, prev.table_id) AS table_id, 
+ curr.row_count AS {new_rdr}, 
+ prev.row_count AS {old_rdr}, 
+ (curr.row_count - prev.row_count) row_diff 
+FROM {project_id}.{new_rdr}.__TABLES__ curr 
+FULL OUTER JOIN {old_rdr}.__TABLES__ prev
```
lrwb-aou, line 98: nit:  Should include the `{project_id}.` here.

data_steward/analytics/cdr_ops/rdr_export_qc.py
```diff
@@ -0,0 +1,308 @@
...
+WHERE prev_code.value IS NULL OR curr_code.value IS NULL
+'''
+pd.read_gbq(query, dialect='standard')
+
+# # Check if observation_source_value vs concept ids
+
+query = f"""
+SELECT 
+    SUM(CASE WHEN observation_source_concept_id IS NULL THEN 1 ELSE 0 END) as n_null_source_concept_id
+    , SUM(CASE WHEN observation_source_concept_id=0 THEN 1 ELSE 0 END) as n_zero_source_concept_id
```
lrwb-aou, line 168: opt:  just a personal preference.  you can leave as is, if you want.  i normally expect the comma to come at the end of the line instead of the start of the next line.

data_steward/analytics/cdr_ops/rdr_export_qc.py
```diff
@@ -0,0 +1,308 @@
...
+query = f"""
+SELECT 
+    SUM(CASE WHEN observation_concept_id IS NULL THEN 1 ELSE 0 END) as n_null_concept_id
+    , SUM(CASE WHEN observation_concept_id=0 THEN 1 ELSE 0 END) as n_zero_concept_id
+FROM `{project_id}.{new_rdr}.observation`
+WHERE observation_source_value IS NOT NULL
+and observation_source_value != ''
+"""
+pd.read_gbq(query,
+            dialect='standard')
```
lrwb-aou, line 185: You could do one of two things here.
1. Just add these SUM/CASE statements to the previous query.
2. Parameterize the previous query and run it twice with different field values.

data_steward/analytics/cdr_ops/rdr_export_qc.py
```diff
@@ -0,0 +1,308 @@
...
+query = f"""
+SELECT 
+    SUM(CASE WHEN value_as_concept_id IS NULL THEN 1 ELSE 0 END) as n_null_concept_id
+    , SUM(CASE WHEN value_as_concept_id=0 THEN 1 ELSE 0 END) as n_zero_concept_id
+FROM `{project_id}.{new_rdr}.observation`
+WHERE value_source_value IS NOT NULL
+and value_source_value != ''
+"""
+pd.read_gbq(query,
+            dialect='standard')
```
lrwb-aou, line 211: You could do one of two things here.
1. Just add these SUM/CASE statements to the previous query.
2. Parameterize the previous query and run it twice with different field values.

data_steward/analytics/cdr_ops/rdr_export_qc.py
```diff
@@ -0,0 +1,308 @@
...
+"""
+pd.read_gbq(query, dialect='standard')
+
+# # Survey version and dates
+
+query = f"""
+SELECT
+    cope_month
+    , MIN(observation_date) AS min_date
+    , MAX(observation_date) AS max_date
```
lrwb-aou, line 303: opt:  put commas as the end of the previous line.

lrwb-aou (commented):
Most of these comments are optional.  Aside from adding the project_id to a query.

data_steward/analytics/cdr_ops/rdr_export_qc.py
mark-velez: Thanks- dunno how that came back

data_steward/analytics/cdr_ops/rdr_export_qc.py
mark-velez: This is a style preference I'm somewhat conflicted about. In a previous life my team would **always** structure queries like this
```sql
SELECT
  col1
 ,col2
 ,col3
FROM `project.dataset.table`
WHERE 1=1    -- wat! yes, this conveniently ensures each condition in the clause is prefixed with "AND ..."
 AND col1 < col3
 AND col2 > col4
```
because it makes it easier to remove items from the selection and the where clause. These are things one typically does when iterating on a query or troubleshooting data issues.

I might be willing to switch all notebook queries to this style- or allow comma to be in _either_ the end or beginning of the line. Thoughts?

The following style is rarely desirable and I think we should discourage it:
```
SELECT col1, col2, col3 FROM `project.dataset.table` WHERE col1 < col3 AND col2 > col4
```

data_steward/analytics/cdr_ops/rdr_export_qc.py
mark-velez: Good idea- I did 1. Also I added the observation_source_value so we know which codes have the problem.

data_steward/analytics/cdr_ops/rdr_export_qc.py
mark-velez: Good idea- I did 1. Also I added the value_source_value so we know which codes have the problem.

data_steward/analytics/cdr_ops/rdr_export_qc.py
mark-velez: Done

mark-velez:
Hey @lrwb-aou I documented the queries I understood. Some of them need to be revised. I would also consider re-ordering the analyses so that the "strong signals" are at the top and "likely noise" is at the bottom- but I think that can be done in the next iteration.

data_steward/analytics/cdr_ops/rdr_export_qc.py
lrwb-aou: If you want to leave the commas at the start of the line, I can support that.  I guess I'm nitpicking about alignment.  I would prefer to keep the alphabetic characters lined up like in your example query above.  So, maybe:
```
     SUM.......
    ,SUM.......
```
instead of:
```
    SUM.....
    , SUM.....
```

mark-velez: Ok I aligned the queries as you mentioned above

lrwb-aou:
> Hey @lrwb-aou I documented the queries I understood. Some of them need to be revised. I would also consider re-ordering the analyses so that the "strong signals" are at the top and "likely noise" is at the bottom- but I think that can be done in the next iteration.

I agree!

lrwb-aou (approved):
Thank you Mark!  I think this is a great starting point for RDR import evaluation.

mark-velez:
@rfrancis1 (do you know Jun's github name?) it might be useful to look at the discussion in this pull request. It should inform some of the conventions we want to follow in our notebooks moving forward.

I'm about to merge this to the `develop` branch. Afterwards please try checking out `develop` locally and running the notebook again and let me know any questions or issues.

rfrancis1:
@mark-velez I don't know if he has one, but I told him to create an account if not.

mark-velez:
Hi @jqian2015 fyi it might be useful to look at this pull request.

jqian2015:
@mark-velez thanks.

PR #675, opened 2020-11-19
RDR QC

rfrancis1:
- Add a QC to check versions and dates
- Add a QC to check answers put in the wrong place (e.g number in strings)


mark-velez:
Thanks @rfrancis1 ! This PR has been merged to `mv/rdr_nb`. Commit IDs will be different because I rebased them onto the target branch.

PR #676, opened 2020-12-01
[DC-1245] Fix failed builds caused by ReplaceWithStandardConceptIdTest

ChaoPang:


data_steward/cdr_cleaner/cleaning_rules/replace_standard_id_in_domain_tables.py
```diff
@@ -98,33 +103,37 @@
...
-        GROUP BY src_id
-        HAVING COUNT(*) > 1 ) b
-    ON 
-        a.src_id = b.src_id AND a.domain_table = '{{table_name}}'
-    CROSS JOIN (
-        SELECT
-            MAX({{table_name}}_id) AS max_id
-        FROM `{{project}}.{{dataset}}.{{table_name}}` 
-    ) src
+    {{unioned_query}}
```
mark-velez, line 136: I'm having trouble seeing what the full query ultimately ends up looking like. Have you considered using a single jinja template like this rather than filling out multiple template fragments and stitching them together?
```suggestion
    {%- for table_name in table_names %}
    SELECT
      a.src_id,
      a.domain_table,
      a.new_concept_id,
      ROW_NUMBER() OVER(ORDER BY a.src_id, a.new_concept_id) + src.max_id AS dest_id
    FROM
      `{{project}}.{{sandbox_dataset}}.{{logging_table}}` AS a
    JOIN (
        SELECT src_id
        FROM `{{project}}.{{sandbox_dataset}}.{{logging_table}}`
        WHERE domain_table = '{{table_name}}'
        GROUP BY src_id
        HAVING COUNT(*) > 1 ) b
    ON a.src_id = b.src_id AND a.domain_table = '{{table_name}}'
    CROSS JOIN (
      SELECT
        MAX({{table_name}}_id) AS max_id
      FROM `{{project}}.{{dataset}}.{{table_name}}`
    ) src
    {% if loop.nextitem is defined %}
    UNION ALL
    {% endif -%}
    {% endfor %}
```
The query would then be rendered like this:
```python
DUPLICATE_ID_UPDATE_QUERY.render(
            project=self.project_id,
            sandbox_dataset=self.sandbox_dataset_id,
            logging_table=SRC_CONCEPT_ID_TABLE_NAME,
            table_names=self.affected_tables)
```
Because there are fewer variables and functions to track it is easier to follow the query generation logic. Less code also means less testing. It's also easier to visualize the full query when it's in one place. Selective use of `-` in jinja blocks enables fine control over whitespace when balancing lean vs pretty SQL.

tests/unit_tests/data_steward/cdr_cleaner/cleaning_rules/replace_standard_id_in_domain_tables_test.py
```diff
@@ -290,11 +258,63 @@ def test_parse_src_concept_id_logging_query(
...
+            logging_table=SRC_CONCEPT_ID_TABLE_NAME,
+            unioned_query=UNION_ALL.join([query_1, query_2]))
+        actual_query = self.rule_instance.parse_duplicate_id_update_query()
+        self.assertEqual(expected_query, actual_query)
+
+        self.assertEqual(mock_parse_duplicate_id_update_subquery.call_count, 2)
+        mock_parse_duplicate_id_update_subquery.assert_any_call(
+            self.condition_table)
+        mock_parse_duplicate_id_update_subquery.assert_called_with(
+            self.procedure_table)
```
mark-velez, line 316: I recommend we remove these tests. At last eng sync we agreed we should eliminate SQL generation stuff from unit tests. The most valuable way to test our SQL will be through integration tests. Sticking to a test-per-method does not usually do us any good and these kinds of tests tend to be circular and change-prone. Your integration tests look good to me. That you were able to implement this optimization without modifying them reinforces that.

mark-velez (changes requested):
This looks correct, but it makes my brain hurt- for each of the modified queries (`DUPLICATE_ID_UPDATE_QUERY` and `SRC_CONCEPT_ID_MAPPING_QUERY`) could you please try to keep generation logic in a single jinja template? See my example below and let me know what you think.

data_steward/cdr_cleaner/cleaning_rules/replace_standard_id_in_domain_tables.py
ChaoPang: I have not considered it but this is certainly much cooler, I will keep that in mind for the future query generation.

ChaoPang: Okay thanks for this suggestion, I haven't used jinja much in the past now I feel like I have another option for quickly generating queries, very cool. In the beginning, I thought it would be tricky to write a single jinja template (SRC_CONCEPT_ID_MAPPING_QUERY) for generating the UNION query for `_logging_standard_concept_id_replacement` because for each table in the loop I had to call two helper functions to get the {domain}_concept_id and {domain}_source_concept_id, and these two variables were also passed into the jinja template. Then it hit me, functions are just objects like variables so I could just pass the helper functions to the jinja template and run it there. This is a good lesson for me, appreciate it.

tests/unit_tests/data_steward/cdr_cleaner/cleaning_rules/replace_standard_id_in_domain_tables_test.py
ChaoPang: This is done

lrwb-aou (approved):
@ChaoPang I've got a small request.  I ran ` PYTHONPATH=./:$PYTHONPATH python cdr_cleaner/reporter.py -d combined -c -f sql jira-issues description name` to create a csv file with the SQL statements.  I see there are a total of 24 statements created by the rule.  The longest is 272 lines, but I think that is well below the query length.  Please update the list of Jira Issues.  While the SQL should have the same effect, it is an SQL change.

ChaoPang:
@lrwb-aou just to make sure I understand your comment completely, I should add DC-1245 to the issue_numbers in the constructor because this is a SQL change and we need to keep track of it, correct?

lrwb-aou:
> @lrwb-aou just to make sure I understand your comment completely, I should add DC-1245 to the issue_numbers in the constructor because this is a SQL change and we need to keep track of it, correct?

Yes.  I am asking to add DC-1245 to the list of issue_numbers in the constructor because this is an SQL change.

mark-velez (approved):

PR #677, opened 2020-12-01
[DC-1169] Integration Test for Invalid Foreign Keys CR

dcampbell-vumc:
* modified main module and unit test because cannot set python code destination object when running the `CREATE OR REPLACE` statement in the `SANDBOX_QUERY` queries
* created integration test to test functionality of main module
* in the integration test only the `person` and `procedure_occurrence` tables are cleaned as between those two all foreign key columns will be tested
* only the mapping tables used in the `INVALID_FOREIGN_KEY_QUERY` for `person` and `procedure_occurrence` are created and supplied with data
* the `get_affected_tables` function from main module is mocked as we do not need all the CDM tables to be iterate through for cleaning, only `person` and `procedure_occurrence`

### DIFF
--- a/tests/unit_tests/data_steward/cdr_cleaner/cleaning_rules/null_invalid_foreign_keys_test.py
+++ b/tests/unit_tests/data_steward/cdr_cleaner/cleaning_rules/null_invalid_foreign_keys_test.py
@@ -178,13 +178,12 @@ class NullInvalidForeignKeys(unittest.TestCase):
 
         sandbox_query = {
             cdr_consts.QUERY:
-                nifk.SANDBOX_QUERY.render(
-                    project_id=self.project_id,
-                    sandbox_dataset_id=self.sandbox_id,
-                    intermediary_table='dc388_person',
-                    dataset_id=self.dataset_id,
-                    table_name=table,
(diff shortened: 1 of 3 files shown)


tests/unit_tests/data_steward/cdr_cleaner/cleaning_rules/null_invalid_foreign_keys_test.py
```diff
@@ -179,16 +179,10 @@ def test_get_query_specs(self, mock_get_affected_tables,
                 nifk.SANDBOX_QUERY.render(
                     project_id=self.project_id,
                     sandbox_dataset_id=self.sandbox_id,
-                    intermediary_table='dc-388_dc-807_person',
+                    intermediary_table='dc388_dc807_person',
```
ksdkalluri, line 182: We can just use the intermediary table with one ticket number. having 2 ticket numbers in the table name will be confusing.

ksdkalluri (changes requested):
Other than the intermediary table name change everything looks good.

ksdkalluri (approved):
Looks Good, Thanks!

PR #678, opened 2020-12-01
[DC-1020] Add script to run tests in GCE VM to curation repo

lrwb-aou:
* adds a developement requirement for python 3.6 needs
* sends test output to stdout isntead of stderr to prevent intermingling of stdout and stderr in the log file
* creates an automation script.  Script should be runnable from a cron job.

### DIFF
--- a/data_steward/dev_requirements.txt
+++ b/data_steward/dev_requirements.txt
@@ -1,7 +1,6 @@
 coverage==4.5.2
 pylint==2.4.4
 modernize==0.7
-mo-future==2.48.19205
 unittest-xml-reporting>=2.5.2,==2.*
 yapf==0.29.0
 lxml==4.5.0
@@ -9,4 +8,3 @@ beautifulsoup4==4.8.2
 soupsieve==1.9.5
 bs4==0.0.1
 pipdeptree==1.0.0
-wheel
--- a/data_steward/dev_requirements.txt
+++ b/data_steward/dev_requirements.txt
@@ -1,6 +1,7 @@
 coverage==4.5.2
 pylint==2.4.4
 modernize==0.7
+mo-future==2.48.19205
 unittest-xml-reporting>=2.5.2,==2.*
 yapf==0.29.0
 lxml==4.5.0
@@ -8,3 +9,4 @@ beautifulsoup4==4.8.2
 soupsieve==1.9.5
 bs4==0.0.1
 pipdeptree==1.0.0
+wheel
--- a/tests/runner.py
+++ b/tests/runner.py
@@ -46,7 +46,7 @@ def print_unsuccessful(function, trace, msg_type):
     print(
         '\n======================================================================'
     )
-    print('{}:  {}'.format(msg_type, function))
+    print(f'{msg_type}:  {function}')
     print(
         '----------------------------------------------------------------------'
     )
@@ -69,7 +69,9 @@ def main(test_path, test_pattern, coverage_filepath):
     start_time = time.time()
     for mod_tests in suite:
         if mod_tests.countTestCases():
-            runner = xmlrunner.XMLTestRunner(output=output_file, verbosity=2)
+            runner = xmlrunner.XMLTestRunner(stream=sys.stdout,
+                                             output=output_file,
+                                             verbosity=2)
             result = runner.run(mod_tests)
             all_results.append(result)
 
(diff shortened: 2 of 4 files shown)


tests/runner.py
```diff
@@ -69,7 +69,9 @@ def main(test_path, test_pattern, coverage_filepath):
     start_time = time.time()
     for mod_tests in suite:
         if mod_tests.countTestCases():
-            runner = xmlrunner.XMLTestRunner(output=output_file, verbosity=2)
+            runner = xmlrunner.XMLTestRunner(stream=sys.stdout,
```
mark-velez, line 72: Is this all you had to do so that all test output would be in the same stream- or do we still need changes e.g. to `print()` in test classes?

data_steward/dev_requirements.txt
```diff
@@ -1,10 +1,12 @@
...
 modernize==0.7
+mo-future==2.48.19205
 unittest-xml-reporting>=2.5.2,==2.*
 yapf==0.29.0
 lxml==4.5.0
 beautifulsoup4==4.8.2
 soupsieve==1.9.5
 bs4==0.0.1
 pipdeptree==1.0.0
+wheel
```
mark-velez, line 12: I assume these changes were needed to support 3.6; could you provide more details as to why?

tests/runner.py
lrwb-aou: Yes.  I did more digging and found that `XMLTestRunner` extends `TextTestRunner`.  `TextTestRunner` sets a `stream` parameter to `stderr` if it isn't defined.  So I defined it and told it to send to `stdout` instead.  Irritated at myself that it took me that long to find this.

tools/automate_all_tests.sh
```diff
@@ -0,0 +1,99 @@
...
+
+if [[ -z "${BRANCH}" ]] || [[ -z "${KEY_FILE}" ]] || [[ -z "${ENV_PATH}" ]]; then
+  echo "${USAGE}"
+  exit 1
+fi
+
+ROOT_DIR=$(git rev-parse --show-toplevel)
+DATA_STEWARD_DIR="${ROOT_DIR}/data_steward"
+DEID_DIR="${DATA_STEWARD_DIR}/deid"
+export GH_USERNAME=$(whoami)
```
mark-velez, line 42: I _don't think_ this would ever be the GitHub username..?
```suggestion
export CURRENT_USER=$(whoami)
```

data_steward/dev_requirements.txt
lrwb-aou: `mo-future` was referenced in the notes [here](https://docs.google.com/document/d/1wb8PRcqjDsPV_Q0WssCsSQpmazCVk-mzqgiI4DyP0c0/edit#heading=h.ooz2v8wi2mun).  Since we aren't using the slimmed down requirements files, I am trying to run the integration tests without these two additions.  The unit tests ran fine.  I'll wait on the integration tests to finish before removing from the PR.

tools/automate_all_tests.sh
lrwb-aou: You're right.  `GH_USERNAME` had to be set to run `curation/data_steward/init.sh` correctly.  It requires a `GH_USERNAME` or `CIRCLE_USERNAME`.

tests/runner.py
mark-velez: You kidding? This is great! :+1:

tools/automate_all_tests.sh
mark-velez: Ah, this is ok then

lrwb-aou: I'll add a comment in here to explain why this was set this way.  This script should primarily be used for automating on the VM.  I strongly discourage anyone use this script to run the tests on their own.

tools/automate_all_tests.sh
```diff
@@ -0,0 +1,99 @@
...
+gcloud config set project aou-res-curation-test
+gcloud config list
+echo $GOOGLE_APPLICATION_CREDENTIALS
+
+# environment, bucket, and dataset setup
+source "${DATA_STEWARD_DIR}/init_env.sh"
+
+# have to reset here because init_env.sh overrides the previous setting
+export GOOGLE_APPLICATION_CREDENTIALS="${KEY_FILE}"
+PYTHONPATH=$PYTHONPATH python "${DATA_STEWARD_DIR}/ci/setup.py"
```
mark-velez, line 91: Should this be 
```suggestion
PYTHONPATH=$PYTHONPATH python "${DATA_STEWARD_DIR}/ci/test_setup.py"
```
?

mark-velez (approved):
I can't wait to have this cron'ed! Then we can at least make sure our build will support our gce environment. @lrwb-aou thanks for the tough work on this.

PR #681, opened 2020-12-04
[DC-1100] Create get_date_columns utility for deactivation retraction

nishanthpp93:
* [DC-1100] Update query and fix tests 
* [DC-1100] Fix tests in related cleaning rule

[DC-1100]: https://precisionmedicineinitiative.atlassian.net/browse/DC-1100
[DC-1100]: https://precisionmedicineinitiative.atlassian.net/browse/DC-1100

### DIFF
--- a/data_steward/retraction/retract_deactivated_pids.py
+++ b/data_steward/retraction/retract_deactivated_pids.py
@@ -1,15 +1,14 @@
 # Python imports
 import argparse
 import logging
-import os
 import re
-from datetime import datetime
 
 # Third party imports
 import pandas as pd
 from google.cloud import bigquery
 
 # Project imports
+from utils import bq, pipeline_logging
 from retraction.retract_utils import DEID_REGEX
 from sandbox import check_and_create_sandbox_dataset
 from constants import bq_utils as bq_consts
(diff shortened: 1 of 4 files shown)


data_steward/retraction/retract_deactivated_pids.py
```diff
@@ -52,6 +33,37 @@ def get_client(project_id):
 WHERE person_id = {{pid}}
 """)
 
+# Queries to create tables in associated sandbox with rows that will be removed per cleaning rule
```
ksdkalluri, line 36: These queries are not being used yet, these should belong in a PR where all the queries are being updated.

ksdkalluri (changes requested):

data_steward/retraction/retract_deactivated_pids.py
nishanthpp93: I need them for the next ticket, like the util functions. I think they can be moved to a new PR for the main ticket DC-963. What are your thoughts?

ksdkalluri: yes, moving this query to that PR will help, and also I think the query will be changed when you are working on it.

ksdkalluri (approved):
Looks good!

PR #688, opened 2020-12-14
[DC-1213] Use the Participant Summary API to retrieve fields of interest for participant validation

dcampbell-vumc:
* Created function `get_site_participant_information` which retrieves all the participants information using the columns listed in `FIELDS_OF_INTEREST_FOR_VALIDATION` 
** uses pagination and count to significantly speed up the time it takes to return records. Extremely helpful when using to get results from sites with a lot of participants as this no longer takes upwards of 10 minutes. Takes about 2 to retrieve 30k+ records.
* Created test function `test_get_site_participant_information`

### DIFF
--- a/data_steward/utils/participant_summary_requests.py
+++ b/data_steward/utils/participant_summary_requests.py
@@ -26,13 +26,15 @@ from google.auth import default
 from utils import auth
 from resources import fields_for
 
-# These fields are coming in from RDR with their naming convention and will be converted
-# to the Curation naming convention in the `get_site_participant_information` function
 FIELDS_OF_INTEREST_FOR_VALIDATION = [
     'participantId', 'firstName', 'middleName', 'lastName', 'streetAddress',
     'streetAddress2', 'city', 'state', 'zipCode', 'phoneNumber', 'email',
     'dateOfBirth', 'sex'
 ]
+"""
+These fields are coming in from RDR with their naming convention and will be converted
+to the Curation naming convention in the `get_site_participant_information` function
+"""
 
 
 def get_access_token():
--- a/data_steward/utils/participant_summary_requests.py
+++ b/data_steward/utils/participant_summary_requests.py
@@ -16,7 +16,6 @@ The intent of the get_participant_information function is to retrieve the inform
 
 # Third party imports
 import re
-import signal
 import pandas
 import requests
 import pandas_gbq
@@ -82,7 +81,6 @@ def get_participant_data(url, headers):
                 link_url = link_obj[0].get('url')
                 next_url = original_url + '&' + link_url[link_url.find('_token'
                                                                       ):]
-                print(next_url)
             else:
                 done = True
 
@@ -278,10 +276,3 @@ def store_participant_data(df, project_id, destination_table):
                              project_id,
                              if_exists="replace",
                              table_schema=table_schema)
-
-
-if __name__ == '__main__':
-    # columns = ['participantId', 'suspensionStatus', 'suspensionTime']
-    # tablename = '_deactivated_participants'
-    # get_deactivated_participants('all-of-us-rdr-prod', 'cdr-20201008-20201201', tablename, columns)
-    get_site_participant_information('all-of-us-rdr-prod', 'PITT')
--- a/data_steward/utils/participant_summary_requests.py
+++ b/data_steward/utils/participant_summary_requests.py
@@ -10,7 +10,7 @@ The intent of the get_participant_information function is to retrieve the inform
     validation for a single site based on the site name (hpo_id). The information necessary for this request as defined
     in the ticket (DC-1213) as well as the Participant Summary Field List
     (https://all-of-us-raw-data-repository.readthedocs.io/en/latest/api_workflows/field_reference/participant_summary_field_list.html)
-    are `participant_id`, `firstName`, `middleName`, `lastName`, `streetAddress`, `streetAddress2`, `city`, `state`,
+    are `participantId`, `firstName`, `middleName`, `lastName`, `streetAddress`, `streetAddress2`, `city`, `state`,
     `zipCode`, `phoneNumber`, `email`, `dateOfBirth`, `sex`
 """
 
@@ -176,7 +176,7 @@ def get_site_participant_information(project_id, hpo_id):
     """
     Fetches the necessary participant information for a particular site.
 
-    :param project_id: The RDR project that contains participant summary data
+    :param project_id: The RDR project hosting the API
     :param hpo_id: awardee name of the site
 
     :return: returns dataframe of participant information
@@ -201,8 +201,13 @@ def get_site_participant_information(project_id, hpo_id):
     #   regardless if there is EHR data uploaded for that participant
     # suspensionStatus=NOT_SUSPENDED and withdrawalStatus=NOT_WITHDRAWN -- ensures only active participants returned
     #   via the API
-    url = 'https://{0}.appspot.com/rdr/v1/ParticipantSummary?awardee={1}&suspensionStatus={2}&consentForElectronicHealthRecords={3}&withdrawalStatus={4}&_sort=participantId&_count=1000'.format(
-        project_id, hpo_id, 'NOT_SUSPENDED', 'SUBMITTED', 'NOT_WITHDRAWN')
+    url = (f'https://{project_id}.appspot.com/rdr/v1/ParticipantSummary'
+           f'?awardee={hpo_id}'
+           f'&suspensionStatus=NOT_SUSPENDED'
+           f'&consentForElectronicHealthRecords=SUBMITTED'
+           f'&withdrawalStatus=NOT_WITHDRAWN'
+           f'&_sort=participantId'
+           f'&_count=1000')
 
     participant_data = get_participant_data(url, headers)
 
@@ -211,7 +216,8 @@ def get_site_participant_information(project_id, hpo_id):
 
     participant_information = []
 
-    # Loop over participant summary records, insert participant data in the same order as participant_information_cols
+    # Loop over participant summary records, insert participant data in
+    # the same order as participant_information_cols
     for entry in participant_data:
         item = []
         for col in participant_information_cols:
--- a/data_steward/utils/participant_summary_requests.py
+++ b/data_steward/utils/participant_summary_requests.py
@@ -178,14 +178,14 @@ def get_deactivated_participants(project_id, dataset_id, tablename, columns):
 
 def get_site_participant_information(project_id, hpo_id):
     """
-    Fetches the necessary participant information for a particular site. RuntimeErrors will occur
-        if parameter are not supplied or are incorrect and a Google TimeoutError will occur if the
-        API call takes longer than 10 minutes to return all the participants from the single site.
+    Fetches the necessary participant information for a particular site.
 
     :param project_id: The RDR project hosting the API
     :param hpo_id: awardee name of the site
 
-    :return: returns dataframe of participant information
+    :return: a dataframe of participant information
+    :raises: RuntimeError if the project_id and hpo_id are not strings
+    :raises: TimeoutError if response takes longer than 10 minutes
     """
     # Parameter checks
     if not isinstance(project_id, str):
--- a/data_steward/utils/participant_summary_requests.py
+++ b/data_steward/utils/participant_summary_requests.py
@@ -1,21 +1,67 @@
 """
 Utility to fetch and store deactivated participants information.
 
-The intent of this module is to call and store deactivated participants information by leveraging the RDR Participant
-    Summary API. Deactivated participant information stored is `participantId`, `suspensionStatus`, and `suspensionTime`
+Original Issues: DC-1213, DC-1042, DC-971, DC-972
+
+The intent of the get_deactivated_participants function is to call and store deactivated participants information by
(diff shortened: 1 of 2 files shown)


data_steward/utils/participant_summary_requests.py
```diff
@@ -41,7 +56,7 @@ def get_access_token():
 
 def get_participant_data(url, headers):
     """
-    Fetches participant data via ParticipantSummary API
+    Fetches participant data via ParticipantSummary API with a retrieval limit of 3 minutes
```
lrwb-aou, line 59: I'm not sure I'd add the bit about the 3 minutes to this comment.  There's nothing inherently limiting this to 3 minutes via the built in URLS.

data_steward/utils/participant_summary_requests.py
```diff
@@ -157,6 +172,73 @@ def get_deactivated_participants(project_id, dataset_id, tablename, columns):
...
+        raise RuntimeError(f'Please specify the RDR project')
+
+    if not isinstance(hpo_id, str):
+        raise RuntimeError(f'Please provide an hpo_id')
+
+    token = get_access_token()
+
+    headers = {
+        'content-type': 'application/json',
+        'Authorization': 'Bearer {0}'.format(token)
```
lrwb-aou, line 195: Could use an f-string for this formatting.
```suggestion
        'Authorization': f'Bearer {token}'
```

data_steward/utils/participant_summary_requests.py
```diff
@@ -157,6 +172,73 @@ def get_deactivated_participants(project_id, dataset_id, tablename, columns):
...
+    }
+
+    # Make request to get API version. This is the current RDR version for reference see
+    # see https://github.com/all-of-us/raw-data-repository/blob/master/opsdataAPI.md for documentation of this API.
+    url = 'https://{0}.appspot.com/rdr/v1/ParticipantSummary?awardee={1}&_sort=participantId&_count=1000'.format(
+        project_id, hpo_id)
```
lrwb-aou, line 201: you can break this into a two lines and using f-strings for readability like
```suggestion
    url = (f'https://{project_id}.appspot.com/rdr/v1/ParticipantSummary'
              # get all participants for a site 1000 at a time.  improves performance
              f'?awardee={hpo_id}&_sort=participantId&_count=1000')
```
This will let you insert comments into the string explaining why fields and values were chosen.

Also, does this successfully get all data for a site with more than 1000 participants?

lrwb-aou: Also, this looks like it is retrieving all data for each participant it is getting from the API.  Is there a way to use the API to only retrieve the fields we are interested in?  Not sure from the documentation yet.  This would reduce the performance time on the system too.

tests/unit_tests/data_steward/utils/participant_summary_requests_test.py
```diff
@@ -159,6 +172,36 @@ def test_get_deactivated_participants(self,
...
+    def test_get_site_participant_information(
+        self, mock_get_site_participant_information,
+        mock_store_participant_data):
+
+        # Pre conditions
+        mock_get_site_participant_information.return_value = self.fake_participant_info_dataframe
+
+        # Tests
+        dataframe_response = psr.get_site_participant_information(
+            self.project_id, self.fake_hpo)
```
lrwb-aou, line 187: Right here, you are storing a mocked object and using it to issue other calls.  If you are testing `get_site_participant_information` you:
1.  shouldn't mock the call to the function you are testing.  you should mock calls that function might make(e.g. to the Participant Summary API).
2. You should check the value of the dataframe it returns is what you expect.  You should not be testing the `store_participant_data` function again.  It should have already been tested.

Mocked objects will almost never fail an assertion, because they are mocks.

tests/unit_tests/data_steward/utils/participant_summary_requests_test.py
```diff
@@ -159,6 +172,36 @@ def test_get_deactivated_participants(self,
 
         self.assertEqual(expected_response, dataset_response)
 
+    @mock.patch('utils.participant_summary_requests.store_participant_data')
```
lrwb-aou, line 175: Not sure you need this at all.

tests/unit_tests/data_steward/utils/participant_summary_requests_test.py
```diff
@@ -159,6 +172,36 @@ def test_get_deactivated_participants(self,
...
+                                                        self.project_id,
+                                                        self.destination_table)
+
+        # Post conditions
+        pandas.testing.assert_frame_equal(
+            dataframe_response,
+            pandas.DataFrame(self.updated_site_participant_info,
+                             columns=psr.FIELDS_OF_INTEREST_FOR_VALIDATION))
+
+        self.assertEqual(expected_response, dataset_response)
```
lrwb-aou, line 203: Not sure you need this stuff either.

data_steward/utils/participant_summary_requests.py
```diff
@@ -157,6 +172,73 @@ def get_deactivated_participants(project_id, dataset_id, tablename, columns):
...
+
+    df = pandas.DataFrame(participant_information,
+                          columns=participant_information_cols)
+
+    # Transforms participantId to an integer string
+    df['participantId'] = df['participantId'].apply(participant_id_to_int)
+
+    # Rename columns to be consistent with the curation software
+    bq_columns = [
+        '_'.join(re.split('(?=[A-Z])', k)).lower()
```
lrwb-aou, line 227: This is a nice condensation of camel case to underscore separation.  👍

lrwb-aou (commented):
This is moving in the right direction.

data_steward/utils/participant_summary_requests.py
dcampbell-vumc: Forgot to remove that from when there was a time limit 🙈

data_steward/utils/participant_summary_requests.py
dcampbell-vumc: I will look into it. Only the columns we are interested in are inserted into the dataframe in the end but if I will see if we can limit the columns on the front end!

dcampbell-vumc: I will also double check that it's getting all the data and not limiting to 1000. At first I thought it would receive a chunk of 1000 participants per URL and then return them all in the end but now I'm not too sure so I will double check!

tests/unit_tests/data_steward/utils/participant_summary_requests_test.py
dcampbell-vumc: I just now realized that this is for the ticket where we store the dataframe so I jumped the gun here 😅

tests/unit_tests/data_steward/utils/participant_summary_requests_test.py
dcampbell-vumc: This is definitely for the ticket where we store the dataframe, I will remove it for now ☺️

data_steward/utils/participant_summary_requests.py
dcampbell-vumc: Added a counter and printed out the participants and made sure that all were being returned. 
Screen Shot 2020-12-15 at 9 42 05 AM

Still looking to see if we can limit to just retrieve info from the columns that we want!

lrwb-aou: Kenny Skaggs indicated the participant summary api does not currently let us specify the fields we are interested in.  It seems to be an all or nothing thing.

What does the above indicate?  Did it return 32751 records?

data_steward/utils/participant_summary_requests.py
```diff
@@ -157,6 +172,77 @@ def get_deactivated_participants(project_id, dataset_id, tablename, columns):
...
+    }
+
+    # Make request to get API version. This is the current RDR version for reference see
+    # see https://github.com/all-of-us/raw-data-repository/blob/master/opsdataAPI.md for documentation of this API.
+    # consentForElectronicHealthRecords=SUBMITTED -- ensures only consenting participants are returned via the API
+    #   regardless if there is EHR data uploaded for that participant
+    # suspensionStatus=NOT_SUSPENDED and withdrawalStatus=NOT_WITHDRAWN -- ensures only active participants returned
+    #   via the API
+    url = 'https://{0}.appspot.com/rdr/v1/ParticipantSummary?awardee={1}&suspensionStatus={2}&consentForElectronicHealthRecords={3}&withdrawalStatus={4}&_sort=participantId&_count=1000'.format(
+        project_id, hpo_id, 'NOT_SUSPENDED', 'SUBMITTED', 'NOT_WITHDRAWN')
```
lrwb-aou, line 205: This line is too long.  Please consider reformatting and using f-strings.
```suggestion
    url = (f'https://{project_id}.appspot.com/rdr/v1/ParticipantSummary'
           f'?awardee={hpo_id}'
           f'&suspensionStatus=NOT_SUSPENDED'
           f'&consentForElectronicHealthRecords=SUBMITTED'
           f'&withdrawalStatus=NOT_WITHDRAWN'
           f'&_sort=participantId'
           f'&_count=1000'
          )
```

data_steward/utils/participant_summary_requests.py
```diff
@@ -157,6 +172,77 @@ def get_deactivated_participants(project_id, dataset_id, tablename, columns):
...
+
+    participant_data = get_participant_data(url, headers)
+
+    # Columns of interest for participants of a desired site
+    participant_information_cols = FIELDS_OF_INTEREST_FOR_VALIDATION
+
+    participant_information = []
+
+    # Loop over participant summary records, insert participant data in the same order as participant_information_cols
```
lrwb-aou, line 214: This line is too long also, please break comment into multiple lines.

tests/unit_tests/data_steward/utils/participant_summary_requests_test.py
```diff
@@ -159,6 +196,32 @@ def test_get_deactivated_participants(self,
...
+        # Tests
+        expected_dataframe = pandas.DataFrame(
+            site_participant_information,
+            columns=psr.FIELDS_OF_INTEREST_FOR_VALIDATION)
+        actual_dataframe = pandas.DataFrame(
+            self.site_participant_information,
+            columns=psr.FIELDS_OF_INTEREST_FOR_VALIDATION)
+
+        # Post conditions
+        pandas.testing.assert_frame_equal(expected_dataframe, actual_dataframe)
```
lrwb-aou, line 223: This unit test is not testing your code right now.  It never executes the `get_site_participant_information` function.  It is comparing identical versions of the dataframe, so it always passes.  I've attached a patch file that runs the function and adds some print statements for debugging purposes.  

[unit_test_patch.txt](https://github.com/all-of-us/curation/files/5712553/unit_test_patch.txt)

lrwb-aou (changes requested):
Please address the conventions (line length).

Must fix the unit test.  If you would like to discuss, let me know.

lrwb-aou (approved):
This is good.  Thank you!  I've added some more suggestions.  This is a nicely written function and unit test.

data_steward/utils/participant_summary_requests.py
```diff
@@ -157,6 +172,83 @@ def get_deactivated_participants(project_id, dataset_id, tablename, columns):
...
 
 
+def get_site_participant_information(project_id, hpo_id):
+    """
+    Fetches the necessary participant information for a particular site.
+
+    :param project_id: The RDR project hosting the API
+    :param hpo_id: awardee name of the site
+
+    :return: returns dataframe of participant information
```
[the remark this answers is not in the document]
dcampbell-vumc: A timeout error, if a site gets so large that it takes longer than 10 minutes to retrieve results from?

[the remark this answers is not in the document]
lrwb-aou: Yes.  A timeout error.  But also that it will raise RuntimeErrors if the parameters are incorrect.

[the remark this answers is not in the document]
dcampbell-vumc: Gotcha! Thanks!

data_steward/utils/participant_summary_requests.py
```diff
@@ -1,22 +1,37 @@
...
+import requests
 import pandas_gbq
+import google.auth.transport.requests as req
+from google.auth import default
 
 # Project imports
 from utils import auth
 from resources import fields_for
 
+FIELDS_OF_INTEREST_FOR_VALIDATION = [
```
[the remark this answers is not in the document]
lrwb-aou: I was unclear.  Thank you for the comment.  I think @mark-velez indicated that a comment in this form after variable usage will make auto-documentation possible.  I think PyCharm encourages its use.

```
FIELDS_OF_INTEREST_FOR_VALIDATION = []
"""
These fields are coming in from RDR with their naming convention and will be converted
to the Curation naming convention in the `get_site_participant_information` function
"""
```

data_steward/utils/participant_summary_requests.py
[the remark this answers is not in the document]
lrwb-aou: I was unclear here too.  I meant it should follow a docstring style.  for example,
```suggestion
    :returns: a dataframe of participant information
    :raises: RuntimeErrors if the project_id and hpo_id are not strings
    :raises: TimeoutError if the response takes longer than 10 minutes.
```

lrwb-aou (approved):
Thank you for all your work!  It looks good!

PR #690, opened 2020-12-22
[DC-1247] Remove enum34 package from requirements

mark-velez:


ksdkalluri (approved):
Pretty quick one! Thanks for looking into this Mark.

PR #692, opened 2020-12-28
[DC-812] Update fill_source_value_free_text_fields CR to New Base Class

ksdkalluri:


mark-velez (changes requested):
@ksdkalluri I **very much** appreciate the careful way you approached refactoring this to use the base class. By keeping the working code as-is you are helping control scope.

My only blockers are minor, I promise:
1.  Naming things. To Lauren's point, "free text" _actually_ caused me to misinterpret what this CR was supposed to do. Could we change the name (e.g. FillSourceValue~Free~TextFields)?
2. Learn from my mistakes- avoid `import *` at all costs!

mark-velez (approved):
Thank you, @ksdkalluri

PR #695, opened 2020-12-29
[DC-956] Pylint Recursion Fix

jp3477:
This is a fix for a bug in pylint that is triggered by exceeding maximum recursion depth.


jp3477:
I set the recursion limit to 8000, but maybe it should be lower? Testing pylint on the following code snippet failed on any recursion depth below 1072, including the default 1000.

`import pandas as pd
pd.merge(None, None)`

ksdkalluri (approved):
We should also be looking for astroid's new release since it seems to fix the issue without increasing the recursion limit.

PR #697, opened 2021-01-05
Papermill test

rfrancis1:
@mark-velez I created a new branch off of it and sent a new PR


data_steward/analytics/cdr_ops/rdr_export_qc.py
```diff
@@ -1,91 +1,34 @@
+# ---
+# jupyter:
+#   jupytext:
+#     text_representation:
+#       extension: .py
+#       format_name: light
+#       format_version: '1.5'
+#       jupytext_version: 1.3.0
```
mark-velez, line 8: After I opened the notebook from the jupyter UI it automatically changes the jupytext version to 1.7.1. I think it should be the same version as what is in `requirements.txt`.
```suggestion
#       jupytext_version: 1.7.1
```
Note: Currently this yaml front-matter is required to be in each notebook. In the future (not now) I would like to find a way to configure this in just one place. That way if we e.g. bump jupytext versions, it would only have to be changed in one place.

data_steward/analytics/cdr_ops/run_rdr_qc_report.py
```diff
@@ -0,0 +1,68 @@
...
+
+NOTEBOOK_TEMPLATE_NAME = "rdr_qc_template.ipynb"
+EXECUTED_NOTEBOOK_NAME = "rdr_qc_export.ipynb"
+
+logging.basicConfig(
+    level = logging.INFO,
+    format = '[%(levelname)s] %(asctime)s - %(message)s',
+    handlers=[logging.StreamHandler()]
+)
+logger = logging.getLogger(__name__)
```
mark-velez, line 15: We should use `pipeline_logging.configure` and relocate this to the main function similar to [here](https://github.com/all-of-us/curation/blob/develop/data_steward/retraction/retract_deactivated_pids.py#L588).

mark-velez (changes requested):
Thanks @rfrancis1 this shows that papermill **can** work! :) In addition to pulling the commit from [mv/papermill](https://github.com/all-of-us/curation/tree/mv/papermill) and testing it out in your environment, there are 2 changes that are needed (see comments below).

FYI after this is merged some refactoring will be needed to enhance reusability. This will be done in a separate ticket and PR as it will need engineers' approval; I'll loop you in on that so you can keep track of where this is going.

mark-velez:
Thanks Francis! I am merging these commits via [this pull request](https://github.com/all-of-us/curation/pull/703) which also addresses the issues mentioned above.

PR #698, opened 2021-01-06
[DC-1287] Add controlled tier cleaning rules lists to clean_cdr

dcampbell-vumc:
* Added empty lists for each of the CT cleaning rule categories (deid, deid_base, deid_clean)
* Associated data stage constant with cleaning rules list placeholder to `DATA_STAGE_RULES_MAPPING`
* Added constants for each new data stage for controlled tier cleaning rule classes to the DataStage class


data_steward/constants/cdr_cleaner/clean_cdr.py
```diff
@@ -7,6 +7,9 @@
 DEID_BASE = 'deid_base'
 DEID_CLEAN = 'deid_clean'
 FITBIT = 'fitbit'
+CONTROLLED_DEID = 'controlled_deid'
```
ksdkalluri, line 10: Since this will be the starting of the controlled tier. I suggest we maintain similar names between the stage names and variable names. if we choose to use **_controlled_tier_*_**  or the **_controlled_*_**. we should aim to use any one of these conventions which will avoid a lot of confusion in the future.

ksdkalluri (changes requested):

data_steward/constants/cdr_cleaner/clean_cdr.py
dcampbell-vumc: I chose to name them all controlled_tier_* to be as explicit as possible!

data_steward/constants/cdr_cleaner/clean_cdr.py
```diff
@@ -7,6 +7,9 @@
 DEID_BASE = 'deid_base'
 DEID_CLEAN = 'deid_clean'
 FITBIT = 'fitbit'
+CONTROLLED_TIER_DEID = 'controlled_deid'
```
ksdkalluri, line 10: I think we should change these strings to `controlled_tier_*` as well. Any thoughts?

ksdkalluri (changes requested):

data_steward/constants/cdr_cleaner/clean_cdr.py
dcampbell-vumc: Oops, forgot to change those. I changed them to `controlled_tier_*` 👍🏾

ksdkalluri (approved):
Looks good. Thanks for updating!

PR #702, opened 2021-01-11
[DC-1297] Update get_client() in utils/bq.py

nishanthpp93:
[DC-1297] Update bq.get_client() to require project and accept creds

[DC-1297]: https://precisionmedicineinitiative.atlassian.net/browse/DC-1297

### DIFF
--- a/data_steward/utils/bq.py
+++ b/data_steward/utils/bq.py
@@ -13,7 +13,6 @@ from google.cloud import bigquery
 from google.auth import default
 
 # Project Imports
-from app_identity import PROJECT_ID
 from utils import auth
 from constants.utils import bq as consts
 from resources import fields_for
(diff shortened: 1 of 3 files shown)


amcgrenera-vumc:
This all looks good to me, I double checked to make sure there are no usages of `get_client()` that do not currently pass `project_id` and I think we are good. I would also maybe add in a test to make sure it breaks if the `project` parameter is not given. The test I used to double check is below, if you want to add in!

``` 
# Tests if no project_id is given
self.assertRaises(TypeError, bq.get_client, project=None) 
```

nishanthpp93:
Thanks for the suggestion, @amcgrenera-vumc! I've added the test.

amcgrenera-vumc (approved):
Thanks Nishanth! This looks good to me!

nishanthpp93:
Thanks Aimee, I had forgotten to update the docstring so added that change!

PR #708, opened 2021-01-22
[DC-822] Update Death Date Validation CR to New Base Class & Correct Pipeline Step

dcampbell-vumc:
* modified `valid_death_dates.py`:
1. module inherits from `BaseCleaningRule`
2. created a sandbox query `SANDBOX_INVALID_DEATH_DATE_ROWS` to sandbox death date rows that occur before the start of the program or after the current date
* modified `cdr_cleaner.py` to use new class in import statement. Removed from all cleaning class lists except for `COMBINED_CLEANING_CLASSES`
* created `valid_death_dates_test.py` to test that the queries are created properly populated with the correct information

### DIFF
--- a/data_steward/cdr_cleaner/cleaning_rules/valid_death_dates.py
+++ b/data_steward/cdr_cleaner/cleaning_rules/valid_death_dates.py
@@ -37,7 +37,7 @@ SANDBOX_INVALID_DEATH_DATE_ROWS = common.JINJA_ENV.from_string("""
 CREATE OR REPLACE TABLE `{{project_id}}.{{sandbox_id}}.{{sandbox_table}}` AS (
 SELECT *
 FROM `{{project_id}}.{{dataset_id}}.{{table}}`
-WHERE {{table}}_date < '{{program_start_date}}' AND {{table}}_date > {{current_date}})
+WHERE {{table}}_date < '{{program_start_date}}' OR {{table}}_date > {{current_date}})
 """)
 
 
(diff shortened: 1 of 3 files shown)


data_steward/cdr_cleaner/cleaning_rules/valid_death_dates.py
```diff
@@ -1,53 +1,142 @@
...
+    def get_sandbox_tablenames(self):
+        """
+        Get the sandbox dataset id for this class instance
+        """
+        return [
+            self.get_sandbox_table_for(affected_table)
+            for affected_table in self.affected_tables
+        ]
+
+    def get_sandbox_table_for(self, affected_table):
```
ksdkalluri, line 121: This is a redundant method, you could use the same logic in the get_sandbox_tablenames() method instead of here. get_sandbox_tablenames() will be refactored as part of using shared sandbox datasets epic. it will be easier then, to update get_sandbox_tablenames() in all the classes.

ksdkalluri (changes requested):

data_steward/cdr_cleaner/cleaning_rules/valid_death_dates.py
dcampbell-vumc: Ah yes, this was pushed before we talked about this and I forgot to make those changes afterwards 😅 I'll modify it and repush 👍🏾

dcampbell-vumc:
Removed the redundant `sandbox_table_for` function!

ksdkalluri (approved):
Looks good to me!

PR #713, opened 2021-01-26
[DC-1343] Add vocabulary qc to rdr qc report

jqian2015:


data_steward/analytics/cdr_ops/rdr_export_qc.py
```diff
@@ -16,6 +16,7 @@
 project_id = ""
 old_rdr = ""
 new_rdr = ""
+#dataset_id=""
```
mark-velez: I will remove this line, it is not needed.

mark-velez (approved):
I removed the unused variable

mark-velez:
@jqian2015 fyi the integration tests are expected to fail and this is ok (at least for now) for these notebooks. Feel free to mention this in the documentation when you get a chance.
image

PR #717, opened 2021-01-27
[DC-1317] Modify run_rdr_qc_report so it can be reused

jp3477:
Created a command line tool for executing parameterized jupyter notebooks.


data_steward/analytics/cdr_ops/README.md
```diff
@@ -4,4 +4,130 @@ Notebooks for operational evaluation of data.
...
+Executing: 100%|█████████████████████████████████████████████████████| 3/3 [00:01<00:00,  2.11cell/s]
+```
+And the outputted html:
+![Executed HTML](./images/html_image.png)
+
+Notice that an extra cell containing the passed parameter values was injected into the notebook.
+
+### Debugging
+
+There are a few sets of errors one may encounter when attempting to execute a notebook.
```
dcampbell-vumc, line 83: As we discussed in the call, in case one were to get the `ModuleNotFoundError: ....` the solution is to set the PYTHONPATH to `<path_to_data_steward>`

Also, we should add here that we need to set the `GOOGLE_APPLICATION_CREDENTIALS` to a data_analytics key file if one gets a `Please visit this URL to authorize this application` error in the notebook, specifically cell 4.

data_steward/analytics/cdr_ops/README.md
```diff
@@ -4,4 +4,130 @@ Notebooks for operational evaluation of data.
...
+Parameters inferred for notebook Example Notebook:
+2021-01-28 17:44:08 - INFO - papermill - Input Notebook:  Example Notebook.ipynb
+  my_unknown_parameter: Unknown type (default "", required=True)
+                                  This parameter has an unknown type.
+  my_string_parameter: str (default "", required=True)
+                                  This is a string parameter.
+  my_int_parameter: int (default 0, required=False)
+                                  This is an integer parameter.
+```
+
```
dcampbell-vumc, line 63: Verified this was working by executing the following command:

`python report_runner.py "rdr_export_qc.py" --help_notebook`

data_steward/analytics/cdr_ops/README.md
```diff
@@ -4,4 +4,130 @@ Notebooks for operational evaluation of data.
...
+Missing the parameter my_string_parameter for notebook Example Notebook
+Parameters inferred for notebook Example Notebook:
+  my_unknown_parameter: Unknown type (default "", required=True)
+                                  This parameter has an unknown type.
+  my_string_parameter: str (default "", required=True)
+                                  This is a string parameter.
+  my_int_parameter: int (default 0, required=False)
+                                  This is an integer parameter.
+```
+
```
dcampbell-vumc, line 103: Verified this by running the following command:

`python report_runner.py "rdr_export_qc.py" -p old_rdr rdr20201201 new_rdr rdr20210104`

This failed successfully because no project_id was supplied!

data_steward/analytics/cdr_ops/README.md
```diff
@@ -4,4 +4,130 @@ Notebooks for operational evaluation of data.
...
+---------------------------------------------------------------------------
+Exception encountered at "In [3]":
+---------------------------------------------------------------------------
+TypeError                                 Traceback (most recent call last)
+<ipython-input-3-6cf2cee3f746> in <module>()
+      1 print(f"My unknown type parameter is {my_unknown_parameter}")
+----> 2 print(f"My string parameter is {my_string_parameter / 0}")
+      3 print(f"My int parameter is {my_int_parameter}")
+
+TypeError: unsupported operand type(s) for /: 'str' and 'int'
```
dcampbell-vumc, line 121: Verified that errors are presented when encountered. I needed to set the google credentials to a data_analytics key file.

dcampbell-vumc (changes requested):
This is great! I love how a html file is generated and I don't have to fuss or mess around with modifying URLs with certain parameters, running individual cells before running the rest, etc! I tested everything out using the `rdr_export_qc.py` notebook.

- I verified that error messages were communicated properly in both the terminal as well as the generated html file. 
- I verified that the `--help_notebook` command was working properly and was able to clearly see the parameters that were necessary for the `rdr_export_qc.py` notebook.
- I verified that the html file would not be generated without the proper parameters by attempting to run the `rdr_export_qc.py` notebook without the required `project_id` parameter.

The only requests I have is in the `README.md`, we add notes about what to do if one gets a `ModuleNotFoundError` and to make sure that the `GOOGLE_APPLICATION_CREDENTIALS` is set to a data_analytics key file!

dcampbell-vumc:
Also, the unit test CircleCI check is failing because `KeyError: 'inferred_type_name'` do you have an idea why this is occurring or should I do some investigating?

data_steward/analytics/cdr_ops/README.md
jp3477: Great, I'll update the README.

dcampbell-vumc (approved):
Thanks for making those changes! The instructions are really clear now! This is good to go once the integration check passes!

dcampbell-vumc (approved):

PR #719, opened 2021-01-29
[DC-811] Update Age CR to New Base Class & Correct Pipeline Step

amcgrenera-vumc:


data_steward/cdr_cleaner/clean_cdr.py
```diff
@@ -166,7 +164,8 @@
     (validate_missing_participants.delete_records_for_non_matching_participants,
     ),
     (CleanMappingExtTables,),
-    (TemporalConsistency,)
+    (TemporalConsistency,),
+    (NegativeAges,)
```
mark-velez, line 168: I understand we need to make `NegativeAges` run only during the combined stage- is there a reason to execute it at the end of the stage as opposed to after `clean_years`?

amcgrenera-vumc: No reason, I just added it to the bottom of the list, let me know if there are certain rules that need to be before/after?

data_steward/cdr_cleaner/cleaning_rules/negative_ages.py
```diff
@@ -27,110 +32,261 @@
...
+        """
+        Return a list of dictionary query specifications.
+
+        :return:  A list of dictionaries. Each dictionary contains a single query
+            and a specification for how to execute that query. The specifications
+            are optional but the query is required.
+        """
+
+        sandbox_queries = []
+        queries = []
```
mark-velez, line 195: I'm not sure it helps to have a separate var for sandbox queries. I think they are distinguished well enough in the loop by the variable `sandbox_query` which can just be appended to `queries`. This could prevent the return type bug (see comment in unit test).

data_steward/cdr_cleaner/cleaning_rules/negative_ages.py
```diff
@@ -27,110 +32,261 @@
...
 
 person = common.PERSON
+death = common.DEATH
 MAX_AGE = 150
+affected_tables = list(date_fields.keys())
+affected_tables.append(death)
+
+#sandbox rows to be removed
+SANDBOX_NEGATIVE_AND_MAX_AGE_QUERY = JINJA_ENV.from_string("""
+CREATE OR REPLACE `{{project_id}}.{{sandbox_id}}.{{intermediary_table}}` AS (
```
mark-velez, line 42:
```suggestion
CREATE OR REPLACE TABLE `{{project_id}}.{{sandbox_id}}.{{intermediary_table}}` AS (
```

data_steward/cdr_cleaner/cleaning_rules/negative_ages.py
```diff
@@ -27,110 +32,261 @@
...
+    EXTRACT(YEAR
+    FROM
+      t.{{table_date}}) - EXTRACT(YEAR
+    FROM
+      p.birth_datetime) > {{MAX_AGE}})
+)
+""")
+
+SANDBOX_NEGATIVE_AGE_DEATH_QUERY = JINJA_ENV.from_string("""
+CREATE OR REPLACE `{{project_id}}.{{sandbox_id}}.{{intermediary_table}}` AS (
```
mark-velez, line 78:
```suggestion
CREATE OR REPLACE TABLE `{{project_id}}.{{sandbox_id}}.{{intermediary_table}}` AS (
```

data_steward/cdr_cleaner/cleaning_rules/negative_ages.py
```diff
@@ -27,110 +32,261 @@
...
+
+        # query for death before birthdate
+        sandbox_query = dict()
+        query = dict()
+        sandbox_query[
+            cdr_consts.QUERY] = SANDBOX_NEGATIVE_AGE_DEATH_QUERY.render(
+                project_id=self.project_id,
+                dataset_id=self.dataset_id,
+                sandbox_id=self.sandbox_dataset_id,
+                intermediary_tatble=self.sandbox_table_for(death),
```
mark-velez, line 240: typo here
```suggestion
                intermediary_table=self.sandbox_table_for(death),
```

data_steward/cdr_cleaner/clean_cdr.py
mark-velez: Not sure about this particular case- all I know is it's safer to preserve the existing order. :man_shrugging: Let me know if you determine that a reordering is needed here.

data_steward/cdr_cleaner/cleaning_rules/negative_ages.py
amcgrenera-vumc: ah good catch, thank you!

data_steward/cdr_cleaner/cleaning_rules/negative_ages.py
```diff
@@ -27,110 +32,261 @@
...
+  JOIN
+    `{{project_id}}.{{dataset_id}}.{{person_table}}` p
+  ON
+    t.person_id = p.person_id
+  WHERE
+    EXTRACT(YEAR
+    FROM
+      t.{{table_date}}) - EXTRACT(YEAR
+    FROM
+      p.birth_datetime) > {{MAX_AGE}})
```
mark-velez, line 73: nit: `UNION ALL` can result in rows being duplicated in the sandbox table. Could we prevent this by using `UNION DISTINCT` or by ORing the criteria in a single statement like this?

```suggestion
SELECT
  *
FROM
  `{{project_id}}.{{dataset_id}}.{{table}}`
WHERE
  {{table}}_id IN (
  SELECT
    t.{{table}}_id
  FROM
    `{{project_id}}.{{dataset_id}}.{{table}}` t
  JOIN
    `{{project_id}}.{{dataset_id}}.{{person_table}}` p
  ON
    t.person_id = p.person_id
  WHERE
    t.{{table_date}} < DATE(p.birth_datetime)
    OR
    EXTRACT(YEAR
    FROM
      t.{{table_date}}) - EXTRACT(YEAR
    FROM
      p.birth_datetime) > {{MAX_AGE}})
```

amcgrenera-vumc: I've updated this to `UNION DISTINCT` please take a look and let me know if that is correct.

amcgrenera-vumc:
@mark-velez adjustments have been made, please take a look at the revisions when you get a chance!

data_steward/cdr_cleaner/clean_cdr.py
amcgrenera-vumc: Yep makes total sense, I've moved it back up to where it originally was, thanks for that suggestion!

mark-velez (approved):
Thanks @amcgrenera-vumc ! Looks good.

PR #720, opened 2021-02-01
[DC-1326] Add table namer property to base cleaning rule

dcampbell-vumc:
* Added table_namer property which returns self._table_namer
* Added table_namer setter which sets table_namer to a default value
  using kwargs if no value is given
  Returns the table_namer if value is given
* Ensured existing cleaning rule tests passed as normal


ksdkalluri (approved):
Looks good to me!

PR #723, opened 2021-02-03
[DC-1325] Update sandbox_table_for in base_cleaning_rule.py

jp3477:


data_steward/utils/sandbox.py
```diff
@@ -42,7 +42,10 @@ def get_sandbox_table_name(table_namer, base_name):
     :param base_name: the name of the cleaning rule
     :return: the concatenated table name
     """
-    return f'{table_namer}_{base_name}'
+    if table_namer is None or table_namer == '':
+        return base_name
+    else:
+        return f'{table_namer}_{base_name}'
```
lrwb-aou, line 48: Feel like we can use python truthy-ness and falsy-ness for this comparison.  I don't recommend changing the function name or signature.
```
def get_name(table_namer, base_name):
    if table_namer is None or table_namer == '':
        return base_name
    else:
        return f'{table_namer}_{base_name}'
 
get_name(None, 'base')        #  returns "base"
get_name('', 'base')          #  returns "base"
get_name(' ', 'base')         #  returns " _base"


# using python truthy-ness                                                                                                                                                                                  
def get_new_name(table_namer, base_name):
  if table_namer and not table_namer.isspace():
    return f'{table_namer}_{base_name}'
  
  return base_name
 
get_new_name(None, 'base')        #  returns "base"
get_new_name('', 'base')          #  returns "base"
get_new_name('   ', 'base')       #  returns "base"
```

tests/unit_tests/data_steward/utils/sandbox_test.py
```diff
@@ -19,3 +19,7 @@ def test_get_sandbox_table_name(self):
         expected = f'{self.table_namer}_{base_name}'
         actual = sandbox.get_sandbox_table_name(self.table_namer, base_name)
         self.assertEqual(expected, actual)
+        self.assertEqual(base_name,
+                         sandbox.get_sandbox_table_name(None, base_name))
+        self.assertEqual(base_name,
+                         sandbox.get_sandbox_table_name('', base_name))
```
lrwb-aou, line 25: Nice unit tests.  Consider adding one to test how you want it to behave if someone passes in a spaces only string.

lrwb-aou (commented):
Nice small PR!  Let me know what you think about the feedback.  I'd prefer to add some handling if a "spaces only" string makes it into the utility function, `get_sandbox_table_name`.  Thoughts?

data_steward/utils/sandbox.py
jp3477: Nice catch, I'll change this.

tests/unit_tests/data_steward/utils/sandbox_test.py
jp3477: Will add this.

data_steward/utils/sandbox.py
jp3477: How does it look now?

lrwb-aou (approved):
This looks good!  Approved (assuming integration tests pass).

PR #724, opened 2021-02-04
[DC-294] Add checks on id range to RDR Export evaluation notebook

ksdkalluri:
* Added a check to rdr_export_qc notebook which checks id range in mapped tables.

### DIFF
--- a/data_steward/analytics/cdr_ops/rdr_export_qc.py
+++ b/data_steward/analytics/cdr_ops/rdr_export_qc.py
@@ -70,6 +70,7 @@ domain_table_list = [
     'measurement', 'note', 'observation', 'procedure_occurrence', 'provider',
     'specimen', 'visit_occurrence'
 ]
+queries = []
 for table in domain_table_list:
     query = f'''
     SELECT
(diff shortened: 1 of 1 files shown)


data_steward/analytics/cdr_ops/rdr_export_qc.py
```diff
@@ -61,6 +61,28 @@
...
+# ## ID range check
+# Combine step may break if any row IDs in the RDR are larger than the added constant(1,000,000,000,000,000).
+# Rows that are greater than 999,999,999,999,999 the will be listed out here.
+
+domain_table_list = [
+    'condition_occurrence', 'device_exposure', 'drug_exposure', 'location',
+    'measurement', 'note', 'observation', 'procedure_occurrence', 'provider',
+    'specimen', 'visit_occurrence'
+]
+for table in domain_table_list:
```
ChaoPang, line 73: Would that be a good idea to combine all queries together since they return results in the same structure?

ChaoPang (changes requested):
One tiny comment!

ChaoPang (approved):
Should we ask Francis to take a look at this?

PR #725, opened 2021-02-04
[DC-1327] Modify clean_dataset to expect a table_namer object

amcgrenera-vumc:


nishanthpp93 (approved):
I think this is good but it will have implications for how `clean_cdr.py` is calling `clean_dataset`. As long as they are accounted for, we shouldn't have issues. Do you happen to know if there's another ticket to handle it?

amcgrenera-vumc:
I don't think any current usages will be affected since we're setting the default to a blank string if it's not passed `table_namer='',`

PR #727, opened 2021-02-05
[DC-1329] create sandbox utility options function

jp3477:
Creates an options clause for defining BigQuery table descriptions and labels.

### DIFF
--- a/data_steward/utils/sandbox.py
+++ b/data_steward/utils/sandbox.py
@@ -1,6 +1,30 @@
 from utils.bq import list_datasets, create_dataset
+from common import JINJA_ENV
+from collections import OrderedDict
 
 SANDBOX_SUFFIX = 'sandbox'
+
+TABLE_LABELS_STRING = JINJA_ENV.from_string("""
+    {%- if labels %}
+    labels=[
+        {% for label_key, label_value in labels.items() %}
+        ("{{ label_key }}", "{{ label_value }}"){{ ", " if not loop.last}}
+        {% endfor %}
+    ]
+    {%- endif %}
+""")
+
+TABLE_DESCRIPTION_STRING = JINJA_ENV.from_string("""
+    {%- if description %}
+    description="{{ description }}"
+    {%- endif %}
+""")
+
+TABLE_OPTIONS_CLAUSE = JINJA_ENV.from_string("""
+OPTIONS(
+{{ contents }}
+)
+""")
 
 
 def create_sandbox_dataset(project_id, dataset_id):
@@ -60,3 +84,34 @@ def check_and_create_sandbox_dataset(project_id, dataset_id):
     if sandbox_dataset not in datasets:
         create_sandbox_dataset(project_id, dataset_id)
     return sandbox_dataset
+
+
+def get_sandbox_labels_string(src_dataset_name,
+                              class_name,
+                              table_tag,
+                              shared_lookup=False):
+    STRING_LABELS = ['src_dataset', 'class_name', 'table_tag']
+
+    labels = OrderedDict([('src_dataset', src_dataset_name),
+                          ('class_name', class_name), ('table_tag', table_tag),
+                          ('shared_lookup', shared_lookup)])
+
+    return TABLE_LABELS_STRING.render(labels=labels)
+
+
+def get_sandbox_table_description_string(description):
+    return TABLE_DESCRIPTION_STRING.render(description=description)
+
+
+def get_sandbox_options(dataset_name,
+                        class_name,
+                        table_tag,
+                        desc,
+                        shared_lookup=False):
+    labels_text = get_sandbox_labels_string(dataset_name, class_name, table_tag,
+                                            shared_lookup)
+    description_text = get_sandbox_table_description_string(desc)
+
+    contents = ',\n'.join([description_text, labels_text])
+
+    return TABLE_OPTIONS_CLAUSE.render(contents=contents)
--- a/data_steward/utils/sandbox.py
+++ b/data_steward/utils/sandbox.py
@@ -141,7 +141,7 @@ def get_sandbox_table_description_string(description):
 
     if not description or description.isspace():
         raise ValueError(
-            f"Description must be a non-empty string. Received teh value '{description}'."
+            f"Description must be a non-empty string. Received the value '{description}'."
         )
 
     return TABLE_DESCRIPTION_STRING.render(description=description)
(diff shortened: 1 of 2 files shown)


data_steward/utils/sandbox.py
```diff
@@ -63,3 +89,76 @@ def check_and_create_sandbox_dataset(project_id, dataset_id):
...
+    return TABLE_LABELS_STRING.render(labels=labels)
+
+
+def get_sandbox_table_description_string(description):
+    """A helper function that returns a formatted description for BigQuery
+
+    :param str description: A table description
+    :return: A formatted table description
+    :rtype: str
+    """
```
amcgrenera-vumc, line 140: Do you know if this is required? Not sure if there should be a check in this function that something was actually passed?

tests/unit_tests/data_steward/utils/sandbox_test.py
```diff
@@ -29,3 +29,42 @@ def test_get_sandbox_table_name(self):
...
+                dataset_name,
+                empty_class_name,
+                table_tag,
+                shared_lookup=non_bool_shared_lookup)
+
+    def test_get_sandbox_table_description_string(self):
+        description = "This is a helpful description."
+
+        actual = sandbox.get_sandbox_table_description_string(description)
+        self.assertIn(f'description="{description}"', actual)
```
amcgrenera-vumc, line 70: Should there be one more unit test to test the `get_sandbox_options` to ensure the `TABLE_OPTIONS_CLAUSE` is filled out correctly?

data_steward/utils/sandbox.py
```diff
@@ -63,3 +89,76 @@ def check_and_create_sandbox_dataset(project_id, dataset_id):
...
+                        class_name,
+                        table_tag,
+                        desc,
+                        shared_lookup=False):
+    """A function that assembles a BigQuery table options clause from labels and descriptions
+
+    :param str dataset_name: A dataset name
+    :param str class_name: A class name
+    :param str table_tag: A table tag
+    :param str description: A table description
```
amcgrenera-vumc, line 154: nit: change to `desc` to match arg

data_steward/utils/sandbox.py
```diff
@@ -63,3 +89,76 @@ def check_and_create_sandbox_dataset(project_id, dataset_id):
...
+    :param str class_name: A class name
+    :param str table_tag: A table tag
+    :param bool shared_lookup: A boolean that indicates if a table is a shared lookup, defaults to False
+    :raises ValueError: Raises if one of first three arguments is whitespace or None
+    :raises ValueError: Raises if the `shared_lookup` is not a boolean
+    :return: A formatted string of the arguments concatenated as labels
+    :rtype: str
+    """
+    STRING_LABELS = ['src_dataset', 'class_name', 'table_tag']
+    BOOL_LABELS = ['shared_lookup']
```
amcgrenera-vumc, line 110: nit: maybe make this lowercase since they are instance variables

jp3477: Sounds good

data_steward/utils/sandbox.py
jp3477: I'm not sure. In the task, it doesn't specifically mention a required check for description. But I can add it if you think that's best.

data_steward/utils/sandbox.py
jp3477: Good catch

tests/unit_tests/data_steward/utils/sandbox_test.py
jp3477: I agree, wasn't sure how I would go about this, but I have an idea now.

data_steward/utils/sandbox.py
```diff
@@ -63,3 +89,82 @@ def check_and_create_sandbox_dataset(project_id, dataset_id):
...
+    """A helper function that returns a formatted description for BigQuery
+
+    :param str description: A table description
+    :return: A formatted table description
+    :rtype: str
+    """
+
+    if not description or description.isspace():
+        raise ValueError(
+            f"Description must be a non-empty string. Received teh value '{description}'."
```
amcgrenera-vumc, line 144: nit: spell check `the`

jp3477: Oops, fixing

tests/unit_tests/data_steward/utils/sandbox_test.py
```diff
@@ -29,3 +30,88 @@ def test_get_sandbox_table_name(self):
...
+
+        #Test description empty
+        description = "    "
+        with self.assertRaises(ValueError):
+            sandbox.get_sandbox_table_description_string(description)
+
+    @patch('utils.sandbox.get_sandbox_labels_string')
+    @patch('utils.sandbox.get_sandbox_table_description_string')
+    def test_get_sandbox_options(self,
```
amcgrenera-vumc, line 80: this looks great, thanks for adding in!

amcgrenera-vumc:
This all looks great. One last thing! I was testing out the option string in BQ to make sure it would run properly and I found that the values for the labels need to be all lower case: https://cloud.google.com/bigquery/docs/reference/rest/v2/tables#labels

> The labels associated with this table. You can use these to organize and group your tables. Label keys and values can be no longer than 63 characters, **can only contain lowercase letters**, numeric characters, underscores and dashes. International characters are allowed. Label values are optional. Label keys must start with a letter and each label in the list must have a different key.

So for the `shared_lookup`, I think we can still keep it a boolean but when adding the value convert to string and  `.lower()`? let me know your thoughts!

jp3477:
> This all looks great. One last thing! I was testing out the option string in BQ to make sure it would run properly and I found that the values for the labels need to be all lower case: https://cloud.google.com/bigquery/docs/reference/rest/v2/tables#labels
> 
> > The labels associated with this table. You can use these to organize and group your tables. Label keys and values can be no longer than 63 characters, **can only contain lowercase letters**, numeric characters, underscores and dashes. International characters are allowed. Label values are optional. Label keys must start with a letter and each label in the list must have a different key.
> 
> So for the `shared_lookup`, I think we can still keep it a boolean but when adding the value convert to string and `.lower()`? let me know your thoughts!

Nice catch, I didn't realize this at all. Sure, that sounds like the best plan.

amcgrenera-vumc (approved):
Looks great, thanks for making the changes! Once the integration test passes good to merge.

PR #729, opened 2021-02-05
[DC-1389] Fixed missing pmids in view

nishanthpp93:
* Add person table to query

### DIFF
--- a/data_steward/tools/generate_ehr_upload_pids.py
+++ b/data_steward/tools/generate_ehr_upload_pids.py
@@ -37,7 +37,7 @@ FROM
     SELECT DISTINCT person_id FROM `{{project_id}}.{{ehr_dataset_id}}.%s_drug_exposure` UNION ALL
     SELECT DISTINCT person_id FROM `{{project_id}}.{{ehr_dataset_id}}.%s_observation` UNION ALL
     SELECT DISTINCT person_id FROM `{{project_id}}.{{ehr_dataset_id}}.%s_visit_occurrence`)""", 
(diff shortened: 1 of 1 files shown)


data_steward/tools/generate_ehr_upload_pids.py
```diff
@@ -31,6 +31,7 @@
   WHERE t.table_id = '%s_person'
   AND m.HPO_ID = '%s'
   AND person_id IN (
+    SELECT DISTINCT person_id FROM `{{project_id}}.{{ehr_dataset_id}}.%s_person` UNION ALL
```
mark-velez, line 34: 1. This will include PIDs that only have person rows. I just want to confirm this is intended. This might've been intentionally excluded originally because we'd had issues where participants would have no other data.

2. nit: I believe 
    ```sql
    SELECT person_id FROM `{{project_id}}.{{ehr_dataset_id}}.%s_person` UNION DISTINCT
    SELECT person_id FROM `{{project_id}}.{{ehr_dataset_id}}.%s_condition_occurrence` UNION DISTINCT
    SELECT person_id FROM `{{project_id}}.{{ehr_dataset_id}}.%s_procedure_occurrence` UNION DISTINCT
    SELECT person_id FROM `{{project_id}}.{{ehr_dataset_id}}.%s_drug_exposure`
    ```
    is equivalent and on most databases performs better.

mark-velez (approved):
Please look into my comment and address as you see fit.

data_steward/tools/generate_ehr_upload_pids.py
nishanthpp93: 1. That is correct, RDR was missing some participants due to the exclusion, but they informed us that they need those participants.
2. It does, thanks for the suggestion. I've made the changes.

PR #736, opened 2021-02-09
[DC-1333] Generalize pid_rid_map cleaning rule

nishanthpp93:
* [DC-1333] Generalize PID RID map clean rule
* [DC-1333] Update tests and remove unnecessary methods

[DC-1333]: https://precisionmedicineinitiative.atlassian.net/browse/DC-1333
[DC-1333]: https://precisionmedicineinitiative.atlassian.net/browse/DC-1333

### DIFF
--- a/data_steward/cdr_cleaner/cleaning_rules/deid/fitbit_pid_rid_map.py
+++ b/data_steward/cdr_cleaner/cleaning_rules/deid/fitbit_pid_rid_map.py
@@ -35,37 +35,6 @@ class FitbitPIDtoRID(PIDtoRID):
                          affected_tables=FITBIT_TABLES,
                          issue_numbers=ISSUE_NUMBERS)
 
-    def get_query_specs(self):
-        """
-        Return a list of dictionary query specifications.
-
-        :return:  A list of dictionaries.  Each dictionary contains a
-            single query and a specification for how to execute that query.
-            The specifications are optional but the query is required.
-        """
-        pass
-
-    def get_sandbox_tablenames(self):
-        return []
-
-    def setup_rule(self, client):
-        """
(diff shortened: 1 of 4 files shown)


data_steward/cdr_cleaner/cleaning_rules/deid/fitbit_pid_rid_map.py
```diff
@@ -0,0 +1,85 @@
...
+# Project imports
+from cdr_cleaner.cleaning_rules.deid.pid_rid_map import PIDtoRID
+from common import FITBIT_TABLES
+
+LOGGER = logging.getLogger(__name__)
+
+ISSUE_NUMBERS = ['DC-1000']
+
+
+class FitbitPIDtoRID(PIDtoRID):
```
dcampbell-vumc, line 16: I'm a little confused at what this module is doing?

nishanthpp93: Sure, this module doesn't do anything other than supply the `affected_tables` and `issue_numbers`. The rest is handled by the parent `PIDtoRID` class. The empty `setup_validation` is because it is not implemented, but I can remove that to make it simple.

dcampbell-vumc: Ah, okay. That makes sense now!

dcampbell-vumc (approved):
Thanks for explaining this to me. I understand now that the fitbit pid-rid mapping will be done using the `pid_rid_map.py` module. The generalization of `pid_rid_map.py` makes sense to me as well. 

The integration test check had failed before but has now passed since your last push so I believe this is good to go!

PR #737, opened 2021-02-09
[DC-1347] Create questionnaire_response_id mapping rule

dcampbell-vumc:
* Created `questionnaire_response_id_map.py` module which:
1. Creates the `_deid_questionnaire_response_map` lookup table if it does not already exists which populates with the `questionnaire_response_id` column from the `observation` table and generates a random value for the `research_response_id` column.
2. Performs the `QRID` to `RID` mapping by replacing the `QRID` in the `observation` table with the `RID` found in the `_deid_questionnaire_response_map` lookup table.

* Unit test created to test the query generation in `questionnaire_response_id_map.py` module

### DIFF
--- a/data_steward/cdr_cleaner/clean_cdr.py
+++ b/data_steward/cdr_cleaner/clean_cdr.py
@@ -69,6 +69,7 @@ from cdr_cleaner.cleaning_rules.update_fields_numbers_as_strings import UpdateFi
 from cdr_cleaner.cleaning_rules.temporal_consistency import TemporalConsistency
 from cdr_cleaner.cleaning_rules.valid_death_dates import ValidDeathDates
 from cdr_cleaner.cleaning_rules.negative_ages import NegativeAges
+from cdr_cleaner.cleaning_rules.deid.questionnaire_response_id_map import QRIDtoRID
 from cdr_cleaner.cleaning_rules.null_person_birthdate import NullPersonBirthdate
 from constants.cdr_cleaner import clean_cdr_engine as ce_consts
 from constants.cdr_cleaner.clean_cdr import DataStage
@@ -193,7 +194,8 @@ DEID_CLEAN_CLEANING_CLASSES = [
     (CleanMappingExtTables,)
 ]
 
-CONTROLLED_TIER_DEID_CLEANING_CLASSES = [(NullPersonBirthdate,)]
+CONTROLLED_TIER_DEID_CLEANING_CLASSES = [(QRIDtoRID,),
+                                         (NullPersonBirthdate,)]
 
 CONTROLLED_TIER_DEID_BASE_CLEANING_CLASSES = []
 
--- a/data_steward/cdr_cleaner/clean_cdr.py
+++ b/data_steward/cdr_cleaner/clean_cdr.py
@@ -194,8 +194,10 @@ DEID_CLEAN_CLEANING_CLASSES = [
     (CleanMappingExtTables,)
 ]
 
-CONTROLLED_TIER_DEID_CLEANING_CLASSES = [(QRIDtoRID,),
-                                         (NullPersonBirthdate,)]
+CONTROLLED_TIER_DEID_CLEANING_CLASSES = [
+    (QRIDtoRID,),  # Should run before any row suppression rules
+    (NullPersonBirthdate,)
+]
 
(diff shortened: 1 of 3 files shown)


data_steward/cdr_cleaner/cleaning_rules/deid/questionnaire_response_id_map.py
```diff
@@ -0,0 +1,153 @@
...
+# Project imports
+from utils import pipeline_logging
+from common import OBSERVATION, JINJA_ENV
+from constants.bq_utils import WRITE_TRUNCATE
+from constants.cdr_cleaner import clean_cdr as cdr_consts
+from cdr_cleaner.cleaning_rules.base_cleaning_rule import BaseCleaningRule
+
+LOGGER = logging.getLogger(__name__)
+
+ISSUE_NUMBERS = ['DC1347']
```
amcgrenera-vumc, line 24: maybe add DC518 to this since it was the original ticket

data_steward/cdr_cleaner/cleaning_rules/deid/questionnaire_response_id_map.py
```diff
@@ -0,0 +1,153 @@
...
+
+# Creates _deid_questionnaire_response_map lookup table and populates with the questionnaire_response_id's
+# from the observation table as well as randomly generates values for hte research_response_id column
+LOOKUP_TABLE_CREATION_QUERY = JINJA_ENV.from_string("""
+CREATE TABLE IF NOT EXISTS `{{project_id}}.{{shared_sandbox_id}}._deid_questionnaire_response_map` 
+(questionnaire_response_id INT64, research_response_id INT64)
+OPTIONS (description='lookup table for questionnaire response ids') AS
+SELECT DISTINCT questionnaire_response_id AS questionnaire_response_id, ROW_NUMBER() OVER(ORDER BY GENERATE_UUID())
```
amcgrenera-vumc, line 32: this is neat/good to know, I had no idea there was this function to use in BQ to create a unique ID!

amcgrenera-vumc: In the original code a lower bound and upper bound for the random number is set, do you know if this needs to be taken into consideration through the query as well? If I remember correctly the `lower_bound` when the pipeline is run was set to 1m.

```# create the deid table.  set upper and lower bounds of the research_id array
    records = table.shape[0]
    upper_bound = lower_bound + (10 * records)
    qr_id_list = table['questionnaire_response_id'].tolist()
    map_table = pd.DataFrame({"questionnaire_response_id": qr_id_list})

    # generate random research_questionnaire_response_ids
    research_id_array = np.random.choice(np.arange(lower_bound, upper_bound),
                                         records,
                                         replace=False)```

Comment (dcampbell-vumc) [reply data_steward/cdr_cleaner/cleaning_rules/deid/questionnaire_response_id_map.py:32]: Yes, Chao helped me create this. `GENERATE_UUID` will create a 16 digit alpha numeric string but combined with the other statements, we got it to create a numeric unique ID 🥳🥳

Comment (dcampbell-vumc) [reply data_steward/cdr_cleaner/cleaning_rules/deid/questionnaire_response_id_map.py:32]: > In the original code a lower bound and upper bound for the random number is set, do you know if this needs to be taken into consideration through the query as well? If I remember correctly the `lower_bound` when the pipeline is run was set to 1m.

I will see if I can modify it so that it'll start on 1m

File: data_steward/cdr_cleaner/clean_cdr.py [hunk 3]
```diff
@@ -192,7 +193,7 @@
     (CleanMappingExtTables,)
 ]
 
-CONTROLLED_TIER_DEID_CLEANING_CLASSES = []
+CONTROLLED_TIER_DEID_CLEANING_CLASSES = [(QRIDtoRID,)]
```
amcgrenera-vumc, line 196: Maybe just add a note here for when future CR's get added (from ticket)

> Should run before any row suppression rules execute on the dataset.

dcampbell-vumc:
I was able to modify the query to start the `research_response_id` values at 1mil as well as fixed the branch conflicts!

amcgrenera-vumc (approved):
This looks great, thanks for making those revisions!

PR #739, opened 2021-02-09
[DC-1330] add base cleaning rule helper function to get SQL options s…

amcgrenera-vumc:
…tring


jp3477 (approved):
@amcgrenera-vumc I tested with the NegativeAges cleaning rule by setting the `table_tag` property and calling its `get_sandbox_options_string` function. Everything looks good to me!

PR #744, opened 2021-02-12
[DC-1360] table suppression

amcgrenera-vumc:


dcampbell-vumc (approved):
This all looks great! Tested the query on my `location` table that I inserted a few rows of data into and verified that all the records were dropped and the schema was maintained! This is good to go!

PR #745, opened 2021-02-12
[DC-1365] Race/Ethnicity Sub-category record suppression

dcampbell-vumc:
* Sandboxes and drops any records with any of these `observation_source_concept_id` 1586151, 1586150, 1586152,
1586153, 1586154, 1586155, 1586156, 1586149
* Unit test
* Integration test


data_steward/cdr_cleaner/cleaning_rules/race_ethnicity_record_suppression.py
```diff
@@ -0,0 +1,149 @@
...
+            and a specification for how to execute that query. The specifications
+            are optional but the query is required.
+        """
+
+        sandbox_records_query = {
+            cdr_consts.QUERY:
+                SANDBOX_RECORDS_QUERY.render(
+                    project_id=self.project_id,
+                    sandbox_id=self.sandbox_dataset_id,
+                    sandbox_table=self.get_sandbox_tablenames(),
```
amcgrenera-vumc, line 88: This all looks great! Only thing, I noticed in the acceptance criteria in the ticket it says:

> It will use the shared sandbox naming convention.

So does this mean to use `get_sandbox_table_name` from `sandbox.py`?

dcampbell-vumc: Hm, you're right, I think that's the one we need to use. @lrwb-aou is `get_sandbox_table_name` from `sandbox.py` the right one to use here? I'm getting `get_sandbox_tablenames` and `get_sandbox_table_name` confused 😩

lrwb-aou: You are right.  They return different things.  One returns a single table name.  The other returns a list/iterable of table names in a sandbox.   We need to use the `get_sandbox_table_name()` from `sandbox.py` to name individual tables in the sandbox.  The other function is used to get an iterable of all available table sandbox table names associated with the cleaning rule.  The iterable returned (`get_sandbox_tablenames`) is used primarily for the report runner and integration tests.

amcgrenera-vumc (approved):
Looks great to me, everything is passing locally for me and outputting as expected. Thanks for making the revision for `sandbox_table_for`!

PR #746, opened 2021-02-15
[DC-1394] Too much survey data being suppressed/cleaned

lrwb-aou:
This changes the run order of several cleaning rules in the RDR data
stage.  This will fix the issue that caused the smoking data to be
suppressed.

It adds comments to help identify some common expected rule running orders.

It removes some useless code from the manual cleaning rules that prevented
the rules from being run as standalone rules.  They ran fine as part of the
RDR list.

### DIFF
--- a/data_steward/cdr_cleaner/manual_cleaning_rules/clean_smoking_ppi.py
+++ b/data_steward/cdr_cleaner/manual_cleaning_rules/clean_smoking_ppi.py
@@ -83,11 +83,11 @@ SELECT
     value_source_concept_id,
     value_source_value,
     questionnaire_response_id
-FROM 
+FROM
 (SELECT
     observation_id,
     person_id,
-    new_observation_concept_id as observation_concept_id, 
+    new_observation_concept_id as observation_concept_id,
     observation_date,
     observation_datetime,
     observation_type_concept_id,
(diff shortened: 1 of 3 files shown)


data_steward/cdr_cleaner/manual_cleaning_rules/clean_smoking_ppi.py
```diff
@@ -104,19 +104,19 @@
     qualifier_source_value,
     new_value_source_concept_id as value_source_concept_id,
     value_source_value,
-    questionnaire_response_id, 
+    questionnaire_response_id,
     ROW_NUMBER() OVER(PARTITION BY person_id, new_observation_source_concept_id ORDER BY rank ASC) AS this_row
 FROM
     `{project_id}.{sandbox_dataset_id}.{smoking_lookup_table}`
-JOIN `{project_id}.{combined_dataset_id}.observation` 
+JOIN `{project_id}.{dataset_id}.observation`
```
nishanthpp93, line 111: This rename makes sense because the `dataset_id` can also be rdr.

data_steward/cdr_cleaner/manual_cleaning_rules/clean_smoking_ppi.py
```diff
@@ -197,46 +197,27 @@ def get_queries_clean_smoking(project_id, dataset_id, sandbox_dataset_id):
...
-    args = parser.default_parse_args([additional_argument])
-    return args
-
-
 if __name__ == '__main__':
+    import cdr_cleaner.args_parser as parser
     import cdr_cleaner.clean_cdr_engine as clean_engine
 
-    ARGS = parse_args()
+    ARGS = parser.get_argument_parser().parse_args()
```
nishanthpp93, line 220: Thanks for updating the parser. Good to get rid of old code.

nishanthpp93 (approved):
The sequence now makes sense. It looks like the cleaning rule to null fields was added around May, but used to generate the RDR a few months after, leaving the August RDR export unaffected.

lrwb-aou:
> The sequence now makes sense. It looks like the cleaning rule to null fields was added around May, but used to generate the RDR a few months after, leaving the August RDR export unaffected.

I know the export itself was fine.  We lost the records when we cleaned it.  It was just an unfortunate rule dependency that didn't get caught.

PR #749, opened 2021-02-16
[DC-1336] Create controlled tier pid_rid_map cleaning rule

nishanthpp93:
* [DC-1336] Add controlled tier pid-rid class
* [DC-1336] Add integration test for controlled tier pid-rid class

[DC-1336]: https://precisionmedicineinitiative.atlassian.net/browse/DC-1336
[DC-1336]: https://precisionmedicineinitiative.atlassian.net/browse/DC-1336


dcampbell-vumc (commented):
The ticket says that there needs to be a test for a participant that doesn't have a remapping but I think all the ones in your integration test had valid mappings. Am I overlooking something? 😅

> 1 participant who does not have a remapping (would expect an log error message or some other notifying behavior)

nishanthpp93:
> The ticket says that there needs to be a test for a participant that doesn't have a remapping but I think all the ones in your integration test had valid mappings. Am I overlooking something? 😅
> 
> > 1 participant who does not have a remapping (would expect an log error message or some other notifying behavior)

I thought the `validate_rule` function was already doing this, but this was not called in the test. I updated it with one unmapped person and logging a warning in a precondition check.

dcampbell-vumc (approved):
Thanks for adding the logging messages! This is good to go!

PR #750, opened 2021-02-17
[DC-1350] Remove registered tier questionnaire_response_id remapping logic

dcampbell-vumc:
Removed original `questionnaire_response_id` remapping rule logic by
* Removing `create_questionnaire_mapping_table` and `map_questionnaire_response_ids` functions from `aou.py`
* Removing any calls to those functions above in `aou.py`
* Removing compute rules in `observation.json` and `config.json`


ChaoPang (approved):
Looks good!

PR #753, opened 2021-02-18
[DC-1431] parameterizing rdr cleaning rule truncation date

lrwb-aou:
* updates the cleaning rule to parameterize the cutoff date
 * updates the bash script to accept a truncation date parameter without requiring the parameter.
 * if not provided, the cleaning rule defaults to using today's date as the truncation date
 * default is implemented in the rule to take the business logic out of the bash script
 * will default correctly if None is the argument value
 * adding unit test to validate the date string

### DIFF
--- a/data_steward/cdr_cleaner/cleaning_rules/truncate_rdr_using_date.py
+++ b/data_steward/cdr_cleaner/cleaning_rules/truncate_rdr_using_date.py
@@ -82,7 +82,7 @@ class TruncateRdrData(BaseCleaningRule):
                          sandbox_dataset_id=sandbox_dataset_id)
         try:
             self.cutoff_date = validate_date_string(truncation_date)
-        except ValueError:
+        except (TypeError, ValueError):
             self.cutoff_date = str(datetime.now().date())
 
     def get_query_specs(self):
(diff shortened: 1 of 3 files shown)


data_steward/cdr_cleaner/cleaning_rules/truncate_rdr_using_date.py
```diff
@@ -142,22 +157,52 @@ def get_sandbox_tablenames(self):
...
+    a ValueError or TypeError.
+
+    :param date_string: The string to validate
+
+    :return:  a valid date string
+    :raises:  A ValueError if the date string is not a valid date or
+        doesn't conform to the specified format.
+    """
+    datetime.strptime(date_string, '%Y-%m-%d')
+    return date_string
```
lrwb-aou, line 174: I think it makes sense to put this in a utility  somewhere but I'm not sure where.  Suggestions are appreciated.

ksdkalluri: the ideal place to have this would be the utils module, which we do not have right now. For now, this can be stored in common I think.

ksdkalluri (approved):
It is a minor change. Looks good! you can merge it as is if you don't want to move it to common for now.

data_steward/cdr_cleaner/cleaning_rules/truncate_rdr_using_date.py
lrwb-aou: I looked in common and didn't see any function definitions in there, so that didn't feel right to me.  This is kind of a BigQuery thing.  Would it make sense to put this in `utils/bq.py`?

PR #755, opened 2021-02-18
[DC-1354] Confirm removal of free standing generate_ext_tables python script

dcampbell-vumc:
* removed bash script invocation in `deid_runner`
* Updated `cleaning_rules/README.md` and removed mentions of `generate_ext_tables.py`


nishanthpp93 (changes requested):
A couple suggestions. It looks like there are a lot of dependencies on `generate_ext_tables.py` which will need to be addressed before it can be removed, so this is good for now.

nishanthpp93 (approved):
Thanks for making the changes, this looks good to go!

PR #758, opened 2021-02-19
[DC-1433] remove unused parameter from ehr/rdr script

lrwb-aou:
* validation_dataset is not needed to run ehr and rdr snapshots
* parameter requirement is removed

### DIFF
--- a/data_steward/tools/create_ehr_and_rdr_snapshot.sh
+++ b/data_steward/tools/create_ehr_and_rdr_snapshot.sh
@@ -7,7 +7,6 @@ Usage: create_ehr_snapshot.sh
   --key_file <path to key file>
   --ehr_dataset <EHR dataset ID>
   --rdr_dataset <RDR dataset ID>
-  --validation_dataset <Validation dataset ID>
   --dataset_release_tag <release tag for the CDR>
   --truncation_date date to truncate the RDR data to. The cleaning rules defaults to the current date if unset.
 "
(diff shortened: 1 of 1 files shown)


data_steward/tools/create_ehr_and_rdr_snapshot.sh
```diff
@@ -46,7 +41,7 @@ while true; do
   esac
 done
 
-if [[ -z "${key_file}" ]] || [[ -z "${ehr_dataset}" ]] || [[ -z "${rdr_dataset}" ]] || [[ -z "${validation_dataset}" ]] || [[ -z "${dataset_release_tag}" ]] ; then
+if [[ -z "${key_file}" ]] || [[ -z "${ehr_dataset}" ]] || [[ -z "${rdr_dataset}" ]] || [[ -z "${dataset_release_tag}" ]] ; then
   echo "Specify the key file location and ehr_dataset ID, rdr_dataset ID, validation_dataset ID and Dataset release tag . $USAGE"
```
mark-velez, line 45: This mentions the validation dataset; might be easier to stay consistent by just echoing usage:
```suggestion
  echo "${USAGE}"
```

mark-velez (approved):
Looks good- just consider changing the output when it's called with missing args (see comment).

data_steward/tools/create_ehr_and_rdr_snapshot.sh
lrwb-aou: Nice catch.  I missed that reference.

PR #759, opened 2021-02-19
[DC-1434] updating the rdr evaluation notebook

lrwb-aou:
* adding some missing backticks about fully qualified table names
* adding leading underscores to fields named based on dataset names.  field names cannot start with numbers.
* add smoking checks query
* add comment to numerics in `value_as_string` check

### DIFF
--- a/data_steward/analytics/cdr_ops/rdr_export_qc.py
+++ b/data_steward/analytics/cdr_ops/rdr_export_qc.py
@@ -16,6 +16,7 @@
 project_id = ""
 old_rdr = ""
 new_rdr = ""
+key_file = ""
 # -
 
 # # QC for RDR Export
@@ -26,40 +27,45 @@ new_rdr = ""
 import urllib
 import pandas as pd
 
+from google.oauth2 import service_account
+
 pd.options.display.max_rows = 120
+
+creds = service_account.Credentials.from_service_account_file(key_file)
+
 # -
 
 # # Table comparison
 # The export should generally contain the same tables from month to month. Tables found only in the old or the new export are listed below.
 
 query = f'''
-SELECT 
+SELECT
   COALESCE(curr.table_id, prev.table_id) AS table_id
- ,curr.row_count AS {new_rdr}
- ,prev.row_count AS {old_rdr}
- ,(curr.row_count - prev.row_count) row_diff
-FROM `{project_id}.{new_rdr}.__TABLES__` curr 
+ ,curr.row_count AS _{new_rdr}
+ ,prev.row_count AS _{old_rdr}
+ ,(curr.row_count - prev.row_count) AS row_diff
+FROM `{project_id}.{new_rdr}.__TABLES__` curr
 FULL OUTER JOIN `{project_id}.{old_rdr}.__TABLES__` prev
   USING (table_id)
 WHERE curr.table_id IS NULL OR prev.table_id IS NULL
 '''
-pd.read_gbq(query, dialect='standard')
+pd.read_gbq(query, dialect='standard', credentials=creds)
 
 # ## Row count comparison
 # Generally the row count of clinical tables should increase from one export to the next.
 
 query = f'''
-SELECT 
+SELECT
   curr.table_id AS table_id
- ,prev.row_count AS {old_rdr}
- ,curr.row_count AS {new_rdr}
+ ,prev.row_count AS _{old_rdr}
+ ,curr.row_count AS _{new_rdr}
  ,(curr.row_count - prev.row_count) row_diff
-FROM `{project_id}.{new_rdr}.__TABLES__` curr 
+FROM `{project_id}.{new_rdr}.__TABLES__` curr
 JOIN `{project_id}.{old_rdr}.__TABLES__` prev
   USING (table_id)
 ORDER BY ABS(curr.row_count - prev.row_count) DESC;
 '''
-pd.read_gbq(query, dialect='standard')
+pd.read_gbq(query, dialect='standard', credentials=creds)
 
 # ## ID range check
 # Combine step may break if any row IDs in the RDR are larger than the added constant(1,000,000,000,000,000).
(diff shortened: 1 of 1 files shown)


data_steward/analytics/cdr_ops/rdr_export_qc.py
```diff
@@ -26,40 +27,45 @@
 import urllib
 import pandas as pd
 
+from google.oauth2 import service_account
+
 pd.options.display.max_rows = 120
+
+creds = service_account.Credentials.from_service_account_file(key_file)
```
mark-velez, line 34: Could we avoid handling credentials and just fall back on application default credentials as we do with other scripts?

lrwb-aou: I added this because I had not run `gcloud auth application-default login` before and did not have any default credentials.  Running from the command line, the script just failed.  It didn't open a browser for me to authenticate with.  I will remove this for now, but I'd like to discuss setting this up to run with impersonated credentials/user provided credentials in the future.

data_steward/analytics/cdr_ops/rdr_export_qc.py
```diff
@@ -26,40 +27,45 @@
...
+ ,curr.row_count AS _{new_rdr}
+ ,prev.row_count AS _{old_rdr}
+ ,(curr.row_count - prev.row_count) AS row_diff
+FROM `{project_id}.{new_rdr}.__TABLES__` curr
 FULL OUTER JOIN `{project_id}.{old_rdr}.__TABLES__` prev
   USING (table_id)
 WHERE curr.table_id IS NULL OR prev.table_id IS NULL
 '''
-pd.read_gbq(query, dialect='standard')
+pd.read_gbq(query, dialect='standard', credentials=creds)
```
mark-velez, line 52: As of `pandas-gbq 0.10.0` the param `dialect` _should_ default to `standard` ([docs](https://pandas-gbq.readthedocs.io/en/stable/changelog.html#changelog-0-10-0)) but for some reason here it defaults to `legacy` even though our [requirements.txt ](https://github.com/all-of-us/curation/blob/develop/data_steward/requirements.txt#L65) says `pandas-gbq==0.11.0`. Any idea what this is about? It would be nice to eliminate duplicate code.

lrwb-aou: @mark-velez According to the [read_gbq docs](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.read_gbq.html), the default for this function is legacy and will become standard in a future release.

data_steward/analytics/cdr_ops/rdr_export_qc.py
lrwb-aou: What am I missing?  Without explicitly creating the credentials, I get the following error when trying to run the notebook.
```
     10 WHERE curr.table_id IS NULL OR prev.table_id IS NULL
     11 '''
---> 12 pd.read_gbq(query, dialect='standard')

... stack trace shortened for readability

/Users/lbiggers/Projects/env37/lib/python3.7/site-packages/ipykernel/kernelbase.py in raw_input(self, prompt)
    697         if not self._allow_stdin:
    698             raise StdinNotImplementedError(
--> 699                 "raw_input was called, but this frontend does not support input requests."
    700             )
    701         return self._input_request(py3compat.cast_unicode_py2(prompt),

StdinNotImplementedError: raw_input was called, but this frontend does not support input requests.
```

lrwb-aou: I was able to use this line to run the notebook.

```
GOOGLE_APPLICATION_CREDENTIALS=~/.keys/XXXXXXXX.json PYTHONPATH=./:$PYTHONPATH  python analytics/cdr_ops/report_runner.py analytics/cdr_ops/rdr_export_qc.py -p project_id <project_value> -p old_rdr <old_value> -p new_rdr <rdr_value>
```
I just really dislike assuming someone has set their path correctly.  This goes back to requiring someone to explicitly choose between prod and test.  It's so easy to set the wrong variable and move forward altering the wrong data.

But, this is a conversation to have outside of this PR.

data_steward/analytics/cdr_ops/rdr_export_qc.py
mark-velez: Thanks @lrwb-aou ; sorry, I'm still confused (not your fault), could you help me understand? Is the `read_gbq` being used different from [this one](https://github.com/pydata/pandas-gbq/blob/0.11.0/pandas_gbq/gbq.py#L856) in `pandas-gbq==0.11.0`? :confused:

data_steward/analytics/cdr_ops/rdr_export_qc.py
mark-velez: I would definitely like to discuss how authentication should work in different use cases and then make everything work consistently. Let's just tackle that as a separate issue please :pray:  because there are docs and things that are affected by such changes.

lrwb-aou: I agree it is too large of an issue to tackle in this PR.  So, I've already removed that code.

data_steward/analytics/cdr_ops/rdr_export_qc.py
lrwb-aou: @mark-velez it's a confusing mess.   But, the link I posted is `pandas.read_gbq` which uses `pandas_gbq.read_gbq`.  Your link is the `pandas_gbq` link.  I didn't write the `dialect='standard'` code, so I don't want to change it without an investigation.  But, looking at the notebook, the code is using the `pandas.read_gbq` function which defaults to legacy and passes that to `pandas_gbq.read_gbq`. 

We could just make the whole thing a lambda function that gets defined once and called multiple times.  But that also feels outside the scope of this PR.  I just wanted the thing to run on rdr datasets whose:
1. names start with numbers
2. add backticks to fully qualified table names
3. add descriptive comments
4. add a smoking rule check to make sure we don't lose that data.

Basically, I just want to increase it's usability right now.

mark-velez (approved):

PR #760, opened 2021-02-19
[DC-1381] Record Suppress some cancer condition concepts

jp3477:
Suppresses a set of cancer related concept codes.

### DIFF
--- a/data_steward/cdr_cleaner/cleaning_rules/cancer_concept_suppression.py
+++ b/data_steward/cdr_cleaner/cleaning_rules/cancer_concept_suppression.py
@@ -34,10 +34,9 @@ WHERE REGEXP_CONTAINS(c.concept_code,
 CONCEPT_SUPPRESSION_QUERY = JINJA_ENV.from_string("""
 SELECT  o.*
 FROM `{{project_id}}.{{dataset_id}}.observation` o
-JOIN `{{project_id}}.{{dataset_id}}.concept` c
-    ON c.concept_id = o.observation_concept_id
-WHERE NOT REGEXP_CONTAINS(c.concept_code, 
-    r'(History_WhichConditions)|(Condition_OtherCancer)|(History_AdditionalDiagnosis)|(OutsideTravel6MonthWhere)'
+LEFT JOIN `{{project_id}}.{{sandbox_id}}.{{sandbox_table}}` AS so
+    ON so.observation_id = o.observation_id
+WHERE so.observation IS NULL
 )
 """)
 
@@ -82,8 +81,11 @@ class CancerConceptSuppression(BaseCleaningRule):
 
         concept_suppression_query = {
             cdr_consts.QUERY:
-                CONCEPT_SUPPRESSION_QUERY.render(project_id=self.project_id,
-                                                 dataset_id=self.dataset_id),
+                CONCEPT_SUPPRESSION_QUERY.render(
+                    project_id=self.project_id,
+                    dataset_id=self.dataset_id,
+                    sandbox_id=self.sandbox_dataset_id,
+                    sandbox_table=self.sandbox_table_for(OBSERVATION)),
             cdr_consts.DESTINATION_TABLE:
                 OBSERVATION,
             cdr_consts.DESTINATION_DATASET:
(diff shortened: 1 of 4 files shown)


data_steward/cdr_cleaner/cleaning_rules/cancer_concept_suppression.py
```diff
@@ -0,0 +1,166 @@
...
+""")
+
+CONCEPT_SUPPRESSION_QUERY = JINJA_ENV.from_string("""
+SELECT  o.*
+FROM `{{project_id}}.{{dataset_id}}.observation` o
+JOIN `{{project_id}}.{{dataset_id}}.concept` c
+    ON c.concept_id = o.observation_concept_id
+WHERE NOT REGEXP_CONTAINS(c.concept_code, 
+    r'(History_WhichConditions)|(Condition_OtherCancer)|(History_AdditionalDiagnosis)|(OutsideTravel6MonthWhere)'
+)
```
ChaoPang, line 41: This query might miss some records because `observation_concept_id` only stores the standard_concept, however, the list generated by the regex expression contains non-standard `concept_ids`, therefore those would be missed. 
1. I think this suppression rule is similar to the other ones where the suppression needs to be applied to any *concept_id field just to be sure we are pulling those records out. 
2. The concepts found by regex contain non-standard ones, we might want to include the corresponding standard concepts as well. 

You could take a look at [AbstractConceptSuppression](https://github.com/all-of-us/curation/blob/03170b8cae4675de53a7491fdb2d1c59e8a69708/data_steward/cdr_cleaner/cleaning_rules/deid/concept_suppression.py#L16) and [AbstractBqLookupTableConceptSuppression](https://github.com/all-of-us/curation/blob/03170b8cae4675de53a7491fdb2d1c59e8a69708/data_steward/cdr_cleaner/cleaning_rules/deid/concept_suppression.py#L141). An implementation of this (similar to this one) can be found at [motor_vehicle_accident_suppression.py](https://github.com/all-of-us/curation/blob/develop/data_steward/cdr_cleaner/cleaning_rules/deid/motor_vehicle_accident_suppression.py). It checks a list of suppressed concepts across all *concept_id fields.

data_steward/cdr_cleaner/cleaning_rules/cancer_concept_suppression.py
```diff
@@ -0,0 +1,166 @@
...
+SANDBOXING_QUERY = JINJA_ENV.from_string("""
+CREATE OR REPLACE TABLE `{{project_id}}.{{sandbox_id}}.{{sandbox_table}}` AS
+SELECT  o.*
+FROM `{{project_id}}.{{dataset_id}}.observation` o
+JOIN `{{project_id}}.{{dataset_id}}.concept` c
+    ON c.concept_id = o.observation_concept_id
+WHERE REGEXP_CONTAINS(c.concept_code, 
+    r'(History_WhichConditions)|(Condition_OtherCancer)|(History_AdditionalDiagnosis)|(OutsideTravel6MonthWhere)'
+)
+""")
```
ChaoPang, line 32: This query might miss some records because `observation_concept_id` only stores the standard_concept, however, the list generated by the regex expression contains non-standard `concept_ids`, therefore those would be missed. 
1. I think this suppression rule is similar to the other ones where the suppression needs to be applied to any *concept_id field just to be sure we are pulling those records out. 
2. The concepts found by regex contain non-standard ones, we might want to include the corresponding standard concepts as well. 

You could take a look at [AbstractConceptSuppression](https://github.com/all-of-us/curation/blob/03170b8cae4675de53a7491fdb2d1c59e8a69708/data_steward/cdr_cleaner/cleaning_rules/deid/concept_suppression.py#L16) and [AbstractBqLookupTableConceptSuppression](https://github.com/all-of-us/curation/blob/03170b8cae4675de53a7491fdb2d1c59e8a69708/data_steward/cdr_cleaner/cleaning_rules/deid/concept_suppression.py#L141). An implementation of this (similar to this one) can be found at [motor_vehicle_accident_suppression.py](https://github.com/all-of-us/curation/blob/develop/data_steward/cdr_cleaner/cleaning_rules/deid/motor_vehicle_accident_suppression.py). It checks a list of suppressed concepts across all *concept_id fields.

data_steward/cdr_cleaner/cleaning_rules/cancer_concept_suppression.py
jp3477: Good idea

ChaoPang (changes requested):
I think suppression should be applied to all *concept_id fields in `observation` because we really want to make sure we are not missing anything. The suppressed concept list can be extended through the `Maps to` relationship to identify a few additional concepts. 

Based on the conversation with @lrwb-aou, the regex expression is sufficient to identify other cancer concepts that might expose sensitive information. There is no need to add the standard concepts to this suppression rule.

ChaoPang (approved):
Looks good!

PR #761, opened 2021-02-19
[DC-1372] Identifying field suppression

amcgrenera-vumc:


data_steward/cdr_cleaner/cleaning_rules/identifying_field_suppression.py
```diff
@@ -0,0 +1,181 @@
...
+                for item in schema:
+                    if item['name'] in fields:
+                        if item['mode'] == 'nullable':
+                            value = 'NULL'
+                        elif item['type'] == 'integer':
+                            value = 0
+                        elif item['type'] == 'string':
+                            value = ''
+                        suppression_statement = REPLACE_STRING.render(
+                            suppression_statement=value, field=item['name'])
```
nishanthpp93, line 122: Changing the directory feels a bit off. You can make use of the helper functions in `utils.bq` for this.
```suggestion
          for table in self.affected_tables:
             schema = bq.get_table_schema(table)
             statements = []
             for item in schema:
                 if item.name in fields:
                     if item.mode == 'nullable':
                         value = 'NULL'
                     elif item.field_type == 'integer':
                         value = 0
                     elif item.field_type == 'string':
                         value = ''
                     else:
                         raise RuntimeError(
                             f"Required field {item.name} needs to be integer or string type to be replaced"
                         )
                     suppression_statement = REPLACE_STRING.render(
                         suppression_statement=value, field=item.name)
```

nishanthpp93 (changes requested):
Works well, added a suggestion to avoid changing directories.

data_steward/cdr_cleaner/cleaning_rules/identifying_field_suppression.py
amcgrenera-vumc: this is much better, thank you for that suggestion!

nishanthpp93 (approved):
Looks good!

PR #763, opened 2021-02-23
[DC-1291] add ability to impersonate service account credentials

lrwb-aou:
* add ability to impersonate credentials with: end user credentials, service account referenced via environment variable, service account referenced via key file path

* add credentials getter to create_tier script
* require impersonation target in create_tier script
* make credentials filepath optional

### DIFF
--- a/data_steward/utils/auth.py
+++ b/data_steward/utils/auth.py
@@ -154,5 +154,5 @@ def get_impersonation_credentials(target_principal,
                 target_scopes=target_scopes,
                 lifetime=IMPERSONATION_LIFETIME)
         except (OSError, ValueError, TypeError, RuntimeError) as exc:
-            print("An unexpected error was encountered in {}")
+            LOGGER.exception("An unexpected error was encountered")
             raise (exc)
--- a/data_steward/utils/auth.py
+++ b/data_steward/utils/auth.py
@@ -100,9 +100,10 @@ def get_impersonation_credentials(target_principal,
         LOGGER.warning(f"Using default scopes: {DEFAULT_SCOPES}")
         target_scopes = DEFAULT_SCOPES
 
-    if not isinstance(target_scopes, list):
-        LOGGER.warning(f"Target scopes requires a list.  Using"
-                       f" default scopes: {DEFAULT_SCOPES}.")
+    if not isinstance(target_scopes, (list, tuple)):
+        LOGGER.warning(f"Target scopes requires an iterable.  Currently "
+                       f"for lists and tuples.  Using "
+                       f"default scopes: {DEFAULT_SCOPES}.")
         target_scopes = DEFAULT_SCOPES
 
     if key_file:
--- a/data_steward/utils/auth.py
+++ b/data_steward/utils/auth.py
@@ -97,12 +97,12 @@ def get_impersonation_credentials(target_principal,
             "Cannot impersonate.  No impersonation target specified.")
 
     if not target_scopes:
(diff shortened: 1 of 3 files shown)


tests/unit_tests/data_steward/tools/create_tier_test.py
```diff
@@ -98,38 +116,74 @@ def test_parse_args(self):
...
+            '--tier',
+            self.tier,
+            '--idataset',
+            self.input_dataset,
+            '--release_tag',
+            self.release_tag,
+            '--deid_stage',
+            'deid_base',
+            '--run_as',
+            self.run_as,
```
lrwb-aou, line 133: FYI:  I really dislike this, but the pre-commit hook made all these changes.

data_steward/utils/auth.py
```diff
@@ -42,3 +63,95 @@ def delegated_credentials(credentials, scopes, subject=None):
...
+        is preferable.
+
+    :returns: valid impersonated credentials
+    """
+    if not target_principal:
+        raise RuntimeError(
+            "Cannot impersonate.  No impersonation target specified.")
+
+    if not target_scopes:
+        LOGGER.warn(f"Using default scopes: {DEFAULT_SCOPES}")
```
ksdkalluri, line 100: LOGGER.warn is deprecated, warning is the updated function signature.

ksdkalluri (commented):
This addition looks good to me, thanks for adding the unit tests for the create_tier. Since @mark-velez is also looking at this PR I'll wait for him to review and add any suggestions.

data_steward/utils/auth.py
```diff
@@ -42,3 +63,95 @@ def delegated_credentials(credentials, scopes, subject=None):
...
+        target_scopes = DEFAULT_SCOPES
+
+    if not isinstance(target_scopes, list):
```
mark-velez, line 103: I tried to set `target_scopes=google.cloud.storage.Client.SCOPE` ([code](https://github.com/googleapis/python-storage/blob/master/google/cloud/storage/client.py#L96-L100)) and it didn't work- maybe something like this instead?
```suggestion
    if not type(target_scopes) in (list, tuple):
```

tests/unit_tests/data_steward/tools/create_tier_test.py
ksdkalluri: YAPF is going this, I guess it's fine.

data_steward/utils/auth.py
mark-velez, line 100: On `python3.7` I get a warning that `LOGGER.warn` is deprecated. It suggests this instead:
```suggestion
        LOGGER.warning(f"Using default scopes: {DEFAULT_SCOPES}")
```

mark-velez (commented):
This appears to work! :fireworks: @lrwb-aou thank you for figuring this out! (Now I want to impersonate all the things... :joy: )

data_steward/utils/auth.py
lrwb-aou: I'll look at this.  I want it to always work!  I was just looking at the scopes defined as strings.  Seems like we should be able to use scopes the way you are using them too.

lrwb-aou: ok.  so this is just a tuple.  it's not an object.  I can get that working pretty quick.

lrwb-aou: @mark-velez check it now to see if it works as you would hope/expect.

mark-velez: Yes- that does it!

ksdkalluri (approved):

PR #767, opened 2021-02-25
[DC-1440] Fix error reporting bug in concept_suppression.py

dcampbell-vumc:
* Modified get_table_in_dataset function to properly check for errors

### DIFF
--- a/data_steward/cdr_cleaner/cleaning_rules/deid/concept_suppression.py
+++ b/data_steward/cdr_cleaner/cleaning_rules/deid/concept_suppression.py
@@ -46,8 +46,8 @@ def get_tables_in_dataset(client: Client, project_id, dataset_id,
                                              table_names=table_names))
     result = query_job.result()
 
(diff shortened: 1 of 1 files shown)


data_steward/cdr_cleaner/cleaning_rules/deid/concept_suppression.py
```diff
@@ -46,8 +46,8 @@ def get_tables_in_dataset(client: Client, project_id, dataset_id,
                                              table_names=table_names))
     result = query_job.result()
 
-    if hasattr(result, 'errors') and result.errors:
```
ChaoPang, line 49: `query_jub.result` could throw GoogleCloudError and TimeoutError, we should put it into a try-catch block

ChaoPang (changes requested):
query_jub.result could throw GoogleCloudError and TimeoutError, we should put it into a try-catch block

ChaoPang (approved):
Looks good!

PR #769, opened 2021-02-26
[DC-1436] Set rules to run on controlled tier clean

amcgrenera-vumc:


jp3477 (approved):
Quick review, everything looks good! 😄

PR #771, opened 2021-03-02
[DC-1441] Remove any extra tables from the controlled tier deid dataset

jp3477:
Cleaning rule to remove extra tables at the end of the controlled tier generation pipeline.

### DIFF
--- a/data_steward/cdr_cleaner/clean_cdr.py
+++ b/data_steward/cdr_cleaner/clean_cdr.py
@@ -229,8 +229,8 @@ CONTROLLED_TIER_DEID_CLEANING_CLASSES = [
     (IDFieldSuppression,),  # Should run after any data remapping
     (GenerateSiteMappingsAndExtTables,),
     (CancerConceptSuppression,),  # Should run after any data remapping rules
-    (AggregateZipCodes,)(RemoveExtraTables,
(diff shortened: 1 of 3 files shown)


nishanthpp93 (approved):
Works well! I noticed that the `list_queries` argument returns an empty list of queries since `setup_rule` is not run, but there is no easy workaround since extra tables can only be known after listing tables in the dataset. I think it's fine. Jobs can take a long time because of multiple queries, but I don't think it should be a significant issue.

jp3477:
> Works well! I noticed that the `list_queries` argument returns an empty list of queries since `setup_rule` is not run, but there is no easy workaround since extra tables can only be known after listing tables in the dataset. I think it's fine. Jobs can take a long time because of multiple queries, but I don't think it should be a significant issue.
> 
> Also, nit: there are some unused imports. They can be removed.

Thanks for the review! I did consider putting the table listing in the class constructor, which would allow the proper queries to be listed. But I decided against it, because `setup_rule` seemed more fitting. And I've removed the unused imports, thanks for pointing those out.

nishanthpp93:
There are some conflicts so you'll need to rebase your branch onto develop, resolve them and force push.

PR #774, opened 2021-03-03
[DC-1445] Enforce Data cutoff of ehr submissions

dcampbell-vumc:
* created cleaning rule to sandbox and drop data in OMOP tables by:
** getting all OMOP tables except for person table for cleaning rule
optimization
** only getting the date and datetime fields from those OMOP tables
** sandbox and drop the records that fall after the cutoff date

* integration test
* unit test
* added cleaning rule to the pipeline

### DIFF
--- a/data_steward/cdr_cleaner/cleaning_rules/ehr_submission_data_cutoff.py
+++ b/data_steward/cdr_cleaner/cleaning_rules/ehr_submission_data_cutoff.py
@@ -26,7 +26,7 @@ CREATE OR REPLACE TABLE `{{project_id}}.{{sandbox_id}}.{{intermediary_table}}` A
 SELECT * FROM `{{project_id}}.{{dataset_id}}.{{cdm_table}}`
 WHERE 
     (coalesce({{date_fields}}) < DATE("{{cutoff_date}}")) 
-    AND coalesce({{datetime_fields}}) < TIMESTAMP("{{cutoff_date}}")))
+    AND coalesce({{datetime_fields}}) < TIMESTAMP("{{cutoff_date}}"))
 """)
 
 DATE_CUTOFF_QUERY = JINJA_ENV.from_string("""
--- a/data_steward/cdr_cleaner/cleaning_rules/ehr_submission_data_cutoff.py
+++ b/data_steward/cdr_cleaner/cleaning_rules/ehr_submission_data_cutoff.py
@@ -53,7 +53,7 @@ class EhrSubmissionDataCutoff(BaseCleaningRule):
         this SQL, append them to the list of Jira Issues.
         DO NOT REMOVE ORIGINAL JIRA ISSUE NUMBERS!
 
-        :params: truncation_date: the last date that should be included in the
+        :params: cutoff_date: the last date that should be included in the
             dataset
         """
         try:
@@ -181,7 +181,7 @@ class EhrSubmissionDataCutoff(BaseCleaningRule):
     def get_sandbox_tablenames(self):
         sandbox_tables = []
         for table in self.affected_tables:
-            sandbox_tables.append(f'{self._issue_numbers[0].lower()}_{table}')
+            sandbox_tables.append(self.sandbox_table_for(table))
         return sandbox_tables
 
 
--- a/data_steward/cdr_cleaner/cleaning_rules/ehr_submission_data_cutoff.py
+++ b/data_steward/cdr_cleaner/cleaning_rules/ehr_submission_data_cutoff.py
@@ -153,7 +153,8 @@ class EhrSubmissionDataCutoff(BaseCleaningRule):
                     cdr_consts.DESTINATION_DATASET:
                         self.dataset_id,
                     cdr_consts.DISPOSITION:
-                        bq_consts.WRITE_TRUNCATE}
+                        bq_consts.WRITE_TRUNCATE
+                }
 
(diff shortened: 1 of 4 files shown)


data_steward/cdr_cleaner/cleaning_rules/ehr_submission_data_cutoff.py
```diff
@@ -0,0 +1,240 @@
...
+import constants.cdr_cleaner.clean_cdr as cdr_consts
+from cdr_cleaner.cleaning_rules.base_cleaning_rule import BaseCleaningRule
+
+LOGGER = logging.getLogger(__name__)
+
+SANDBOX_QUERY = JINJA_ENV.from_string("""
+CREATE OR REPLACE TABLE `{{project_id}}.{{sandbox_id}}.{{intermediary_table}}` AS (
+SELECT * FROM `{{project_id}}.{{dataset_id}}.{{cdm_table}}`
+WHERE 
+    (coalesce({{date_fields}}) < DATE("{{cutoff_date}}"))
```
jp3477, line 28: So this would check the first non-null date and non-null datetime field for each row. Is the first non-null field reliably always the correct one to use?

data_steward/cdr_cleaner/cleaning_rules/ehr_submission_data_cutoff.py
```diff
@@ -0,0 +1,240 @@
...
+    def validate_rule(self, client):
+        """
+        Validates the cleaning rule which deletes or updates the data from the tables
+        """
+        raise NotImplementedError("Please fix me.")
+
+    def get_sandbox_tablenames(self):
+        sandbox_tables = []
+        for table in self.affected_tables:
+            sandbox_tables.append(f'{self._issue_numbers[0].lower()}_{table}')
```
jp3477, line 184: You could use the `sandbox_table_for` helper function here, I think.
```suggestion
            sandbox_tables.append(self.sandbox_table_for(table))
```

data_steward/cdr_cleaner/cleaning_rules/ehr_submission_data_cutoff.py
```diff
@@ -0,0 +1,240 @@
...
+                 sandbox_dataset_id,
+                 cutoff_date=None):
+        """
+        Initialize the class with proper information.
+
+        Set the issue numbers, description and affected datasets. As other tickets may affect
+        this SQL, append them to the list of Jira Issues.
+        DO NOT REMOVE ORIGINAL JIRA ISSUE NUMBERS!
+
+        :params: truncation_date: the last date that should be included in the
```
jp3477, line 56: Changing docstring param name from `truncation_date` to `cutoff_date`
```suggestion
        :params: cutoff_date: the last date that should be included in the
```

jp3477 (commented):
This looks good, and the tests ran well for me. I just left a few comments.

data_steward/cdr_cleaner/cleaning_rules/ehr_submission_data_cutoff.py
dcampbell-vumc: All the date and DATETIME fields of each OMOP table (excluding the `person` table) are retrieved from the  snippet below and then rendered as the `{{date_fields}}` or `{{datetime_fields}}` in the query:

``` 
  # gets all fields from the affected table
  fields = fields_for(table)

  date_fields = []
  datetime_fields = []

  for field in fields:
      # appends only the date columns to the date_fields list
      if field['type'] in ['date']:
          date_fields.append(field['name'])

      # appends only the datetime columns to the datetime_fields list
      if field['type'] in ['timestamp']:
          datetime_fields.append(field['name'])
```

The `fields_for` function returns all the fields for the given table. For each field in the table if it has a `date` or `timestamp` `type`, it will append it to a `date_fields` or `datetime_fields` list. The contents of those lists are what is passed in as the `{{date_fields}}` and `{{datetime_fields}}` of the `SANDBOX_QUERY`

Let me know if that made sense, I'm happy to hop on a call if you need more explanation ☺️☺️

data_steward/cdr_cleaner/cleaning_rules/ehr_submission_data_cutoff.py
dcampbell-vumc: I changed it and both the unit and integration tests passed 👍🏾

data_steward/cdr_cleaner/cleaning_rules/ehr_submission_data_cutoff.py
ksdkalluri: @dcampbell-vumc looking at the query again, It is MAX(Date fields/datetime fields) we should be using instead of coalesce.

dcampbell-vumc: @ksdkalluri can you explain this change? As it stands, it doesn't work in big query. I'm confused because instead of having separate lists of `date_fields` and `datetime_fields` this has it in one and I don't think we can compare `datetime` fields with a `date` type cutoff date.

ksdkalluri: Like we discussed it's GREATEST() we have to use instead of MAX().

jp3477 (approved):
Ah, I see why `GREATEST` works instead of `MAX`. Approved!

PR #775, opened 2021-03-03
[DC-1447] Motor vehicle accidents concept suppression Cleaning rule fails on non aou required tables

ksdkalluri:


data_steward/resources.py
```diff
@@ -374,3 +374,15 @@ def get_concept_id_fields(table_name) -> List[str]:
...
+
+def has_domain_table_id(table_name):
+    """
+    Determines if a table has domain_table_id
+    
+    :param table_name: Name of a cdm domain table
+    :return: True/False if domain_table_id is available is table fields
+    """
+    return f'{table_name}_id' in [
+        field['name'] for field in fields_for(table_name)
```
lrwb-aou, line 387: I'd prefer to use `field.get('name', '')` to  shield against errors if `'name'` doesn't exist.  It should always exist, but I wouldn't count on it.

data_steward/cdr_cleaner/cleaning_rules/deid/motor_vehicle_accident_suppression.py
```diff
@@ -131,7 +131,7 @@ def __init__(self, project_id, dataset_id, sandbox_dataset_id):
             project_id=project_id,
             dataset_id=dataset_id,
             sandbox_dataset_id=sandbox_dataset_id,
-            affected_tables=resources.CDM_TABLES,
+            affected_tables=AOU_REQUIRED,
```
lrwb-aou, line 134: The `AOU_REQUIRED` tables are the ones we require sites to submit.  Correct?

data_steward/cdr_cleaner/cleaning_rules/deid/concept_suppression.py
```diff
@@ -146,14 +146,16 @@ def get_query_specs(self, *args, **keyword_args) -> query_spec_list:
         sandbox_queries = [
             self.get_sandbox_query(table_name)
             for table_name in self.affected_tables
-            if get_concept_id_fields(table_name)
+            if (get_concept_id_fields(table_name) and
+                has_domain_table_id(table_name))
         ]
```
lrwb-aou, line 151: Could we do this once in initialization of `AbstractConceptSuppression`?  Like:

```
# in class initialization
affected_tables = [table_name for table_name in affected_tables 
                               if (get_concept_id_field(table_name) and has_domain_table_id(table_name))]

super().__init__(issue_numbers=issue_numbers,
                         description=description,
                         affected_datasets=affected_datasets,
                         project_id=project_id,
                         dataset_id=dataset_id,
                         sandbox_dataset_id=sandbox_dataset_id,
                         affected_tables=affected_tables)


# then these lines would just be
for table_name in self.affected_tables
```

Primary benefits:  one place  to update instead of two and `self.affected_tables` reflects them tables that are actually affected.

lrwb-aou (commented):
Looks good.  Minor comments attached.

data_steward/cdr_cleaner/cleaning_rules/deid/motor_vehicle_accident_suppression.py
ksdkalluri: yes, but these are the main cdm tables that will have concept_id fields.

lrwb-aou (approved):
Looks good!

PR #776, opened 2021-03-04
[DC-773] Fix query variable in create_deid_map from dict to list

amcgrenera-vumc:


data_steward/retraction/create_deid_map.py
```diff
@@ -142,10 +142,10 @@ def run_queries(queries):
     :param queries: list of queries
     """
     query_job_ids = []
-    for query_dict in queries:
+    for query in queries:
         logging.info('Creating or renaming _deid_map table with query: %s' %
-                     (query_dict['query']))
-        job_results = bq_utils.query(q=query_dict['query'], batch=True)
+                     (query))
```
dcampbell-vumc, line 147: nit: The redundant parenthesis can be removed.
```suggestion
                     query)
```

dcampbell-vumc (changes requested):
Just a little nit pick change and then this will be good to go!

data_steward/retraction/create_deid_map.py
amcgrenera-vumc: good catch, thank you! I've taken it out!

amcgrenera-vumc:
@dcampbell-vumc the integration test is failing, I'm going to investigate.

lrwb-aou:
> @dcampbell-vumc the integration test is failing, I'm going to investigate.

@amcgrenera-vumc and @dcampbell-vumc This is failing because of [Line 173 in retract_deactivated_pids.py](https://github.com/all-of-us/curation/blob/develop/data_steward/retraction/retract_deactivated_pids.py#L173) and the branch name.  Take `deid` out of your branch name and it will work just fine.  Line 173 is looking for either `deid` in the dataset name or a `de_identified: true` dataset label.  Test datasets are built using your branch name.  So, it's hitting this and failing because the test isn't providing a pid_rid mapping table name.

PR #778, opened 2021-03-04
[DC-1448] NullPersonBirthdate changes schema of person table

nishanthpp93:
* [DC-1448] Use UPDATE instead of SELECT *


lrwb-aou (approved):
looks good!

PR #779, opened 2021-03-04
[DC-542] PMI_PreferNotToAnswer appear as an observation_source_concept_id

dcampbell-vumc:
* Suppresses any records in the `observation` table with an `observation_source_concept_id` as `903079`
* Integration test
* Added cleaning rule to the pipeline


ksdkalluri (approved):
Looks good!

PR #780, opened 2021-03-05
[DC-1140] Snapshot_by_query does not copy person_ext from staging

jp3477:
Correcting a bug where person_ext table is not loaded from the staging dataset.


ksdkalluri (approved):
Looks Good! Thanks for working on this.

PR #781, opened 2021-03-08
[DC-1454] Update ct_pid_rid_map.py to store _deid_map table in sandbox dataset

ksdkalluri:


lrwb-aou (approved):
looks good.  tests are passing.

PR #782, opened 2021-03-08
[DC-1455] COPE survey_version_concept info is not updated in observation_ext table

amcgrenera-vumc:


dcampbell-vumc (approved):
Ran the integration test and it passes as normal. This looks good to go!

PR #783, opened 2021-03-08
[DC-1458] unit normalization should store _unit_mapping in sandbox dataset

lrwb-aou:
* updates unit normalization code
* updates existing integration test

### DIFF
--- a/data_steward/cdr_cleaner/cleaning_rules/unit_normalization.py
+++ b/data_steward/cdr_cleaner/cleaning_rules/unit_normalization.py
@@ -32,13 +32,13 @@ UNIT_MAPPING_TABLE_DISPOSITION = bq.bigquery.job.WriteDisposition.WRITE_EMPTY
 
 SANDBOX_UNITS_QUERY = JINJA_ENV.from_string("""
 CREATE OR REPLACE TABLE
-    `{{project_id}}.{{sandbox_dataset}}.{{intermediary_table}}` AS(
+    `{{project_id}}.{{sandbox_dataset_id}}.{{intermediary_table}}` AS(
 SELECT
     m.*
   FROM
   `{{project_id}}.{{dataset_id}}.{{measurement_table}}` as m
 INNER JOIN
-  `{{project_id}}.{{dataset_id}}.{{unit_table_name}}` as um
+  `{{project_id}}.{{sandbox_dataset_id}}.{{unit_table_name}}` as um
 USING
   (measurement_concept_id,
     unit_concept_id))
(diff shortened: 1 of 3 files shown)


data_steward/cdr_cleaner/cleaning_rules/unit_normalization.py
```diff
@@ -175,14 +178,14 @@ def setup_rule(self, client=None):
         # Uploading data to _unit_mapping table
         unit_mappings_csv_path = os.path.join(resources.resource_files_path,
                                               UNIT_MAPPING_FILE)
-        job = bq.upload_csv_data_to_bq_table(client, self.dataset_id,
+        job = bq.upload_csv_data_to_bq_table(client, self.sandbox_dataset_id,
```
ChaoPang, line 181: This is not within the scope of the ticket, however, [`bq.upload_csv_data_to_bq_table` ](https://github.com/all-of-us/curation/blob/6049de9db6ad8ff5db1ad80a086421e3d9c8944e/data_steward/utils/bq.py#L117)returns the job results object instead of the job object

ChaoPang (changes requested):
It looks good except for one minor comment below.

data_steward/cdr_cleaner/cleaning_rules/unit_normalization.py
lrwb-aou: @ChaoPang I added a doc string about the errors that could be raised via the `bq.upload_csv_data_to_bq_table` helper function.  I changed the variable name `job` to `result`.  I removed the erroneous call to `job.result()`.

ChaoPang (approved):
Looks good! However, the corresponding integration test didn't pass for some reason. I approved it but please make sure the integration test passes.

data_steward/cdr_cleaner/cleaning_rules/unit_normalization.py
ChaoPang: @lrwb-aou thanks for adding the docstring!

ChaoPang (changes requested):
Looks good! 

But the corresponding integration test didn't pass because `_unit_mapping` already existed in the test `sandbox_dataset` created by CircleCI. It seems that `unit_normalization.py` assumes `_unit_mapping` doesn't exist in the `sandbox_dataset`, and fails if `_unit_mapping` has been created before. Do we want to add `exists_ok=True` to `create_tables` for creating `_unit_mapping`?

data_steward/cdr_cleaner/cleaning_rules/unit_normalization.py
```diff
@@ -163,10 +164,16 @@ def setup_rule(self, client=None):
...
+            the load job fails. Error raised from bq.upload_csv_data_to_bq_table
+            helper function.
         """
 
         # creating _unit_mapping table
-        unit_mapping_table = f'{self.project_id}.{self.dataset_id}.{UNIT_MAPPING_TABLE}'
+        unit_mapping_table = (f'{self.project_id}.'
+                              f'{self.sandbox_dataset_id}.'
+                              f'{UNIT_MAPPING_TABLE}')
         bq.create_tables(
```
ChaoPang, line 177: @lrwb-aou do we want to add exists_ok=True to create_tables for creating _unit_mapping?

ChaoPang (approved):
Looks good!

PR #784, opened 2021-03-08
[DC-1459] Add _cdr_metadata table QA handoff date if creating deid_base dataset

dcampbell-vumc:
* Added add_argument for qa_handoff_date which will default to today's
  date
* Added *args to create_tier function

### DIFF
--- a/data_steward/tools/create_tier.py
+++ b/data_steward/tools/create_tier.py
@@ -2,6 +2,7 @@
 import argparse
 import logging
 import re
+from datetime import datetime
 
 # Third party imports
 
@@ -17,6 +18,8 @@ LOGGER = logging.getLogger(__name__)
 
 TIER_LIST = ['controlled', 'registered']
 DEID_STAGE_LIST = ['deid', 'deid_base', 'deid_clean']
+BASE = 'base'
+
 SCOPES = [
     'https://www.googleapis.com/auth/bigquery',
     'https://www.googleapis.com/auth/devstorage.read_write',
@@ -300,6 +303,13 @@ def parse_deid_args(args=None):
                         action='store_true',
                         required=False,
                         help='Log to the console as well as to a file.')
+    parser.add_argument(
+        '-qa',
+        '--qa_handoff_date',
+        dest='qa_handoff_date',
+        help='Will default to today\'s date for QA dataset handoff',
+        default=datetime.today(),
+        required=False)
 
     common_args, unknown_args = parser.parse_known_args(args)
     custom_args = clean_cdr._get_kwargs(unknown_args)
--- a/data_steward/tools/create_tier.py
+++ b/data_steward/tools/create_tier.py
@@ -9,6 +9,7 @@ from datetime import datetime
 # Project imports
 from utils import auth
 from utils import bq
+from common import JINJA_ENV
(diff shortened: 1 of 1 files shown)


data_steward/tools/create_tier.py
```diff
@@ -15,6 +16,7 @@
 
 TIER_LIST = ['controlled', 'registered']
 DEID_STAGE_LIST = ['deid', 'base', 'clean']
+BASE = 'base'
```
lrwb-aou, line 19: @dcampbell-vumc You need to fetch the latest changes and rebase, the values in `DEID_STAGE_LIST` are stale.

data_steward/tools/create_tier.py
```diff
@@ -254,16 +256,29 @@ def parse_deid_args(args=None):
                         action='store_true',
                         required=False,
                         help='Log to the console as well as to a file.')
+    parser.add_argument(
+        '-qa',
+        '--qa_handoff_date',
+        dest='qa_handoff_date',
+        help='Will default to today\'s date for QA dataset handoff',
+        default=datetime.today(),
+        required=False)
```
lrwb-aou, line 265: You don't need to input the date from the command line.  You can just use the date from a python utility.  This is an interpreter example code.

```
>>> from datetime import datetime
>>> datetime.now()
datetime.datetime(2021, 3, 8, 17, 12, 14, 553973)
>>> # one line to get the current date
>>> datetime.strftime(datetime.now(), '%Y-%m-%d')
'2021-03-08'
```

data_steward/tools/create_tier.py
```diff
@@ -254,16 +256,29 @@ def parse_deid_args(args=None):
...
-                args.target_principal)
+
+    if args.deid_stage == BASE:
+        create_tier(args.credentials_filepath, args.project_id, args.tier,
+                    args.idataset, args.release_tag, args.deid_stage,
+                    args.target_principal, args.qa_handoff_date)
+    else:
+        create_tier(args.credentials_filepath, args.project_id, args.tier,
+                    args.idataset, args.release_tag, args.deid_stage,
+                    args.target_principal)
```
lrwb-aou, line 281: There is no need to change lines 264-266.  You can just get the date in `create_tier()` if the datastage contains `base` in it's string name.  No need for this conditional.

lrwb-aou (changes requested):
You're on the right track.  But I think you need to update your branch and make some changes.  We can talk about it tomorrow if you would like.

data_steward/tools/create_tier.py
```diff
@@ -17,11 +19,18 @@
...
+BASE = 'base'
+
 SCOPES = [
     'https://www.googleapis.com/auth/bigquery',
     'https://www.googleapis.com/auth/devstorage.read_write',
 ]
 
+ADD_ETL_METADATA_QUERY = JINJA_ENV.from_string("""
+INSERT INTO `{project_id}.{dataset_id}._cdr_metadata` (qa_handoff_date)
+VALUES ({field_value})
```
lrwb-aou, line 31: I think you need to wrap this value in single quotes.
```suggestion
VALUES ('{field_value}')
```

lrwb-aou (approved):
looks good!

PR #785, opened 2021-03-09
[DC-1463] Exclude cope survey tables from CDM tables

nishanthpp93:
* [DC-1463] Exclude cope survey tables from CDM tables and add test


jp3477 (approved):
This all looks good!

PR #786, opened 2021-03-09
[DC-1271] updated remove_records_with_wrong_date.py to use cutoff_date instead of current_date

ksdkalluri:


amcgrenera-vumc (commented):
This looks good to me and is passing locally when I run the unit tests. Only concern is if this affects anything that will run since it's adding in a required argument?

amcgrenera-vumc (approved):
This looks good, thanks for checking and adding the parameter in the bash scripts. Once integration tests pass you are good to merge.

PR #787, opened 2021-03-10
[DC-1466] [P0 - domain alignment] some records are not rerouted but dropped

ChaoPang:


jp3477 (approved):
Approved! Thanks for the walkthrough

PR #788, opened 2021-03-10
[DC-1050] [P2] Domain Alignment Cleaning Rule needs to update Mapping tables when records are moved to another table

ChaoPang:


ksdkalluri (approved):
looks good!

PR #789, opened 2021-03-11
[DC-1471] Fix Missing back-ticks in ehr_union.py and add arguments destination

ksdkalluri:


dcampbell-vumc (approved):
Went through and verified back-ticks were present where necessary.

PR #792, opened 2021-03-12
[DC-1478] Remove deactivated participants script fails for tables without datetime fields

ksdkalluri:
Fix the retract utils script works with staging dataset and add a condition on fields that are missing end_datetime

### DIFF
--- a/data_steward/retraction/retract_utils.py
+++ b/data_steward/retraction/retract_utils.py
@@ -127,8 +127,8 @@ def get_datasets_list(project_id, dataset_ids_list):
     dataset_ids = [
         dataset_id for dataset_id in dataset_ids
         if get_dataset_type(dataset_id) != common.OTHER and
-        not is_sandbox_dataset(dataset_id) and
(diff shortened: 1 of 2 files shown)


data_steward/retraction/retract_utils.py
```diff
@@ -127,8 +127,8 @@ def get_datasets_list(project_id, dataset_ids_list):
     dataset_ids = [
         dataset_id for dataset_id in dataset_ids
         if get_dataset_type(dataset_id) != common.OTHER and
-        not is_sandbox_dataset(dataset_id) and
-        not is_staging_dataset(dataset_id)
+        not is_sandbox_dataset(dataset_id)
+        # and not is_staging_dataset(dataset_id)
```
nishanthpp93, line 131: I think you can remove this line.

nishanthpp93 (approved):
I think I remember this from the CDR run. Looks good! Added a note that you can remove the comment

PR #793, opened 2021-03-12
[DC-1479] EHR data cutoff invalid query for tables without datetime fields

ksdkalluri:
Fix submission cutoff date comparison by using '1900-01-01' for null values


dcampbell-vumc (approved):
Unit and integration tests are both passing locally for me! This is good to go!

PR #794, opened 2021-03-12
[DC-1480] Remove records with wrong date fails due to cutoff_date param position

ksdkalluri:
Fix Cutoff_date param in remove_records_with_wrong_date.py


nishanthpp93 (approved):
Looks good

PR #796, opened 2021-03-12
[DC-1646] Remove staging_sandbox creation and update its usages with unioned_ehr_sandbox

ksdkalluri:
Remove staging_sandbox creation and update its usages with unioned_ehr_sandbox

* move ehr_submission_data_cutoff to top of the unioned_ehr cleaning_rules

### DIFF
--- a/data_steward/cdr_cleaner/clean_cdr.py
+++ b/data_steward/cdr_cleaner/clean_cdr.py
@@ -108,6 +108,8 @@ EHR_CLEANING_CLASSES = [
 ]
 
 UNIONED_EHR_CLEANING_CLASSES = [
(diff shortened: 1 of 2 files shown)


jp3477 (approved):
Approved! But I am curious why you moved up the run order of the EhrSubmissionDataCutoff rule?

ksdkalluri:
> Approved! But I am curious why you moved up the run order of the EhrSubmissionDataCutoff rule?

This is done while running the pipeline since we do not have a process in place to get only data submitted till a cut off in EHR submissions so we thought that should be the first cleaning rule we should be running for the unioned_ehr so the other cleaning rules should not have to clean the unnecessary data.

PR #798, opened 2021-03-15
[DC-1487] Fixes for remove_non_matching_participant

mark-velez:
* Return query dict rather than list for sandbox queries
 * Make ehr_dataset_id, validation_dataset_id args required
 * In generate_combined_dataset.sh add ehr_dataset param


dcampbell-vumc (approved):
Looks good and the unit test passes as expected 👍🏾

PR #801, opened 2021-03-19
[DC-1378] update add_cdr_metadata.py to be called from another modules

dcampbell-vumc:
In `add_cdr_metadata.py` module:
* Updated queries to use jinja template
* Updated/created doc strings for functions
* Added clarifying comments
* Updated query generation to use render instead of format
* Split `__name__ == '__main__'` to two functions to allow other module to call `add_cdr_metadata.py`
     * Created `parser_cdr_metadata_args` and `main`
* `__name__ == '__main__'` only calls `main()` function now

In `add_cdr_metadata_test.py` module:
* Added variables to `setUp` function to be used in `test_parse_cdr_metadata_args` function
* Updated query generation to use render
* Fixed `actual_query` in `test_etl_metadata_query` and `test_copy_query` to not have extra carriage return
* Created `test_parser_cdr_metadata_args` function to:
    * Test if incorrect parameters are given, `project_id` and `component`
      are required args
    * Test if incorrect choice for component are given, component can
      only be `COPY`, `INSERT`, or `CREATE`
    * Test if overall, all the correct parameters are given

### DIFF
--- a/data_steward/tools/add_cdr_metadata.py
+++ b/data_steward/tools/add_cdr_metadata.py
@@ -1,8 +1,20 @@
+"""
+This module will populate the cdr_metadata table with CDR metadata in order to version control the pipeline for each
+    release. The metadata to be included is etl_version, ehr_source, ehr_cutoff_date, rdr_source, rdr_export_date,
+    cdr_generation_date, qa_handoff_date, and vocabulary_version.
+
+Original Issues: DC-1378, DC-347
+"""
+
+# Python imports
 import argparse
 
+# Project imports
 import bq_utils
 import resources
 from utils import bq
+from common import JINJA_ENV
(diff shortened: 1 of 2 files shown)


data_steward/tools/add_cdr_metadata.py
```diff
@@ -162,7 +207,11 @@ def add_metadata(dataset_id, project_id, table_fields, field_values=None):
             args.rdr_source, args.rdr_export_date, args.cdr_generation_date,
             args.qa_handoff_date, args.vocabulary_version
         ]))
-    print(field_values_dict)
+
```
ChaoPang, line 210: nit: move `field_values_dict` into `if args.component == INSERT:` because it's only used there.

ChaoPang (approved):
The rest looks good to me

PR #802, opened 2021-03-22
[DC-1506] Use Double Brackets in Jinja Template

dcampbell-vumc:
* Updated `ADD_ETL_METADATA_QUERY` to have double brackets
* Updated the `dataset_id` in the rendering of `ADD_ETL_METADATA_QUERY`
  so that it will use the STAGING dataset
* Updated the `field_value` variable name to match the one used in
  `ADD_ETL_METADATA_QUERY`


ChaoPang (approved):
Looks good!

PR #803, opened 2021-03-23
[DC-1513] CleanMappingExtTables rule introduces duplicates

nishanthpp93:
* [DC-1513] Update mapping queries
* [DC-1513] Use single query template


dcampbell-vumc (commented):
These changes make sense to me and how this module won't end up duplicating the possibly already duplicated OMOP records. 
The ticket does say to update the integration test for the `clean_mapping.py` module however one doesn't exist yet. Has there been a ticket created to go back and create this integration test?

nishanthpp93:
Good point, I don't think it exists yet. I'll reach out to Heather so it can be added along with the rest of the integration tests.

dcampbell-vumc (approved):
Approved!

PR #805, opened 2021-03-24
[DC-1015] Organize the table schemas directory

ksdkalluri:


data_steward/tools/import_rdr_omop.sh
```diff
@@ -85,7 +85,7 @@ for file in $cdm_files; do
   table_name=${filename%.*}
   echo "Importing ${RDR_DATASET}.${table_name}..."
   CLUSTERING_ARGS=""
-  if grep -q person_id resource_files/fields/"${table_name}".json; then
+  if grep -q person_id resource_files/schemas/"${table_name}".json; then
```
mark-velez, line 88: There are also references to `additional_rdr_tables/` in this file that need to be changed to `rdr`. In these cases it's better to declare `RDR_FIELDS_DIR` so we can change e.g.
`${FIELDS_DIR}/additional_rdr_tables/${table_name}.json` to `${RDR_FIELDS_DIR}/${table_name}.json`.

mark-velez, line 88: This will yield a file not found error. Is it ok to use `schemas/cdm/clinical/` or are there other files we need to load?

mark-velez:
There is an invalid schema path in [load_vocab.sh](https://github.com/all-of-us/curation/blob/kk/dc_1015/data_steward/tools/load_vocab.sh#L98) that should be updated

data_steward/tools/import_rdr_omop.sh
```diff
@@ -83,19 +84,16 @@ fi
 for file in $cdm_files; do
   filename=$(basename ${file})
   table_name=${filename%.*}
+  schema_file="$(find "${FIELDS_DIR}" -type f -iname "${table_name}".json)"
```
mark-velez, line 87: Looks like you figured it out! I would just suggest that it check that one (and only one) schema is located or otherwise fail.

```suggestion
  # Locate the schema file (ensuring one and only one is found)
  schema_file=$(find "${FIELDS_DIR}" -type f -iname "${table_name}.json")
  schema_count=$(echo "$schema_file" | wc -l);
  if [[ 1 -ne schema_count ]]; then
    echo "Import failed: ${schema_count} schemas were found for table ${table_name}."
    exit 1
  fi
```

data_steward/tools/import_rdr_omop.sh
```diff
@@ -54,7 +54,8 @@ fi
 ROOT_DIR=$(git rev-parse --show-toplevel)
 DATA_STEWARD_DIR="${ROOT_DIR}/data_steward"
 TOOLS_DIR="${DATA_STEWARD_DIR}/tools"
-FIELDS_DIR="${DATA_STEWARD_DIR}/resource_files/fields"
+FIELDS_DIR="${DATA_STEWARD_DIR}/resource_files/schemas"
+RDR_FIELDS_DIR="${FIELDS_DIR}/rdr/"
```
mark-velez, line 58: There's an extra trailing slash here
```suggestion
RDR_FIELDS_DIR="${FIELDS_DIR}/rdr"
```

data_steward/tools/import_rdr_omop.sh
```diff
@@ -72,8 +73,8 @@ bq mk -f --description "RDR DUMP loaded from ${RDR_DIRECTORY} dated ${RDR_UPLOAD
 
 python "${DATA_STEWARD_DIR}/cdm.py" "${RDR_DATASET}"
 # accommodate new file "pid_rid_mapping" by RDR
-bq mk --table "${GOOGLE_CLOUD_PROJECT}:${RDR_DATASET}.pid_rid_mapping" "${FIELDS_DIR}/additional_rdr_tables/pid_rid_mapping.json"
-bq mk --table "${GOOGLE_CLOUD_PROJECT}:${RDR_DATASET}.cope_survey_semantic_version_map" "${FIELDS_DIR}/additional_rdr_tables/cope_survey_semantic_version_map.json"
+bq mk --table "${GOOGLE_CLOUD_PROJECT}:${RDR_DATASET}.pid_rid_mapping" "${FIELDS_DIR}/pid_rid_mapping.json"
```
mark-velez, line 76:
```suggestion
bq mk --table "${GOOGLE_CLOUD_PROJECT}:${RDR_DATASET}.pid_rid_mapping" "${RDR_FIELDS_DIR}/pid_rid_mapping.json"
```

mark-velez (approved):

PR #806, opened 2021-03-24
[DC-1198] Integration Test for Missing Race Responses CR

jp3477:
Converted this integration to the new base class format.

### DIFF
--- a/tests/integration_tests/data_steward/cdr_cleaner/cleaning_rules/repopulate_person_post_deid_test.py
+++ b/tests/integration_tests/data_steward/cdr_cleaner/cleaning_rules/repopulate_person_post_deid_test.py
@@ -1,3 +1,31 @@
+"""
+Integration test for repopulate_person_post_deid module
+
+The de-id scripts removes all fields in the person table except for the person_id and the birthdate_time field.
+Before CDR handoff to the Workbench team, we need to repopulate the following fields with demographic information
+from the observation table.
+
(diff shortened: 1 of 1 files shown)


tests/integration_tests/data_steward/cdr_cleaner/cleaning_rules/repopulate_person_post_deid_test.py
```diff
@@ -1,161 +1,174 @@
...
+
+    def test_repopulate_person_post_deid(self):
+        """
+        Tests that the specifications for queries perform as designed.
+
+        Validates pre conditions, tests execution, and post conditions based on the load
+        statements and the tables_and_counts variable.        
+        """
+
+        queries = []
```
dcampbell-vumc, line 105: nit: this line can be removed because of `queries` on line 147

dcampbell-vumc (changes requested):
Everything looks great here! There is a small nit that can be changed.

tests/integration_tests/data_steward/cdr_cleaner/cleaning_rules/repopulate_person_post_deid_test.py
jp3477: Sounds good, just pushed

dcampbell-vumc (approved):
Good to go!

PR #808, opened 2021-03-25
[DC-1204] Integration Test for Temporal Data Point CR

nishanthpp93:
* [DC-1204] Create integration test
* [DC-1204] Update integration test and fix CR


dcampbell-vumc (approved):
Integration test passes as normal! This is good to go!

PR #809, opened 2021-03-25
[DC-1524] Manually run pid/rid for CT tables

amcgrenera-vumc:


tests/integration_tests/data_steward/cdr_cleaner/cleaning_rules/deid/ct_pid_rid_map_test.py
```diff
@@ -139,7 +139,9 @@ def test_field_cleaning(self):
                 (50003, 345, 500, datetime.fromisoformat('2020-08-17').date(),
                  datetime.fromisoformat('2020-08-17 13:00:00+00:00'), 12),
                 (50004, 789, 800, datetime.fromisoformat('2020-08-17').date(),
-                 datetime.fromisoformat('2020-08-17 12:00:00+00:00'), 13)
+                 datetime.fromisoformat('2020-08-17 12:00:00+00:00'), 13),
+                (50005, 3456, 1000, datetime.fromisoformat('2020-08-17').date(),
+                 datetime.fromisoformat('2020-08-17 11:00:00+00:00'), 14)
```
lrwb-aou, line 144: Does this mean that `person_id = 3456` continues to exist in the data after this rule executes?  This means this rule heavily depends on a correct notebook to find any problems.  But a test does exist for a warning log message?

data_steward/cdr_cleaner/cleaning_rules/deid/pid_rid_map.py
```diff
@@ -17,12 +17,10 @@
...
-SELECT
-    t.* EXCEPT (person_id),
-    d.research_id as person_id
-FROM `{{input_table.project}}.{{input_table.dataset_id}}.{{input_table.table_id}}` t
-JOIN `{{deid_map.project}}.{{deid_map.dataset_id}}.{{deid_map.table_id}}` d
-ON t.person_id = d.person_id
+UPDATE `{{input_table.project}}.{{input_table.dataset_id}}.{{input_table.table_id}}` t
+SET person_id = d.research_id
+FROM `{{deid_map.project}}.{{deid_map.dataset_id}}.{{deid_map.table_id}}` d
+WHERE t.person_id = d.person_id
```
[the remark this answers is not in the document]
amcgrenera-vumc: good catch, I don't see this changing anything, other than being more explicit. Updating now!

tests/integration_tests/data_steward/cdr_cleaner/cleaning_rules/deid/ct_pid_rid_map_test.py
amcgrenera-vumc: @dcampbell-vumc @nishanthpp93 correct me if I'm wrong. We added this in because while it should raise the `WARNING` it should also still remain in the table. This record doesn't have a matching record in the `pid_rid_mapping` table, so it remains the same. However, should there be something in place where if a record doesn't have a match in the pid_rid table it should be removed?

lrwb-aou: @amcgrenera-vumc That is how RT handles unmapped pids.  I think this may be fine for the moment and something to consider.  We do have a CT notebook that checks for pid/rid mappings.  It's just good for this to be a conscious choice instead of just a curious byproduct of the SQL.

lrwb-aou (approved):
Looks good.

tests/integration_tests/data_steward/cdr_cleaner/cleaning_rules/deid/ct_pid_rid_map_test.py
lrwb-aou: After talking with @mark-velez , we think the default should be to drop records where the pid/rid can't be mapped.  I think we should merge/close this issue as it is and do this work in a different PR.

amcgrenera-vumc: Ok great, this is good to go then, I will close this out!

PR #811, opened 2021-03-29
[DC-1510] Prevent overwriting generalized zip code data in CT deid_base

jp3477:
Restrict cleaning rule operation on zip code observations.

### DIFF
--- a/data_steward/cdr_cleaner/cleaning_rules/fill_source_value_text_fields.py
+++ b/data_steward/cdr_cleaner/cleaning_rules/fill_source_value_text_fields.py
@@ -52,7 +52,7 @@ import logging
 
 import resources
 from cdr_cleaner.cleaning_rules.base_cleaning_rule import BaseCleaningRule, query_spec_list
-from common import JINJA_ENV
+from common import JINJA_ENV, OBSERVATION
 from constants import bq_utils as bq_consts
 from constants.cdr_cleaner import clean_cdr as cdr_consts
 
@@ -146,7 +146,7 @@ def get_fields_dict(table_name, fields):
     return fields_to_replace
 
 
-def get_modified_columns(fields, fields_to_replace):
+def get_modified_columns(fields, fields_to_replace, table=''):
     """
 
     This method updates the columns by adding prefix to each column if the column is being replaced and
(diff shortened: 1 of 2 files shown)


data_steward/cdr_cleaner/cleaning_rules/fill_source_value_text_fields.py
```diff
@@ -146,22 +146,30 @@ def get_fields_dict(table_name, fields):
     return fields_to_replace
 
 
-def get_modified_columns(fields, fields_to_replace):
+def get_modified_columns(fields, fields_to_replace, table=''):
```
ChaoPang, line 149: nit: can we use `None` instead an empty string, not sure if this would work though. 
```suggestion
def get_modified_columns(fields, fields_to_replace, table=None):
```

ChaoPang (changes requested):
The change looks good, however, the integration test fails due to the missing `care_site` table. The reason is that `fill_source_value_text_fields` iterates through all CDM tables and generates a query for every single table including `care_site`, it seems that the integration test only creates `observation` and `concept`. There are two potential solutions for this, 
1. create all empty CDM tables in the integration test
2. remove tables that do not exist from `self.affect_tables` in `setup` in the cleaning rule, this can be achieved by using ` get_tables_in_dataset`, refer to this example https://github.com/all-of-us/curation/blob/c55a75abd812aac654b367c108d2a8bd7358199b/data_steward/cdr_cleaner/cleaning_rules/id_deduplicate.py#L122

data_steward/cdr_cleaner/cleaning_rules/fill_source_value_text_fields.py
jp3477: Yes we can

jp3477:
> The change looks good, however, the integration test fails due to the missing `care_site` table. The reason is that `fill_source_value_text_fields` iterates through all CDM tables and generates a query for every single table including `care_site`, it seems that the integration test only creates `observation` and `concept`. There are two potential solutions for this,
> 
> 1. create all empty CDM tables in the integration test
> 2. remove tables that do not exist from `self.affect_tables` in `setup` in the cleaning rule, this can be achieved by using ` get_tables_in_dataset`, refer to this example https://github.com/all-of-us/curation/blob/c55a75abd812aac654b367c108d2a8bd7358199b/data_steward/cdr_cleaner/cleaning_rules/id_deduplicate.py#L122

I went with the first option.

ChaoPang (approved):
Approved. Please merge after the integration test passes.

PR #813, opened 2021-04-02
[DC-1511] Insert instead of Update is used to add cdr metadata

dcampbell-vumc:
* Updated logic to update `qa_handoff_date` if base is in the deid_tier
* Created `test_qa_handoff_date_update` function to test changes in
  create_tier.py


ChaoPang (changes requested):
It looks good but the unit test is broken due to the changes made to `create_tier`. I will approve after the test passes

ChaoPang (approved):
Approved.

PR #814, opened 2021-04-02
[DC-1518] [DC-1543] create primary pid/rid/shift storage script

lrwb-aou:
Script saves only new mappings to primary mapping table and adds randomized shift integer between 1 and 365 inclusive.
Corrects an f-string in the auth module.
Adds variable definitions to common module.
Adds table definition for pipeline_tables.pid_rid_mapping.
updates unit test because of name clash with newly added table definition.
simplifies rdr_additional_table in resources module.


jp3477 (approved):
Everything worked as expected. Just noted a small typo which is NIT.

PR #816, opened 2021-04-05
[DC-1514] person_ext table copy fails

amcgrenera-vumc:


data_steward/cdr_cleaner/cleaning_rules/create_person_ext_table.py
```diff
@@ -56,7 +70,7 @@ def __init__(self, project_id, dataset_id, sandbox_dataset_id):
         DO NOT REMOVE ORIGINAL JIRA ISSUE NUMBERS!
```
lrwb-aou, line 70: @amcgrenera-vumc One thing, add the `table_namer` parameter to the initialization signature on line 64 and pass it to the super class in line 73 call.  So we can start using shared sandboxes.

tests/integration_tests/data_steward/cdr_cleaner/cleaning_rules/create_person_ext_table_test.py
```diff
@@ -0,0 +1,201 @@
...
+                concept_id INT64,
+                concept_code STRING,
+                concept_name STRING);
+            INSERT INTO
+              `{{project_id}}.{{dataset_id}}.concept` (
+                concept_id,
+                concept_code,
+                concept_name)
+            VALUES
+              (1585266, '', 'PII State: CA'),
```
lrwb-aou, line 140: Possibly related, there is no `concept_code` defined for the state here.  Just the name, which isn't normally used as part of cleaning rule criteria.  
```suggestion
             (1585266, 'PIIState_CA', 'PII State: CA'),
```

amcgrenera-vumc: adjusted!

tests/integration_tests/data_steward/cdr_cleaner/cleaning_rules/create_person_ext_table_test.py
```diff
@@ -0,0 +1,201 @@
...
+                                45878463, 1585847, 'SexAtBirth_Female'),
+                               (345, 'PPI/PM', 1585266, 'PII State: CA',
+                                45878463, 1585847, 'SexAtBirth_Female'),
+                               (678, 'PPI/PM', 1585266, 'PII State: CA',
+                                45878463, 1585847, 'SexAtBirth_Female'),
+                               (910, 'PPI/PM', 1585266, 'PII State: CA',
+                                45878463, 1585847, 'SexAtBirth_Female')]
+        }]
+
+        self.default_test(tables_and_counts)
```
lrwb-aou, line 199: @amcgrenera-vumc You should add a `tearDown(self):` function to this test.  It isn't cleaning up after itself because it is not relying on defining all the fully qualified tables required to set this test up.  Meaning it leaves a dirty test environment for the next test.  And the next test shouldn't have to worry about cleaning up from the previous test.  Your SQL statements `DROP TABLE IF EXISTS...` are symptomatic of the missing `tearDown()`.  Consider adding:

```suggestion
        self.default_test(tables_and_counts)
            
    def tearDown(self):
        tables = ['observation', 'observation_ext', 'concept', 'person']
        for table in tables:
            self.client.delete_table('.'.join([self.fq_dataset_name, table]), not_found_ok=True)
          
        super().tearDown() 
```

tests/integration_tests/data_steward/cdr_cleaner/cleaning_rules/create_person_ext_table_test.py
```diff
@@ -0,0 +1,201 @@
...
+              (1122, 123, 1585847, 45878463, 1585845),
+              (3344, 345, 1585847, 45878463, 1585845),
+              (5566, 678, 1585847, 45878463, 1585845),
+              (7788, 910, 1585847, 45878463, 1585845)
+            """).render(project_id=self.project_id, dataset_id=self.dataset_id)
+
+        # test data for concept table
+        concept_data_query = self.jinja_env.from_string("""
+            DROP TABLE IF EXISTS
+              `{{project_id}}.{{dataset_id}}.concept`;
```
lrwb-aou, line 130: These are symptomatic of a test setup/initialization/tearDown problem.

tests/integration_tests/data_steward/cdr_cleaner/cleaning_rules/create_person_ext_table_test.py
amcgrenera-vumc: good catch, I will add this in, thank you!

data_steward/cdr_cleaner/cleaning_rules/create_person_ext_table.py
amcgrenera-vumc: Let me know if I've added correctly.

tests/integration_tests/data_steward/cdr_cleaner/cleaning_rules/create_person_ext_table_test.py
amcgrenera-vumc: Do you suggest I take out `DROP TABLE` and just have it be `CREATE TABLE`, since I've added in the `tearDown`?

lrwb-aou (changes requested):
[create_person_ext_test_patch.txt](https://github.com/all-of-us/curation/files/6267196/create_person_ext_test_patch.txt)
Try this patch.  It uses the test base to create the tables with functional schemas and clean them up.  You're having to do a lot of extra stuff the base class is able to do for you.

amcgrenera-vumc:
@lrwb-aou thank you for sending that patch! I've added it, I also left in the `tearDown`, but let me know if that's not necessary anymore?

lrwb-aou:
> @lrwb-aou thank you for sending that patch! I've added it, I also left in the `tearDown`, but let me know if that's not necessary anymore?

@amcgrenera-vumc The `tearDown` is not necessary anymore.  It is handled by the base class because all those tables are now part of the fully qualified class table names.

tests/integration_tests/data_steward/cdr_cleaner/cleaning_rules/create_person_ext_table_test.py
```diff
@@ -0,0 +1,194 @@
...
+
+        self.default_test(tables_and_counts)
+
+    def tearDown(self):
+        tables = ['observation', 'observation_ext', 'concept', 'person']
+        for table in tables:
+            self.client.delete_table('.'.join([self.fq_dataset_name, table]),
+                                     not_found_ok=True)
+
+        super().tearDown()
```
lrwb-aou, line 194: This is repeating stuff done in `curation/tests/integration_tests/data_steward/cdr_cleaner/cleaning_rules/bigquery_tests_base.py`.  I suggested adding it before because you had not previously added those four tables to `cls.fq_table_names` in the `setUpClass` method.

amcgrenera-vumc: ah ok that makes sense, removing now!

tests/integration_tests/data_steward/cdr_cleaner/cleaning_rules/create_person_ext_table_test.py
```diff
@@ -0,0 +1,186 @@
...
+sex_at_birth_source_concept_id: value_source_concept_id in observation where observation_source_concept_id = 1585845
+sex_at_birth_source_value: concept_code in the concept table where joining from observation where 
+observation_source_concept_id = 1585845
+"""
+
+# Python imports
+import os
+
+# Project imports
+from app_identity import PROJECT_ID
```
lrwb-aou, line 27: nit:  this works, but is repeating work that is already done by the app_identity module.

```suggestion
from app_identity import get_application_identity
```

tests/integration_tests/data_steward/cdr_cleaner/cleaning_rules/create_person_ext_table_test.py
```diff
@@ -0,0 +1,186 @@
...
+    @classmethod
+    def setUpClass(cls):
+        print('**************************************************************')
+        print(cls.__name__)
+        print('**************************************************************')
+
+        super().initialize_class_vars()
+
+        # set the test project identifier
+        project_id = os.environ.get(PROJECT_ID)
```
lrwb-aou, line 43: if the import changes, this should use:

```suggestion
        project_id = get_application_id()
```

lrwb-aou (approved):
I added an optional suggestion a few days ago.  Just didn't approve it at the time.  Do you have any concerns or suggestions?

amcgrenera-vumc:
@lrwb-aou my bad, realized i never pushed, made that adjustment!

lrwb-aou:
no worries.  I just wanted to push the PR forward.

amcgrenera-vumc:
once the integration test passes I can merge, if all good on your end!

lrwb-aou:
@amcgrenera-vumc I've approved it.  It's good.

PR #817, opened 2021-04-05
[DC-1512] Person records are duplicated in observation and observation_ext table

jp3477:
Adds an integer representing the HPO site to the observation id in order to eliminate duplicates.


ksdkalluri (changes requested):
Hi, @jp3477  this approach looks good. is it possible to add a unit test to test the constant factor generation?

jp3477:
> Hi, @jp3477 this approach looks good. is it possible to add a unit test to test the constant factor generation?

Sure, will add one.

ksdkalluri (approved):
.Looks good to me! but the integration tests are failing for ehr_union_test and this is good to merge upon fixing them.

ksdkalluri (approved):

PR #819, opened 2021-04-06
[DC-1215] add project name to log files

dcarbone:
Apologies for the extended dev cycle down the wrong path.  I believe the following is sufficient to produce the desired result.


data_steward/utils/pipeline_logging.py
```diff
@@ -23,7 +23,7 @@
 
 LOG_FORMAT = '%(asctime)s - %(levelname)s - %(name)s - %(message)s'
 LOG_DATEFMT = '%Y-%m-%d %H:%M:%S'
-FILENAME_FMT = '%Y%m%d'
+FILENAME_FMT = '%Y%m%d-' + os.getenv('GOOGLE_CLOUD_PROJECT', 'UNDEFINED_GOOGLE_PROJECT')
```
lrwb-aou, line 26: A few notes:

1. We generally use f-strings where possible.  I think this is newish in python 3.X.
2. We already get the project_id via an import variable in https://github.com/all-of-us/curation/blob/develop/data_steward/app_identity.py#L6

With both of those considerations (but I don't think f-strings apply in this special situation), we should probably be doing:
```suggestion
FILENAME_FMT = '%Y%m%d-' + get_application_id()
```
and `from app_identity import get_application_id`.

lrwb-aou: Also, the linting check will fail if the PR title doesn't conform to https://github.com/all-of-us/curation/blob/develop/tools/validate_pr_title.sh .  Ticket number should be in brackets followed by a space.

data_steward/utils/pipeline_logging.py
```diff
@@ -16,14 +16,16 @@
...
 
 DEFAULT_LOG_DIR = os.path.join(resources.base_path, 'logs')
 """Default location for log files"""
 DEFAULT_LOG_LEVEL = logging.INFO
 
 LOG_FORMAT = '%(asctime)s - %(levelname)s - %(name)s - %(message)s'
 LOG_DATEFMT = '%Y-%m-%d %H:%M:%S'
-FILENAME_FMT = '%Y%m%d-' + os.getenv('GOOGLE_CLOUD_PROJECT', 'UNDEFINED_GOOGLE_PROJECT')
+app_id = get_application_id()
+FILENAME_FMT = f'%Y%m%d-{app_id}'
```
lrwb-aou, line 28: This would also work.
```
FILENAME_FMT = f'%Y%m%d-{get_application_id()}'
```

lrwb-aou (approved):
This is good.  I was fine with the config file changes if you need those as part of your environment setup too.

lrwb-aou:
@dcarbone I didn't look closely at the unit tests until now.  They are failing because app_identity is raising the RuntimeError when `GOOGLE_CLOUD_PROJECT` is not set.

data_steward/utils/pipeline_logging.py
dcarbone: I actually tried both, verified they both work, but in random googling I found more examples of this usage, so I assumed it was more "pythonic"

dcarbone:
> @dcarbone I didn't look closely at the unit tests until now. They are failing because app_identity is raising the RuntimeError when `GOOGLE_CLOUD_PROJECT` is not set.

Right, I assumed that is what was desired.  I wanted to add a new test for this specific feature anyway.

Off the type of my head, I feel there are 3 options:

1. Update tests to require a project before running
2. Provide the ability for the logger to be constructed properly if there is some code to be run outside a context where it needs a google project
3. Create a new handler that can be added explicitly when this file-naming functionality is desired.

Personally I like option 2.  If the project name isn't defined, simply don't bother trying to annotate the log filename.

Does that sound acceptable?  In this case, could also create an initial log entry that is like `WARN: GOOGLE_PROJECT_NAME not set at runtime!`, or something like that.

lrwb-aou:
> > @dcarbone I didn't look closely at the unit tests until now. They are failing because app_identity is raising the RuntimeError when `GOOGLE_CLOUD_PROJECT` is not set.
> 
> Right, I assumed that is what was desired. I wanted to add a new test for this specific feature anyway.
> 
> Off the type of my head, I feel there are 3 options:
> 
>     1. Update tests to require a project before running
> 
>     2. Provide the ability for the logger to be constructed properly if there is some code to be run outside a context where it needs a google project
> 
>     3. Create a new handler that can be added explicitly when this file-naming functionality is desired.
> 
> 
> Personally I like option 2. If the project name isn't defined, simply don't bother trying to annotate the log filename.
> 
> Does that sound acceptable? In this case, could also create an initial log entry that is like `WARN: GOOGLE_PROJECT_NAME not set at runtime!`, or something like that.

Option 2 sounds good.  I'd probably just use a try/except/else statement to achieve that.

data_steward/utils/pipeline_logging.py
```diff
@@ -16,27 +16,46 @@
...
+
+def _get_log_filename() -> str:
+    """
+    Construct runtime-specific log filename
+    """
+    try:
+        # attempt to add application id suffix (google project name)
+        app_id = get_application_id()
+        filename_suffix = f'-{app_id}'
+    except Exception:
```
lrwb-aou, line 51: I'd suggest explicitly catching `RuntimeError`, since that's what `get_application_id()` might raise.

data_steward/app_identity.py
```diff
@@ -12,7 +12,11 @@ def get_application_id():
...
 
-    if project_id:
-        return project_id
-    else:
-        raise RuntimeError('{} is not set.  Set and retry.'.format(PROJECT_ID))
+    # ensure project id is a non-empty string
+    if isinstance(project_id, str):
+        project_id = project_id.strip()
+        if project_id is not '':
+            return project_id
```
lrwb-aou, line 19: nit:  This works, but I'm not convinced it's necessary.  Environment variables are set when that first space is met.  So below, `FOO` is set to an empty string, `BAR` is good, and `BAZ= bad` results in a command not found error.  So, I don't think it's possible to have `FOO=<multiple blanks>` from an environment variable.  Also, environment variables are always going to give us [strings](https://docs.python.org/3/library/os.html#os.environ).
```
(env37) [Thu Apr 08 11:50:13] [~/Projects/curation/data_steward] lbiggers $ FOO=     BAR=good BAZ= bad python
-bash: bad: command not found
(env37) [Thu Apr 08 11:50:31] [~/Projects/curation/data_steward] lbiggers $ FOO=     BAR=good BAZ=bad  python
Python 3.7.2 (default, Feb 12 2019, 08:15:36) 
[Clang 10.0.0 (clang-1000.11.45.5)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> import os
>>> os.environ.get('FOO')
''
>>> os.environ.get('BAR')
'good'
>>> os.environ.get('BAZ')
'bad'
>>> 
```
The only way this could get messed up is if someone starts injecting environment variables in through python code.  Which could happen, but it **_really_** shouldn't.  If somebody is injecting keys/values, I would argue they should be injecting them into a [Namespace](https://docs.python.org/3/tutorial/classes.html#python-scopes-and-namespaces) not the environment dictionary.
```
>>> import os
>>> os.environ['GOOF']='     '
>>> foo(os.environ.get('GOOF'))
bar is :     
```
Finally, an empty string receives the same Truthy/Falsy treatment as None and zero.
```
(env37) [Thu Apr 08 11:54:56] [~/Projects/curation/data_steward] lbiggers $ python
Python 3.7.2 (default, Feb 12 2019, 08:15:36) 
[Clang 10.0.0 (clang-1000.11.45.5)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> def foo(bar):
...   if bar:
...     print(f'bar is :{bar}')
...   else:
...     print('bar is falsy')
... 
>>> foo('')
bar is falsy
>>> foo(None)
bar is falsy
>>> foo(0)
bar is falsy
>>> foo('  ')
bar is :  
>>> # but the above can't happen with true environment variables
... 
>>> foo('23')
bar is :23
>>> foo(23)
bar is :23
```

dcarbone: > nit: This works, but I'm not convinced it's necessary. Environment variables are set when that first space is met. So below, `FOO` is set to an empty string, `BAR` is good, and `BAZ= bad` results in a command not found error. So, I don't think it's possible to have `FOO=<multiple blanks>` from an environment variable. Also, environment variables are always going to give us [strings](https://docs.python.org/3/library/os.html#os.environ).
> 
> ```
> (env37) [Thu Apr 08 11:50:13] [~/Projects/curation/data_steward] lbiggers $ FOO=     BAR=good BAZ= bad python
> -bash: bad: command not found
> (env37) [Thu Apr 08 11:50:31] [~/Projects/curation/data_steward] lbiggers $ FOO=     BAR=good BAZ=bad  python
> Python 3.7.2 (default, Feb 12 2019, 08:15:36) 
> [Clang 10.0.0 (clang-1000.11.45.5)] on darwin
> Type "help", "copyright", "credits" or "license" for more information.
> >>> import os
> >>> os.environ.get('FOO')
> ''
> >>> os.environ.get('BAR')
> 'good'
> >>> os.environ.get('BAZ')
> 'bad'
> >>> 
> ```
>
> The only way this could get messed up is if someone starts injecting environment variables in through python code. Which could happen, but it **_really_** shouldn't. If somebody is injecting keys/values, I would argue they should be injecting them into a [Namespace](https://docs.python.org/3/tutorial/classes.html#python-scopes-and-namespaces) not the environment dictionary.
> 
> ```
> >>> import os
> >>> os.environ['GOOF']='     '
> >>> foo(os.environ.get('GOOF'))
> bar is :     
> ```
> 
> Finally, an empty string receives the same Truthy/Falsy treatment as None and zero.
> 
> ```
> (env37) [Thu Apr 08 11:54:56] [~/Projects/curation/data_steward] lbiggers $ python
> Python 3.7.2 (default, Feb 12 2019, 08:15:36) 
> [Clang 10.0.0 (clang-1000.11.45.5)] on darwin
> Type "help", "copyright", "credits" or "license" for more information.
> >>> def foo(bar):
> ...   if bar:
> ...     print(f'bar is :{bar}')
> ...   else:
> ...     print('bar is falsy')
> ... 
> >>> foo('')
> bar is falsy
> >>> foo(None)
> bar is falsy
> >>> foo(0)
> bar is falsy
> >>> foo('  ')
> bar is :  
> >>> # but the above can't happen with true environment variables
> ... 
> >>> foo('23')
> bar is :23
> >>> foo(23)
> bar is :23
> ```

I hear you on testing the "falsey" nature of the retrieved variable, however few small corrections:

1. it is very possible to have an empty string or space-prefix / suffixed envvar: 
    ```shell
    carbone@MANUS:~$ export PADDING='   padded    '
    dcarbone@MANUS:~$ export NO_PADDING='nopadding'
    dcarbone@MANUS:~$ printenv | grep PADDING
    PADDING=   padded    
    NO_PADDING=nopadding
    ```
2. when undefined, the return value from `os.getenv()` is actually `None`, not an empty str, and I did not want to try calling `.strip()` off a non-str object:
    ```python
    import os

    if __name__ == '__main__':
        print(os.getenv('nobody knows the trouble i seeeeeee').strip())
    ```
    the above errors with `AttributeError: 'NoneType' object has no attribute 'strip'`
    * The same goes when utilizing the `.get()` method directly off any dict type, which is ultimately what `os.environ` is
    * We could remediate this by using `os.getenv('{envname}', '')`, at which point we would get an empty string
        ```shell
        import os

        if __name__ == '__main__':
            print(os.getenv('nobody knows the trouble i seeeeeee', '').strip())
        ```

I think an acceptable update would be:

```python
    project_id = os.getenv.get(PROJECT_ID, '').strip()

    # ensure project id is a non-empty string
    if project_id:
        return project_id

    # in all other cases, OBJECTION!
    raise RuntimeError(f'{PROJECT_ID} is not set.  Set and retry.')
```

What do you think?

lrwb-aou: You're right.   I forgot about the possibility of having spaces with quoted environment values.
I did know that `.get()` returned `None` for an undefined dictionary key.  The previous code counted on its falsy nature.  I prefer using the dictionary default value when you have a desired default type.  I like the code above.

dcarbone: Ah, sure, I should have enhanced my reply by saying:

> With the need to strip the envvar in mind, i need to ensure its a string before calling .strip()...

etc.  Did not mean to come across as preachy :)

lrwb-aou:
https://8086-95246777-gh.circle-artifacts.com/0/test_results/unit/html/data_steward_utils_pipeline_logging_py.html 
and
https://8086-95246777-gh.circle-artifacts.com/0/test_results/unit/html/data_steward_app_identity_py.html
indicate the changes are fully unit tested.  
Once we get the integration tests working for you, we'll know it's safe to merge.  (I suspect it is safe to merge it, as is.)

lrwb-aou:
I suspect that failure is due to this:  https://precisionmedicineinitiative.atlassian.net/browse/DC-1532  .  Has more to do with a bad test than your particular code.

lrwb-aou (approved):
The test that failed is not affected by your code change.  This looks good.

tests/unit_tests/data_steward/utils/pipeline_logging_test.py
```diff
@@ -60,16 +62,30 @@ def assert_logs_handled(self, current_level, mock_file_emit,
...
-        :param mock_open: mock to prevent side effect of opening file
+        Verifies that the provided logging.FileHandler instance would be writing to a file whose name matches
+        expected_basename
+        :param file_handler: logging.FileHandler instance to test
+        :param expected_basename: expected basename of log file
+        """
+        log_filename = file_handler.baseFilename
+        log_basename = os.path.basename(log_filename)
+        # use endswith as opposed to .assertEquals so we don't care about slash type.
+        self.assertTrue(log_filename.endswith(expected_basename))
```
mark-velez, line 76: The variable `log_basename` is unused. Maybe you intended to do this (below) but then decided against it?
```suggestion
        self.assertEquals(log_filename, expected_basename)
```
It seems to me that it'd better account for a scenario like this, no?
```
log_filename='abc20210409-project.log'
expected_basename ='20210409-project.log'
```

mark-velez (approved):
Looks good! Please consider my comments, nothing blocking.

data_steward/app_identity.py
```diff
@@ -10,9 +10,11 @@ def get_application_id():
     :return:
     """
     # NOTE: Google interchangeably refers to this identifier as application_id or project_id
-    project_id = os.environ.get(PROJECT_ID)
+    project_id = os.getenv(PROJECT_ID, '').strip()
```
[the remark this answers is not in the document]
dcarbone: That is a very valid point.  I can tend to get over-zealous with input sanitation. I am OK with removing the call to `strip()`

tests/unit_tests/data_steward/utils/pipeline_logging_test.py
dcarbone: You are correct, it was an abandoned attempt at doing a direct filename match.  It turns out python's basename methods don't properly handle paths in some cases: https://stackoverflow.com/questions/678236/how-to-get-the-filename-without-the-extension-from-a-path-in-python.  I ran into this in my own testing, locally.

To keep this code simple and to not have to deal with path markup differences, I chose to simply compare using `.endswith()`.

dcarbone: I have updated the test to be slightly more intelligent about its assertions.

PR #820, opened 2021-04-07
[DC-1551] Update import_rdr_omop.sh

dcampbell-vumc:
* Add condition to prevent clustering for the `pid_rid_mapping` table


amcgrenera-vumc (approved):
Looks good to me!

PR #826, opened 2021-04-09
[DC-1518] partition primary_pid_rid_mapping table

lrwb-aou:
* change mapping table name to primary_pid_rid_mapping
* update name where needed
* insert partition information into storage queries
* allow setting partition date in storage queries


data_steward/cdr_cleaner/cleaning_rules/store_pid_rid_mappings.py
```diff
@@ -171,6 +188,14 @@ def validate_rule(self):
     import cdr_cleaner.clean_cdr_engine as clean_engine
 
     ext_parser = parser.get_argument_parser()
+    ext_parser.add_argument(
+        '-e',
+        '--export_date',
+        action='store',
+        dest='export_date',
+        help=('Date of the RDR export. Should adhere to '
```
ChaoPang, line 196: Should we add a type validation on the format of `export_date`?

ChaoPang (approved):
Everything looked good, and I successfully ran the script using a few fake datasets created in the test environment, the script behaved as expected. 

I approved the PR. Although there is just one tiny comment I had regarding adding type validation to the `export_date ` argument, I don't think it's a mandatory change so I leave it up to you.

data_steward/cdr_cleaner/cleaning_rules/store_pid_rid_mappings.py
lrwb-aou: @ChaoPang I considered this and could be swayed to add one.  I initially decided to put the date parsing/validation code into the class initialization so it was checked at one location and dealt with accordingly whether we run the script as a stand alone script or as an import to another module. If the input can't be parsed as a date, the class initializer does raise an error.  I'll see about adding some earlier type checking to it.

ChaoPang (approved):
I went through the changes the second time, it looks good!

PR #828, opened 2021-04-12
[DC-1529] Record suppress two missed concepts(organ transplant - other)

ksdkalluri:
Adds cleaning rule to suppress Organ transplant(other) concepts


ChaoPang (changes requested):
The cleaning rule looks good and it behaves as expected. However,

1. It needs to be added to `CONTROLLED_TIER_DEID_CLEANING_CLASSES` in `clean_cdr.py` 
2. The ticket mentioned "but any record in any table with one one of these concept_ids in a *_concept_id field should be removed", does that mean this rule should be applied to all CDM tables?

ChaoPang (approved):
Looks good!

PR #830, opened 2021-04-14
[DC-1500] Alter the GenerateSiteMappingsAndExtTables cleaning rule to read the site mapping table in pipeline_tables

ksdkalluri:


jp3477 (approved):
Looks good, approved.

PR #832, opened 2021-04-14
[DC-1537] Drop records without a valid pid/rid mapping

nishanthpp93:
* [DC-1537] Add delete statement and update warning log

[DC-1537]: https://precisionmedicineinitiative.atlassian.net/browse/DC-1537


amcgrenera-vumc (approved):
This looks good to me. Integration test is working locally as expected!

PR #833, opened 2021-04-14
[DC-1542] integration test for storing pid/rid/shift mappings in primary table

lrwb-aou:
* fixes one bug in the cleaning rule
* mocks the hard coded PIPELINE_TABLES primary dataset reference
* makes sure existing values are not overwritten.
* makes sure new values are added along with random shift elements and partition dates

### DIFF
--- a/tests/integration_tests/data_steward/cdr_cleaner/cleaning_rules/store_pid_rid_mappings_test.py
+++ b/tests/integration_tests/data_steward/cdr_cleaner/cleaning_rules/store_pid_rid_mappings_test.py
@@ -4,7 +4,7 @@ import os
 import mock
 
 from app_identity import get_application_id
-from common import PID_RID_MAPPING, PIPELINE_TABLES, PRIMARY_PID_RID_MAPPING
+from common import PID_RID_MAPPING, PRIMARY_PID_RID_MAPPING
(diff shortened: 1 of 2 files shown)


tests/integration_tests/data_steward/cdr_cleaner/cleaning_rules/store_pid_rid_mappings_test.py
```diff
@@ -0,0 +1,116 @@
+from datetime import datetime
+import os
+
+import mock
+
+from app_identity import get_application_id
+from common import PID_RID_MAPPING, PRIMARY_PID_RID_MAPPING
+from resources import fields_for
```
ksdkalluri, line 8: looks good!
nit: there are 2 unused Imports in the test script.

ksdkalluri (approved):

tests/integration_tests/data_steward/cdr_cleaner/cleaning_rules/store_pid_rid_mappings_test.py
lrwb-aou: @ksdkalluri nice catch!  Just removed them.  I'll wait until tests pass, and then merge with your approval.

PR #834, opened 2021-04-14
[DC-1555] Analytics: RDR notebook class of PPI Concept

amcgrenera-vumc:


lrwb-aou (changes requested):
@amcgrenera-vumc This SQL looks good.  Would you mind adding something to this query description about how we don't consider Qualifier Value or Question mapping to either Topic or PPI Modifier as something this query is intended to flag?  It might make it easier to understand in the future.

lrwb-aou (approved):
Looks great!  Nice description!

PR #837, opened 2021-04-19
[DC-1532] Clean Height Weight Integration Test Fix

dcampbell-vumc:
* Added test_cleaning_results test to utilize subTests in order for
  setUp to be called once

### DIFF
--- a/tests/integration_tests/data_steward/cdr_cleaner/cleaning_rules/clean_height_weight_test.py
+++ b/tests/integration_tests/data_steward/cdr_cleaner/cleaning_rules/clean_height_weight_test.py
@@ -388,7 +388,7 @@ class CleanHeightWeightTest(BaseTest.CleaningRulesTestBase):
                 {concept_ancestor_data_query}'''
         ])
 
-    def assert_test_get_query_specs(self):
+    def assert_get_query_specs(self):
         """
         Tests that queries run successfully and sandbox tables are generated
         Note: This does NOT validate logic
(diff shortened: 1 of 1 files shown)


tests/integration_tests/data_steward/cdr_cleaner/cleaning_rules/clean_height_weight_test.py
```diff
@@ -388,7 +388,7 @@ def setUp(self):
                 {concept_ancestor_data_query}'''
         ])
```
nishanthpp93, line 389: I wonder if these can be broken up to prevent them all running at the same time? I'm not sure if/how much it helps, but it seems to work for other tests running on the same table.
```suggestion
self.load_test_data([
            person_data_query, measurement_data_query,
            measurement_ext_data_query, condition_data_query,
            concept_data_query, concept_ancestor_data_query
         ])
```

nishanthpp93 (commented):
I think this should be good. Added a small suggestion that could reduce the number of queries running at the same time.

tests/integration_tests/data_steward/cdr_cleaner/cleaning_rules/clean_height_weight_test.py
dcampbell-vumc: I split this into three different loads to reduce the number of queries running simultaneously. This is ready for another looksie.

nishanthpp93: I see. I think best would be to send them as a list of six individual queries without the f-strings in a single call to `load_test_data`, so they all run separately. But this could work as well.

nishanthpp93 (approved):

tests/integration_tests/data_steward/cdr_cleaner/cleaning_rules/clean_height_weight_test.py
dcampbell-vumc: @nishanthpp93 I have them as the separate queries now instead of the f-strings so this is ready for another quick review before I merge

nishanthpp93 (approved):
Looks good to me!

PR #838, opened 2021-04-19
[DC-1544] Require all AOU_REQUIRED files before processing EHR data

dcarbone:
mostly opening to trigger integration tests


data_steward/validation/main.py
```diff
@@ -795,6 +870,11 @@ def _get_submission_folder(bucket, bucket_items, force_process=False):
         has been processed and force_process is False or no submission
         directory exists
     """
+
+    # fast exit
+    if bucket_items is None or len(bucket_items) is 0:
```
lrwb-aou, line 875:
```suggestion
    if not bucket_items:
```

Wanted to put this here while I was thinking about it.  I think it would be "more pythonic" to say "if not bucket_items" than to do the existing double conditional here.

data_steward/validation/main.py
```diff
@@ -751,29 +752,103 @@ def list_submitted_bucket_items(folder_bucketitems):
...
+    If the returned list is greater than 1, the consumer should assume the
+    targeted set is invalid or incomplete
+
+    :param submitted_bucket_items: list of string "file" names from bucket
+    :return: list of strings of missing require file names
+    """
+
+    # if the provided input is empty, state that all required files are missing
+    if submitted_bucket_items is None or len(submitted_bucket_items) is 0:
+        return common.AOU_REQUIRED_FILES
```
lrwb-aou, line 845: I would also suggest using 
```suggestion
    if not submitted_bucket_items:
```
for this check too.  I don't think you actually care if the identity value of the `len()` is the same identity as `0` either.  For that length check I would recommend, `len(submitted_bucket_items) == 0`.

data_steward/validation/main.py
```diff
@@ -824,9 +904,19 @@ def _get_submission_folder(bucket, bucket_items, force_process=False):
...
             if item['name'].startswith(folder_name)
         ]
+
+        # filter files based on age
         submitted_bucket_items = list_submitted_bucket_items(
             folder_bucket_items)
 
+        # determine if the above is missing any of the "required" files
+        missing_required = _get_missing_required_files(submitted_bucket_items)
+        if len(missing_required) > 0:
```
lrwb-aou, line 914: My thought process is the same here.  I would suggest:
```suggestion
    if missing_required:
```
It seems clearer, especially with the well named variable.

dcarbone:
@lrwb-aou: :+1: thanks for the pointers, i agree the suggestions are more "pythonic".

I'm working on a few small updates to deal with tz stuff, should have it ready this afternoon.

data_steward/validation/main.py
dcarbone: Good point, I still conflate `is` and `==` in python.

PR #839, opened 2021-04-19
[DC-1521] Shared Sandbox Name Fix

dcampbell-vumc:
* Updated `create_tier.create_datasets` to drop the tier from the sandbox_dataset_id
* Updated `create_tier_test.test_create_datasets` to test the changes
  made in `create_tier.create_datasets`

### DIFF
--- a/tests/unit_tests/data_steward/tools/create_tier_test.py
+++ b/tests/unit_tests/data_steward/tools/create_tier_test.py
@@ -294,20 +294,24 @@ class CreateTierTest(unittest.TestCase):
         }
 
         # Tests if incorrect parameters are given
-        self.assertRaises(RuntimeError, create_datasets, None, self.dataset_name,
-                          self.input_dataset, self.tier, self.release_tag)
+        self.assertRaises(RuntimeError, create_datasets, None,
+                          self.dataset_name, self.input_dataset, self.tier,
+                          self.release_tag)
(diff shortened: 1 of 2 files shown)


tests/unit_tests/data_steward/tools/create_tier_test.py
```diff
@@ -288,26 +288,30 @@ def test_create_dataset(self, mock_client, mock_define_dataset,
...
+                          self.dataset_name, self.input_dataset, None,
+                          self.release_tag)
+        self.assertRaises(RuntimeError, create_datasets, client,
+                          self.dataset_name, self.input_dataset, self.tier,
+                          None)
 
         # Test
-        expected = create_datasets(client, self.name, self.input_dataset,
-                                   self.tier, self.release_tag)
+        expected = create_datasets(client, self.dataset_name,
```
jp3477, line 312: NIT: this variable should probably be called actual

jp3477 (changes requested):
Would you be able to repeat the same test case for a registered tier dataset? May be redundant, but I think it's in the ticket acceptance criteria. Other than that, everything looks great.

tests/unit_tests/data_steward/tools/create_tier_test.py
dcampbell-vumc: You're right! I changed it

jp3477 (approved):
Great, approved!

PR #841, opened 2021-04-19
[DC-1570] Update create_person_ext_table Cleaning rule to add 0 and "No Matching Concept" when the values are not available via join

ksdkalluri:


amcgrenera-vumc:
@ksdkalluri the query looks good to me! Under the AC in the ticket, it does say to add an integration test. I added some data to verify no NULL values show, see attached patch file.
[[DC-1570]_update_integration_test.patch.zip](https://github.com/all-of-us/curation/files/6379200/DC-1570._update_integration_test.patch.zip)

amcgrenera-vumc (approved):
thanks for adding the patch, this is approved!

PR #842, opened 2021-04-20
[DC-1575] [P0] drop null or zero concept ids failure

ksdkalluri:


lrwb-aou (approved):
This looks good.  Nice to change to the DML statement.

PR #844, opened 2021-04-22
[DC-1477] Suppress care_site table in RT deid

dcampbell-vumc:
* Removed `care_site` from `DEID_TABLES` list
* Added `care_site` to `SUPPRESSED_TABLES` list

### DIFF
--- a/data_steward/tools/run_deid.py
+++ b/data_steward/tools/run_deid.py
@@ -26,7 +26,7 @@ LOGGER = logging.getLogger(__name__)
 DEID_TABLES = [
     'person', 'observation', 'visit_occurrence', 'condition_occurrence',
     'drug_exposure', 'procedure_occurrence', 'device_exposure', 'death',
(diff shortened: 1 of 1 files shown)


data_steward/tools/run_deid.py
```diff
@@ -26,11 +26,11 @@
 DEID_TABLES = [
     'person', 'observation', 'visit_occurrence', 'condition_occurrence',
     'drug_exposure', 'procedure_occurrence', 'device_exposure', 'death',
-    'measurement', 'location', 'care_site', 'specimen', 'observation_period'
+    'measurement', 'location', 'specimen', 'observation_period'
```
lrwb-aou, line 29: @dcampbell-vumc I think the scope is wrong on this ticket.  You should leave `care_site` as a `DEID_TABLES` item.  Just add it to the `SUPPRESSED_TABLES` list.

lrwb-aou (approved):
This is good to merge if integration tests pass.  I think they should.  :)

PR #847, opened 2021-04-26
[DC-1588] Update participant summary test to use bq client library

nishanthpp93:
* [DC-1588] Use client library for storing participant data instead of pd_gbq

[DC-1588]: https://precisionmedicineinitiative.atlassian.net/browse/DC-1588


dcampbell-vumc (approved):
Ran the unit and integration test and they both pass as expected! This is good to go!

PR #848, opened 2021-04-26
[DC-1590] Reduce table creation and drop for ID deduplicate test

nishanthpp93:
* [DC-1590] Reduce table creation-drop instances to just once

[DC-1590]: https://precisionmedicineinitiative.atlassian.net/browse/DC-1590

### DIFF
--- a/tests/integration_tests/data_steward/cdr_cleaner/cleaning_rules/id_de_duplicate_test.py
+++ b/tests/integration_tests/data_steward/cdr_cleaner/cleaning_rules/id_de_duplicate_test.py
@@ -7,6 +7,7 @@ The intent is remove duplicate primary keys all domain tables excluding person t
 
 # Python Imports
 import os
+from datetime import date, datetime
 
 # Project Imports
 from common import CONDITION_OCCURRENCE, OBSERVATION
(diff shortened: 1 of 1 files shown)


amcgrenera-vumc:
This looks good to me and is passing locally, only question, how come you added the additional fields? And should those be included in the `fields` and `cleaned_values` in `tables_and_counts`?

tests/integration_tests/data_steward/cdr_cleaner/cleaning_rules/id_de_duplicate_test.py
```diff
@@ -39,11 +39,11 @@ def setUpClass(cls):
                                                 cls.sandbox_id)
 
         # Generates list of fully qualified table names
-        for table_name in [CONDITION_OCCURRENCE, OBSERVATION]:
+        for table_name in cls.affected_tables:
             cls.fq_table_names.append(
```
amcgrenera-vumc, line 43: Good call on using this instead of explicitly listing!

nishanthpp93:
> This looks good to me and is passing locally, should the additional required fields you added be included in the `fields` and `cleaned_values` in `tables_and_counts`?

That's a good point. I know the cleaning rule should not affect them, but it is worth checking and making sure that they are unaffected.

amcgrenera-vumc (approved):
thanks for adding that in! this looks good to me

PR #849, opened 2021-04-26
[DC-1589] Union domain alignment logging table queries to a single query

nishanthpp93:
* [DC-1589] Use single unioned query for domain alignment

[DC-1589]: https://precisionmedicineinitiative.atlassian.net/browse/DC-1589


jp3477 (approved):
Code looks good and tests pass.

PR #850, opened 2021-04-26
[DC-1591] Use circle project in bucket and dataset creation to reduce conflicts

nishanthpp93:
* [DC-1591] Update bash script to include project username

[DC-1591]: https://precisionmedicineinitiative.atlassian.net/browse/DC-1591


jp3477 (approved):
Looks good. I appreciate the double quotes.

PR #853, opened 2021-04-29
[DC-1495] Add a date conformance check to the RDR notebook

amcgrenera-vumc:


dcampbell-vumc (approved):
This is great! I was able to run the notebook and see the Date Conformance Check section. I inserted some dummy values in my test dataset that would fail this check to verify that it is working properly.  

Screen Shot 2021-04-29 at 11 30 56

PR #854, opened 2021-05-03
[DC-1602] Incorrect Name for DEID_MAP_TABLE in run_deid.py

dcampbell-vumc:
* Changed `DEID_MAP_TABLE` name to `primary_pid_rid_mapping`

### DIFF
--- a/data_steward/tools/run_deid.py
+++ b/data_steward/tools/run_deid.py
@@ -36,7 +36,7 @@ VOCABULARY_TABLES = [
     'relationship', 'concept_synonym', 'concept_ancestor',
     'source_to_concept_map', 'drug_strength'
 ]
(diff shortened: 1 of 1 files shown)


amcgrenera-vumc (approved):
This looks good to me and the unit test is passing locally!

My only question, and this might be a topic for eng sync and we can address at a later time, but should these variables be defined in `common` rather than string literals?

dcampbell-vumc:
> This looks good to me and the unit test is passing locally!
> 
> My only question, and this might be a topic for eng sync and we can address at a later time, but should these variables be defined in `common` rather than string literals?

I agree with you that they should be. However we're planning to do a refactor of deid sometime in the future which will take this into account.

amcgrenera-vumc:
Sounds good! This is approved!

PR #856, opened 2021-05-03
[DC-1604] SITE_TABLE_ID Name Incorrect In `generate_ext_tables.py`

dcampbell-vumc:
* Updated `SITE_TABLE_ID` name to `site_maskings`


nishanthpp93 (commented):
There seem to be more references to `_site_mappings` in the codebase. Should they be updated as well? https://github.com/all-of-us/curation/search?q=_site_mappings

dcampbell-vumc:
> There seem to be more references to `_site_mappings` in the codebase. Should they be updated as well? https://github.com/all-of-us/curation/search?q=_site_mappings

The `_site_mappings` in `generate_ext_tables` is being updated in another RT bug ticket. The other ones did not trigger any issues when generating RT deid so I didn't consider them for this ticket. What are your thoughts? After looking at the other references, one is in a CSV file for CT and the other a unit test. I can update the unit test no problem but I'm not sure about the CSV file.

nishanthpp93:
That's a good question. I think updating all references would make the transition more complete. The CSV file seems to look for the `_site_mappings` table for the references for the qc notebook, but since we have transitioned to `site_maskings`, I think it would need to be updated as well. But important to note that past CDRs that use `_site_mappings` will not work, but I guess that was accounted for when first transitioning from `_site_mappings` to `site_maskings`.

dcampbell-vumc:
> That's a good question. I think updating all references would make the transition more complete. The CSV file seems to look for the `_site_mappings` table for the references for the qc notebook, but since we have transitioned to `site_maskings`, I think it would need to be updated as well. But important to note that past CDRs that use `_site_mappings` will not work, but I guess that was accounted for when first transitioning from `_site_mappings` to `site_maskings`.

Okie dokie, I will update the rest of those references to `_site_mappings` to `site_maskings`

dcampbell-vumc:
@nishanthpp93 I updated all references to `_site_mapping` to `site_mappings` this is ready for another look ☺️

nishanthpp93 (approved):
Looks good!

PR #857, opened 2021-05-03
[DC-1605] Use Shared Sandbox In `survey_version_info.py` Manual Cleaning Rule

dcampbell-vumc:
* Updated `qrid_map_dataset_id` to be the shared sandbox dataset id
* Updated the integration test to reflect the changes in the module


dcampbell-vumc:
Branch name is breaking integration test check, changing branch name, closing PR

PR #858, opened 2021-05-03
[DC-1606] Modify `get_mapping_table_ids` Function In `generate_ext_tables`

dcampbell-vumc:
* Updated the `get_mapping_table_ids` function to use bq.list_tables as
  bq_utils.list_tables is being deprecated


dcampbell-vumc:
Branch name breaking integration test check, changing branch name, and closing this PR

PR #861, opened 2021-05-03
[DC-1595] Update BQ retraction to use Jinja and client library

nishanthpp93:
* [DC-1595] Use jinja and update EHR retraction
* [DC-1595] Reuse client
* [DC-1595] Simplify queries used
* [DC-1595] Update retraction queries
* [DC-1595] Use single function instead of separate functions for datasets
* [DC-1595] Update integration test
* [DC-1595] Update unit tests and fix bugs


ksdkalluri (approved):
I ran the tests locally, and the changes look good!

PR #862, opened 2021-05-04
[DC-1608] _deid_map Table Should Be Created and Loaded Instead Of primary_pid_rid_mapping Table in run_deid.py

dcampbell-vumc:
* Changed `deid_map_table` f-string to be the `_deid_map` table


lrwb-aou (approved):
Looks good!

PR #863, opened 2021-05-04
[DC-1609] Add `compute` Logic Back To config.json File

dcampbell-vumc:
* Add compute logic back to config.json file

### DIFF
--- a/data_steward/deid/config/ids/config.json
+++ b/data_steward/deid/config/ids/config.json
@@ -202,6 +202,22 @@
         "datetime": "TIMESTAMP_SUB( CAST(:FIELD AS TIMESTAMP), INTERVAL (:SHIFT) DAY) AS :FIELD"
     },
     {
+         "_id": "compute",
+         "id": ["(SELECT research_id FROM :idataset._deid_map WHERE _deid_map.person_id = :value_field) as :FIELD"],
+         "response_id": [
+             "(SELECT ",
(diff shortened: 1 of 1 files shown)


data_steward/deid/config/ids/config.json
```diff
@@ -201,6 +201,22 @@
...
+         "id": ["(SELECT research_id FROM :idataset._deid_map WHERE _deid_map.person_id = :value_field) as :FIELD"],
+         "response_id": [
+             "(SELECT ",
+             "CASE WHEN :value_field IS NULL THEN NULL ",
+             "ELSE (",
+                 "SELECT dqrm.research_response_id ",
+                 "FROM :idataset._deid_questionnaire_response_map AS dqrm ",
+                 "WHERE dqrm.questionnaire_response_id = :value_field) ",
+             "END ) AS :FIELD"
+         ],
```
lrwb-aou, line 215: I see what happened with the previous patch.  Lines 207-215 needed to be removed, but the entire dictionary was removed instead.  If you can remove these lines, this will be good to merge.

lrwb-aou (changes requested):
See comment.  Let me know if you have questions.

data_steward/deid/config/ids/config.json
dcampbell-vumc: Ah, I see. I will remove those and repush!

lrwb-aou (approved):
This looks good!

PR #864, opened 2021-05-04
[DC-1610] Miscategorized Error Message In rules.py

dcampbell-vumc:
* Changed logging message from error to info


dcampbell-vumc:
Branch name breaks integration test check, changing name, reissuing PR

PR #865, opened 2021-05-04
[DC-1610] FMiscategorized Error Message In rules.py

dcampbell-vumc:
* Changed logging message from error to info


dcampbell-vumc:
Branch name breaking integration test check still, renaming branch again, reissuing PR.

PR #869, opened 2021-05-07
[DC-18] Duplicated cleaning rule introduces suppression bug

jp3477:
Removing the `PreferNotToAnswerSuppression` cleaning rule from codebase.


lrwb-aou (approved):
The changes are correct.  Pylint is failing because the commit message isn't formatted correctly.  Could you try amending this commit message and pushing the message change?

jp3477:
> The changes are correct. Pylint is failing because the commit message isn't formatted correctly. Could you try amending this commit message and pushing the message change?

Thanks, that fixed it.

PR #870, opened 2021-05-07
[DC-1619] Convert observation_source_concept_id cleaning rule to DML statement

jp3477:
Converted a write truncate select statement to a dml delete.


lrwb-aou (approved):
This is so much easier to read and understand!

PR #874, opened 2021-05-14
[DC-1600] May RDR import and cleaning

lrwb-aou:
* replaced bash RDR import and cleaning scripts with python.
* uses google client objects to complete work
* uses impersonation
* uses client object to do schemaed table creation with a query.  placed into utils.bq
* loads data directly from RDR bucket
* ensures AOU_REQUIRED files are built for RDR import (even if left empty)


ksdkalluri (approved):
@lrwb-aou Thanks for taking that initiation. The changes were perfect and a good starting point for our bash to python conversion epic. Added utils were good and are working well. I am hoping there might be some minor changes/additions to the script when planned utils are added to bq in the bash to python epic. This is good to merge.

PR #877, opened 2021-05-17
[DC-1527] Organ and Geolocation QC for controlled tier

rfrancis1:


data_steward/analytics/cdr_ops/controlled_tier_qc/check_controlled_tier.py
```diff
@@ -20,15 +20,19 @@
...
 
+pd.set_option('display.max_colwidth', -1)
+pd.set_option('display.width', None)
+# -
+
 # to_include = ['DC-1370', 'DC-1377', 'DC-1346', 'DC-1348', 'DC-1355', 'DC-1357', 'DC-1359', 
 #             'DC-1362', 'DC-1364', 'DC-1366', 'DC-1368', 'DC-1373',
 #             'DC-1380', 'DC-1382', 'DC-1388', 'DC-1496']
-to_include = []
+to_include = ['DC-1527', 'DC-1535', 'DC-1370', 'DC-1377', 'DC-1362']
```
mark-velez, line 35: Please set `to_include = []` so that all rules will run by default. Engs typically run this from the command line and will not think to edit this line.

mark-velez (changes requested):
It works now, thanks @rfrancis1 ! Please just update `check_controlled_tier` so that all rules run by default and then we can merge this.

mark-velez (approved):

PR #879, opened 2021-05-20
[DC-1633] Create cleaning rule to update invalid ZIP entries

dcampbell-vumc:
* Added `zip3_lookup` table constant to `common.py`
* Created `update_invalid_zip_codes.py` module to update any zip codes that are less than 5 digits and/or don't match a value in the `zip3_lookup` view.
* Created `update_invalid_zip_codes_test.py` module to test functionality of the main module
* Created a schema file for `zip3_lookup` table which will only be used for the integration test module since it is actually a view in the pipeline tables dataset
* Added CR to `RDR_CLEANING_RULES` list

### DIFF
--- a/data_steward/cdr_cleaner/cleaning_rules/update_invalid_zip_codes.py
+++ b/data_steward/cdr_cleaner/cleaning_rules/update_invalid_zip_codes.py
@@ -3,9 +3,12 @@ Sandbox and update invalid zip codes found in the observation table.
 
 Original Issues: DC-1633, DC-1645
 
-The intent of this cleaning rule is to sandbox and update any invalid zip code in the observation table. A zip code is
-    considered invalid if it does not match a zip3 code in the zip3 master lookup table or if it less than 5 digits.
-    The record is sandboxed and updated to have the following information:
+The intent of this cleaning rule is to remove any leading/trailing whitespace in the zip code string then sandbox and
+update any invalid zip code in the observation table. A zip code is considered invalid if it:
+        Is less than 5 digits in length
+        Is alpha-numeric
+        Does not match any zip3 code in the master zip3 lookup table
+If zip code is deemed invalid the record is sandboxed and updated to have the following information:
         value_as_string and value_source_value = 'Response removed due to invalid value'
         value_as_number = 0
         value_source_concept_id = 2000000010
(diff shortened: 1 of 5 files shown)


data_steward/cdr_cleaner/cleaning_rules/update_invalid_zip_codes.py
```diff
@@ -0,0 +1,153 @@
...
+from constants.cdr_cleaner import clean_cdr as cdr_consts
+from common import JINJA_ENV, OBSERVATION, PIPELINE_TABLES, ZIP3_LOOKUP
+from cdr_cleaner.cleaning_rules.base_cleaning_rule import BaseCleaningRule
+
+LOGGER = logging.getLogger(__name__)
+
+# Creates sandbox that contains all invalid zip codes which are deemed invalid because they:
+# 1. Are not 5 digits in length
+# 2. Do not match a zip3 code in the master zip3 lookup table
+SANDBOX_INVALID_ZIP_CODES = JINJA_ENV.from_string("""
```
jp3477, line 28: I think there are 3 scenarios where this wouldn't give the desired results from the ticket specs:

1. In the case where the zip code has 5 digits, but the first 3 digits are not in the zip3_lookup table, the zip code will not be sandboxed as intended (e.g.  '00501'). The second LEFT JOIN in the query will not get a chance to run. Maybe replacing the joins with a UNION DISTINCT would work instead.

2. 
> # only use first 5 characters, and ensure they're only numeric in value

In the case where the zip code has 5 characters and the first 3 characters are a valid zip3 while the last two characters are nonnumeric, the record is not sandboxed as intended (e.g. '123ab').

3. 
> # get answer, trim value
I believe zip values should be trimmed of whitespace before other checks are run. Right now, a string like '    12345 ' would be sandboxed as invalid.

tests/integration_tests/data_steward/cdr_cleaner/cleaning_rules/update_invalid_zip_codes_test.py
```diff
@@ -0,0 +1,159 @@
...
+                
+                -- invalid zip codes, will be sandboxed and updated in observation table --
+                (3, 3, 0, date('2019-03-03'), 0, 1585250, null, '1234', null, null), 
+                (4, 4, 0, date('2019-03-03'), 0, 1585250, null, '123', null, null),
+                (5, 5, 0, date('2019-03-03'), 0, 1585250, null, '12', null, null),
+                (6, 6, 0, date('2019-03-03'), 0, 1585250, null, '1', null, null),
+                
+                -- not zip codes (not zip code concept id), will not be affected --
+                (7, 7, 0, date('2019-03-03'), 0, 1111111, null, '111', null, null),
+                (8, 8, 0, date('2019-03-03'), 0, 2222222, null, '222', null, null)
```
jp3477, line 118: I added these extra test scenarios corresponding to the cases mentioned in the cleaning rule feedback. [No need to commit unless you want to]
```suggestion
                (8, 8, 0, date('2019-03-03'), 0, 2222222, null, '222', null, null),
                
                -- invalid zip codes with 5 characters and unknown zip 3 --
                (9, 9, 0, date('2019-03-03'), 0, 1585250, null, '00501', null, null),

                -- invalid zip codes with 5 characters, last 2 characters non-numeric --
                (10, 10, 0, date('2019-03-03'), 0, 1585250, null, '123ab', null, null),

                -- valid zip codes with 5 numeric characters but extra whitespace --
                (11, 11, 0, date('2019-03-03'), 0, 1585250, null, '    12345 ', null, null)                
```

tests/integration_tests/data_steward/cdr_cleaner/cleaning_rules/update_invalid_zip_codes_test.py
```diff
@@ -0,0 +1,159 @@
...
+            'fq_sandbox_table_name':
+                self.fq_sandbox_table_names[0],
+            'fields': [
+                'observation_id', 'person_id', 'observation_concept_id',
+                'observation_date', 'observation_type_concept_id',
+                'observation_source_concept_id', 'value_as_number',
+                'value_as_string', 'value_source_concept_id',
+                'value_source_value'
+            ],
+            'loaded_ids': [1, 2, 3, 4, 5, 6, 7, 8],
```
jp3477, line 135: I added these extra test scenarios for the cases mentioned in the cleaning rule feedback. [No need to commit unless you want to]
```suggestion
            'loaded_ids': [1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11],
```

tests/integration_tests/data_steward/cdr_cleaner/cleaning_rules/update_invalid_zip_codes_test.py
```diff
@@ -0,0 +1,159 @@
...
+            'sandboxed_ids': [3, 4, 5, 6],
```
jp3477, line 136: I added these extra test scenarios for the cases mentioned in the cleaning rule feedback. [No need to commit unless you want to]
```suggestion
            'sandboxed_ids': [3, 4, 5, 6, 9, 10],
```

tests/integration_tests/data_steward/cdr_cleaner/cleaning_rules/update_invalid_zip_codes_test.py
```diff
@@ -0,0 +1,159 @@
...
+                (3, 3, 0, self.date, 0, 1585250, 0.0, self.invalid_zip_response,
+                 self.invalid_zip_concept, self.invalid_zip_response),
+                (4, 4, 0, self.date, 0, 1585250, 0.0, self.invalid_zip_response,
+                 self.invalid_zip_concept, self.invalid_zip_response),
+                (5, 5, 0, self.date, 0, 1585250, 0.0, self.invalid_zip_response,
+                 self.invalid_zip_concept, self.invalid_zip_response),
+                (6, 6, 0, self.date, 0, 1585250, 0.0, self.invalid_zip_response,
+                 self.invalid_zip_concept, self.invalid_zip_response),
+                (7, 7, 0, self.date, 0, 1111111, None, '111', None, None),
+                (8, 8, 0, self.date, 0, 2222222, None, '222', None, None)
```
jp3477, line 150: I added these extra test scenarios for the cases mentioned in the cleaning rule feedback. [No need to commit unless you want to]
```suggestion
                (8, 8, 0, self.date, 0, 2222222, None, '222', None, None),
                (9, 9, 0, self.date, 0, 1585250, 0.0, self.invalid_zip_response,
                 self.invalid_zip_concept, self.invalid_zip_response),
                (10, 10, 0, self.date, 0, 1585250, 0.0,
                 self.invalid_zip_response, self.invalid_zip_concept,
                 self.invalid_zip_response),
                (11, 11, 0, self.date, 0, 1585250, None, '    12345 ', None,
```

jp3477 (changes requested):
I tried to outline a few of the conditions in the ticket specs that I think wouldn't be met. Let me know if this makes sense.

data_steward/cdr_cleaner/cleaning_rules/update_invalid_zip_codes.py
```diff
@@ -0,0 +1,197 @@
...
+SELECT o.* FROM `{{project_id}}.{{dataset_id}}.{{obs_table}}` o
+LEFT JOIN `{{project_id}}.{{pipeline_tables}}.{{zip3_lookup}}` z
+ON SUBSTR(o.value_as_string, 1, 3) = CAST(z.zip3 AS STRING)
+WHERE (observation_source_concept_id = 1585250 AND z.zip3 IS NULL)))
+""")
+
+CLEAN_ZIPS_OF_WHITESPACE = JINJA_ENV.from_string("""
+UPDATE `{{project_id}}.{{dataset_id}}.{{obs_table}}` SET
+value_as_string = TRIM(value_as_string)
+WHERE person_id IN (
```
jp3477, line 61: Right now, this is modifying all observations that the person has, rather than just the zip code observation. Changing the `WHERE` clause to the following should fix it:

```
WHERE observation_id IN (
SELECT observation_id FROM `{{project_id}}.{{sandbox_dataset_id}}.{{whitespace_sandbox}}`)
```

data_steward/cdr_cleaner/cleaning_rules/update_invalid_zip_codes.py
```diff
@@ -0,0 +1,197 @@
...
+SELECT person_id FROM `{{project_id}}.{{sandbox_dataset_id}}.{{whitespace_sandbox}}`)
+""")
+
+UPDATE_INVALID_ZIP_CODES = JINJA_ENV.from_string("""
+UPDATE `{{project_id}}.{{dataset_id}}.{{obs_table}}` SET
+value_as_string = 'Response removed due to invalid value',
+value_source_value = 'Response removed due to invalid value',
+value_as_number = 0,
+value_source_concept_id = 2000000010
+WHERE person_id IN (
```
jp3477, line 71: Same problem as above. This should fix it:
```
WHERE observation_id IN (
SELECT observation_id FROM `{{project_id}}.{{sandbox_dataset_id}}.{{sandbox_table}}`)
```

jp3477 (changes requested):
Thanks for the changes. I noticed a couple other things. Once those are resolved, I think it's good to go.

data_steward/cdr_cleaner/cleaning_rules/update_invalid_zip_codes.py
dcampbell-vumc: Nice catch!

dcampbell-vumc:
Made those changes!

jp3477 (approved):
Good to go!

PR #882, opened 2021-05-24
[DC-1661] Exclude death table from cleaning rule to drop zero concept ids

nishanthpp93:
* [DC-1661] Exclude death table

[DC-1661]: https://precisionmedicineinitiative.atlassian.net/browse/DC-1661

### DIFF
--- a/data_steward/cdr_cleaner/cleaning_rules/drop_zero_concept_ids.py
+++ b/data_steward/cdr_cleaner/cleaning_rules/drop_zero_concept_ids.py
@@ -26,12 +26,13 @@ observation
 measurement
     - measurement_source_concept_id
     - measurement_concept_id
-death
-    - cause_source_concept_id
-    - cause_concept_id
 
 Remove those rows from the clean dataset
 Archive/sandbox those rows
+
+As of DC-1661, the death table has been removed.
+This allows the death table with suppressed cause_concept_id and
+cause_source_concept_id to persist without being deleted.9
 """
 
 # Python imports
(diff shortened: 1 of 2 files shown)


data_steward/cdr_cleaner/cleaning_rules/drop_zero_concept_ids.py
```diff
@@ -26,12 +26,13 @@
...
-death
-    - cause_source_concept_id
-    - cause_concept_id
 
 Remove those rows from the clean dataset
 Archive/sandbox those rows
+
+As of DC-1661, the death table has been removed.
+This allows the death table with suppressed cause_concept_id and
+cause_source_concept_id to persist without being deleted.9
```
dcampbell-vumc, line 35: Small typo at end of sentence

```suggestion
cause_source_concept_id to persist without being deleted.
```

dcampbell-vumc (changes requested):
Small typo in the doc string comment. The unit test runs as expected however I think we should update the unit test doc string and remove the reference to the death table as well as add the comment you put in the cleaning rule doc string about the change reasoning. That could prevent any future confusion.

data_steward/cdr_cleaner/cleaning_rules/drop_zero_concept_ids.py
nishanthpp93: Thanks for catching this, I updated it

dcampbell-vumc (approved):
Thanks for making those changes! Unit test ran successfully so this is good to go!

PR #885, opened 2021-05-26
[DC-1644] Remove EHR data where EHR consent cannot be located

ksdkalluri:
* Added Cleaning rule to remove EHR data where EHR consent cannot be located
* Added Integration tests


jp3477 (approved):
Looks great and does what it's supposed to!

PR #889, opened 2021-05-28
[DC-1698] observation_source_value Values Not Nulled In R2021q2r1_deid

dcampbell-vumc:
* Add cleaning rule to REGISTERED_TIER_DEID_CLEANING_CLASSES list
* Updated affected datasets in cleaning rule module

### DIFF
--- a/data_steward/cdr_cleaner/clean_cdr.py
+++ b/data_steward/cdr_cleaner/clean_cdr.py
@@ -274,7 +274,7 @@ CONTROLLED_TIER_DEID_CLEAN_CLEANING_CLASSES = [
 
 REGISTERED_TIER_DEID_CLEANING_CLASSES = [
     (QRIDtoRID,),  # Should run before any row suppression rules
-    (FillSourceValueTextFields,),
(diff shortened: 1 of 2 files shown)


data_steward/cdr_cleaner/clean_cdr.py
```diff
@@ -274,6 +274,7 @@
 
 REGISTERED_TIER_DEID_CLEANING_CLASSES = [
     (QRIDtoRID,),  # Should run before any row suppression rules
+    (FillSourceValueTextFields,),
```
lrwb-aou, line 277: This should be:
```suggestion
    (StringFieldsSuppression,),
``` 
This is the wrong class.

lrwb-aou (changes requested):

data_steward/cdr_cleaner/clean_cdr.py
dcampbell-vumc: 🙈😹😅

lrwb-aou (approved):
This is approved!  :)  Looks good now.

PR #890, opened 2021-06-01
[DC-1643] ValidDeathDates should remove death_dates occurring before any PPI response

jp3477:
excluded records that have death before any ppi records.

### DIFF
--- a/data_steward/cdr_cleaner/cleaning_rules/valid_death_dates.py
+++ b/data_steward/cdr_cleaner/cleaning_rules/valid_death_dates.py
@@ -39,18 +39,18 @@ SELECT d.*
 FROM `{{project_id}}.{{dataset_id}}.{{table}}` d
 LEFT JOIN (
         SELECT
-            person_id, MIN(o.observation_date) first_ppi_date
+            person_id, MAX(o.observation_date) last_ppi_date
         FROM `{{project_id}}.{{dataset_id}}.observation` o
         JOIN `{{project_id}}.{{dataset_id}}.concept` c
             ON c.concept_id =  o.observation_source_concept_id
         WHERE c.vocabulary_id = 'PPI'
         GROUP BY person_id
-) first_ppi_date
-    ON first_ppi_date.person_id = d.person_id
+) last_ppi_date
+    ON last_ppi_date.person_id = d.person_id
 WHERE {{table}}_date < '{{program_start_date}}' OR {{table}}_date > {{current_date}}
     OR (
-        first_ppi_date.person_id IS NULL 
-        OR DATE_DIFF(first_ppi_date.first_ppi_date, d.{{table}}_date, DAY) >= 1
+        last_ppi_date.person_id IS NULL 
+        OR DATE_DIFF(last_ppi_date.last_ppi_date, d.{{table}}_date, DAY) >= 1
     ) 
 )
 """)
--- a/data_steward/cdr_cleaner/cleaning_rules/valid_death_dates.py
+++ b/data_steward/cdr_cleaner/cleaning_rules/valid_death_dates.py
@@ -37,6 +37,7 @@ SANDBOX_INVALID_DEATH_DATE_ROWS = common.JINJA_ENV.from_string("""
 CREATE OR REPLACE TABLE `{{project_id}}.{{sandbox_id}}.{{sandbox_table}}` AS (
 SELECT d.*
 FROM `{{project_id}}.{{dataset_id}}.{{table}}` d
+-- Find the latest PPI observation for each person --
 LEFT JOIN (
         SELECT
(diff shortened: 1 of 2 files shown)


data_steward/cdr_cleaner/cleaning_rules/valid_death_dates.py
```diff
@@ -35,9 +35,24 @@
...
 SANDBOX_INVALID_DEATH_DATE_ROWS = common.JINJA_ENV.from_string("""
 CREATE OR REPLACE TABLE `{{project_id}}.{{sandbox_id}}.{{sandbox_table}}` AS (
-SELECT *
-FROM `{{project_id}}.{{dataset_id}}.{{table}}`
-WHERE {{table}}_date < '{{program_start_date}}' OR {{table}}_date > {{current_date}})
+SELECT d.*
+FROM `{{project_id}}.{{dataset_id}}.{{table}}` d
+LEFT JOIN (
+        SELECT
+            person_id, MIN(o.observation_date) first_ppi_date
```
lrwb-aou, line 42: This looks like it will only find the earliest date.  If I interpret 1643 correctly, we want to sandbox any death records that exist before any PPI records.  Added the following test code to check the current implementation when the death date falls between PPI records.  It fails with the existing implementation.
[1643_patch.txt](https://github.com/all-of-us/curation/files/6580336/1643_patch.txt)

data_steward/cdr_cleaner/cleaning_rules/valid_death_dates.py
```diff
@@ -35,9 +35,24 @@
 # or after the current date.
 SANDBOX_INVALID_DEATH_DATE_ROWS = common.JINJA_ENV.from_string("""
 CREATE OR REPLACE TABLE `{{project_id}}.{{sandbox_id}}.{{sandbox_table}}` AS (
-SELECT *
-FROM `{{project_id}}.{{dataset_id}}.{{table}}`
-WHERE {{table}}_date < '{{program_start_date}}' OR {{table}}_date > {{current_date}})
+SELECT d.*
+FROM `{{project_id}}.{{dataset_id}}.{{table}}` d
+LEFT JOIN (
```
lrwb-aou, line 40: This is a really nice query.  Consider using Jinja `-- comment --` format to describe in plain English what this inner join is accomplishing.

lrwb-aou (changes requested):
The updated test code fails the scenario where we have PPI record, death record, another PPI record.

data_steward/cdr_cleaner/cleaning_rules/valid_death_dates.py
jp3477: Sounds good, will make that change.

data_steward/cdr_cleaner/cleaning_rules/valid_death_dates.py
jp3477: Will add that, thanks

data_steward/cdr_cleaner/cleaning_rules/valid_death_dates.py
jp3477: Made the change. I think just flipping the MIN to a MAX did the trick.

tests/integration_tests/data_steward/cdr_cleaner/cleaning_rules/valid_death_dates_test.py
```diff
@@ -89,8 +119,16 @@ def setUp(self):
         fq_dataset_name = self.fq_table_names[0].split('.')
         self.fq_dataset_name = '.'.join(fq_dataset_name[:-1])
 
+        self.copy_vocab_tables()
+
         super().setUp()
 
+    def copy_vocab_tables(self):
```
lrwb-aou, line 126: nit:  This is an optional thing, but I'm curious about why you are creating a function to run a loop instead of just running the loop.

jp3477: I had copied this method from Nishanth, so you would have to ask him. But if I had to guess, it would be because it's more portable.

lrwb-aou: I would suggest that this isn't portable because you're not importing this function and using it, you're redefining it.  I would recommend just using the loop and adding a comment above it about what it's doing.  This is adding a bit of extra complexity for just 2 lines of logical code.

jp3477: Sounds good, will do that

lrwb-aou (approved):
The integration test here passed.  🎉   This is approved when the test cycle finishes.

PR #893, opened 2021-06-03
[DC-1692] Suppress Covid EHR vaccine records

jp3477:
Suppresses COVID EHR vaccine records in all CDM tables.


ksdkalluri (approved):
The code looks good and doing what is supposed to be done. I verified all the concepts identified by the query and all the concepts belong to EHR-related vocabularies.

PR #898, opened 2021-06-08
[DC-1639] Update pid_rid_mapping comparison in rdr_qc notebook to use USING instead of ON

ksdkalluri:


lrwb-aou (approved):

data_steward/tools/create_rdr_snapshot.py
```diff
@@ -103,7 +111,9 @@ def main(raw_args=None):
     cleaning_args = [
         '-p', args.curation_project_id, '-d',
         datasets.get('staging', 'UNSET'), '-b',
-        datasets.get('sandbox', 'UNSET'), '--data_stage', 'rdr'
+        datasets.get('sandbox',
+                     'UNSET'), '--data_stage', 'rdr', '--truncation_date',
+        args.truncation_date, '--export_date', args.export_date
```
lrwb-aou, line 116: nice addition of the export_date.

lrwb-aou:
@ksdkalluri Need to run yapf on `data_steward/cdr_cleaner/cleaning_rules/truncate_rdr_using_date.py`.  

Are you going to add the CDM table fix to this one, or is that going in a different PR?

ksdkalluri:
> @ksdkalluri Need to run yapf on `data_steward/cdr_cleaner/cleaning_rules/truncate_rdr_using_date.py`.
> 
> Are you going to add the CDM table fix to this one, or is that going in a different PR?

I did add the CDM table fix here.

data_steward/tools/import_rdr_omop.py
```diff
@@ -86,6 +86,8 @@ def create_rdr_tables(client, rdr_dataset, bucket):
     for table, schema in schema_dict.items():
         schema_list = bq.get_table_schema(table, schema)
         table_id = f'{project}.{rdr_dataset}.{table}'
+        table = bigquery.Table(table_id, schema=schema_list)
+        table = client.create_table(table)
```
lrwb-aou, line 90: I'm not crazy about the way this loop is structured now.  Proposed lines 89 and 90 are a functional duplicate of current lines 126 and 127. With this addition, we create each table, then try to write truncate the table on load, and if the file doesn't exist, try to create the table again in line 129 (in the patch), which will probably raise a Table Already Exists error because these lines executed.

I think we should just remove the conditional logic at line 123/125 (current line/patched line) and move the entire conditional block over.
```
            if table in AOU_REQUIRED:
``` 
to achieve the same effect and ensure we are only creating each table once.  Instead of adding these two lines.

The current patched approach is to create the table, then recreate the table at either file load or file load failure.
This changes the original approach of "create the table once, either at file load or as an empty table".

data_steward/tools/import_rdr_omop.py
```diff
@@ -120,12 +120,12 @@ def create_rdr_tables(client, rdr_dataset, bucket):
...
-                LOGGER.info(f'Created empty table `{dest_table.full_table_id}`')
+
+            LOGGER.info(f'Creating empty CDM table, `{table}`')
+            dest_table = f'{client.project}.{rdr_dataset}.{table}'
+            dest_table = bigquery.Table(dest_table, schema=schema_list)
+            dest_table = client.create_table(dest_table)
+            LOGGER.info(f'Created empty table `{dest_table.full_table_id}`')
         else:
             destination_table = client.get_table(
                 table_id)  # Make an API request.
```
mark-velez, line 131: nit: I think `table_id` can be reused here. Also the same info log could apply following the try-except-else block (note un-indent).
```suggestion
            destination_table = bigquery.Table(table_id, schema=schema_list)
            destination_table = client.create_table(destination_table)
        else:
            destination_table = client.get_table(
                table_id)  # Make an API request.
        LOGGER.info(f'Loaded {destination_table.num_rows} rows into '
                    f'`{destination_table.full_table_id}`.')
```

mark-velez (approved):
Looks good to me, @ksdkalluri . Consider my optional suggestions.

PR #899, opened 2021-06-09
[DC-1702] Update Google API dependencies

dcarbone:
This PR updates the underlying google api sdk's used to the latest versions, with the exception of https://pypi.org/project/google-cloud-logging/, as versions >= 2 produced an error I do not have time to resolve.


nishanthpp93 (commented):
I am assuming all the dependencies showed up via pip freeze after a fresh install of the requirements. This is very helpful and in the past we've missed dependencies, causing conflicts during deployment. Has this been deployed to the test project? If no issues show up, this might be good but I'll let others pitch in as well.

tests/integration_tests/data_steward/cdr_cleaner/cleaning_rules/bigquery_tests_base.py
```diff
@@ -152,11 +152,14 @@ def assertTableDoesNotExist(self, fq_table_name):
 
             :param fq_table_name: a fully qualified table name
             """
-            response = self.client.query(
-                f"select count(*) from `{fq_table_name}`")
-            self.assertRaises(gc_exc.GoogleCloudError,
-                              response.result,
-                              timeout=15)
+            stmt = f"select count(*) from `{fq_table_name}`"
+            # response = self.client.query(stmt)
```
lrwb-aou, line 156: dead code statement?

tests/unit_tests/data_steward/validation/participants/identity_match_test.py
```diff
@@ -410,8 +411,8 @@ def test_match_participants_same_participant_simulate_location_pii_read_errors(
 
     def test_match_participants_same_participant_simulate_pii_read_errors(self):
         # pre conditions
-        self.mock_pii_values.side_effect = googleapiclient.errors.HttpError(
-            500, b'bar', b'baz')
+        self.mock_pii_values.side_effect = test_util.mock_google_http_error(
+            status_code=500, content=b'baz', reason='bar')
```
lrwb-aou, line 415: why change this mock to use a test_util object instead of an actual expected error?

tests/integration_tests/data_steward/cdr_cleaner/cleaning_rules/identifying_field_suppression_test.py
```diff
@@ -155,8 +155,8 @@ def test_identifying_field_suppression(self):
                     'measurement_type_concept_id', 'provider_id'
                 ],
                 'loaded_ids': [321, 123],
-                'cleaned_values': [(321, 12345, 111, 444, None),
-                                   (123, 6789, 222, 555, None)]
+                'cleaned_values': [(321, 12345, 111, 444, 0),
+                                   (123, 6789, 222, 555, 0)]
```
lrwb-aou, line 159: Why change these from `None` (which evaluates like `null` in the response) to 0?  `provider_id` is a nullable field.

tests/integration_tests/data_steward/cdr_cleaner/cleaning_rules/identifying_field_suppression_test.py
```diff
@@ -167,10 +167,8 @@ def test_identifying_field_suppression(self):
                     'provider_id', 'care_site_id'
                 ],
                 'loaded_ids': [12345, 6789],
-                'cleaned_values': [
-                    (12345, 1, 1990, None, None, None, None, None),
-                    (6789, 2, 1980, None, None, None, None, None)
-                ]
+                'cleaned_values': [(12345, 1, 1990, 0, 0, 0, 0, 0),
+                                   (6789, 2, 1980, 0, 0, 0, 0, 0)]
```
lrwb-aou, line 171: similar to above, these are all nullable fields.  during deidentification, we set nullable fields to `null` if the field data is to be removed.  This test does not seem to meet that expectation.

tests/unit_tests/data_steward/validation/participants/identity_match_test.py
```diff
@@ -369,8 +370,8 @@ def test_match_participants_same_participant_simulate_write_errors(self):
     def test_match_participants_same_participant_simulate_location_pii_read_errors(
         self):
         # pre conditions
-        self.mock_location_pii.side_effect = googleapiclient.errors.HttpError(
-            500, b'bar', b'baz')
+        self.mock_location_pii.side_effect = test_util.mock_google_http_error(
+            status_code=500, content=b'baz', reason='bar')
```
lrwb-aou, line 374: what does this `test_util` object give us that is better than just using python mock objects?  This same comment applies to the other areas replacing mocks that are set to actual objects with this faked test_util object.

lrwb-aou (commented):
I've got some questions.  
1. Did you ever load this into the test environment like Nishanth asked?
2. You updated the requirements.txt file only.  Now it's got the mock, pylint, pipdeptree, coverage, etc (all dev_requirements.txt definitions) packages lumped in with production requirements packages.  Was this intentional?
3. I thought this was primarily about updating dependencies, but it goes into multiple testing updates.  Why?

tests/integration_tests/data_steward/cdr_cleaner/cleaning_rules/bigquery_tests_base.py
dcarbone: Yeah, that can be removed.  I'll do a cleanup push.

tests/integration_tests/data_steward/cdr_cleaner/cleaning_rules/identifying_field_suppression_test.py
dcarbone: Typing issues.  The tests failed as the value was not compatible when set to 0.

tests/integration_tests/data_steward/cdr_cleaner/cleaning_rules/identifying_field_suppression_test.py
dcarbone: As with above, when running the integration tests they were failing due to `None` not being a valid value.  I don't remember if it was due to the column definition itself in the test tables or due to the way the tests themselves are run.

tests/unit_tests/data_steward/validation/participants/identity_match_test.py
dcarbone: I'm not sure I understand the question.  The `HttpError` object now must be given a `requests.Response` instance at construction, preventing all existing definitions from working.  To that end, this func builds an upstream `HttpError` instance with a valid `requests.Response` instance that can be used with the mock funcs, as is demonstrated here.

Perhaps we should have a discussion as to what you would propose as an alternative?

tests/unit_tests/data_steward/validation/participants/identity_match_test.py
dcarbone: That test util func returns an actual object that is constructed with a `requests.Response` type as is now required.

It was previous not required, and I did not want to manually build an `HttpError` instance _and_ a `requests.Response` instance every time.

PR #901, opened 2021-06-14
[DC-1580] Create python script to run dataset output process

jp3477:
Created a python script from transferring tables from one project to another.

### DIFF
--- a/data_steward/requirements.txt
+++ b/data_steward/requirements.txt
@@ -21,6 +21,7 @@ google-auth==1.11.0
 google-auth-httplib2==0.0.3
 google-auth-oauthlib==0.4.1
 google-cloud-bigquery==1.24.0
+google-cloud-bigquery-datatransfer==3.1.1
 google-cloud-core==1.2.0
 google-cloud-logging==1.14.0
 google-cloud-storage==1.29.0
--- a/data_steward/requirements.txt
+++ b/data_steward/requirements.txt
@@ -21,7 +21,6 @@ google-auth==1.11.0
 google-auth-httplib2==0.0.3
 google-auth-oauthlib==0.4.1
 google-cloud-bigquery==1.24.0
-google-cloud-bigquery-datatransfer==3.1.1
 google-cloud-core==1.2.0
 google-cloud-logging==1.14.0
 google-cloud-storage==1.29.0
--- a/data_steward/tools/copy_dataset_to_output_prod.py
+++ b/data_steward/tools/copy_dataset_to_output_prod.py
@@ -9,7 +9,8 @@ import argparse
 
 # Project imports
 from utils import bq, auth, pipeline_logging
-from tools.create_tier import (TIER_LIST, DEID_STAGE_LIST, get_dataset_name)
+from tools.create_tier import (TIER_LIST, DEID_STAGE_LIST,
+                               validate_create_tier_args)
 from tools.recreate_person import update_person
 
 SCOPES = [
@@ -18,6 +19,29 @@ SCOPES = [
     'https://www.googleapis.com/auth/cloud-platform'
 ]
 LOGGER = logging.getLogger(__name__)
+
+
+def get_dataset_name(tier, release_tag, deid_stage):
+    """
+    Helper function to create the output prod dataset name based on the given criteria
+    This function should return a name for the final dataset only (not all steps along the way)
+    The function returns a string in the form: [C|R]{release_tag}[_base|_clean]
+
+    :param tier: controlled or registered tier intended for the output dataset
+    :param release_tag: release tag for dataset in the format of YYYYq#r#
+    :param deid_stage: deid stage (deid, base or clean)
+    :return: a string for the dataset name
+    """
+    # validate parameters
+    validate_create_tier_args(tier, deid_stage, release_tag)
+
+    tier = tier[0].upper()
+    release_tag = release_tag.upper()
+    deid_stage = deid_stage.replace('deid', '')
+
+    dataset_name = f"{tier}{release_tag}{deid_stage}"
+
+    return dataset_name
 
 
 def get_arg_parser() -> argparse.ArgumentParser:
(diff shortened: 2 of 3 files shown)


data_steward/requirements.txt
```diff
@@ -21,6 +21,7 @@ google-auth==1.11.0
 google-auth-httplib2==0.0.3
 google-auth-oauthlib==0.4.1
 google-cloud-bigquery==1.24.0
+google-cloud-bigquery-datatransfer==3.1.1
```
nishanthpp93, line 24: This can be removed

data_steward/tools/copy_dataset_to_output_prod.py
```diff
@@ -0,0 +1,120 @@
...
+                                 dest='src_dataset_id',
+                                 action='store',
+                                 help='The source dataset to copy.',
+                                 required=True)
+    argument_parser.add_argument(
+        '-r',
+        '--release_tag',
+        action='store',
+        dest='release_tag',
+        help='release tag for dataset in the format of YYYYq#r#',
```
nishanthpp93, line 58: I believe the release tag for output-prod is different from the naming used in prod and should be all caps - this can be a check during parsing (check if isupper()) for that argument type https://docs.python.org/3/library/argparse.html#type.

nishanthpp93 (commented):
Looks great! Added a couple suggestions

data_steward/tools/copy_dataset_to_output_prod.py
jp3477: Looking at the `C2021Q2R1` as an example, it looks like the dataset name is all caps but the release tag label (`2021q2r1`) is lowercase. I'm thinking I can still accept the lowercase tag as an arg and then convert it to caps for the dataset name. Is that fine?

jp3477: I created a new `get_dataset_name` that parameterizes the name properly. It also removes "deid" from the dataset name.

nishanthpp93: Not always. This means the user could type in the internal release tag by mistake (eg `2021q3r4`), instead of the output release tag (eg `2021Q3R1`). Output release number should be set to 1 unless there's a hotfix, regardless of the internal release number, so converting it to upper may cause issues.

jp3477: Ok I see, I'll change the arguments then.

data_steward/tools/copy_dataset_to_output_prod.py
```diff
@@ -0,0 +1,204 @@
...
+                                 dest='src_dataset_id',
+                                 action='store',
+                                 help='The source dataset to copy.',
+                                 required=True)
+    argument_parser.add_argument(
+        '-r',
+        '--release_tag',
+        action='store',
+        dest='release_tag',
+        help='release tag for dataset in the format of YYYYQ#R#',
```
nishanthpp93, line 142: You can also add `type=validate_release_tag_param` so it exits immediately if arg is not in the right format (can be unit tested)

data_steward/tools/copy_dataset_to_output_prod.py
```diff
@@ -0,0 +1,204 @@
...
+    User defined helper function to validate that the tier, deid_stage, release_tag parameter
+     follows the correct naming convention
+    :param tier: tier parameter passed through from either a list or command line argument
+    :param stage: deid_stage parameter passed through from either a list or command line argument
+    :param tag: release tag parameter passed through either the command line arguments
+    :return: None, breaks if not valid params are passed
+    """
+    validate_tier_param(tier)
+    validate_deid_stage_param(stage)
+    validate_release_tag_param(tag)
```
nishanthpp93, line 81: This may not be necessary, argparse can validate choices and the string when the args are passed.

data_steward/tools/copy_dataset_to_output_prod.py
```diff
@@ -0,0 +1,204 @@
...
+    helper function to validate the deid_stage parameter passed is correct, must be 'deid', 'base' or 'clean'
+
+    :param deid_stage: deid_stage parameter passed through from either a list or command line argument
+    :return: nothing, breaks if not valid
+    """
+    if deid_stage not in DEID_STAGE_LIST:
+        msg = f"Parameter ERROR: {deid_stage} is an incorrect input for the deid_stage parameter, accepted: deid, " \
+              f"base, clean"
+        LOGGER.error(msg)
+        raise argparse.ArgumentTypeError(msg)
```
nishanthpp93, line 51: This should already be validated since you're using choices in argparse. A unit test would help validate this.

data_steward/tools/copy_dataset_to_output_prod.py
```diff
@@ -0,0 +1,204 @@
...
+    :return: a string for the dataset name
+    """
+    # validate parameters
+    validate_create_tier_args(tier, deid_stage, release_tag)
+
+    tier = tier[0].upper()
+    release_tag = release_tag.upper()
+    deid_stage = '' if release_tag == 'deid' else f'_{deid_stage}'
+
+    dataset_name = f"{tier}{release_tag}{deid_stage}"
```
nishanthpp93, line 102: Another quirk about the output-prod project: The clean dataset won't have `_clean`, it will be just the tier and release tag. The base dataset will have `_base`

nishanthpp93 (commented):
Thanks for adding the previous suggestions, added some more. I think a unit test for the arg parse would help a lot. There are existing tests you can reference which check for SystemExit. Let me know if you run into any issues while testing.

data_steward/tools/copy_dataset_to_output_prod.py
jp3477: Sounds good, will work on all these suggestions.

nishanthpp93 (approved):
Looks great! Thanks for making all the changes

PR #902, opened 2021-06-16
[DC-1708] Create fitbit dataset

nishanthpp93:
* [DC-1708] Create fitbit package

[DC-1708]: https://precisionmedicineinitiative.atlassian.net/browse/DC-1708


dcampbell-vumc (changes requested):
All the unit and integration tests are passing as normals. Only thing that needs to be updated is the class doc string in `truncate_fitbit_data`. It is referencing truncating all dates after 11/26/2019, this should be removed since we are now going to be passing in the truncation date 👍🏾

dcampbell-vumc (approved):
This looks good!

dcampbell-vumc (approved):
This is good to go!

PR #903, opened 2021-06-21
[DC-1319] RDR eval: Add check for basics survey module

dcampbell-vumc:
* Added code to check that participants with data in other survey modules also have data from the basics survey module to the RDR notebook

### DIFF
--- a/data_steward/analytics/cdr_ops/rdr_export_qc.py
+++ b/data_steward/analytics/cdr_ops/rdr_export_qc.py
@@ -432,9 +432,10 @@ pd.read_gbq(query, dialect='standard')
 # associated with the basics survey module.
 # In ideal circumstances, this query will not return any results.
 
-query = f'''SELECT DISTINCT person_id FROM `{project_id}.{new_rdr}.person` 
-WHERE person_id NOT IN (
-SELECT DISTINCT person_id FROM `{project_id}.{new_rdr}.concept`  
+query = f'''SELECT DISTINCT person_id FROM `{project_id}.{new_rdr}.observation` 
+JOIN `{project_id}.{new_rdr}.concept` on (observation_source_concept_id=concept_id)
+WHERE vocabulary_id = 'PPI' AND person_id NOT IN (
+SELECT DISTINCT person_id FROM `{project_id}.{new_rdr}.rdr20191203.concept`  
 JOIN `{project_id}.{new_rdr}.concept_ancestor` on (concept_id=ancestor_concept_id)
 JOIN `{project_id}.{new_rdr}.observation` on (descendant_concept_id=observation_concept_id)
 WHERE concept_class_id='Module'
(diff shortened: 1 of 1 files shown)


data_steward/analytics/cdr_ops/rdr_export_qc.py
```diff
@@ -425,3 +425,20 @@
...
 pd.read_gbq(query, dialect='standard')
+
+# Checks for basics survey module
+# Participants with data in other survey modules must also have data from the basics survey module.
+# This check identifies survey responses associated with participants that do not have any responses
+# associated with the basics survey module.
+# In ideal circumstances, this query will not return any results.
+
+query = f'''SELECT DISTINCT person_id FROM `{project_id}.{new_rdr}.person` 
+WHERE NOT EXISTS (
```
jp3477, line 436:
```suggestion
WHERE person_id NOT IN (
```
I think this is the correct way to find persons who are not selected from the subquery. `WHERE NOT EXISTS` will always return no rows if at least 1 row is returned by the subquery.

jp3477 (changes requested):
Added one suggestion below.

Also, reading through the ticket description, I can't tell if the check should be for persons with no Basics AND at least 1 other survey response (i.e. excluding persons with no survey responses at all). What do you think about that?

data_steward/analytics/cdr_ops/rdr_export_qc.py
dcampbell-vumc: Nice catch! I'll make that change and repush!

dcampbell-vumc:
> Added one suggestion below.
> 
> Also, reading through the ticket description, I can't tell if the check should be for persons with no Basics AND at least 1 other survey response (i.e. excluding persons with no survey responses at all). What do you think about that?

I think so as well. If a participant doesn't have any basics data but has other survey data that needs to be flagged. If the participant has no basics data and no other survey data at all, that's fine because they may have just signed up for the program and nothing has been filled out other than the preliminary stuff.

jp3477:
> > Added one suggestion below.
> > Also, reading through the ticket description, I can't tell if the check should be for persons with no Basics AND at least 1 other survey response (i.e. excluding persons with no survey responses at all). What do you think about that?
> 
> I think so. If a participant doesn't have any basics data but has other survey data that needs to be flagged. If the participant has no basics data and no other survey data at all, that's fine because they may have just signed up for the program and nothing has been filled out other than the preliminary stuff.

Gotcha, that makes sense. I think the query might need another join to `observation` then, in order to find the additional survey records. Currently it is flagging the latter case where he participants have no records.

jp3477 (changes requested):
Thanks for making the last change. As mentioned in my previous comment, I think another join to `observation` is needed to exclude persons with no other survey responses.

dcampbell-vumc:
@jp3477 I validated that the participants generated from the query have survey data but no basics survey data. I also ensured that if a participant has NO survey data, they are not in the list of participants that meet the above criteria. This is ready for another review!

jp3477 (approved):
Thanks for the changes. Good to go!

PR #904, opened 2021-06-23
[DC-1566] Add CT fitbit cleaning classes

jp3477:
Added CT fitbit cleaning classes to clean_cdr.


ksdkalluri (approved):
Changes look good. The yapf linting check failed as the commit message is missing [DC-1566] Jira tag in it. Please fix the commit message and merge when all the checks pass.

PR #905, opened 2021-06-25
[DC-1675] Remove usage of staging_sandbox and drop steps that delete staging datasets

ksdkalluri:
Update clean_cdr to use combined_sandbox instead of combined_staging_sandbox

* remove steps that drop staging and staging_sandbox datasets


nishanthpp93 (approved):
Glad to see staging sandbox removed!

PR #906, opened 2021-06-25
[DC-1732] Deduplicate rows based on columns except id

nishanthpp93:
* Row deduplicate

### DIFF
--- a/data_steward/cdr_cleaner/cleaning_rules/remove_row_duplicates.py
+++ b/data_steward/cdr_cleaner/cleaning_rules/remove_row_duplicates.py
@@ -5,6 +5,7 @@ import logging
 
 # Project imports
 import cdm
+from utils.bq import fields_for
 from constants.cdr_cleaner import clean_cdr as cdr_consts
 from constants.bq_utils import WRITE_TRUNCATE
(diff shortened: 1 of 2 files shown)


data_steward/cdr_cleaner/cleaning_rules/remove_row_duplicates.py
```diff
@@ -0,0 +1,136 @@
...
+        this SQL, append them to the list of Jira Issues.
+        DO NOT REMOVE ORIGINAL JIRA ISSUE NUMBERS!
+        """
+        desc = (
+            'Remove the duplicate columns excluding ID from OMOP tables in a given dataset'
+        )
+
+        super().__init__(issue_numbers=JIRA_ISSUE_NUMBERS,
+                         description=desc,
+                         affected_datasets=[cdr_consts.UNIONED],
```
ksdkalluri, line 56: @nishanthpp93 This needs to be in combined cleaning classes right?

ksdkalluri (changes requested):

data_steward/cdr_cleaner/cleaning_rules/remove_row_duplicates.py
nishanthpp93: You're right, it is in combined.

ksdkalluri (approved):
Looks good!

PR #907, opened 2021-06-25
[DC-1729] Fix non person_id tending up in AFFECTED_TABLES list

ksdkalluri:


dcampbell-vumc (approved):
I verified that all the tables returned from the `get_person_id_tables` function in `data_stewad/resources.py` contained the `person_id` field. This is good to go!

PR #908, opened 2021-06-25
[DC-1730] Fix sandbox_dataset_id name generation logic usage in sandbox_and_remove_pids.py

ksdkalluri:


dcampbell-vumc (commented):
Quick question about a TODO item in the doc string:

dcampbell-vumc (approved):
Perfect! This is good to go!

PR #909, opened 2021-06-25
[DC-1733] Simplify and fix query in temp_consistency

nishanthpp93:
* [DC-1733] Remove left joins

[DC-1733]: https://precisionmedicineinitiative.atlassian.net/browse/DC-1733

### DIFF
--- a/data_steward/cdr_cleaner/cleaning_rules/temporal_consistency.py
+++ b/data_steward/cdr_cleaner/cleaning_rules/temporal_consistency.py
@@ -88,15 +88,15 @@ FROM (
     vo.*
   FROM
     `{{project_id}}.{{dataset_id}}.visit_occurrence` vo
-  LEFT JOIN
+  JOIN
     `{{project_id}}.{{dataset_id}}.condition_occurrence` co
   ON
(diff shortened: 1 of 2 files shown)


lrwb-aou (commented):
Consider adding the following test patch changes.  It should fail on the current develop branch, but pass here.  Just to exercise this patch and make sure we include a test case that exercises this scenario.

[temporal_consistency_test_patch.txt](https://github.com/all-of-us/curation/files/6719068/temporal_consistency_test_patch.txt)

ksdkalluri:
@nishanthpp93 can you update DC-1731 to reflect these changes in the scope and make link this PR.

nishanthpp93:
@ksdkalluri I think it's better to not change the scope and keep them separate. DC-1731 should be the investigation ticket and DC-1733 the fix. I've linked the tickets so they can be referenced easily.

nishanthpp93:
@lrwb-aou Thank you for updating the test to verify! I have applied the patch and it passes successfully.

lrwb-aou (approved):
The `TemporalConsistencyTest` has passed now.  It all looks good to merge!

PR #913, opened 2021-07-01
[DC-1727] Remove snap codes from the rdr qc notebook

rfrancis1:


mark-velez (approved):
Thanks for this pull request! If possible, could you add a notice about `snap_codes` in the markdown (i.e. [here](https://github.com/all-of-us/curation/blob/cc2983154648a1dba1f7a27d95a7c726786b1dec/data_steward/analytics/cdr_ops/rdr_export_qc.py#L136)) like this?

```
Note: Snap codes are not modeled in the vocabulary but may be used in the RDR export. 
They are excluded here by filtering out snap codes in the Public PPI Codebook 
which were loaded into `curation_sandbox.snap_codes`.
```

Fyi I added spreadsheet info to the table description in bigquery.

rfrancis1:
@mark-velez the notes were added.

PR #914, opened 2021-07-01
[DC-1214] Store Participant Summary data in a table

jp3477:
Calls and stores data from the Participant Summary API into a BQ dataset.


ksdkalluri (approved):
Good updates Jason. Looks good to me!

PR #919, opened 2021-07-07
[DC-1720] Analytics notebook for combined dataset

mark-velez:


data_steward/analytics/cdr_ops/combined.py
```diff
@@ -0,0 +1,479 @@
...
+#     language: python
+#     name: python3
+# ---
+
+# ## Notebook parameters
+
+# + tags=["parameters"]
+PROJECT_ID = ''  # identifies the project containing the datasets
+DATASET_ID = ''  # the dataset to evaluate
+BASELINE_DATASET_ID = ''  # a baseline dataset for metrics comparison
```
mark-velez, line 20: From @rfrancis1 : Add a comment saying that this should be an older combined dataset

data_steward/analytics/cdr_ops/combined.py
```diff
@@ -0,0 +1,479 @@
...
+    EHR_CONSENT_PERMISSION_CONCEPT_ID=EHR_CONSENT_PERMISSION_CONCEPT_ID,
+    YES_CONCEPT_ID=YES_CONCEPT_ID)
+q(query)
+# -
+
+# ## Date and datetime fields should have the same date
+# The date represented by associated `_date` and `_datetime` fields of the same row should be the same. If there any discrepancies, there may be a bug in the pipeline (i.e. `ensure_date_datetime_consistency`). It may also be useful to report discrepancies to EHR Ops.
+#
+# ### Implementation notes
+# For each associated date/timestamp pair in the dataset the script that follows assembles a universal subquery which can be applied to all pertinent tables and whose results can be compiled together into a single result set that provides relevant troubleshooting information. For legibility, a maximum of 10 problem rows is returned for each table.
```
lrwb-aou, line 269: nit:  I would suggest breaking the line into smaller line lengths.

data_steward/analytics/cdr_ops/combined.py
```diff
@@ -0,0 +1,479 @@
...
+# ,observation_date       AS date_value
+# ,"observation_datetime" AS timestamp_field
+# ,observation_datetime   AS timestamp_value
+# FROM `DATASET_ID.observation`
+# WHERE observation_date<>DATE(observation_datetime)
+# LIMIT 10
+# ```
+#
+
+q(f'''
```
lrwb-aou, line 287: this style seems to depart from the rest of the style of this script.  By that, I think you've been defining the query string and then calling `q()` with the defined value.  Not a big thing, but seems to depart from the convention you started with.

data_steward/analytics/cdr_ops/combined.py
```diff
@@ -0,0 +1,479 @@
...
+UNION ALL
+
+SELECT
+ project_id
+,dataset_id
+,table_id
+,row_count
+FROM `2021q3r1_combined_release.__TABLES__` 
+WHERE table_id = 'note'
+''')
```
lrwb-aou, line 344: Was this meant to be removed?

data_steward/analytics/cdr_ops/combined.py
```diff
@@ -0,0 +1,479 @@
...
+     || ',' || column_name  || ' AS concept_id '
+     || ', COUNT(1) AS row_count '
+     -- domain table joined with its mapping table --
+     || 'FROM `' || table_schema || '.' || table_name || '` t'
+     || ' JOIN `' || table_schema || '.' || '_mapping_' || table_name || '` m'
+     || '  USING (' || c.table_name || '_id) '
+     || ' LEFT JOIN `{{DATASET_ID}}.concept` c '
+     || '  ON ' || column_name || ' = c.concept_id '
+     
+     -- invalid concept_id --
```
lrwb-aou, line 430: Thank you for these comments!  Makes reading the SQL much more manageable.

data_steward/analytics/cdr_ops/combined.py
```diff
@@ -0,0 +1,479 @@
...
+);
+EXECUTE IMMEDIATE (
+ 'INSERT `{{FQN_INVALID_CONCEPT}}` (dataset_id, table_name, column_name, src_hpo_id, concept_id, row_count) ' || query);
+''')
+q(tpl.render(DATASET_ID=DATASET_ID, FQN_INVALID_CONCEPT=FQN_INVALID_CONCEPT))
+q(
+    tpl.render(DATASET_ID=BASELINE_DATASET_ID,
+               FQN_INVALID_CONCEPT=FQN_INVALID_CONCEPT))
+
+q(f'''
```
lrwb-aou, line 460: This is another instance of using `q()` directly instead of defining the string and using it.  It's an optional fix, but seems to go against the code style in the rest of this script.

data_steward/analytics/cdr_ops/combined.py
```diff
@@ -0,0 +1,479 @@
...
+
+# ## Invalid concept prevalence
+# EHR submissions may provide source `concept_id`s which we assume would be
+# successfully mapped to standard `concept_id`s. This compares the frequency
+# of invalid values in required concept_id fields per hpo between the current
+# combined dataset and a baseline combined dataset. The frequency of invalid
+# values in this dataset should generally be non-increasing compared to baseline.
+# If the prevalence of invalid concepts is increasing, EHR Ops should be alerted.
+
+FQN_INVALID_CONCEPT = 'curation_sandbox.invalid_concept'
```
lrwb-aou, line 394: I'm not sure this should be hard coded.  Why couldn't it use the combined dataset sandbox?  You could define it with `f'{DATSET_ID}_sandbox.invalid_concept`.

data_steward/analytics/cdr_ops/combined.py
```diff
@@ -0,0 +1,479 @@
...
+
+CLIENT = bigquery.Client(project=PROJECT_ID)
+
+pd.options.display.max_rows = 1000
+pd.options.display.max_colwidth = 0
+pd.options.display.max_columns = None
+pd.options.display.width = None
+
+
+def q(query):
```
lrwb-aou, line 38: nit:  `q` is very short and not super descriptive.  Maybe expand the function name.  Objections?

lrwb-aou (changes requested):
Most of the comments are stylistic in nature.  A few regard dead code or hard coded parameters.

data_steward/analytics/cdr_ops/combined.py
```diff
@@ -0,0 +1,479 @@
...
+{% endif %}
+{% endfor %}
+''')
+query = tpl.render(dataset_id=DATASET_ID, date_fields=date_fields)
+q(query)
+
+# ## Participants must have basics data
+# Identify any participants who have don't have any responses
+# to questions in the basics survey module (see [DC-706](https://precisionmedicineinitiative.atlassian.net/browse/DC-706)). These should be
+# reported to the RDR as they are supposed to be filtered out
```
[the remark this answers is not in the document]
mark-velez: My albeit hazy understanding is that historically at least some of these records were dummy data which should be filtered out by RDR. There might also be some "complete participant" designation (or something like that) which should also be part of their filters. Nonetheless advising the user to consult RDR about any records that turn up here might be the most prudent and therefore the advice seems appropriate to me. Would you agree?

data_steward/analytics/cdr_ops/combined.py
```diff
@@ -0,0 +1,479 @@
...
+    
+    :param query: the query to execute
+    """
+    print(query)
+    return CLIENT.query(query).to_dataframe()
+
+
+# -
+
+# ## Check for duplicates in observation
```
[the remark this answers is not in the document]
mark-velez: Done. I wonder if there are technical terms to clearly distinguish the different types of duplicates. Duplicate keys vs duplicate data?

data_steward/analytics/cdr_ops/combined.py
```diff
@@ -0,0 +1,479 @@
...
+  USING (person_id)
+WHERE 
+(
+ -- age <= 0y --
+ t.{{date_field}} < DATE(p.birth_datetime)
+
+ -- age >= 150y --
+ OR DATE_DIFF(t.{{date_field}}, DATE(p.birth_datetime), YEAR) >= 150 
+)
+{% if not loop.last -%}
```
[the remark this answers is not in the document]
mark-velez: It is pretty neat. The syntax never seems to stick though- I always have to look it up.

data_steward/analytics/cdr_ops/combined.py
mark-velez: Thank you! This resolved qualms I had in my use of `CREATE OR REPLACE`; with the table in the combined sandbox `CREATE IF NOT EXISTS` looks like a clearly better choice. So I made that change.

data_steward/analytics/cdr_ops/combined.py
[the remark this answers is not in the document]
lrwb-aou: If we're being technical, I'd call it "duplicate primary key" vs. "data duplicates (or duplicate data)".

data_steward/analytics/cdr_ops/combined.py
mark-velez: I spend more time than one might think on trying to make SQL intelligible. I'm so glad you noticed!

data_steward/analytics/cdr_ops/combined.py
```diff
@@ -0,0 +1,470 @@
...
+# ## Verify dataset properties
+# The dataset metadata must have
+# 1. a description indicating the version of the pipeline run and the input datasets
+# 2. labels `phase`, `release_tag`, `de_identified` with appropriate values
+#
+# In the future this should try to find the dataset using labels supplied via notebook
+# parameters (e.g. `combined`, `release_tag=2021q3r1`, `phase=clean`). Currently
+# there is no way to specify `combined`.
+
+query = f'''
```
mark-velez, line 350: Per @jp3477: look into retrieving labels from dataset object instead

data_steward/analytics/cdr_ops/combined.py
```diff
@@ -0,0 +1,470 @@
...
+
+  -- Only store pids where the latest response is YES --
+  SELECT 
+   person_id
+  FROM pid_last_consent_response
+  WHERE 
+      last_value_source_concept_id = {{YES_CONCEPT_ID}}
+);
+
+EXECUTE IMMEDIATE query;
```
mark-velez, line 259: Per @lrwb-aou , `DECLARE` the variable `query` at the script start, and `SET` it **after** creating the temp table; it's easier to follow that way.

```suggestion
DECLARE query STRING;

-- PIDs whose last EHR consent response was affirmative --
CREATE OR REPLACE TEMP TABLE consented AS
( 
  WITH 

  -- For each pid, the latest EHR consent record --
  pid_last_consent_response AS
  (SELECT DISTINCT 
     p.person_id
    ,LAST_VALUE(o.value_source_concept_id)
      OVER (
       PARTITION BY o.person_id
       ORDER BY o.observation_datetime
       ROWS BETWEEN UNBOUNDED PRECEDING AND UNBOUNDED FOLLOWING
      ) AS last_value_source_concept_id
   FROM `{{DATASET_ID}}.person` p
    JOIN `{{DATASET_ID}}.observation` o
     ON p.person_id = o.person_id
   WHERE observation_source_concept_id = {{EHR_CONSENT_PERMISSION_CONCEPT_ID}})

  -- Only store pids where the latest response is YES --
  SELECT 
   person_id
  FROM pid_last_consent_response
  WHERE 
      last_value_source_concept_id = {{YES_CONCEPT_ID}}
);

SET query = (
SELECT
 -- Generate a subquery for any table containing HPO-submitted data                              --
 -- which finds any HPO-submitted rows whose pids are not in the consent lookup table. --

 STRING_AGG(
     '(SELECT DISTINCT '
      || '"' || table_name || '" AS table_name '
      || ',t.person_id           AS person_id '
      || ',m.src_hpo_id          AS hpo_id '
      || 'FROM `' || table_schema || '.' || table_name || '` t '
      || 'JOIN `' || table_schema || '.' || table_id || '` m ' 
      || 'USING (' || table_name ||'_id) '
      || 'LEFT JOIN consented c '
      || ' USING (person_id)'
      || 'WHERE m.src_hpo_id <> "rdr" AND c.person_id IS NULL)'
   , ' UNION ALL ')
 FROM `{{DATASET_ID}}.INFORMATION_SCHEMA.COLUMNS` c
 JOIN `{{DATASET_ID}}.__TABLES__` t
  ON t.table_id = '_mapping_' || c.table_name
 WHERE column_name = 'person_id'
   AND t.row_count > 0
);

EXECUTE IMMEDIATE query;
```

data_steward/analytics/cdr_ops/combined.py
lrwb-aou: maybe `exe` or `q_exe`?

data_steward/analytics/cdr_ops/combined.py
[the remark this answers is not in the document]
lrwb-aou: yes I can agree.

data_steward/analytics/cdr_ops/combined.py
mark-velez: I changed it to `execute()` 😊

data_steward/analytics/cdr_ops/combined.py
```diff
@@ -0,0 +1,446 @@
...
+# 1. a description indicating the version of the pipeline run and the input datasets
+# 2. labels `phase`, `release_tag`, `de_identified` with appropriate values
+#
+# In the future this should try to find the dataset using labels supplied via notebook
+# parameters (e.g. `combined`, `release_tag=2021q3r1`, `phase=clean`). Currently
+# there is no way to specify `combined`.
+
+dataset = CLIENT.get_dataset(dataset_ref=f'{PROJECT_ID}.{DATASET_ID}')
+
+print(dataset.labels)
```
lrwb-aou, line 357: This is much cleaner than querying for the labels.  It was a nice suggestion @jp3477 .  Is it feasible to just build the label checks in right here?  That way, no one has to think about this too deeply.
```
try:
    phase = dataset.labels['phase']
    deidentified = dataset.labels['de_identified']
    release_tag = dataset.labels['release_tag']
except KeyError as ke:
    print(f'dataset label validation failed because keys were missing entirely')
    print(ke.message)
else:
    invalid_labels = []
    if phase != 'clean':
        invalid_labels.append('phase')
    elif deidentified != 'false':
        invalid_labels.append('de_identified')
    elif !DATASET_ID.contains(release_tag):
        invalid_labels.append('release_tag')

    if invalid_labels:
        print(f'`{", ".join(invalid_labels)}` label(s) is not set to the expected value(s).  did a failure occur in the combined process?')
    else:
        print("labels pass basic inspection")
```

lrwb-aou (approved):
This looks good as is.  I left one suggestion but left it as a suggestion.  It may be scope creep.

PR #924, opened 2021-07-08
[DC-1723] Add a check for participants under 18

jqian2015:


mark-velez (approved):
Looks good, thanks @jqian2015 !

PR #926, opened 2021-07-12
[DC-1748] fix state generalization by population rule

lrwb-aou:
* limit the update statement to only impact state participant survey records
* update unit test to validate it only updates records where observation_source_concept_id = 1585249


ksdkalluri (approved):
The query works. I ran the integration test locally, It worked. You can merge it when Circle tests pass.

PR #927, opened 2021-07-12
[DC-1743] Add `note` table to affected_tables in EnsureDateDatetimeConsistency cleaning class

ksdkalluri:


tests/integration_tests/data_steward/cdr_cleaner/cleaning_rules/ensure_date_datetime_consistency_test.py
```diff
@@ -140,8 +142,23 @@ def test_required_field_datetime_pairs(self):
...
+                  (103, 444444, date('2016-05-01'), timestamp('2015-05-07 11:00:00'), 0, 0, '', '', 0, 0),
+                  (104, 555555, date('2016-05-01'), timestamp('2016-05-01 11:00:00'), 0, 0, '', '', 0, 0),
+                  (105, 666666, date('2016-05-01'), timestamp('2016-05-01 11:00:00'), 0, 0, '', '', 0, 0),
+                  (106, 777777, date('2016-05-01'), timestamp('2016-05-01 11:00:00'), 0, 0, '', '', 0, 0),
+                  (107, 888888, date('2016-05-01'), timestamp('2016-05-01 11:00:00'), 0, 0, '', '', 0, 0)
+                """)
+        condition_query = condition_tmpl.render(
+            fq_dataset_name=self.fq_dataset_name)
+        note_query = note_tmpl.render(fq_dataset_name=self.fq_dataset_name)
+        self.load_test_data([condition_query, note_query])
```
lrwb-aou, line 161: The test for `note` really only needs to have two entries.  This can be simplified with the given patch.  
[ensure_date_datetime_consistency_test_patch.txt](https://github.com/all-of-us/curation/files/6804094/ensure_date_datetime_consistency_test_patch.txt)
This patch tests the two fields in `note` to with  "control" and "expected change" records.

lrwb-aou (commented):
I think the test can be further simplified.  I'm curious why we broke with the OMOP specification and decided to require `note_datetime` back in DC-253.

lrwb-aou (approved):
This is nice!

PR #929, opened 2021-07-12
[DC-1216] Create or Update a Site's DRC Identity Match Table

danyel-campbell:
* Added functions to generate SQL compliant schema definitions from JSON schema object/files to `bq.py` from the `base_cleaning_rule.py` to reduce ticket complexity
** This work is part of [DC-1697 Remove get_bq_fiels_sql & Related Functionality Out Of base_cleaning_rule.py](https://precisionmedicineinitiative.atlassian.net/browse/DC-1697)
* Added identity match constant to `identity_match.py`
* Created drc validation table for a site
* Copy values from ps_value table generated from participant summary api
* Update values in drc validation table to either `missing_rdr` if
* missing all data or `missing_ehr` if missing a value for a column


data_steward/tools/create_update_drc_id_match_table.py
```diff
@@ -0,0 +1,229 @@
...
+    set_expression = []
+    schema_list = bq.get_table_schema(IDENTITY_MATCH_TABLE)
+    ', '.join([item.name for item in schema_list])
+
+    for item in schema_list:
+        set_expression.append(SET_EXPRESSION.render(col=item.name))
+
+    # Work around to remove person_id from the fields list since person_id is required
+    # and will never need to be updated to missing_rdr
+    set_expression.remove("\nperson_id = 'missing_rdr'")
```
jp3477, line 122: Is it possible to do this removal beforehand, maybe in the for loop? Matching this string may get finicky, especially with the newline character.

data_steward/tools/create_update_drc_id_match_table.py
```diff
@@ -0,0 +1,229 @@
...
+    """
+    schema_list = bq.get_table_schema(IDENTITY_MATCH_TABLE)
+    fields_name_str = ',\n'.join([item.name for item in schema_list])
+
+    # person_id will not need to be updated since it will always be in table since it
+    # is a required field
+    fields = fields_name_str.replace('person_id,', '')
+    id_match_table_id = table_id
+
+    # Updates missing rows to contain 'missing_rdr'
```
jp3477, line 144: From my understanding of the ticket, I thought missing column values, not entire rows, should be replaced with "missing_rdr". Is that true?

data_steward/tools/create_update_drc_id_match_table.py
```diff
@@ -0,0 +1,229 @@
...
+MISSING_EHR_QUERY = JINJA_ENV.from_string("""
+UPDATE `{{project_id}}.{{drc_dataset_id}}.{{id_match_table_id}}`
+SET {{col}} = 'missing_ehr'
+WHERE {{col}} IS NULL
+""")
+
+MISSING_RDR_QUERY = JINJA_ENV.from_string("""
+UPDATE `{{project_id}}.{{drc_dataset_id}}.{{id_match_table_id}}`
+SET {{set_expression}}
+WHERE COALESCE({{fields}}) IS NULL
```
jp3477, line 52: This would set "missing_rdr" only if all fields are null/missing. Isn't the intention to set "missing_rdr" field by field?

data_steward/tools/create_update_drc_id_match_table.py
```diff
@@ -0,0 +1,229 @@
+"""
+Utility to create or update a site's DRC identity match table.
+
+There should be a record for each participant and the record should be filled with default values of `missing_rdr` or
+    `missing_ehr`. Each record should contain data for the fields: person_id, first_name, middle_name, last_name,
+    phone_number, email, address_1, address_2, city, state, zip, birth_date, sex, and algorithm.
+
+The record for each of the above fields should default to `missing_rdr` if the joined record in the
+    ps_api_values_<hpo_id> table does not contain any information otherwise, it should default to `missing_rdr`
```
jp3477, line 9:
```suggestion
    ps_api_values_<hpo_id> table does not contain any information otherwise, it should default to `missing_ehr`
```

data_steward/tools/create_update_drc_id_match_table.py
```diff
@@ -0,0 +1,229 @@
...
+                        action='store',
+                        dest='project_id',
+                        help='Project associated with the input datasets',
+                        required=True)
+    parser.add_argument('--hpo_id',
+                        action='store',
+                        dest='hpo_id',
+                        help='awardee name of the site',
+                        required=True)
+    parser.add_argument('-s',
```
jp3477, line 190: This parameter is never used. You can either remove it or add in some logging, which I think may be useful.

jp3477 (changes requested):
Had a few suggestions/questions. Let me know what you think.

data_steward/tools/create_update_drc_id_match_table.py
dcampbell-vumc: This is what the ticket says at the end of the Scope paragraph:
> The record for each field should default to missing_rdr if the joined record in the ps_api_values_<hpo_id> table does not contain any information.  Otherwise, default to missing_ehr

I'm interpreting that as, if there is no other value in the record (which is a row right?) than the `person_id` since that is a required field, it should say `missing_rdr` because that would mean that we didn't get any information from the RDR participant summary API.

data_steward/tools/create_update_drc_id_match_table.py
dcampbell-vumc: I thought it was `missing_ehr` for missing information field by field...

data_steward/tools/create_update_drc_id_match_table.py
dcampbell-vumc: You're right, I will remove that!

data_steward/tools/create_update_drc_id_match_table.py
jp3477: I see, it could be that. But what if all fields are available except one. @lrwb-aou can you weigh in?

lrwb-aou: @jp3477 and @dcampbell-vumc Good questions!  I had to review the ticket to see what is going on.  I think @jp3477's interpretation is correct based on the third acceptance criteria and the commented conversation with Caroline.  

> The record defaults to missing_rdr if no RDR data exists for that person **in that field**.  Otherwise defaults to missing_ehr.

The premise is that for each field we have two possible default values and the default value chosen is based on the "importance" of the missing data.  If we are missing the RDR data, we can't compare anything that is submitted for that field to anything, because we just don't have the data.  

For example, a participant doesn't provide their middle name through the survey portal but their middle name is in their ehr record.  This should record the default value for the field "middle_name" as "missing_rdr" to indicate we can't compare the ehr record data for middle name to the rdr record data for middle name because we are "missing_rdr" data for this field for this participant.

This is in contrast to having middle name data provided by the participant through survey data and not having middle name information available from ehr records.  In this case, the default value would be "missing_ehr" because we can't compare the two values since we are missing the ehr data for this participant for this field.

In the case where middle name data is missing from both ehr and rdr data, we would record the field as "missing_rdr".  Survey data is considered more reliable as it comes directly from the participant.

But maybe, this is not what the question is about.  The default value should be set for each field in the record.  The presence or absence of first name data should not affect the default value of middle name data.  I think the quoted above should be changed to 

> Each field should default to missing_rdr if the joined record in the ps_api_values_<hpo_id> table does not contain any information. Otherwise, default to missing_ehr

The default value should be set based on the data available for each field in the participant record.  It is improbable that we will be missing an entire row of data from the survey/rdr side.  It is likely that we will be missing some fields.

We are interested in defaulting field values, not entire records.

jp3477: Got it, that makes sense.

tests/integration_tests/data_steward/tools/create_update_drc_id_match_table_test.py
```diff
@@ -0,0 +1,136 @@
...
+        # Pre conditions
+        mock_table_schema.return_value = self.schema
+
+        expected = f'\nCASE WHEN first_name IS NULL THEN \'missing_rdr\' ELSE \'missing_ehr\' END AS first_name, , \n' \
+                   f'CASE WHEN last_name IS NULL THEN \'missing_rdr\' ELSE \'missing_ehr\' END AS last_name'
+
+        # Test
+        actual = id_validation.get_case_statements()
+
+        self.assertEqual(actual, expected)
```
jp3477, line 85: This test seems to be failing. I think there's an extra comma at the end of the first line of "expected".

jp3477 (changes requested):
Thanks for all these changes, they look good. I just pointed out one integration test that is failing.

jp3477 (approved):
Found the issue, @danyel-campbell. Nice work!

danyel-campbell:
Yay! Glad that it’s finally done 💚💚

PR #931, opened 2021-07-12
[DC-1615] domain alignment rule is dropping some PPI concepts

jp3477:
Excluded PPI concepts from being sandboxed during domain alignment.


ksdkalluri (approved):
I ran the integration test locally, it passes, and updates look good. Try re-running the integration test in CircleCI and merge once it finishes successfully.

PR #936, opened 2021-07-21
[DC-1774] changes to generalize code 1332843

lrwb-aou:
* generalize the answers for a question that is similar to a perviously generalized question
* treat COPE concept same as a Basics concept


mark-velez (approved):
I reviewed this mostly by looking very carefully in Athena that the basics and cope concept IDs were correctly associated. I also verified the output dataset had the same answer keys for both questions.

PR #937, opened 2021-07-21
[DC-1738] Comments in `clean_ppi_numeric_fields_using_parameters.py` SQL are backward

nishanthpp93:
* [DC-1738] Update jinja comments with correct tickets

[DC-1738]: https://precisionmedicineinitiative.atlassian.net/browse/DC-1738


ksdkalluri (approved):
Quick one. Looks Good!

PR #938, opened 2021-07-22
[DC-1777] Update `cdm_schemas()` logic to include `source_to_concept_map` in vocabulary tables

ksdkalluri:


data_steward/resources.py
```diff
@@ -250,6 +251,19 @@ def is_id_match(table_id):
     return table_id.startswith('identity_')
 
 
+def vocabulary_schemas():
+    vocabulare_schemas = dict()
```
nishanthpp93, line 255: I thought this was a typo, you can use `vocab_schemas`.

nishanthpp93 (commented):
I think this looks good but it depends on integration tests passing so nothing breaks. If you push the suggestion I made below, it should trigger the tests.

ksdkalluri:
@nishanthpp93 Integration tests passed. Is this good to merge?

data_steward/resources.py
```diff
@@ -250,6 +251,19 @@ def is_id_match(table_id):
...
+    vocab_schemas = dict()
+    for dir_path, _, files in os.walk(vocabulary_fields_path):
+        for f in files:
+            file_path = os.path.join(dir_path, f)
+            with open(file_path, 'r', encoding='utf-8') as file_path:
+                file_name = os.path.basename(f)
+                table_name = file_name.split('.')[0]
+                schema = json.load(file_path)
+                vocab_schemas[table_name] = schema
+    return vocab_schemas
```
nishanthpp93, line 264: This seems to be returning schemas but we only need table_names, right?

nishanthpp93 (changes requested):
One more change

data_steward/resources.py
ksdkalluri: We can return just names true, but it's not gonna make any difference in the loop as checking if the string exists in a list or a dictionary works the same as when we are checking in the dictionaries it will compare with the keys. If we need all the vocabulary schemas anywhere in the repo we can use this utility.

nishanthpp93: Yes, that's true. It seemed like overkill. If it can be used somewhere else, then we can use it. If it's not used anywhere, it may be unnecessary.

ksdkalluri: so should I change it to just return the names?

ksdkalluri: updated it.

nishanthpp93 (approved):
Thanks for updating @ksdkalluri. Looks good and works as expected. Once integration tests pass, this is good to go.

PR #939, opened 2021-07-22
[DC-1769] Script to remove developer datasets

lrwb-aou:
* add script to identify and remove developer datasets
* allow script runner to back out of deletion


data_steward/utils/bq.py
```diff
@@ -421,28 +421,6 @@ def update_labels_and_tags(dataset_id,
     return {**existing_labels_or_tags, **updates}
 
 
-def delete_dataset(project_id,
-                   dataset_id,
-                   delete_contents=True,
-                   not_found_ok=True):
```
nishanthpp93, line 427: Thanks, this is good to remove as well! We can use the client directly. I checked for usages too and did not find any. I noticed list_datasets is redundant as well, but a different ticket can be created to remove it along with usages.

nishanthpp93 (commented):
This looks good overall. I added a couple of suggestions, let me know your thoughts.

data_steward/tools/clean_project_datasets.py
```diff
@@ -8,68 +14,79 @@
...
     ]
-    datasets_with_substrings = get_datasets_with_substrings(
-        all_datasets, name_substrings)
-    logging.info(f'Datasets marked for deletion: {datasets_with_substrings}')
-    logging.info('Proceed?')
+
+    datasets_with_substrings = [
+        dataset for dataset in all_datasets for substring in name_substrings
+        if substring in dataset
+    ]
```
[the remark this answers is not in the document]
lrwb-aou: Thanks!

data_steward/tools/clean_project_datasets.py
```diff
@@ -8,68 +14,79 @@
...
     :param datasets_to_delete_list: list of dataset_ids to delete
+    :param offboarding_dev: boolean set to true to ensure if you are deleting 
+        datasets outside of test environment, you really intend to do that
     :return:
     """
+    # a fail safe to prevent deleting datasets in environments other than test.
+    if 'test' not in client.project and test_env:
+        raise RuntimeError(
+            "Attempting to delete datasets out of non-test environment "
+            "while not explicitly making this choice!!")
```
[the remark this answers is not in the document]
lrwb-aou: I considered that and actually prefer it.  I just moved the check close to where we are actually deleting datasets to prevent someone from importing just this function and skipping the check if it was in a previous function.  I'm fine with moving this forward in the script if you're agreeable to making this function "private" so someone would have to intentionally name mangle the import to get to it.

data_steward/utils/bq.py
lrwb-aou: 👍 I created two different tickets yesterday.  They are in Daniel's Technical Debt epic.  See tickets 1778 and 1779.

data_steward/tools/clean_project_datasets.py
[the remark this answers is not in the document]
nishanthpp93: Thanks, I hadn't considered that. Sounds good to me.

data_steward/tools/clean_project_datasets.py
```diff
@@ -98,9 +124,26 @@ def get_response():
...
+                        action='store_true',
+                        help='Send logs to console')
+
+    return parser.parse_args(raw_args)
+
+
+def main(raw_args=None):
+    args = get_arguments(raw_args)
+
+    pipeline_logging.configure(add_console_handler=args.console_log)
```
nishanthpp93, line 143: I'd suggest making this `add_console_handler=True`. The datasets to delete do not show up otherwise.

nishanthpp93 (commented):
Thanks for the update, the changes look good! Added one more minor update.

data_steward/tools/clean_project_datasets.py
lrwb-aou: @nishanthpp93 there's a couple of ways to go about this.  

1. We can hard code it (which I am strongly against).  I don't want to hard code it because then we have to make code changes when we start changing where this code is run.  It's fine while we are manually running the code, but it becomes a nuissance if we were to ever import this into some other piece of code and call it from there.  We won't want to log everything to the console at that point.
2. We can default the parser argument to True instead of False, so we have to explicitly turn off the logging.  Then if we do import this into another module and call this main, we can turn the console logs off without requiring a code change.
3. We get used to using the command line to send the logs to the console as well as the log file.  The code is set up for easier programmatic reuse.  The datasets to delete  are being logged in the log file, even if they lines aren't sent to the console.

I generally try to develop so that there is less to change when something becomes part of an automated process.  That's why I generally use option 3.  I develop for the programmatic usage and expect myself to remember the extra command line flags.

Thoughts?

nishanthpp93: @lrwb-aou Sure, I would prefer 3 as well, and follow that for other scripts. However, in this script, the question for the prompt does not show up. All I see is "Please press Y/n" - it does not inform about what I am consenting to, rather just demands consent - which can be dangerous. If just the logs for lines 79-85 which set up and pose the question are marked for printing to stdout regardless of the arg for console log, that works for me.

lrwb-aou: @nishanthpp93 How about option 2 then?  I just changed the default value to be `True` for this.  So someone must explicitly turn off console loggig.

nishanthpp93: @lrwb-aou Sure, I should've mentioned another option, which is to store the text for lines 79-85 in an f-string and passed as `get_response(question_text)` so that the `input` call prints all of the question text instead of just "Please press Y/n". Option 2 works, but can be counter-intuitive compared to the other scripts we have.

nishanthpp93: I just realized - option 2 does not handle invariance - we need the lines 79-85 to be printed regardless of the arg for console log. This can be handled by passing the text to the `input` call like I mentioned above. I'm open to other ways to achieve this as well.

lrwb-aou: @nishanthpp93 I made the suggested changes.  Let me know if you see anymore "gotchas".

data_steward/tools/clean_project_datasets.py
```diff
@@ -58,31 +51,59 @@ def run_deletion(project_id, name_substrings):
...
+    LOGGER.info(f'{len(datasets_with_substrings)} Datasets marked for '
+                f'deletion in project `{project_id}`: ')
+    for dataset in datasets_with_substrings:
+        LOGGER.info(f'\t{dataset}')
+
+    msg = (f'After reviewing datasets, proceed?\nThis action '
+           f'cannot be reversed.\n'
+           f'[Y/y/N/n]:  ')
+    LOGGER.info(msg)
+    response = input(msg)
```
nishanthpp93, line 91: This doesn't show the datasets to review, so I expanded the `msg`. However, this is not a required change since there is a warning for setting console log in the beginning.
```suggestion
    msg = (f'{len(datasets_with_substrings)} Datasets marked for '
           f'deletion in project `{project_id}`: \n' +
           '\n'.join(f'\t{dataset}' for dataset in datasets_with_substrings) +
           f'\nAfter reviewing datasets, proceed?\nThis action '
           f'cannot be reversed.\n'
           f'[Y/y/N/n]:  ')
    response = input(msg)
```

data_steward/tools/clean_project_datasets.py
```diff
@@ -58,31 +51,59 @@ def run_deletion(project_id, name_substrings):
     :param name_substrings: Identifies substrings that help identify datasets to delete
     :return:
     """
+    # make the developer running this script approve the environment.
+    msg = (f'This will remove datasets from the `{project_id}` '
+           f'environment.\nAre you sure you with to proceed?  '
```
nishanthpp93, line 56: nit: typo for `wish`

nishanthpp93 (approved):
Thanks for making the changes @lrwb-aou. I made some final comments, but they are not required for merging and this is good to merge as-is.

data_steward/tools/clean_project_datasets.py
lrwb-aou: I'm not making this change.  We're already repeating the messages via logs and input messages.  It feels like we are trying to force this to be an always command line tool that needs to have someone watching it.

PR #941, opened 2021-07-26
[DC-1776] Add checks for missing questionnaire_response_id in survey

rfrancis1:


data_steward/analytics/cdr_ops/rdr_export_qc.py
```diff
@@ -467,3 +467,19 @@
...
+
+# Survey data in the RDR export should all have **questionnaire_response_id** [DC-1776](https://precisionmedicineinitiative.atlassian.net/browse/DC-1776). Any violations should be reported to the RDR team.
+
+query = f"""
+SELECT 
+    observation_id, 
+    person_id, 
+    observation_concept_id
+FROM `{project_id}.{new_rdr}.observation`
+WHERE questionnaire_response_id IS NULL
```
nishanthpp93, line 481: @rfrancis1 My understanding is the `questionnaire_response_id` not being null only applies to surveys, but can be null for PM data. So this may be identifying PM data along with survey rows with `questionnaire_response_id=NULL`. Is that correct?

nishanthpp93 (commented):
Thanks for adding the check, @rfrancis1. I'm not completely familiar with the PM data but added a comment. Let me know your thoughts.

data_steward/analytics/cdr_ops/rdr_export_qc.py
rfrancis1: Thank you @nishanthpp93. I believe that PM data would go to measurement and not in observation.

nishanthpp93: Thanks, right. I shouldn't have said PM data. It is data related to PM, or more specifically, why PM data was not measured. I believe this is separate from survey data, but added to observation at the time of taking PM so does not need to have a questionnaire_response_id. I used a `SELECT DISTINCT value_source_value` for this. However, let me know if this information is recorded via a survey instead.

rfrancis1: Oh I see what you mean. thanks, Nishanth.

rfrancis1: Nishanth, it's changed to be restricted to PPI 1-3. @lrwb-aou I am only checking PPI 1-3, but let me know if we need to include PPI 4-6

nishanthpp93: Thanks, this is great. Interestingly, I see a lot of data without questionnaire_reponse_id in the cleaned RDR datasets (used for the CDR), but nothing in the monthly RDR export datasets.

rfrancis1: Yes, if we (curation) touched the cleaned RDR, then it's because of the pmi_backfill. I think the export should all have questionnaire_response_id, so this shouldn't return anything, but I guess it's a nice-to-have check.

nishanthpp93 (approved):

PR #946, opened 2021-07-29
[DC-1772] adding cdr metadata

lrwb-aou:
* add rdr_export_date to cdr_metadata table.
* field was already required by the bash script.  it just wasn't used.
* [ci skip] as this is a change to a bash script and no tests exist


jp3477 (approved):
Looks good. Confirmed field was set in BQ.

PR #953, opened 2021-08-10
[DC-1798] Measurement Record Suppression rule in clean data set is not cleaning as expected

nishanthpp93:
* [DC-1798] Use UPDATE statement
* [DC-1798] Fix tests


jp3477 (approved):
Looks good, and tests pass!

PR #955, opened 2021-08-12
[DC-1787] Create cleaning rule to remove records for pids not in CDR

nishanthpp93:
* [DC-1787] Create CR to remove data for PIDs not in CDR
* [DC-1787] Update CR and add test

[DC-1787]: https://precisionmedicineinitiative.atlassian.net/browse/DC-1787
[DC-1787]: https://precisionmedicineinitiative.atlassian.net/browse/DC-1787

### DIFF
--- a/data_steward/cdr_cleaner/cleaning_rules/remove_non_existing_pids.py
+++ b/data_steward/cdr_cleaner/cleaning_rules/remove_non_existing_pids.py
@@ -1,0 +1,119 @@
+"""
+Dataset (currently limited to fitbit dataset) should only contain pids that exist in the corresponding CDR
+"""
+import logging
+
+# Project imports
+from common import JINJA_ENV, FITBIT_TABLES
+from constants.cdr_cleaner import clean_cdr as cdr_consts
+from cdr_cleaner.cleaning_rules.base_cleaning_rule import BaseCleaningRule, query_spec_list
+from cdr_cleaner.clean_cdr_utils import get_tables_in_dataset
+
+LOGGER = logging.getLogger(__name__)
+
+JIRA_ISSUE_NUMBERS = ['DC1787']
+
+SANDBOX_NON_EXISTING_PIDS = JINJA_ENV.from_string("""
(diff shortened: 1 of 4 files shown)


data_steward/cdr_cleaner/cleaning_rules/remove_non_existing_pids.py
```diff
@@ -0,0 +1,138 @@
...
+WHERE person_id NOT IN
+(SELECT person_id
+FROM `{{project_id}}.{{reference_dataset_id}}.person`)
+""")
+
+
+class RemoveNonExistingPids(BaseCleaningRule):
+
+    def __init__(self, project_id, dataset_id, sandbox_dataset_id,
+                 reference_dataset_id):
```
jp3477, line 37: How would clean_cdr know what reference_dataset_id to pass? Would it be worth making `reference_dataset_id` a keyword param?

jp3477 (commented):
Mostly looks good, just left a question.

data_steward/cdr_cleaner/cleaning_rules/remove_non_existing_pids.py
nishanthpp93: Good question, that part is handled by the clean engine [here](https://github.com/all-of-us/curation/blob/242b53e1467baad726a4ff6a1c851f67c17e8be6/data_steward/cdr_cleaner/clean_cdr_engine.py#L169). Do you mean adding a default value? It ideally shouldn't need it, and you can see another example in [survey_version_info.py](https://github.com/all-of-us/curation/blob/develop/data_steward/cdr_cleaner/manual_cleaning_rules/survey_version_info.py). Not sure if that answers your question, let me know.

jp3477: Ah I see, that makes sense. Thanks for sharing.

jp3477 (approved):
Nice job!

PR #957, opened 2021-08-13
[DC-844] [DC-1195] update family history cleaning rule to base class

lrwb-aou:
* update the cleaning rule to a base class rule
* remove unnecessary unit test
* add integration test
* adding validation checks to the cleaning rule


data_steward/cdr_cleaner/cleaning_rules/update_family_history_qa_codes.py
```diff
@@ -24,47 +37,172 @@
...
+    def setup_validation(self, client):
+        """
+        Run required steps for validation setup
+        """
+        self.init_counts = self._get_counts(client)
+
+        if self.init_counts.get('total_rows') == 0:
+            raise RuntimeError('NO DATA EXISTS IN OBSERVATION TABLE')
+
+    def validate_rule(self, client):
```
ksdkalluri, line 127: Adding Validation is a good idea!!

ksdkalluri (approved):
Ran the integration tests locally, worked fine. Thanks for adding the validation in the CR.

data_steward/cdr_cleaner/cleaning_rules/update_family_history_qa_codes.py
lrwb-aou: 👍

PR #960, opened 2021-08-19
[DC-1811] circleci image update

lrwb-aou:
* update circleci docker image tag.  previous image was deprecated
    * attempting to get google sdk to install without using repos
    * forced to install google-cloud-sdk without apt, as it isn't in the new image yet
    * source the gcloud tools before use
    * changing " to ' for setting path variable


lrwb-aou:
closing this PR.  some of this will be handled by DC-1820.

PR #962, opened 2021-08-24
[DC-1724] Cleaning rule to remove participants who appear under 18

ksdkalluri:


tests/integration_tests/data_steward/cdr_cleaner/cleaning_rules/remove_participants_under_18years.py
```diff
@@ -0,0 +1,150 @@
+"""
+Integration test for remove_participants_under_18years module
```
jp3477, line 2: This file should be renamed `remove_participants_under_18years_test.py ` to properly run the integration test

tests/integration_tests/data_steward/cdr_cleaner/cleaning_rules/remove_participants_under_18years.py
```diff
@@ -0,0 +1,150 @@
...
+    BaseTest
+
+PERSON_DATA_TEMPLATE = JINJA_ENV.from_string("""
+INSERT INTO `{{project_id}}.{{dataset_id}}.person`
+(person_id, birth_datetime, gender_concept_id, year_of_birth, race_concept_id, ethnicity_concept_id)
+VALUES
+      (1, '1970-01-01 00:00:00 EST', 0, 1970, 0, 0),
+      (2, '2002-01-01 00:00:00 EST', 0, 2002, 0, 0),
+      (3, '2003-01-01 00:00:00 EST', 0, 2003, 0, 0),
+      (4, '2015-01-01 00:00:00 EST', 0, 2015, 0, 0)
```
jp3477, line 28: NIT: Some poor engineer 12 years from now will be confused why their tests are failing 😁

jp3477 (changes requested):
The integration test file needs to be renamed as mentioned below. But other than that, everything looks good and the tests pass.

jp3477 (approved):
Thanks for making that change. And I had missed the clean_cdr, nice catch.

tests/integration_tests/data_steward/cdr_cleaner/cleaning_rules/remove_participants_under_18years.py
ksdkalluri: @jp3477 Now that poor engineer can avoid it for 5 more years 😉. TBH I couldn't think of another approach to avoid this.

PR #963, opened 2021-08-26
[DC-1801] Removes cdm.csv from resource_files and all references to it

ksdkalluri:


nishanthpp93 (approved):
The changes look good. Could not find any reference to cdm.csv and the tests pass. I think I have a different issue with my tests that cause `test_generate_metrics` and `test_process_hpo_ignore_dirs` where `extract_date_from_rdr_dataset_id` is not mocked, but if those pass for you this is good to go.

ksdkalluri:
@nishanthpp93 These tests were passing locally for me. I am merging this PR.

PR #964, opened 2021-08-30
[DC-1812] create a new notebook to compare CT_vs_RT in CDR

jqian2015:


data_steward/analytics/cross_ct_rt/CDR_QC_RT_vs_CT.py
```diff
@@ -0,0 +1,497 @@
...
+LOOP
+  SET i = i + 1;
+  IF i > ARRAY_LENGTH(person_tables) THEN
+    LEAVE;
+  END IF;
+  EXECUTE IMMEDIATE '''
+INSERT ct_person_id
+SELECT distinct person_id
+FROM `{project_id}.{ct_dataset}.''' || person_tables[ORDINAL(i)] || '''`
+''';
```
nishanthpp93, line 396: This looks like it is appending person_ids from every table into a single table. There can be a lot of repeated person_ids from different tables since the distinct is limited to the current table. I believe these can be simplified to not use a loop, but we can discuss more if you have time tomorrow. Let me know your thoughts.

nishanthpp93 (commented):
I think the queries run fine, but added some comments. Let me know if you have questions.

data_steward/analytics/cross_ct_rt/readme.md
```diff
@@ -0,0 +1,13 @@
+
+# how to run these notebooks:
+
+python report_runner.py "CDR_QC_RT_vs_CT.py" -p project_id "project_id" -p rt_dataset "rt_deid" -p ct_dataset "ct_deid" -p maximum_age "89" --output_path CDR_QC_RT_vs_CT.html
```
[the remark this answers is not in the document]
jqian2015: yes, will do.

data_steward/analytics/cross_ct_rt/CDR_QC_RT_vs_CT.py
```diff
@@ -0,0 +1,497 @@
...
+CASE WHEN
+  countif((c.questionnaire_response_id IS NOT NULL AND r.questionnaire_response_id IS NOT NULL) AND c.questionnaire_response_id != r.questionnaire_response_id) > 0 
+  THEN 'FAILURE' ELSE 'ok'
+END
+ as success
+FROM `{project_id}.{ct_dataset}.observation` c
+FULL OUTER JOIN `{project_id}.{rt_dataset}.observation` r
+USING (observation_id)
+'''
+df1=pd.read_gbq(query, dialect='standard')
```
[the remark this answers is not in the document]
jqian2015: yes, i will try this.

data_steward/analytics/cross_ct_rt/CDR_QC_RT_vs_CT.py
jqian2015: @lrwb-aou @nishanthpp93  Lauren will look into this. thanks.

lrwb-aou: @nishanthpp93 What do you suggest?  This was done in a hurry and could possibly be optimized.  I will document my initial decisions for these queries here so we can discuss.
1. I didn't want to hard code all the tables with `person_id` as a column and wanted to read this from the available tables in the RT dataset since that is the target dataset for this verification.
2. I wanted to get all person_ids without reading them from the `person` table because an error affecting the `person` table may hide an error with the RT/CT cross dataset verification.
3. I did try to limit to reading the distinct `person_id`s for each OMOP table to make this a less expensive process.  I know there will be some duplicates due the person_id existing in multiple tables.  I felt this was ok because it was, at most, #_of_participants * #_of_tables_with_personid_column.  These tables are also not being stored, but are temporary.
4. I think line 403 should be changed to `WHERE person_id not in (select distinct person_id from ct_person_id);`

nishanthpp93: @lrwb-aou thanks for the context. I think 4 should definitely help. Do you think line 396 can have the same where clause while appending to the table?

lrwb-aou: @nishanthpp93 I think that could work.  It could be done on 396 and 381.  But we would want to check to make sure it is not already in the temporary table instead of the other tier table.

data_steward/analytics/cross_ct_rt/CDR_QC_RT_vs_CT.py
```diff
@@ -0,0 +1,497 @@
...
+INSERT ct_person_id
+SELECT distinct person_id
+FROM `{project_id}.{ct_dataset}.''' || person_tables[ORDINAL(i)] || '''`
+''';
+
+END LOOP;
+
+SELECT distinct person_id,
+(2021 - year_of_birth) as age,
+CASE WHEN (2021 - year_of_birth) < {maximum_age} THEN 1 ELSE 0 END as Failure
```
nishanthpp93, line 469: One more change: Can this extract the year from current date/datetime?

lrwb-aou: @nishanthpp93 good catch!   It should use the current year and programmatically extract it.

jqian2015: will do. thanks.

data_steward/analytics/cross_ct_rt/CDR_QC_RT_vs_CT.py
jqian2015: @nishanthpp93 @lrwb-aou I am not sure what exactly we have to do about line 396 and 381, it will be helpful if this can be finalized. thanks.

lrwb-aou: @jqian2015 This would be adding
```
WHERE person_id not in (select distinct person_id from rt_person_id)
``` 
to the `INSERT rt_person_id` SQL and
```
WHERE person_id not in (select distinct person_id from ct_person_id)
``` 
to the `INSERT ct_person_id` SQL.  The statement should be added as an additional line before the line containing `''';`

jqian2015: will do. thanks.

data_steward/analytics/cross_ct_rt/CDR_QC_RT_vs_CT.py
nishanthpp93, line 469:
```suggestion
(EXTRACT (YEAR FROM CURRENT_DATE()) - year_of_birth) as age,
CASE WHEN (EXTRACT (YEAR FROM CURRENT_DATE()) - year_of_birth) < {maximum_age} THEN 1 ELSE 0 END as Failure
```

data_steward/analytics/cross_ct_rt/readme.md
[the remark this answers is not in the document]
nishanthpp93: Let me know if you ran into any issues while moving this.

nishanthpp93 (commented):
Added more specific changes for this, since I might've missed the updates. Let me know if you have any questions.

nishanthpp93 (approved):
Looks good! Thanks for making the changes @jqian2015!

PR #965, opened 2021-09-02
[DC-1735] Update Case statement logic to respect data type for required fields in `domain_alignment` cleaning rule

ksdkalluri:


jp3477 (approved):
Nice, this works well!

PR #970, opened 2021-09-13
[DC-1235] Add structure for queries and DRC Email Comparisons

jp3477:
This is the first of several tickets for participant validation. It also includes some modifications to the `create_update_drc_id_match_table.py` script.

### DIFF
--- a/data_steward/participant_validation/validate_email.py
+++ b/data_steward/participant_validation/validate_email.py
@@ -1,0 +1,119 @@
+# Python imports
+import logging
+import argparse
+
+# Project imports
+from utils import bq, pipeline_logging, auth
+from tools.create_tier import SCOPES
+from common import JINJA_ENV, PS_API_VALUES, DRC_OPS
+
+LOGGER = logging.getLogger(__name__)
+
+EHR_OPS = 'ehr_ops'
+
+CREATE_EMAIL_COMPARISON_FUNCTION = JINJA_ENV.from_string("""
+    CREATE FUNCTION IF NOT EXISTS `{{project_id}}.{{drc_dataset_id}}.CompareEmail`(rdr_email string, ehr_email string)
+    RETURNS string
+    AS ((
+        WITH normalized_rdr_email AS (
+            SELECT LOWER(TRIM(rdr_email)) AS rdr_email
+        )
+        , normalized_ehr_email AS (
+            SELECT LOWER(TRIM(ehr_email)) AS ehr_email
+        )
+        SELECT
+            CASE 
+                WHEN normalized_rdr_email.rdr_email = normalized_ehr_email.ehr_email
+                    AND REGEXP_CONTAINS(normalized_rdr_email.rdr_email, '@') THEN 'match'
+                WHEN normalized_rdr_email.rdr_email IS NOT NULL AND normalized_ehr_email.ehr_email IS NOT NULL THEN 'no_match'
+                WHEN normalized_rdr_email.rdr_email IS NULL THEN 'missing_rdr'
+                ELSE 'missing_ehr'
+            END AS email
+        FROM normalized_rdr_email, normalized_ehr_email 
+
+    ));
+""")
+
+MATCH_EMAIL_QUERY = JINJA_ENV.from_string("""
+    UPDATE `{{project_id}}.{{drc_dataset_id}}.{{id_match_table_id}}` upd
+    SET upd.email = `{{project_id}}.{{drc_dataset_id}}.CompareEmail`(ps.email, ehr.email)
+    FROM `{{project_id}}.{{drc_dataset_id}}.{{id_match_table_id}}` id_match
+    LEFT JOIN `{{project_id}}.{{drc_dataset_id}}.{{ps_api_table_id}}` ps
+        ON ps.person_id = id_match.person_id
+    LEFT JOIN `{{project_id}}.{{ehr_ops_dataset_id}}.{{hpo_pii_email_table_id}}` ehr
+        ON ehr.person_id = id_match.person_id
+    WHERE upd.person_id = id_match.person_id
+""")
+
+
+def identify_rdr_ehr_email_match(client, project_id, hpo_id):
+
+    id_match_table_id = f'drc_identity_match_{hpo_id}'
+    hpo_pii_email_table_id = f'{hpo_id}_pii_email'
+    ps_api_table_id = f'{PS_API_VALUES}_{hpo_id}'
+
+    create_email_query = CREATE_EMAIL_COMPARISON_FUNCTION.render(
+        project_id=project_id, drc_dataset_id=DRC_OPS)
+
+    match_email_query = MATCH_EMAIL_QUERY.render(
+        project_id=project_id,
+        id_match_table_id=id_match_table_id,
+        hpo_pii_email_table_id=hpo_pii_email_table_id,
+        ps_api_table_id=ps_api_table_id,
+        drc_dataset_id=DRC_OPS,
+        ehr_ops_dataset_id=EHR_OPS)
+
+    LOGGER.info("Creating `CompareEmail` function if doesn't exist.")
+    job = client.query(create_email_query)
+    job.result()
+
+    LOGGER.info(f"Matching email field for {hpo_id}.")
+    job = client.query(match_email_query)
+    job.result()
+
+
+def get_arg_parser():
+    parser = argparse.ArgumentParser(
+        description=""" Create and update DRC match table for hpo sites.""")
+    parser.add_argument('-p',
+                        '--project_id',
+                        action='store',
+                        dest='project_id',
+                        help='Project associated with the input datasets',
+                        required=True)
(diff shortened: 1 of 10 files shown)


data_steward/participant_validation/validate_email.py
```diff
@@ -0,0 +1,121 @@
...
+MATCH_EMAIL_QUERY = JINJA_ENV.from_string("""
+    UPDATE `{{project_id}}.{{drc_dataset_id}}.{{id_match_table_id}}` upd
+    SET upd.email = `{{project_id}}.{{drc_dataset_id}}.CompareEmail`(ps.email, ehr.email),
+        upd.algorithm = 'yes'
+    FROM `{{project_id}}.{{drc_dataset_id}}.{{id_match_table_id}}` id_match
+    LEFT JOIN `{{project_id}}.{{drc_dataset_id}}.{{ps_api_table_id}}` ps
+        ON ps.person_id = id_match.person_id
+    LEFT JOIN `{{project_id}}.{{ehr_ops_dataset_id}}.{{hpo_pii_email_table_id}}` ehr
+        ON ehr.person_id = id_match.person_id
+    WHERE upd.person_id = id_match.person_id
```
nishanthpp93, line 46: I think this might need to be a [MERGE](https://cloud.google.com/bigquery/docs/reference/standard-sql/dml-syntax#merge_statement) since new participants can be added by ps or the site. I'm trying to test out some things to see how it works for this instance.

nishanthpp93 (commented):
Thanks for adding the integration test @jp3477. I  added a comment but maybe I'm not understanding the process correctly.

data_steward/validation/participants/validate.py
```diff
@@ -0,0 +1,126 @@
...
+    LEFT JOIN `{{project_id}}.{{ehr_ops_dataset_id}}.{{hpo_pii_email_table_id}}` ehr
+        ON ehr.person_id = ps.person_id
+    WHERE upd.person_id = ps.person_id
+        AND upd._PARTITIONTIME = ps._PARTITIONTIME
+""")
+
+
+def identify_rdr_ehr_match(client, project_id, hpo_id, ehr_ops_dataset_id):
+
+    id_match_table_id = f'drc_identity_match_{hpo_id}'
```
nishanthpp93, line 49: This might need to be just `identity_match_{hpo_id}`, we have schema for it. Is this a different table from that?

tests/integration_tests/data_steward/validation/participants/validate_test.py
```diff
@@ -0,0 +1,173 @@
...
+
+    @classmethod
+    def setUpClass(cls):
+        print('**************************************************************')
+        print(cls.__name__)
+        print('**************************************************************')
+
+    def setUp(self):
+        self.project_id = os.environ.get(PROJECT_ID)
+        self.dataset_id = 'drc_ops'
```
nishanthpp93, line 60: This might need to be a new env var 'DRC_DATASET_ID'

nishanthpp93 (approved):
This looks great! Thanks for this @jp3477. I like the tests and the infrastructure script. Anyone who adds a new function would find it much much easier to work on!

tests/integration_tests/data_steward/validation/participants/validate_test.py
nishanthpp93: Actually, I think this is fine as is.

data_steward/validation/participants/validate.py
```diff
@@ -0,0 +1,126 @@
...
+
+MATCH_FIELDS_QUERY = JINJA_ENV.from_string("""
+    UPDATE `{{project_id}}.{{drc_dataset_id}}.{{id_match_table_id}}` upd
+    SET upd.email = `{{project_id}}.{{drc_dataset_id}}.CompareEmail`(ps.email, ehr.email),
+        upd.algorithm = 'yes'
+    FROM `{{project_id}}.{{drc_dataset_id}}.{{ps_api_table_id}}` ps
+    LEFT JOIN `{{project_id}}.{{ehr_ops_dataset_id}}.{{hpo_pii_email_table_id}}` ehr
+        ON ehr.person_id = ps.person_id
+    WHERE upd.person_id = ps.person_id
+        AND upd._PARTITIONTIME = ps._PARTITIONTIME
```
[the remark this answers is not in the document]
jp3477: Yeah, theoretically they shouldn't if they're being triggered in GAE.

data_steward/validation/participants/validate.py
jp3477: I think they reference the same thing, but the ticket requests 'drc_identity_match'. I guess it's a rebranding?

data_steward/tools/create_update_drc_id_match_table.py
```diff
@@ -34,12 +50,12 @@
 POPULATE_VALIDATION_TABLE = JINJA_ENV.from_string("""
 INSERT INTO `{{project_id}}.{{drc_dataset_id}}.{{id_match_table_id}}` (_PARTITIONTIME, {{fields}}) 
 SELECT TIMESTAMP_TRUNC(CURRENT_TIMESTAMP, HOUR), person_id,
```
nishanthpp93, line 52: I think this might need to be changed to something like
```suggestion
SELECT 
    CASE WHEN ABS(TIMESTAMP_TRUNC(CURRENT_TIMESTAMP, HOUR) - _PARTITIONTIME) < 2 THEN _PARTITIONTIME 
    ELSE TIMESTAMP_TRUNC(CURRENT_TIMESTAMP, HOUR) AS _PARTITIONTIME,
    person_id, 
```
My main concern is that the partition for PS will be created with a certain hour partition. But loading it with PS API data will take time due to API call limits. The id_match partition will only be created after the ps partition is fully populated. Is that correct? If so, the partition hour may be different, which can lead to problems while joining on the partition time. I think there may be more elegant ways to keep them in sync, so open to ideas. Let me know your thoughts.

nishanthpp93 (commented):
Added one last comment.

data_steward/validation/participants/validate.py
nishanthpp93: Thanks for updating this!

data_steward/tools/create_update_drc_id_match_table.py
jp3477: Interesting, that could be a problem. Yes, id_match is only created after ps_api load is done. We might be able to set the intended partition at the beginning of the python process (generate_metrics?) where they are both called, and then pass that partition explicitly into the SQL.

nishanthpp93 (commented):
Looks like BQ copies the whole table and not a single partition, so we need another filter. I suggested a small change but open to ideas. Let me know your thoughts. Also, try to rebase onto develop and force push. I think no other integration tests should break, but it's good to update.

tests/integration_tests/data_steward/tools/store_participant_summary_results_test.py
```diff
@@ -0,0 +1,114 @@
...
+class StoreParticipantSummaryResultsTest(TestCase):
+
+    @classmethod
+    def setUpClass(cls):
+        print('**************************************************************')
+        print(cls.__name__)
+        print('**************************************************************')
+
+        cls.project_id = os.environ.get(PROJECT_ID)
+        # cls.dataset_id = os.environ.get('COMBINED_DATASET_ID')
```
nishanthpp93, line 43: nit: this can be removed

tests/integration_tests/data_steward/validation/participants/validate_test.py
```diff
@@ -0,0 +1,177 @@
...
+        ps_values_fields = [
+            dict(name='person_id', type='integer', mode='nullable'),
+            dict(name='email', type='string', mode='nullable'),
+            dict(name='algorithm', type='string', mode='nullable')
+        ]
+
+        bq_utils.create_table(self.ps_values_table_id,
+                              ps_values_fields,
+                              drop_existing=True,
+                              dataset_id=self.dataset_id)
```
nishanthpp93, line 81: nit: Standard table schemas can sometimes prevent differences between test and prod behavior.
The following needs a couple of imports - fields_for from resources and Table from bigquery.
```suggestion
        schema = resources.fields_for(f'{PS_API_VALUES}')
        table = Table(f'{self.project_id}.{self.dataset_id}.{self.ps_values_table_id}', schema=schema)
        table = self.client.create_table(table)
```

nishanthpp93 (approved):
Looks great! Thanks for updating, @jp3477! Added a couple nits, but they are not required for merge in since it works as is.

tests/integration_tests/data_steward/validation/participants/validate_test.py
jp3477: Just pushed some changes to address these. In the process, I modified create_table statements to create partitions during testing.

nishanthpp93: Amazing, that mimics prod way better, thanks for doing this!

PR #973, opened 2021-09-14
[DC-1839] yapf linting failure fixes

lrwb-aou:
* yapf run on only one file that is causing lint failures


lrwb-aou:
Merging the linting fix.  A new PR will be created for the integration test fix.

PR #975, opened 2021-09-21
[DC-1860] Update `rdr_export_qc` notebook to use `notebook_utils.execute` instead of `pandas_gbq`

ksdkalluri:


nishanthpp93 (approved):
Looks good!

PR #976, opened 2021-09-21
[DC-1855] Provide pids associated with sites dropped in 2021q2r1 and 2021q3r2

nishanthpp93:
* [DC-1855] Create script to load submission from archive

[DC-1855]: https://precisionmedicineinitiative.atlassian.net/browse/DC-1855

### DIFF
--- a/data_steward/tools/load_archived_submission.py
+++ b/data_steward/tools/load_archived_submission.py
@@ -158,7 +158,8 @@ def main(project_id, dataset_id, bucket_name, hpo_id, folder_name):
     prefix = f'{hpo_id}/{site_bucket}/{folder_name}'
     LOGGER.info(
         f'Starting jobs for loading {bucket_name}/{prefix} into {dataset_id}')
-    load_folder(dataset_id, bq_client, bucket_name, prefix, gcs_client, hpo_id)
+    _ = load_folder(dataset_id, bq_client, bucket_name, prefix, gcs_client,
+                    hpo_id)
     LOGGER.info(f'Successfully loaded {bucket_name}/{prefix} into {dataset_id}')
 
 
--- a/data_steward/tools/load_archived_submission.py
+++ b/data_steward/tools/load_archived_submission.py
@@ -4,7 +4,7 @@ Loads a submission into bq from the archive
 import logging
 import argparse
 from typing import List
-from concurrent.futures import TimeoutError
+from concurrent.futures import TimeoutError as TOError
 
 from google.cloud.bigquery import LoadJobConfig, LoadJob, Table, Client as BQClient
 from google.cloud.storage import Client as GCSClient
@@ -47,7 +47,7 @@ def get_bucket(client: BQClient, hpo_id: str) -> str:
                 f'Found more than one bucket name for site {hpo_id}: {bucket_names}'
             )
         bucket_name = bucket_names[0]
-    except (GoogleCloudError, TimeoutError) as e:
+    except (GoogleCloudError, TOError) as e:
         LOGGER.error(f'Job failed with error {str(e)}')
         raise e
     return bucket_name
@@ -95,10 +95,11 @@ def load_folder(dst_dataset: str, bq_client: BQClient, bucket_name: str,
         job_config.skip_leading_rows = 1
         job_config.source_format = 'CSV'
         source_uri = f'gs://{bucket_name}/{blob.name}'
-        load_job = bq_client.load_table_from_uri(source_uri,
-                                                 destination,
-                                                 job_config=job_config,
-                                                 job_id_prefix=__name__)
+        load_job = bq_client.load_table_from_uri(
+            source_uri,
+            destination,
+            job_config=job_config,
+            job_id_prefix=f"{__file__.split('/')[-1].split('.')[0]}_")
         LOGGER.info(f'table:{destination} job_id:{load_job.job_id}')
         load_jobs.append(load_job)
         load_job.result()
--- a/data_steward/tools/load_submission.py
+++ b/data_steward/tools/load_submission.py
@@ -4,11 +4,11 @@ Loads a submission into bq from the archive
 import logging
 import argparse
 from typing import List
+from concurrent.futures import TimeoutError
 
 from google.cloud.bigquery import LoadJobConfig, LoadJob, Table, Client as BQClient
 from google.cloud.storage import Client as GCSClient
 from google.cloud.exceptions import GoogleCloudError
-from concurrent.futures import TimeoutError
 
 from common import AOU_REQUIRED, JINJA_ENV
 from utils.bq import get_table_schema, get_client
@@ -63,9 +63,9 @@ def _filename_to_table_name(filename: str) -> str:
     return filename.split('/')[-1].replace('.csv', '').lower()
 
 
-def load_stage(dst_dataset: str, bq_client: BQClient, bucket_name: str,
-               prefix: str, gcs_client: GCSClient,
-               hpo_id: str) -> List[LoadJob]:
+def load_folder(dst_dataset: str, bq_client: BQClient, bucket_name: str,
+                prefix: str, gcs_client: GCSClient,
+                hpo_id: str) -> List[LoadJob]:
     """
     Stage files from a bucket to a dataset
 
@@ -94,7 +94,6 @@ def load_stage(dst_dataset: str, bq_client: BQClient, bucket_name: str,
         job_config.schema = schema
         job_config.skip_leading_rows = 1
         job_config.source_format = 'CSV'
-        job_config
         source_uri = f'gs://{bucket_name}/{blob.name}'
         load_job = bq_client.load_table_from_uri(source_uri,
                                                  destination,
@@ -159,7 +158,7 @@ def main(project_id, dataset_id, bucket_name, hpo_id, folder_name):
     prefix = f'{hpo_id}/{site_bucket}/{folder_name}'
     LOGGER.info(
         f'Starting jobs for loading {bucket_name}/{prefix} into {dataset_id}')
-    load_stage(dataset_id, bq_client, bucket_name, prefix, gcs_client, hpo_id)
+    load_folder(dataset_id, bq_client, bucket_name, prefix, gcs_client, hpo_id)
     LOGGER.info(f'Successfully loaded {bucket_name}/{prefix} into {dataset_id}')
 
 
--- a/data_steward/tools/load_submission.py
+++ b/data_steward/tools/load_submission.py
@@ -89,15 +89,17 @@ def load_stage(dst_dataset: str, bq_client: BQClient, bucket_name: str,
         hpo_table_name = f'{hpo_id}_{table_name}'
         fq_hpo_table = f'{bq_client.project}.{dst_dataset}.{hpo_table_name}'
         destination = Table(fq_hpo_table, schema=schema)
-        destination = bq_client.create_table(destination, exists_ok=True)
+        destination = bq_client.create_table(destination)
         job_config = LoadJobConfig()
         job_config.schema = schema
         job_config.skip_leading_rows = 1
         job_config.source_format = 'CSV'
+        job_config
         source_uri = f'gs://{bucket_name}/{blob.name}'
         load_job = bq_client.load_table_from_uri(source_uri,
                                                  destination,
-                                                 job_config=job_config)
+                                                 job_config=job_config,
+                                                 job_id_prefix=__name__)
         LOGGER.info(f'table:{destination} job_id:{load_job.job_id}')
         load_jobs.append(load_job)
         load_job.result()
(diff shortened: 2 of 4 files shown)


data_steward/tools/load_submission.py
```diff
@@ -0,0 +1,124 @@
...
+Loads a submission into bq from the archive
+"""
+import logging
+import argparse
+
+from google.cloud.bigquery import LoadJobConfig, Table
+from google.cloud.storage import Client
+
+from common import AOU_REQUIRED, JINJA_ENV
+from utils import pipeline_logging
```
lrwb-aou, line 11: :nit:  I would probably import this under `if __name__=='__main__':` so that it is only imported if run as the main script.

data_steward/tools/load_submission.py
```diff
@@ -0,0 +1,124 @@
...
+        bucket_names_table=bq_consts.HPO_ID_BUCKET_NAME_TABLE_ID,
+        hpo_id=hpo_id.upper())
+
+    bucket_name_result = client.query(bucket_name_query).result()
+    bucket_name = bucket_name_result.to_dataframe()['bucket_name'].to_list()[0]
+    return bucket_name
+
+
+def _filename_to_table_name(filename: str) -> str:
+    return filename.split('/')[-1].replace('.csv', '').lower()
```
lrwb-aou, line 37: :nit:  I understand what is happening here.  A doc string with an example input and output would be useful for new team members.

data_steward/tools/load_submission.py
```diff
@@ -0,0 +1,124 @@
...
+
+
+def load_stage(dst_dataset, bq_client, bucket_name, prefix, gcs_client, hpo_id):
```
lrwb-aou, line 40: You added IDE hints in the `_filename_to_table_name` signature.  They would be nice to have here too, for consistency.

data_steward/tools/load_submission.py
```diff
@@ -0,0 +1,124 @@
...
+    :param gcs_client: a Cloud Storage client object
+    :return: list of completed load jobs
+    """
+    blobs = list(gcs_client.list_blobs(bucket_name, prefix=prefix))
+
+    load_jobs = []
+    for blob in blobs:
+        table_name = _filename_to_table_name(blob.name)
+        if table_name not in AOU_REQUIRED:
+            continue
```
lrwb-aou, line 56: I would prefer to log what we are skipping before continuing.  It could be a `debug` level message.  But it would be nice for this continuation to not be completely silent.  Thoughts?

data_steward/tools/load_submission.py
```diff
@@ -0,0 +1,124 @@
...
+        schema = get_table_schema(table_name)
+        hpo_table_name = f'{hpo_id}_{_filename_to_table_name(blob.name)}'
```
lrwb-aou, line 58: Why not just
```suggestion
        hpo_table_name = f'{hpo_id}_{table_name}'
```

data_steward/tools/load_submission.py
```diff
@@ -0,0 +1,124 @@
...
+        fq_hpo_table = f'{bq_client.project}.{dst_dataset}.{hpo_table_name}'
+        destination = Table(fq_hpo_table, schema=schema)
+        destination = bq_client.create_table(destination, exists_ok=True)
```
lrwb-aou, line 61: What is your rationale for `exists_ok=True` here?  I would think we wouldn't want to overwrite an existing table if we are reloading from the archive.  I would expect an error to be thrown to prevent this.

data_steward/tools/load_submission.py
```diff
@@ -0,0 +1,124 @@
...
+        job_config = LoadJobConfig()
+        job_config.schema = schema
+        job_config.skip_leading_rows = 1
+        job_config.source_format = 'CSV'
```
lrwb-aou, line 65: nice!  Have you considered using `job_id` or `job_id_prefix` to help group jobs created by this script?  I think google has some decent advice [here](https://cloud.google.com/bigquery/docs/running-jobs#generate-jobid).  What do you think?

data_steward/tools/load_submission.py
```diff
@@ -0,0 +1,124 @@
...
+        source_uri = f'gs://{bucket_name}/{blob.name}'
+        load_job = bq_client.load_table_from_uri(source_uri,
+                                                 destination,
+                                                 job_config=job_config)
+        LOGGER.info(f'table:{destination} job_id:{load_job.job_id}')
```
lrwb-aou, line 70: 👍

data_steward/tools/load_submission.py
```diff
@@ -0,0 +1,124 @@
...
+                                 required=True)
+    return argument_parser
+
+
+def main(project_id, dataset_id, bucket_name, hpo_id, folder_name):
+    bq_client = get_client(project_id)
+    gcs_client = Client(project_id)
+    site_bucket = get_bucket(bq_client, hpo_id)
+    prefix = f'{hpo_id}/{site_bucket}/{folder_name}'
+    load_stage(dataset_id, bq_client, bucket_name, prefix, gcs_client, hpo_id)
```
lrwb-aou, line 117: Is there a reason `load_stage` is returning an object but the script isn't using the returned object?

lrwb-aou, line 117: nit:  It would be nice to log the procedure starting and finishing successfully.

data_steward/tools/load_submission.py
```diff
@@ -0,0 +1,124 @@
...
+                                 help='Name of the submission folder to load',
+                                 required=True)
+    return argument_parser
+
+
+def main(project_id, dataset_id, bucket_name, hpo_id, folder_name):
+    bq_client = get_client(project_id)
+    gcs_client = Client(project_id)
+    site_bucket = get_bucket(bq_client, hpo_id)
+    prefix = f'{hpo_id}/{site_bucket}/{folder_name}'
```
lrwb-aou, line 116: Do we need to do this `get_bucket` if we've already looked up the folder_name and provided it via the command line?  Am I missing something?  If I've looked up the `folder_name`, don't I know the `site_bucket` and the full prefix I'm interested in?  I would think this might be more useful, as is,  if I were providing the hpo_id and the date the ehr snapshot ran.  What do you think?

lrwb-aou (commented):
@nishanthpp93 this is a nice PR.  I think we can discuss some conventions we might be interested in pursuing.  Let me know if you'd like to talk about these comments with a call.

data_steward/tools/load_submission.py
```diff
@@ -0,0 +1,124 @@
...
+    return filename.split('/')[-1].replace('.csv', '').lower()
+
+
+def load_stage(dst_dataset, bq_client, bucket_name, prefix, gcs_client, hpo_id):
+    """
+    Stage files from a bucket to a dataset
+
+    :param dst_dataset: reference to destination dataset object
+    :param bq_client: a BigQuery client object
+    :param bucket_name: the location in GCS containing the vocabulary files
```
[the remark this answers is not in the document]
nishanthpp93: You're right, it should be the archive.

data_steward/tools/load_submission.py
nishanthpp93: I think by default it [appends](https://cloud.google.com/bigquery/docs/reference/rest/v2/Job#jobconfigurationload) but you're right in that this wouldn't help us. I'll change it to throw an error instead.

data_steward/tools/load_submission.py
nishanthpp93: I wasn't sure what to do with the list of job_ids. They are logged inside load_stage, but let me know if you have ideas on using them here.

data_steward/tools/load_submission.py
nishanthpp93: You're right, I could get it from the user but didn't want to add another parameter. I initially designed it to start with just the hpo_id and the date, but there were a lot of variations in the folder names and a manual review was necessary (this was before we were enforcing the naming convention). We can discuss this further over a call.

data_steward/tools/load_submission.py
```diff
@@ -0,0 +1,157 @@
...
+Loads a submission into bq from the archive
+"""
+import logging
+import argparse
+from typing import List
+
+from google.cloud.bigquery import LoadJobConfig, LoadJob, Table, Client as BQClient
+from google.cloud.storage import Client as GCSClient
+from google.cloud.exceptions import GoogleCloudError
+from concurrent.futures import TimeoutError
```
lrwb-aou, line 11: Ran pylint and got this.
```
tools/load_submission.py:11:0: C0411: standard import "from concurrent.futures import TimeoutError" should be placed before "from google.cloud.bigquery import LoadJobConfig, LoadJob, Table, Client as BQClient" (wrong-import-order)
```
It's just asking to change the import order.

data_steward/tools/load_submission.py
lrwb-aou: I think the `load_stage` function is good.  I just don't think we need to return a value if there are no plans to use it.  Since the job_ids are not needed in this script, do you think `load_stage` will/should ever be used outside this module allowing a different script to run this function and get the job_ids from it?  Or would you prefer if a different script ran this one to funnel it through the `main`?

I think what to do depends on other possible uses.

data_steward/tools/load_submission.py
nishanthpp93: I've used __name__ as the prefix but let me know of more info to add.

data_steward/tools/load_submission.py
nishanthpp93: I can see this being used to load other files from the archive too, but they may not have the prefix similar to the site files (e.g. vocabulary). I think in such cases the `load_stage` function needs to be used directly to allow for that flexibility.

lrwb-aou: Alright.  In the interest of python's "Explicit is better than implicit", would you consider using
```
_ = load_stage(...)
```

data_steward/tools/load_submission.py
```diff
@@ -0,0 +1,157 @@
+"""
```
lrwb-aou, line 1: This is likely my last comment, but I think `load_archived_submission` is a clearer name than `load_submission`.  Someone new looking at the code base may not realize that `load_submission` is pulling from a different source than the `import_rdr_omop` script.

tests/unit_tests/data_steward/tools/load_archived_submission_test.py
```diff
@@ -0,0 +1,64 @@
+from unittest import TestCase, mock
+from unittest.mock import ANY
+
+from tools import load_archived_submission as ls
+import common
+
+
+class LoadVocabTest(TestCase):
```
lrwb-aou, line 8: nit:  Class name doesn't reflect what you're testing.

nishanthpp93: You're right, I just added this using the vocabulary test

data_steward/tools/load_archived_submission.py
```diff
@@ -0,0 +1,172 @@
...
+        destination = bq_client.create_table(destination)
+        job_config = LoadJobConfig()
+        job_config.schema = schema
+        job_config.skip_leading_rows = 1
+        job_config.source_format = 'CSV'
+        source_uri = f'gs://{bucket_name}/{blob.name}'
+        load_job = bq_client.load_table_from_uri(source_uri,
+                                                 destination,
+                                                 job_config=job_config,
+                                                 job_id_prefix=__name__)
```
lrwb-aou, line 101: I like this, but I think when run as a stand alone script this will prefix the job with `__main__`.  You can view some other samples with this quick script.
```python
import os

print(f'name -> \t{__name__}')

print(f'file ->\t {__file__}')

#print(f'qualname -> \t {__qualname__}')

#print(f'module ->\t{__module__}')

def in_func():
    print(f'----------------------------------------------------')
    print(f'name -> \t{__name__}')

    print(f'file ->\t {__file__}')
    print(f'file ->\t {os.path.basename(__file__)[0:-3]}')
    print(f'file ->\t {__file__.split("/")[-1][0:-3]}')

    # These are intended to be called on objects and not as stand alone built-ins. 
#    print(f'qualname -> \t {__qualname__}')

#    print(f'module ->\t {__module__}')                                                                                                                                                                                                   


in_func()
```

nishanthpp93: Thanks, I was thinking the main script which is calling the function should be highlighted. This would result in the same job prefix if the caller was itself or a different script. It is definitely more informative than '__main__', so I can use `__file__` and extract the filename for the prefix.

nishanthpp93: Actually, I tested this a bit more and it looks like `__name__` just returns the imported string from the caller. So it would serve the same purpose anyway. I'll just use `__file__`.

tests/unit_tests/data_steward/tools/load_archived_submission_test.py
```diff
@@ -0,0 +1,64 @@
...
+        expected = common.CONDITION_OCCURRENCE
+        actual = ls._filename_to_table_name(self.files[0])
+        self.assertEqual(expected, actual)
+        expected = common.OBSERVATION
+        actual = ls._filename_to_table_name(self.files[3])
+        self.assertEqual(expected, actual)
+
+    def test_load_folder(self):
+        """"
+        Verify the load function is called with the right URIs
```
lrwb-aou, line 44: 👍

lrwb-aou (approved):
This is good.  I think it will get used frequently.

PR #977, opened 2021-09-22
[DC-1847] Fix broken achilles integration tests

mark-velez:
By loading csv files from a test bucket

### DIFF
--- a/tests/test_util.py
+++ b/tests/test_util.py
@@ -219,49 +219,6 @@ def delete_all_tables(dataset_id):
     return deleted
 
 
-def download_file_from_google_drive(id, destination):
-    URL = "https://docs.google.com/uc?export=download"
-
-    session = requests.Session()
-
-    response = session.get(URL, params={'id': id}, stream=True)
-    token = get_confirm_token(response)
-
-    if token:
-        params = {'id': id, 'confirm': token}
-        response = session.get(URL, params=params, stream=True)
-
-    save_response_content(response, destination)
-
-
-def get_confirm_token(response):
-    for key, value in response.cookies.items():
-        if key.startswith('download_warning'):
-            return value
-
-    return None
-
-
-def save_response_content(response, destination):
-    CHUNK_SIZE = 32768
-
-    with open(destination, "wb") as f:
-        for chunk in response.iter_content(CHUNK_SIZE):
-            if chunk:  # filter out keep-alive new chunks
-                f.write(chunk)
-
-
-def get_synpuf_results_files():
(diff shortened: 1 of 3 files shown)


tests/test_util.py
```diff
@@ -13,6 +13,7 @@
 import gcs_utils
 import resources
 
+TEST_RESOURCES_BUCKET = 'aou-res-curation-test-resources'
```
ksdkalluri, line 16: @mark-velez can we store the bucket name in the environment variables? or is it okay to be here?

mark-velez: Yes, I discussed this with Lauren- this is not great, but it is no worse than the current approach of hardcoding a file_id. I'm reluctant to add another environment variable. Let me know if you can think of any easy alternatives.

ksdkalluri: @mark-velez  we have project_id already in envireonment_variables, we can leverage it '{project_id}-resources` . let me know your thoughts.

mark-velez:
@ksdkalluri does this approach to getting the bucket look ok? (Excuse the failed builds I added while trying to figure out my lint setup.)

tests/test_util.py
```diff
@@ -13,6 +13,7 @@
 import gcs_utils
 import resources
 
+RESOURCES_BUCKET_FMT = '{project_id}-resources'
```
ksdkalluri, line 16: @mark-velez project_id is not yet available yet here.

tests/test_util.py
```diff
@@ -281,33 +239,22 @@ def write_cloud_fp(bucket, name, fp):
     return gcs_utils.upload_object(bucket, name, fp)
 
 
-def populate_achilles(hpo_bucket, hpo_id=FAKE_HPO_ID, include_heel=True):
+def populate_achilles(hpo_id=FAKE_HPO_ID, include_heel=True):
     from validation import achilles, achilles_heel
     import app_identity
 
     app_id = app_identity.get_application_id()
```
ksdkalluri, line 246: @mark-velez this app_id can be used in place of `project_id`.

ksdkalluri (changes requested):

tests/test_util.py
mark-velez: This is a format string where `project_id` is a named index..?

tests/test_util.py
mark-velez: That's what is happening in the line below..?

ksdkalluri: yes, my bad thats' true. All the tests were passing. Thanks for updating this.

ksdkalluri (approved):

PR #978, opened 2021-09-23
[DC-1889] Reformat JSON schema files (no functional changes)

mark-velez:
This is to provide for a clean diff after the JSON files are updated

### DIFF
--- a/data_steward/resource_files/schemas/cdm/achilles/achilles_analysis.json
+++ b/data_steward/resource_files/schemas/cdm/achilles/achilles_analysis.json
@@ -1,37 +1,37 @@
 [
-    {
-        "type": "integer", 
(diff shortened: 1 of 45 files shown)


nishanthpp93 (commented):
Thanks @mark-velez, looks like all trailing whitespaces are removed and indents are 2 spaces, which will help highlight actual schema changes later on. I think the linting checks are failing due to the missing ticket number. We can add this as a subtask to the schema upgrade ticket and use that number. Let me know your thoughts.

mark-velez:
Thanks @nishanthpp93 . I'd only referenced the ticket in the commit- didn't realize I had to do it in the pull request title. I updated the title and it passes the linting check now.

nishanthpp93 (approved):

PR #983, opened 2021-09-28
[DC-1820] Development on Docker efforts

dcarbone:
This PR encompasses two major shifts for the Curation team:

1. All execution of tests can now be done within containers
2. The Circle-CI config has been updated:
  1. It is now based on a VM
  2. Docker layer caching has been enabled
  3. Tests are executed within the same container as devs can now use locally


dcarbone:
fwiw, codacy is lying.  `--chmod` is absolutely a valid flag, it's just behind the times.

docker/develop/Dockerfile
```diff
@@ -0,0 +1,188 @@
...
+# expected to be defined via either docker-compose or as docker build --build-arg
+ARG UID
+ARG GID
+
+# this will be the name of our in-container user
+ARG CURATION_USER="curation"
+ARG CURATION_HOME="/home/${CURATION_USER}"
+
+# gsdk args
+ARG GSDK_VERSION
```
MichaelSchmidt82, line 44: Was there a particular reason defualts were excluded (Understanding that `docker-compose.yaml`'s values take precedence over what would be defaulted)?

MichaelSchmidt82 (approved):
Solid work. I __need__ this.

docker/develop/Dockerfile
dcarbone: Hmm, are you asking why I didn't define a default for some of the ARG's?

It boils down to this:

1. This container was expected to be run using compose
2. Having it defined statically inside the dockerfile AND inside the compose yaml means that updating this to a newer version now requires updating two files rather than one.

IMO if it were moved into the Dockerfile, it should be removed from the compose yaml.

lrwb-aou (approved):
Please merge this when you're ready.  We've said, "when 1820 is merged..." pretty often.

PR #985, opened 2021-10-01
[DC-1897] Refactoring reporter.py:

dcarbone:
- moving much of the business logic for class-based cleaning rules into "get_class_info" func
- no longer relying on existence of "get_query_specs" to determine if instance is subclass of base cleaning rule
- no longer attempting to execute each function

@lrwb-aou I need guidance on two pieces:

1. What should be done in the scenario where an object _is_ provided but it is _not_ a subclass of the base cleaning rule?  Should it throw an exception?
2. What was the original intent behind executing each inspected method?  It triggered DC-1896 the same as the problem this PR was originally intending to fix.  I replaced it with `callable()`, but I don't think that is exactly what you had intended.

### DIFF
--- a/data_steward/cdr_cleaner/reporter.py
+++ b/data_steward/cdr_cleaner/reporter.py
@@ -3,15 +3,10 @@ A package to generate a csv file type report for cleaning rules.
 """
 # Python imports
 import csv
+import inspect
 import logging
 import os
 from copy import copy
-from inspect import signature
-
-# Third party imports
-from googleapiclient.errors import HttpError
-from google.api_core.exceptions import BadRequest
-from google.auth.exceptions import DefaultCredentialsError
 
 # Project imports
 import cdr_cleaner.args_parser as cleaning_parser
@@ -19,6 +14,7 @@ import cdr_cleaner.clean_cdr as control
 import cdr_cleaner.clean_cdr_engine as engine
 import constants.cdr_cleaner.clean_cdr as cdr_consts
 import constants.cdr_cleaner.reporter as report_consts
+from cdr_cleaner.cleaning_rules.base_cleaning_rule import BaseCleaningRule
 
 LOGGER = logging.getLogger(__name__)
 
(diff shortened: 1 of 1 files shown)


lrwb-aou:
>     * moving much of the business logic for class-based cleaning rules into "get_class_info" func
> 
>     * no longer relying on existence of "get_query_specs" to determine if instance is subclass of base cleaning rule
> 
>     * no longer attempting to execute each function
> 
> 
> @lrwb-aou I need guidance on two pieces:
> 
>     1. What should be done in the scenario where an object _is_ provided but it is _not_ a subclass of the base cleaning rule?  Should it throw an exception?
> 
>     2. What was the original intent behind executing each inspected method?  It triggered DC-1896 the same as the problem this PR was originally intending to fix.  I replaced it with `callable()`, but I don't think that is exactly what you had intended.


The original intent of the `reporter.py` script is to make it easier to generate documentation at the end of the pipeline run.  One of the first questions I was asked after the 2019 run is, what's different.  For each data stage, the script can print any of the various class attributes, like jira issue number, sandbox tables created, sql run for cleaning, module or class name, etc. out to a csv file.  The reporting capability is a nice side effect of the cleaning rule structure.

It has also been used to provide rule information to analysts without needing to run the pipeline (just recently used to generate the list of sandbox table names each rule created to aid statistical analses).  The reporter is intended to be called with dummy data to prevent actually calling any real project resources.  This is because:
1. We don't want to accidentally change production data to get rule information.
2. We don't always want engineers to be running this feature.  It would be nice to hand this to someone who is uncredentialed in the project (analysts, PMs, etc) and let them run the reporter to get the data they want to investigate.

Obviously, this doesn't work for rules that are not class based.  In that situation, the reporter gives the information it can provide via logs and the generated csv file.  The logs provide feedback that the rule is not a class and cannot provide the requested information.  The script adds the module name to the csv output file, and adds blank attributes for the other requested fields since it cannot provide that information.  It should not generate an error that will stop execution because rule order is dependent on its placement in the list, not the time it was added to a datastage.  Meaning, newer class based rules might exist after older ad-hoc rules and it's best to get as much requested info as possible to make whatever analysis or documentation is needed easier to finish.

To my knowledge instantiated and called to use the class functions to provide the requested data.

Could you use a mock to intercept the PSAPI call until the class is properly fixed?

lrwb-aou (approved):
@dcarbone This is a cleaner version of the original.  The original was trying to use duck typing.  This is nicer.

PR #989, opened 2021-10-05
[DC-1840] Modify `snapshot_by_query.py` to copy a 5.2 cdm dataset to 5.3.1 cdm dataset

ksdkalluri:


data_steward/tools/snapshot_by_query.py
```diff
@@ -37,33 +37,64 @@ def create_empty_cdm_tables(snapshot_dataset_id):
     cdm.create_vocabulary_tables(snapshot_dataset_id)
 
 
-def get_field_cast_expr(field, data_type):
+def get_field_cast_expr(field, data_type, required_mode, existing_fields):
```
mark-velez, line 40: Please accept the entire field object (named something like `src_field`) here rather than three separate params and indicate in docstring the type. Rename the last parameter something like `dst_fields`.

data_steward/tools/snapshot_by_query.py
```diff
@@ -37,33 +37,64 @@ def create_empty_cdm_tables(snapshot_dataset_id):
...
+        'integer': 0,
+        'float': 0.0,
+        'string': '\'\'',
+        'date': 'DATE(\'1970-01-01\')',
+        'timestamp': 'TIMESTAMP(\'1970-01-01\')'
+    }
+    if field not in existing_fields and required_mode == 'required':
+        col = f'CAST({replacement_values[data_type.lower()]} AS {bigquery_data_types[data_type.lower()]}) AS {field}'
+    elif field not in existing_fields and required_mode == 'nullable':
+        col = f'CAST(NULL as {bigquery_data_types[data_type.lower()]}) AS {field}'
```
mark-velez, line 67: As discussed, please allow it to fail loudly if a required field is missing in the source. Something like this (untested, just for illustration):
```suggestion
    if dst_field not in existing_fields:
        if required_mode == 'required':
            raise RuntimeError(f'Unable to load the field "{dst_field}" which is required in the destination table and missing from the source table')
        else:
            col = f'CAST(NULL as {bigquery_data_types[data_type.lower()]}) AS {field}'
```

data_steward/tools/snapshot_by_query.py
```diff
@@ -37,33 +37,64 @@ def create_empty_cdm_tables(snapshot_dataset_id):
...
+        col = f'CAST(COALESCE(CAST({field} AS {bigquery_data_types[data_type.lower()]}), NULL) AS {bigquery_data_types[data_type.lower()]}) AS {field}'
+    return col
 
-    if data_type not in ['integer', 'float']:
-        col = f'CAST({field} AS {data_type.upper()}) AS {field}'
-    else:
-        col = f'CAST({field} AS {bigquery_int_float[data_type]}) AS {field}'
 
-    return col
+def get_existing_field_names(table_id, dataset_id):
```
mark-velez, line 75: Please prefer the standard bq client library over `bq_utils`. I'll put an example in a `get_copy_table_query()` comment below.

data_steward/tools/snapshot_by_query.py
```diff
@@ -37,33 +37,64 @@ def create_empty_cdm_tables(snapshot_dataset_id):
...
     try:
         table_name = table_id
         fields = resources.fields_for(table_name)
-        fields_with_datatypes = [
-            get_field_cast_expr(field['name'], field['type'])
-            for field in fields
+        existing_field_names = get_existing_field_names(table_id, dataset_id)
+        col_cast_exprs = [
+            get_field_cast_expr(field['name'], field['type'], field['mode'],
+                                existing_field_names) for field in fields
```
mark-velez, line 95: As discussed lets use the standard bq client rather than the `bq_utils`. I think it'd be a bit easier to follow too: 

```suggestion
        client = bq.get_client(project_id)
        table = client.get_table(f'{dataset_id}.{project_id}')
        dst_fields = resources.fields_for(table_id)
        col_cast_exprs = (get_field_cast_expr(field, table.schema) for field in dst_fields)
```

mark-velez (changes requested):
Thanks @ksdkalluri ! I summarized the points we discussed in the comments.

PR #991, opened 2021-10-08
[DC-1896] make retract deactivated pids cleaning rule conform to cleaning rule conventions

lrwb-aou:
* impements the table_namer parameter
* implements getting a list of sandbox names
* rule should be idempotent, not changing production data to report statistics
* fixes reporter bug introduced that returned bool for sql field instead of actual sql
* adds logging to the clean engine for identifying rules that do not use table_namer yet
* fixes `get_sandbox_table_name()` signature of 
* removes pre-existing unit test
* add integration test
* updates `retract_deactivated_pids` to use schema files to generate sql when a client is unavailable.
* fixes the reporter script to provide sql instead of a boolean value


data_steward/retraction/retract_deactivated_pids.py
```diff
@@ -21,7 +23,8 @@
 
 TABLE_INFORMATION_SCHEMA = JINJA_ENV.from_string(  # language=JINJA2
     """
-SELECT *
+SELECT * except(is_generated, generation_expression, is_stored, is_hidden,
+is_updatable, is_system_defined, clustering_ordinal_position)
 FROM `{{project}}.{{dataset}}.INFORMATION_SCHEMA.COLUMNS`
```
nishanthpp93, line 28: Thank you. I believe this makes the result identical for both client and no-client situations?

data_steward/retraction/retract_deactivated_pids.py
```diff
@@ -210,6 +234,14 @@ def generate_queries(client,
     return sandbox_queries + clean_queries
 
 
+def create_and_get_sandbox_table_name(table, data_stage=None):
+    """
+    Return formatted sandbox table name.
+    """
+    base_name = f"{'_'.join(ISSUE_NUMBERS).lower().replace('-', '_')}_{table}"
+    return sb.get_sandbox_table_name(data_stage, base_name)
```
nishanthpp93, line 242: I thought this function was creating sandbox tables but it looks like this is not doing much except concatenating. Can the function name be changed and maybe simplified to:
```suggestion
    return f`{data_stage}_{base_name}`
```

data_steward/retraction/retract_utils.py
```diff
@@ -159,6 +159,9 @@ def is_labeled_deid(client, project_id, dataset_id):
     :param dataset_id: Identifies the dataset
     :return: Boolean indicating if the dataset is labeled a deid dataset or None if unlabeled
     """
+    if not client:
+        return None
+
```
nishanthpp93, line 164: This function states that `None` indicates the dataset to be unlabeled, but an undefined client can lead to the same response. Should this raise an error instead?

tests/integration_tests/__init__.py
```diff
@@ -0,0 +1,4 @@
+import unittest
+
+# Hides first line of function docstring in test output
+unittest.TestCase.shortDescription = lambda x: '\t'
```
nishanthpp93, line 4: Is this for the `"""\n` part? I've noticed almost all our docstrings to start with an empty line and it's become a somewhat unspoken convention. I think this will help standardize that convention.

tests/integration_tests/data_steward/cdr_cleaner/cleaning_rules/remove_participant_data_past_deactivation_date_test.py
```diff
@@ -0,0 +1,143 @@
...
+        INSERT INTO `{{fq_table_name}}`
+        (observation_id, person_id, observation_concept_id, observation_date,
+         observation_type_concept_id, observation_source_concept_id)
+        VALUES
+        -- Values to exist after running the cleaning rule --
+          (801, 1, 1585899, date('2019-05-01'), 45905771, 111111),
+          (802, 2, 1585899, date('2019-05-01'), 45905771, 222222),
+        -- Values that should be removed by the cleaning rule --
+          (804, 1, 1585899, date('2020-05-01'), 45905771, null),
+          (805, 3, 1585899, date('2020-05-01'), 45905771, 45)
```
nishanthpp93, line 94: Thank you, I think these are helpful. I think it would also be helpful to indicate the reasons. E.g. 801 is retained because it is before 1's deactivation, 2 is not deactivated, 804 is after 1 deactivated and 805 is after 3 deactivated.

nishanthpp93 (changes requested):
Looks good! I think it achieves what it sets out to, but wanted some clarification on some parts along with some suggestions. Let me know your thoughts

data_steward/cdr_cleaner/cleaning_rules/remove_participant_data_past_deactivation_date.py
```diff
@@ -29,85 +29,112 @@
...
+        Return a list table names created to backup deleted data.
         """
-        return []
+        LOGGER.info("Generating static list of possible sandbox table names")
+
+        if self.client:
+            LOGGER.info(f"Getting live table names from: `{self.dataset_id}`")
+            tables_list = self.client.list_tables(self.dataset_id)
+
+            table_names = [table_item.table_id for table_item in tables_list]
```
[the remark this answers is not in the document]
lrwb-aou: In this case `self.affected_tables = common.CDM_TABLES + common.FITBIT_TABLES`.  The current implementation is consistent with the original implementation.  It does not assume that `self.affected_tables` is the source of truth for this rule.  When a client is available, it assumes the live dataset is the source of truth.  Since we are trying to make sure to remove all post deactivation data, I think it makes more sense to use the ultimate source of truth.  

But maybe I could make a change to the value of `self.affected_tables` when `setup_rule` is run to make this better.  I'll try it.

data_steward/retraction/retract_deactivated_pids.py
nishanthpp93: I think this is fine. As long as the extra columns aren't used anywhere, this works as expected.

tests/integration_tests/__init__.py
nishanthpp93: Thanks, I see the intent now. It is odd, but the tab doesn't seem to solve it, rather pushes things to a newline. I think the `lambda x: None` is good.

data_steward/retraction/retract_utils.py
nishanthpp93: I see, this makes sense. I think it would be helpful to add this as a comment saying this returns a None for no-client to trigger a name check.

lrwb-aou: 👍 Yes, I will be happy to add that.

nishanthpp93 (approved):
Looks good! Thanks for incorporating the changes. This is good to merge once tests pass.

PR #996, opened 2021-10-12
[DC-1844] Update csv files for testing

mark-velez:
This pull request is just to help see the diffs between the files @jp3477 and I are using for testing vs the files used for integration tests. It might be useful to use these same files in integration tests ( assuming these will also pass :man_shrugging: )- @ratuagga and @nishanthpp93 let's discuss.


mark-velez:
Closing as this was issued for demonstration purposes and the feature has been applied by another PR already

PR #1006, opened 2021-10-27
[DC-1893] Invalid Zip Code concept code applied to wrong field

jp3477:
Changed set field values of invalid zip code concepts.


ratuagga (dismissed):
For the listing error, just adding an empty commit with commit message starting with '[DC-1893]' should fix it.

lrwb-aou:
@jp3477 I think @ratuagga is correct, you would need to either add or amend your commit message to get past the linting check.  I ran the cleaning rule on some mocked data and it performs as expected.  I did notice that the integration tests didn't actually run, so I kicked them off.  I don't expect any issues, but will wait until they complete.

jp3477:
@ratuagga and @lrwb-aou Thanks for the suggestion, I added the new commit. The tests are running again.

lrwb-aou (approved):
Coverage results indicate the tests were run.  This will be good to merge once all tests finish.

PR #1007, opened 2021-10-27
[DC-1623] EHR Ops raising error trying to process completeness metrics for site

jp3477:


jp3477:
@nishanthpp93 I don't believe the empty completeness query method will work because a list of dictionaries is returned by the `query_rows` function. Without any rows, there are no dictionary keys to use to generate the table headers in the results.html. What do you suggest? I could revert back to not displaying the table if there are empty rows.

nishanthpp93 (approved):
Thanks, I see what's happening. This new query returns no rows so the key in `report_data` is not set, which I think is fine and possibly the expected way to disable this section. I was thinking the headers were required even when it's empty, but looks like it's not since `hpo_report.html` has a check for each key being `false-y`. I think this does the job. We can talk more on this tomorrow too.

PR #1008, opened 2021-10-28
[DC-1793] Update `clean_engine` to utilize `pipeline_logging` utility

ratuagga:
Updated add_console_logging function to use pipeline_logging.

### DIFF
--- a/data_steward/cdr_cleaner/clean_cdr_engine.py
+++ b/data_steward/cdr_cleaner/clean_cdr_engine.py
@@ -9,6 +9,7 @@ from google.cloud.exceptions import GoogleCloudError
 
 # Project imports
 from utils import bq
+from utils.pipeline_logging import configure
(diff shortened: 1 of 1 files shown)


data_steward/cdr_cleaner/clean_cdr_engine.py
```diff
@@ -24,19 +25,7 @@ def add_console_logging(add_handler=True):
     until later.  Useful for debugging.
 
     """
-    logging.basicConfig(
```
lrwb-aou, line 27: @ratuagga This is a nice change.  Consider updating the docstring for this function too.  It doesn't seem to be as relevant with this change.

ratuagga: Sure @lrwb-aou.

lrwb-aou (approved):
Looks good.  I verified it ran as expected locally too.

PR #1010, opened 2021-10-30
[DC-1472] Branch names with deid can cause deactivation test to fail

ratuagga:
Initial Commit to reproduce errors that need fixing.


nishanthpp93 (approved):
Looks good!

PR #1012, opened 2021-11-01
[DC-1890] Create dependency between PPI branching and concept remap and reorder rules

nishanthpp93:
* [DC-1890] Update order of cleaning rules

[DC-1890]: https://precisionmedicineinitiative.atlassian.net/browse/DC-1890?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/data_steward/cdr_cleaner/clean_cdr.py
+++ b/data_steward/cdr_cleaner/clean_cdr.py
@@ -140,6 +140,14 @@ RDR_CLEANING_CLASSES = [
     (StoreNewPidRidMappings,),
     (TruncateRdrData,),
     (RemoveParticipantsUnder18Years,),
+    # execute map_questions_answers_to_omop before PpiBranching gets executed
+    # since PpiBranching relies on fully mapped concepts
+    # trying to load a table while creating query strings,
+    # won't work with mocked strings.  should use base class
+    # setup_query_execution function to load dependencies before query execution
+    (
+        map_questions_answers_to_omop.
+        get_update_questions_answers_not_mapped_to_omop,),
     (PpiBranching,),
     # execute FixUnmappedSurveyAnswers before the dropping responses rules get executed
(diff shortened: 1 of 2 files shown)


data_steward/cdr_cleaner/clean_cdr.py
```diff
@@ -140,6 +140,14 @@
     (StoreNewPidRidMappings,),
     (TruncateRdrData,),
     (RemoveParticipantsUnder18Years,),
+    # execute map_questions_answers_to_omop before PpiBranching gets executed
```
ksdkalluri, line 143: @nishanthpp93 This is good, but we also should add the dependency in the PPI_branching rule dependency list.

ksdkalluri (changes requested):

data_steward/cdr_cleaner/clean_cdr.py
nishanthpp93: I see, thanks. Will add it in the cleaning rule class.

nishanthpp93: @ksdkalluri I am looking at the BaseCleaningRule class and it looks like depends_on can raise an error because this cleaning rule is not base classed. Let me know if that's okay.

nishanthpp93: Just tested this and it looks like the PpiBranching class cannot be instantiated with it. It runs into the same error.

nishanthpp93: I can instead add a comment there about the dependency along with a TODO. Let me know your thoughts.

ksdkalluri: okay, that's a bummer.  Just add a TODO item about the dependency and it will be good for now.

ksdkalluri (approved):

PR #1013, opened 2021-11-02
[DC-1598] [Cloud Storage] Define Google Storage Client

MichaelSchmidt82:


dcarbone:
@MichaelSchmidt82: This looks great!  Thanks for putting this together.

Looks like the test failed as we are currently not activating gcloud creds in the unit test suite.

Have a look [at the integration test](https://github.com/all-of-us/curation/blob/develop/.circleci/config.yml#L158) to see how you can get this activated in the unit tests.

MichaelSchmidt82:
I mocked the default auth for cases where the env var GOOGLE_APPLICATION_CREDENTIALS is not set AND no credentials are passed.

tests/unit_tests/data_steward/gcloud/gcs/gcs_test.py
```diff
@@ -0,0 +1,58 @@
...
+
+    @classmethod
+    def setUpClass(cls):
+        print('**************************************************************')
+        print(cls.__name__)
+        print('**************************************************************')
+
+    def setUp(self):
+        # Input parameters expected by the class
+        self.client = StorageClient()
```
lrwb-aou, line 24: @MichaelSchmidt82 this is causing your test to fail.  Can you describe what you are wanting to do here so we can figure out the best fix?

data_steward/gcloud/gcs/__init__.py
```diff
@@ -0,0 +1,47 @@
...
+        if not prefix.endswith('/'):
+            prefix += '/'
+
+        extra_params: dict = {
+            'projection': 'noAcl',
+            'prefix': prefix,
+            'delimiter': '/'
+        }
+
+        path: str = f'/b/{bucket}/o'
```
lrwb-aou, line 37: @MichaelSchmidt82 why do have the leading `/b` and trailing `/o`?  I'm not sure we have any paths that will conform to that specification.  To be consistent, we should keep this similar to, https://github.com/all-of-us/curation/blob/develop/data_steward/gcs_utils.py#L131.  This will require updating the unit test.

lrwb-aou (changes requested):

data_steward/gcloud/gcs/__init__.py
MichaelSchmidt82: This was used in the original test found in develop.  It's copied and pasted over.  I'll updte the code/test and retest.

MichaelSchmidt82: also: https://github.com/all-of-us/curation/blob/b7d0b5123a97a9c18526edba99cd95277ef0ffc9/data_steward/gcs_utils.py#L64 is not referenced in the project (unused function)

Where as: https://github.com/all-of-us/curation/blob/b7d0b5123a97a9c18526edba99cd95277ef0ffc9/data_steward/utils/gcs.py#L7 (which contains the `f'/b/{bucket}/o'`) is referenced twice.

lrwb-aou: @nishanthpp93 Can you weigh in here?  You added the code referenced [here.](https://github.com/all-of-us/curation/blob/b7d0b5123a97a9c18526edba99cd95277ef0ffc9/data_steward/utils/gcs.py#L7)  If the `/b/` and `/o` are important to the intended function, then I will withdraw my comment about removing them from the path string.

nishanthpp93: @lrwb-aou Sure, I believe it is the [JSON API request form](https://cloud.google.com/storage/docs/folders#rest-apis) of the bucket path. This function is slightly lower level so this form is needed here.

MichaelSchmidt82: Are their any more changes or items to discuss on this PR?

nishanthpp93 (commented):
I think @lrwb-aou already addressed some concerns. I just realized that the previous function didn't have any integration tests added. Since this new method will be used going forward, could you add an integration test? For the setup, it would just initialize the new storage client and create a file inside a specific path in a bucket (say BUCKET_NAME_FAKE env var). For the validation, it would make sure that the subfolder is returned given the parent folder. Let's discuss this over a call.

tests/integration_tests/data_steward/gcloud/gcs_client_test.py
```diff
@@ -0,0 +1,56 @@
...
+    def test_list_sub_prefixes(self):
+        self._empty_bucket()
+        self._stage_bucket()
+
+        items = self.client.list_sub_prefixes(self.bucket_name, self.prefix)
+        print(items)
+
+        self.assertEqual(len(self.sub_prefixes), len(items))
+        for item in items:
+            self.assertIn(item[:-1], self.sub_prefixes)
```
nishanthpp93, line 40: nit: One of the most helpful unit test assertions IMO is `assertCountEqual`. It can compare unhashable types and is order agnostic. Despite its name, it verifies that all items in one exist in the other (essentially counts of each item are the same in both).
```suggestion
        items = self.client.list_sub_prefixes(self.bucket_name, self.prefix)

        self.assertCountEqual(items, self.sub_prefixes)
```
This will need minor adjustments like adding `/` to each sub_prefix, so I am not making this a required change.
You can also get rid of the print statement.

nishanthpp93 (approved):
Quick and brilliant work on the test! I ran it on my end and it looks good. Added a suggestion but not required for merge. Thanks again for adding this in!

tests/integration_tests/data_steward/gcloud/gcs_client_test.py
MichaelSchmidt82: First: good to know about `assertCountEqual`!  I'll keep that in mind for future tests.

Second: The `print()` statement was removed.  It should not have made it in.

Third:  I considered adding the trailing `/` into the sub_prefixes so that the `self.assertIn()` call wouldn't have the splice `[:-1]`, but ultimately I felt this made the code less readable in the upload section:

`bucket.blob(f'{sub_prefix}obj.txt').upload_from_string(self.data)`

That line without the `/` seperating the filename/blob name looked odd, so I opted for the string splice.

Thanks for the positive feedback!

nishanthpp93: You're right, it does look odd. The `sub_prefix` would then need the splice, but interestingly, it seems to work even with the `/` added without the splice as well.

PR #1014, opened 2021-11-03
[DC-1942] Delete note_nlp table from AOU_REQUIRED and MAPPED_CLINICAL tables

nishanthpp93:
* [DC-1942] Remove note_nlp from table lists

[DC-1942]: https://precisionmedicineinitiative.atlassian.net/browse/DC-1942?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


ratuagga (commented):
It was nice adding 'note_nlp' in 'other_cdm_tables' will still include it in 'cdm_tables' list for the integration tests. I did not know that excluding it in 'clinical_data_tables' will be fine, but if there is no mapping table for 'note_nlp', can it be included in the 'unmapped_clinical_data_tables' along with 'death'? So, it can be part of 'clinical_data_tables' list.. 
I'm not sure if it'll add any value or not.. please correct me if I'm wrong.

data_steward/common.py
```diff
@@ -52,8 +50,7 @@
 # Clinical tables which do not have a corresponding mapping table.
```
ratuagga, line 50: I think this comment should be 'Clinical tables which have a corresponding mapping tables.'
I just noticed it.

nishanthpp93:
> It was nice adding 'note_nlp' in 'other_cdm_tables' will still include it in 'cdm_tables' list for the integration tests. I did not know that excluding it in 'clinical_data_tables' will be fine, but if there is no mapping table for 'note_nlp', can it be included in the 'unmapped_clinical_data_tables' along with 'death'? So, it can be part of 'clinical_data_tables' list..
> I'm not sure if it'll add any value or not.. please correct me if I'm wrong.

This is great feedback and I think makes sense! Will update this.

ratuagga (approved):

PR #1015, opened 2021-11-11
[DC-1810] Apply Additional Row Suppression Rules to RT CDR

ratuagga:
added cancer and organTransplant concepts in data suppression list.

### DIFF
--- a/data_steward/cdr_cleaner/clean_cdr.py
+++ b/data_steward/cdr_cleaner/clean_cdr.py
@@ -319,6 +319,8 @@ REGISTERED_TIER_DEID_CLEANING_CLASSES = [
     (StringFieldsSuppression,),
     (SectionParticipationConceptSuppression,),
     (RegisteredCopeSurveyQuestionsSuppression,),
+    (CancerConceptSuppression,),
+    (OrganTransplantConceptSuppression,),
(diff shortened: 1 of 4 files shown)


jp3477 (changes requested):
These additions look good. Can you also add `cdr_consts.REGISTERED_TIER_DEID` to the `affected_datasets` parameters in the cleaning classes, [here](https://github.com/all-of-us/curation/blob/ra/data_suppression_list/data_steward/cdr_cleaner/cleaning_rules/cancer_concept_suppression.py#L52) and [here](https://github.com/all-of-us/curation/blob/ra/data_suppression_list/data_steward/cdr_cleaner/cleaning_rules/deid/organ_transplant_concept_suppression.py#L38).

You can follow the example of another cleaning rule that applies to both tiers, [generate_ext_tables](https://github.com/all-of-us/curation/blob/ra/data_suppression_list/data_steward/cdr_cleaner/cleaning_rules/generate_ext_tables.py#L88).

ratuagga:
Hi @jp3477,
I made the additions you suggested. Thanks for that.
I did not find the UT for organ_transplant_concept_suppression though

jp3477 (approved):
Thanks for these changes! Good to merge, assuming integration tests pass.

PR #1018, opened 2021-11-12
[DC-1951] Deprecate discovery based GCS utils with a warning

MichaelSchmidt82:
Includes:

1) Added Deprecated library for future uses (BQ Utils 👀).
2) Emit a `DeprecationWarning()` for function calls that utilize the GC Discovery client.

See [DC-1951](https://precisionmedicineinitiative.atlassian.net/browse/DC-1951)


dcarbone (approved):
This looks good to me!

PR #1020, opened 2021-11-16
[DC-1970] Define empty_bucket(name: str) helper method for gcloud.gcs.StorageClient()

MichaelSchmidt82:
A helper function called empty_bucket() has been created.  Any additional kwargs that may be supplied list_blob() are passed along for deletion as well (e.g.  To empty the bucket starting at a given prefix: `StorageClient.empty_bucket('bucket', prefix='prefix')`).

This work includes integration and unit tests.

### DIFF
--- a/data_steward/gcloud/gcs/__init__.py
+++ b/data_steward/gcloud/gcs/__init__.py
@@ -26,9 +26,10 @@ class StorageClient(Client):
         (i.e gsutil rm -r gs://bucket/prefix/)
         """
 
-        blobs = self.list_blobs(bucket, **kwargs)
-        for blob in blobs:
-            blob.delete()
+        pages = self.list_blobs(bucket, **kwargs).pages
+        for page in pages:
+            for blob in page:
+                blob.delete()
 
     def list_sub_prefixes(self, bucket: str, prefix: str) -> None:
         """
--- a/data_steward/gcloud/gcs/__init__.py
+++ b/data_steward/gcloud/gcs/__init__.py
@@ -18,15 +18,15 @@ class StorageClient(Client):
 
     def empty_bucket(self, bucket: str, **kwargs) -> None:
         """
-        Delete all/partial blobs in a bucket.
+        Delete all blobs in a bucket.
         :param name: A GCS bucket name.
 
         Some common keyword arguments:
         :param prefix: (Optional) Prefix used to filter blobs.
-        (i.e gsutil rm -r gs://bucket/prefix)
+        (i.e gsutil rm -r gs://bucket/prefix/)
         """
-        bucket = self.get_bucket(bucket)
-        blobs = bucket.list_blobs(**kwargs)
+
+        blobs = self.list_blobs(bucket, **kwargs)
         for blob in blobs:
             blob.delete()
 
--- a/data_steward/gcloud/gcs/__init__.py
+++ b/data_steward/gcloud/gcs/__init__.py
@@ -16,7 +16,21 @@ class StorageClient(Client):
     See https://googleapis.dev/python/storage/latest/client.html
     """
 
-    def list_sub_prefixes(self, bucket: str, prefix: str):
+    def empty_bucket(self, bucket: str, **kwargs) -> None:
+        """
+        Delete all/partial blobs in a bucket.
+        :param name: A GCS bucket name.
+
+        Some common keyword arguments:
+        :param prefix: (Optional) Prefix used to filter blobs.
+        (i.e gsutil rm -r gs://bucket/prefix)
+        """
+        bucket = self.get_bucket(bucket)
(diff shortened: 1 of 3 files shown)


data_steward/gcloud/gcs/__init__.py
```diff
@@ -16,7 +16,21 @@ class StorageClient(Client):
...
+        """
+        Delete all blobs in a bucket.
+        :param name: A GCS bucket name.
+
+        Some common keyword arguments:
+        :param prefix: (Optional) Prefix used to filter blobs.
+        (i.e gsutil rm -r gs://bucket/prefix/)
+        """
+
+        blobs = self.list_blobs(bucket, **kwargs)
```
dcarbone, line 29: One thing to consider here is how to handle buckets with potentially thousands, if not hundreds of thousands of records.  Attempting to retrieve the entire list of blogs in a single request could run us into issues if we try to operate on a bucket of sufficient size.

To that end, I'd recommend having a look at the `page` and `page_size` parameters, see how we could chunk up the total list by only having a few specific ones in memory at a time.

tests/unit_tests/data_steward/gcloud/gcs/gcs_test.py
```diff
@@ -19,31 +28,42 @@ def setUpClass(cls):
         print('**************************************************************')
 
     def setUp(self):
-        self.client = StorageClient.create_anonymous_client()
```
dcarbone, line 22: Why did you switch away from using the anonymous client constructor?

dcarbone (changes requested):
@MichaelSchmidt82: Looks good for the most part, just a question and a suggestion :)

data_steward/gcloud/gcs/__init__.py
MichaelSchmidt82: OUTSTANDING point!  There is a max_results kwarg that maybe passed to address this 🙂
So for these use cases, you would probably provide a loop:
```
sc = StorageClient()
while sc.list_blobs('humongous_bucket', max_results=1):
    sc.empty_bucket('humongous_bucket', max_results=10000)
```

See: https://googleapis.dev/python/storage/latest/client.html#google.cloud.storage.client.Client.list_blobs for additional args

you might also address this using offsets to "page" through results:

start_offset (str) – (Optional) Filter results to objects whose names are lexicographically equal to or after startOffset. If endOffset is also set, the objects listed will have names between startOffset (inclusive) and endOffset (exclusive).

end_offset (str) – (Optional) Filter results to objects whose names are lexicographically before endOffset. If startOffset is also set, the objects listed will have names between startOffset (inclusive) and endOffset (exclusive).

tests/unit_tests/data_steward/gcloud/gcs/gcs_test.py
MichaelSchmidt82: It became apparent that this was not really addressing a critical point of pain when designing tests:  Auth happens on object creation.

So whenever a StorageClient is created, the upstream google client is initialized and an auth happens during this phase.  When creating an anonymous client, that auth is sort of "delayed" until an auth is determined to be appropriate.  When Nishanth and I were examining the tests, we found using google's create_anonymous_client() was causing errors with "auth tokens that couldn't be refreshed" or something to that effect between tests.  Mocking the auth process didn't work for us or for StackOverflow.

What we really needed was to create the client, but not call the upstream __init__ that forces the auth process, so the DummyClient() was created.   Using this dummy client, we bypass the __init__ call using a pass statement, and can then mock the upstream client's methods (list_blobs, etc.).

data_steward/gcloud/gcs/__init__.py
dcarbone: `max_results` _could_ do what we need if we assume that the page resets each time we delete a set of them.

I would also recommend having a look at the `batch()` functionality of the client: https://googleapis.dev/python/storage/latest/client.html#google.cloud.storage.client.Client.batch

This would help us execute fewer requests in total, as well.

tests/unit_tests/data_steward/gcloud/gcs/gcs_test.py
dcarbone: thanks for the explanation :)

data_steward/gcloud/gcs/__init__.py
MichaelSchmidt82: I'm not sure what needs to change.  Do you want the function to perform deletions as part of a batch operation/page approach?

MichaelSchmidt82: Ah, I see this:
https://github.com/googleapis/google-cloud-python/issues/727

and went with the pages you recommended above.  good catch.

dcarbone (approved):
lgtm!

PR #1023, opened 2021-11-17
[DC-1965] updated bucket_access_configured(bucket_name) to use StorageClient()

lledezma:
function utilizes IAM test permission checking instead of list access.

### DIFF
--- a/data_steward/tools/add_hpo.py
+++ b/data_steward/tools/add_hpo.py
@@ -17,6 +17,7 @@ import app_identity
 import bq_utils
 import constants.bq_utils as bq_consts
 import gcs_utils
+from gcloud.gcs import StorageClient
 import resources
 from tools import cli_util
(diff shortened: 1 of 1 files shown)


data_steward/tools/add_hpo.py
```diff
@@ -17,6 +17,7 @@
 import bq_utils
 import constants.bq_utils as bq_consts
 import gcs_utils
```
MichaelSchmidt82, line 19: Is gcs_utils still being used?

lledezma: Good catch.

dcarbone (approved):
lgtm!

MichaelSchmidt82 (commented):
Just a comment, not a change request.  consider longer or better descriptive variable names to improve readability.

PR #1024, opened 2021-11-17
[DC-1964] Update field_mappings.csv to include newly added fields for `domain_allignment` CR

ksdkalluri:


jp3477 (approved):
Nice, double checked that all of the mappings are here.

ksdkalluri:
> Nice, double checked that all of the mappings are here.

@jp3477 can you check this again. I had to force push after fixing merge conflicts.

PR #1026, opened 2021-11-18
[DC-1973] Update export_test.py to use gcloud.gcs.StorageClient()

MichaelSchmidt82:
1) Removed helper function _empty_bucket(). Now uses helper from StorageClient instead (created by DC-1970).
2) Removed gcs_util.get_object() and use gcloud.gcs.StorageClient calls instead.
3) Added some type hints for primitive types.
4) Changed some string concaternations into f-strings for readability/space saving.


tests/integration_tests/data_steward/validation/export_test.py
```diff
@@ -86,22 +85,26 @@ def test_export_achillesheel(self, mock_is_hpo_id):
     def test_run_export(self, mock_is_hpo_id):
```
nishanthpp93, line 85: I was hoping the upgrade would reduce the number of 'Resource Warning' messages about unclosed sockets etc. But a lot of them seem to persist. I think this is primarily due to bq_utils being called, but hopefully there was a drop in the number due to replacing gcs_utils. Did you notice such a reduction?

nishanthpp93 (commented):
Looks good! Added a couple comments. Let me know your thoughts. Also, could you add this ticket to the sprint so I can mark myself as reviewer?

tests/integration_tests/data_steward/validation/export_test.py
```diff
@@ -4,8 +4,11 @@
 
 import mock
 
+from google.cloud import storage
+
 import bq_utils
 import common
+from gcloud.gcs import StorageClient
 import gcs_utils
```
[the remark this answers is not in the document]
MichaelSchmidt82: No.  This is defined at line 34 and used 4 times in this test.  However, `from google.cloud import storage` should NOT be used, but is imported.  It should be removed.

tests/integration_tests/data_steward/validation/export_test.py
MichaelSchmidt82: Actually, it's probably better to add this in so I'll do this now.

tests/integration_tests/data_steward/validation/export_test.py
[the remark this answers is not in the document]
nishanthpp93: I see the usages for gcs_utils. They all seem to be for the bucket retrieval. I guess that's fine and can be replaced later.

tests/integration_tests/data_steward/validation/export_test.py
nishanthpp93: Thanks! I think this is good to go.

nishanthpp93 (approved):

PR #1029, opened 2021-11-22
[DC-1985] update `Remove_non_matching_participant` to skip for sites where identity_match table is not available.

ksdkalluri:


data_steward/cdr_cleaner/cleaning_rules/remove_non_matching_participant.py
```diff
@@ -93,33 +104,37 @@ def get_list_non_match_participants(project_id, validation_dataset_id, hpo_id):
...
-    incomplete_jobs = bq_utils.wait_on_jobs([query_job_id])
-    if incomplete_jobs:
-        raise bq_utils.BigQueryJobWaitError(incomplete_jobs)
-
-    # return the person_ids only
-    return [row[PERSON_ID_FIELD] for row in bq_utils.response2rows(results)]
+        try:
+            LOGGER.info(
+                'Identifying non-match participants in {dataset_id}.{identity_match_table}'
+                .format(dataset_id=validation_dataset_id,
```
nishanthpp93, line 115: Is this showing up in git diff because of the indent? You could skip the indenting by returning an empty list in the conditional and changing it to `if not exists`.

data_steward/cdr_cleaner/cleaning_rules/remove_non_matching_participant.py
```diff
@@ -67,6 +67,17 @@ def exist_participant_match(ehr_dataset_id, hpo_id):
...
 
+def exist_identity_match(table_id, dataset_id):
+    """
+    This function checks if the hpo has valid the identity_match table
+
+    :param table_id:
+    :param dataset_id:
+    :return:
+    """
+    return bq_utils.table_exists(table_id, dataset_id)
```
nishanthpp93, line 78: `bq_utils` will be deprecated. You could use [this example](https://cloud.google.com/bigquery/docs/samples/bigquery-table-exists#bigquery_table_exists-python) for the same check. This will require creating a client but it's not a big change.

nishanthpp93 (changes requested):
Looks good but added suggestions for a cleaner git diff. Let me know your thoughts.

data_steward/cdr_cleaner/cleaning_rules/remove_non_matching_participant.py
ksdkalluri: That is not just the indent, the whole chunk of code is moved to process out of the loop.

nishanthpp93 (approved):
Looks good! Can be merged when tests pass.

PR #1030, opened 2021-11-22
[DC-1986] `snapshot_by_query` is failing to cast 'bool' data_type.

ksdkalluri:


jp3477 (approved):
Nice, this looks good.

jp3477:
An integration test is failing, though, so take a look at that before merging.

PR #1031, opened 2021-11-24
[DC-1992] Fix migrate_cdm52_to_cdv531.py `LF/CRLF` issue.

ksdkalluri:


nishanthpp93 (approved):
Verified that it is LF.

PR #1033, opened 2021-11-24
[DC-1993] ehr_union_test failure

ratuagga:
removed list to set conversion to retain list sequence.

### DIFF
--- a/data_steward/validation/ehr_union.py
+++ b/data_steward/validation/ehr_union.py
@@ -757,9 +757,7 @@ def main(input_dataset_id, output_dataset_id, project_id, hpo_ids_ex=None):
     # Get all hpo_ids.
     hpo_ids = [item['hpo_id'] for item in bq_utils.get_hpo_info()]
     if hpo_ids_ex:
-        for ex_id in hpo_ids_ex:
-            if ex_id in hpo_ids:
-                hpo_ids.remove(ex_id)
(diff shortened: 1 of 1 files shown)


data_steward/validation/ehr_union.py
```diff
@@ -757,7 +757,9 @@ def main(input_dataset_id, output_dataset_id, project_id, hpo_ids_ex=None):
     # Get all hpo_ids.
     hpo_ids = [item['hpo_id'] for item in bq_utils.get_hpo_info()]
     if hpo_ids_ex:
-        hpo_ids = list(set(hpo_ids) - set(hpo_ids_ex))
+        for ex_id in hpo_ids_ex:
+            if ex_id in hpo_ids:
+                hpo_ids.remove(ex_id)
```
nishanthpp93, line 762: Nice catch. You can also do list comprehension.
```suggestion
        hpo_ids = [hpo_id for hpo_id in hpo_ids if hpo_id not in hpo_ids_ex]
```

nishanthpp93 (approved):
Added a suggestion. I'd say try it and merge only after tests pass.

data_steward/validation/ehr_union.py
ratuagga: that would work too to preserve the order, yes.

PR #1034, opened 2021-11-29
[DC-1953] Update `UpdateFieldsNumbersAsStrings` CR to only sandbox modified records

ksdkalluri:


nishanthpp93 (approved):
Looks good! Do you think OBSERVATION_SOURCE_VALUES can be wrapped in triple quotes? It would be more readable and easier to copy into BQ. Approving since it still works.

PR #1035, opened 2021-11-29
[DC-1234] Perform DRC Name Comparisons

jp3477:
Added checks for first, middle, and last name to the refactored participant validation module.

### DIFF
--- a/data_steward/validation/participants/participant_validation_queries.py
+++ b/data_steward/validation/participants/participant_validation_queries.py
@@ -1,6 +1,28 @@
 # Project imports
 from common import JINJA_ENV
 
+CREATE_NAME_COMPARISON_FUNCTION = JINJA_ENV.from_string("""
+    CREATE FUNCTION IF NOT EXISTS `{{project_id}}.{{drc_dataset_id}}.CompareName`(rdr_name string, ehr_name string)
+    RETURNS string
+    AS ((
+        WITH normalized_rdr_name AS (
+            SELECT REGEXP_REPLACE(LOWER(TRIM(rdr_name)), '[^A-Za-z]', '') AS rdr_name
+        )
+        , normalized_ehr_name AS (
+            SELECT REGEXP_REPLACE(LOWER(TRIM(ehr_name)), '[^A-Za-z]', '') AS ehr_name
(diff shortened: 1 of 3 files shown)


data_steward/validation/participants/participant_validation_queries.py
```diff
@@ -1,6 +1,28 @@
 # Project imports
 from common import JINJA_ENV
 
+CREATE_NAME_COMPARISON_FUNCTION = JINJA_ENV.from_string("""
+    CREATE FUNCTION IF NOT EXISTS `{{project_id}}.{{drc_dataset_id}}.CompareName`(rdr_name string, ehr_name string)
```
ksdkalluri, line 5: @jp3477 Add a docstring to this function about the functionality and limitations of this function, as it will be easier to get around these in the future in order to expand the scope.

data_steward/validation/participants/participant_validation_queries.py
```diff
@@ -1,6 +1,28 @@
...
+    RETURNS string
+    AS ((
+        WITH normalized_rdr_name AS (
+            SELECT REGEXP_REPLACE(LOWER(TRIM(rdr_name)), '[^A-Za-z]', '') AS rdr_name
```
ksdkalluri, line 9: @jp3477 this comparison meets the criteria within the scope of the ticket, but double-check with the name_matching in the existing implementation so that we are not missing any edge cases for eg. normalizing the names, etc.

ksdkalluri (changes requested):

data_steward/validation/participants/participant_validation_queries.py
jp3477: I confirmed that this function has the same functionality as the existing [python function](https://github.com/all-of-us/curation/blob/develop/data_steward/validation/participants/normalizers.py#L226).

data_steward/validation/participants/participant_validation_queries.py
jp3477: How does it look now?

ksdkalluri (approved):
good to merge. Thanks for working on this!

PR #1036, opened 2021-11-30
[DC-2028] Add secondary contact state codes to csv file

brendagutman:


data_steward/resource_files/old_map_short_codes.csv
```diff
@@ -93,4 +93,6 @@ type,pmi_code,short_code
 "Answer","HowOldWereYouOtherMentalHealthSubstanceUse_Adolescent","MentalHealthCondition_OtherAdolescent"
 "Answer","HowOldWereYouOtherMentalHealthSubstanceUse_OlderAdult","MentalHealthCondition_OtherOlderAdult"
 "Answer","HowOldWereYouOtherMentalHealthSubstanceUse_Elderly","MentalHealthCondition_OtherElderly"
-"Question","MentalHealth_RxMedsforOtherMentalHealthSubstanceUse","MentalHealth_RxMedsOtherCondition"
\ No newline at end of file
+"Question","MentalHealth_RxMedsforOtherMentalHealthSubstanceUse","MentalHealth_RxMedsOtherCondition"
+"Answer","secondcontactstate_ak","secondcontactstate_alaska"
```
mark-velez, line 97: @brendagutman I think the casing might need to be changed here. And I believe there were some more `secondcontactstate_*` codes we include as well.

mark-velez (changes requested):
Thanks @brendagutman! Please add all the `secondcontact` codes.

data_steward/resource_files/old_map_short_codes.csv
mark-velez: @brendagutman

PR #1025, opened 2021-11-17
[Draft - please do not review yet] [DC-1899] [CircleCI] Run a nightly job to remove stale test buckets

hiro-mishima:

PR #1038, opened 2021-12-01
[DC-1901] [CircleCI] Run a nightly job to remove stale test environment datasets

hiro-mishima:
This is for removing stale **datasets**. Removing stale **buckets** is done in (https://github.com/all-of-us/curation/pull/1025).


data_steward/tools/delete_stale_test_datasets.py
```diff
@@ -0,0 +1,88 @@
...
+    """
+    Check if the project is set to test.
+
+    :param bq_client: Client
+    :return: None if the project is set properly
+    :raise: ValueError if project is not 'aou-res-curation-test'
+    """
+    test_project = 'aou-res-curation-test'
+
+    if bq_client.project != test_project:
```
nishanthpp93, line 24: After today's request from security about identifiers in Github, I wonder if we should be scrubbing instances of full project names from the project.
```suggestion
    if 'test' not in bq_client.project:
```

data_steward/tools/delete_stale_test_datasets.py
```diff
@@ -0,0 +1,88 @@
...
+
+    for dataset_name in dataset_names:
+
+        if first_n and n >= first_n:
+            break
+
+        dataset_created = bq_client.get_dataset(dataset_name).created
+
+        if (now - dataset_created).days <= 90:
+            continue
```
nishanthpp93, line 58: I know the ticket mentions this and you are doing as it directs, but there are some things that we might have missed while grooming it. I think creation time may not be the most accurate way to identify stale datasets. I have some test datasets that I use occasionally but they were created a long time ago and I delete any tables after using them. It would be a false positive under these criteria. Let's find a time to discuss this tomorrow.

tests/unit_tests/data_steward/tools/delete_stale_test_datasets_test.py
```diff
@@ -0,0 +1,129 @@
...
+    def test_check_project_error(self, mock_bq):
+        mock_bq.get_client.project = 'aou-res-curation-test-wrong-name'
+
+        with self.assertRaises(ValueError):
+            delete_stale_test_datasets._check_project(mock_bq.get_client)
+
+    @patch('tools.delete_stale_test_datasets.bq')
+    def test_filter_stale_datasets_all_not_empty(self, mock_bq):
+        """Test case: All datasets are NOT empty.
+        """
```
nishanthpp93, line 75: I appreciate the descriptive tests!

nishanthpp93 (changes requested):
Great implementation of what is expected in the ticket! I have some concerns I'd like to discuss more about. I would also like to test deploy this in circle to see how it runs. Added some suggestions as well. Let's discuss tomorrow.

data_steward/tools/delete_stale_test_datasets.py
hiro-mishima: Thanks for your review, @nishanthpp93 ! Tomorrow I will be offline here and there, so let me set up a meeting on Monday. Regarding your concern on this, does it also apply to stale buckets? The PR for stale buckets is this one: 
https://github.com/all-of-us/curation/pull/1025

nishanthpp93: I'm not sure, let's discuss in the meeting.

data_steward/tools/delete_stale_test_datasets.py
```diff
@@ -0,0 +1,103 @@
...
+    pipeline_logging.configure(logging.INFO, add_console_handler=True)
+
+    bq_client = bq.get_client(os.environ.get('GOOGLE_CLOUD_PROJECT'))
+
+    _check_project(bq_client)
+
+    stale_datasets = _filter_stale_datasets(bq_client, first_n=100)
+
+    for stale_dataset in stale_datasets:
+        LOGGER.info(
```
hiro-mishima, line 93: `INFO` is enough for creating output to the console.

image

https://app.circleci.com/pipelines/github/all-of-us/curation/4064/workflows/60fe0c49-37ad-4e39-a3c4-d4154e877f2c/jobs/12135

data_steward/tools/delete_stale_test_datasets.py
```diff
@@ -0,0 +1,103 @@
...
+        LOGGER.critical(
+            f"{dataset_name}, {dir(bq_client.list_tables(dataset_name))}")
+        LOGGER.critical(
+            f"{dataset_name}, {bq_client.list_tables(dataset_name).pages}")
+        LOGGER.critical(
+            f"{dataset_name}, {dir(bq_client.list_tables(dataset_name).item_to_value)}"
+        )
+        #if next(bq_client.list_tables(dataset_name), None):
+        #    continue
+        if list(bq_client.list_tables(dataset_name)) == 0:
```
hiro-mishima, line 70: `bq_client.list_tables()` returns `HTTPIterator`. This is not an iterator. If used as iterator (used with `next()` for example), it throws `TypeError: 'HTTPIterator' object is not an iterator` with CircleCi execution. For some reason, when I run it locally, this error does not happen. Some settings in my environment seem to interpret `HTTPIterator` as an iterator.

To avoid `TypeError`, StackOverflow told me I can cast that to a list:
https://stackoverflow.com/questions/67427634/how-to-get-the-next-blob-in-python-google-cloud-storage-library-method-list-blob

The mentioned solution is used in the following script in our repository as well. 
https://github.com/all-of-us/curation/blob/ac9f38b2f4580ae806121dd929293159132c7d2a/data_steward/utils/gcs.py#L37

So, I am using `list(bq_client.list_tables())` as workaround here.

Related page in Google API doc:
https://googleapis.dev/python/google-api-core/latest/page_iterator.html

nishanthpp93 (changes requested):
Minor update for the new format. Integration test looks good! Thanks for updating it.

hiro-mishima:
@nishanthpp93 
Because of the new configuration, the integration tests' dataset naming rule changed for my execution.

The datasets used to be created like:
`all_of_us_curation_hm_dc_1901_delete_stale_datasets_xyz`
But with this configuration, the datasets are named like:
`all_of_us_hiro_mishima_hm_dc_1901_delete_stale_datasets_xyz`

Since the dataset names got longer, the integration tests are failing for me now. As a workaround, I will rename my branch from `hm/dc_1901_delete_stale_datasets` to `hm/dc_1901`. It might force me to close this pull request and create a new pull request with the new branch. FYI!

PR #1039, opened 2021-12-01
[DC-1975] replaced google.cloud.storage with StorageClient

lledezma:
replaced google.cloud.storage with StorageClient, removed _empty_bucket() helper method and replaced it with the new empty_bucket() helper method, and replaced string concatenation with f-string.

### DIFF
--- a/tests/integration_tests/data_steward/retraction/retract_data_gcs_test.py
+++ b/tests/integration_tests/data_steward/retraction/retract_data_gcs_test.py
@@ -20,12 +20,13 @@ class RetractDataGcsTest(TestCase):
         print('**************************************************************')
 
     def setUp(self):
+        self.project_id = app_identity.get_application_id()
         self.hpo_id = test_util.FAKE_HPO_ID
         self.bucket = os.environ.get(f'BUCKET_NAME_FAKE')
         self.site_bucket = 'test_bucket'
         self.folder_1 = '2019-01-01-v1/'
         self.folder_2 = '2019-02-02-v2/'
-        self.client = StorageClient()
+        self.client = StorageClient(self.project_id)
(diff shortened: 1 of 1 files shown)


tests/integration_tests/data_steward/retraction/retract_data_gcs_test.py
```diff
@@ -22,26 +20,21 @@ def setUpClass(cls):
...
-        self.project_id = app_identity.get_application_id()
         self.hpo_id = test_util.FAKE_HPO_ID
         self.bucket = os.environ.get(f'BUCKET_NAME_FAKE')
         self.site_bucket = 'test_bucket'
         self.folder_1 = '2019-01-01-v1/'
         self.folder_2 = '2019-02-02-v2/'
-        self.client = storage.Client(self.project_id)
-        self.folder_prefix_1 = self.hpo_id + '/' + self.site_bucket + '/' + self.folder_1
-        self.folder_prefix_2 = self.hpo_id + '/' + self.site_bucket + '/' + self.folder_2
+        self.client = StorageClient()
```
MichaelSchmidt82, line 28: The original call to `storage.Client(self.project_id)` specified the project.  [see here](https://googleapis.dev/python/storage/latest/client.html).  And since `StorageClient()` inherits from storage.Client(), StorageClient() needs to be updated to target this project_id.  Therefore, StorageClient(self.project_id) is the correct way to initialize the object

MichaelSchmidt82 (changes requested):
See comment

tests/integration_tests/data_steward/retraction/retract_data_gcs_test.py
lledezma: I thought the project_id was pulled from the environment automatically with the new StorageClient(), but thanks for the explanation. I have submitted the correct changes.

MichaelSchmidt82 (approved):

PR #1040, opened 2021-12-02
[DC-2006] `ehr_union_test.py` helper function `_load_datasets()` does not load from RDR_PATH

ratuagga:
Bug fix for ehr_union integration test.


nishanthpp93 (changes requested):
There are some conflicts that need to be resolved. Try to rebase onto the new develop to resolve them. It should be straightforward since your changes are minor. Let me know if you run into any issues.

nishanthpp93 (approved):
Looks good!

PR #1041, opened 2021-12-02
[DC-1969] Update tests.integration_tests.bq_utils_tests to use gcloud.gcs.StorageClient()

lledezma:
The following functions have been updated to utilize the new StorageClient and its functionality.
test_load_csv()
test_load_cdm_csv()
test_query_result()
test_load_ehr_observation()

### DIFF
--- a/tests/integration_tests/data_steward/bq_utils_test.py
+++ b/tests/integration_tests/data_steward/bq_utils_test.py
@@ -9,6 +9,7 @@ import bq_utils
 import common
 import app_identity
 import gcs_utils
+from gcloud.gcs import StorageClient
 import resources
 from tests import test_util
 from tests.test_util import (FAKE_HPO_ID, FIVE_PERSONS_PERSON_CSV,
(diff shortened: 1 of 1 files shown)


MichaelSchmidt82 (changes requested):
There is a blob.upload_from_filename, which is a convenient method so that you don't need to pass in a file-like object.  [See here](https://googleapis.dev/python/storage/latest/blobs.html#google.cloud.storage.blob.Blob.upload_from_filename)

you can remove the `with` and statement `fp` variable.  It'll have better exception handling I would imagine as well.  you would just use `blob.upload_from_filename(csv_file_name)`

I'm marking this a request change only because I think it improves readability.

MichaelSchmidt82:
after review, using the `with open('some_file', 'rb') as pf:` is the better approach.

I think its worthwhile to move some of the SC client code out of the `with` statement and only use the core fp upload_from_file(fp) as keeping this code under `with` can be misleading due Python scoping.

MichaelSchmidt82 (approved):

PR #1042, opened 2021-12-03
[DC-1971] update combine_ehr_rdr_test.py to use gcloud.gcs.StorageClient()

lledezma:
Update the following method:
@staticmethod
def _upload_file_to_bucket(bucket, dataset_id, path, table):
and replace any calls to test_util.empty_bucket()
to use gcloud.gcs.StorageClient()

### DIFF
--- a/tests/integration_tests/data_steward/tools/combine_ehr_rdr_test.py
+++ b/tests/integration_tests/data_steward/tools/combine_ehr_rdr_test.py
@@ -45,7 +45,8 @@ class CombineEhrRdrTest(unittest.TestCase):
     @staticmethod
     def load_dataset_from_files(dataset_id, path, mappings=False):
         bucket = gcs_utils.get_hpo_bucket(test_util.FAKE_HPO_ID)
-        test_util.empty_bucket(bucket)
+        storage_client = StorageClient()
+        storage_client.empty_bucket(bucket)
         job_ids = []
         for table in resources.CDM_TABLES:
             job_ids.append(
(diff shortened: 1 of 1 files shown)


tests/integration_tests/data_steward/tools/combine_ehr_rdr_test.py
```diff
@@ -45,7 +45,8 @@ def setUpClass(cls):
     @staticmethod
     def load_dataset_from_files(dataset_id, path, mappings=False):
         bucket = gcs_utils.get_hpo_bucket(test_util.FAKE_HPO_ID)
-        test_util.empty_bucket(bucket)
+        storage_client = StorageClient()
```
ksdkalluri, line 48: hey @lledezma, quick question is there any reason for creating StorageClient() multiple times in each test case? Ideally, it can be created once in the setUpClass() and can be reused in all the test cases right.

ksdkalluri (changes requested):

tests/integration_tests/data_steward/tools/combine_ehr_rdr_test.py
lledezma: I decided to create one for each because they are static methods.

ksdkalluri: Even though they are static methods when they were called all the class methods will be instantiated and as this static method is inside the class and inherits all the class properties. Right now it will be fine as this class has only 2 methods where we are using the Storage Client, but as the number of static methods(test cases) increases the test running time increases as it tries to create a storage client each and every time.

lledezma: I see. So, do you suggest creating a StorageClient once in the setUpClass() or setUp(self) ?

lledezma: Ok. I will leave it as is. If more static methods are added to the class, then necessary updates will be made to only call StorageClient once. As you mentioned, right now it is fine as this class only has two methods.

ksdkalluri (approved):
Looks good!

PR #1043, opened 2021-12-07
[DC-2028] Add the secondcontactstate code mappings

brendagutman:


hiro-mishima (approved):
Since there is no integration test available for this part of the process yet, Krishna, Brenda, and I gathered and manually confirmed that this updated CSV file is working as expected:
https://precisionmedicineinitiative.atlassian.net/browse/DC-2028?focusedCommentId=104125

Thanks, Brenda and Krishna!

brendagutman:
After reviewing the concepts. They do not need to be added to old_map_short_codes

PR #1045, opened 2021-12-08
[DC-2022] Parallelize integration tests in Circle CI

nishanthpp93:
* [DC-2022] Update parallelism to 4
* [DC-2022] Add circle node index identifier

[DC-2022]: https://precisionmedicineinitiative.atlassian.net/browse/DC-2022?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[DC-2022]: https://precisionmedicineinitiative.atlassian.net/browse/DC-2022?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


curation.sh
```diff
@@ -49,6 +49,14 @@ dkr_run_args=(
   "$(pwd)/.circleci:/home/curation/project/curation/.circleci:z"
 )
 
+# If running specific tests using env var
+if [[ -n "${CURATION_TESTS_FILEPATH}" && -s "${CURATION_TESTS_FILEPATH}" ]]; then
+  dkr_run_args+=("-v")
+  dkr_run_args+=("$(realpath "${CURATION_TESTS_FILEPATH}"):/tests-to-run.txt:z")
```
dcarbone, line 55: This can be `:ro`, as we're not going to be editing this file :)

data_steward/cdr_cleaner/cleaning_rules/negative_ages.py
```diff
@@ -267,7 +267,7 @@ def validate_rule(self, client):
         raise NotImplementedError("Please fix me.")
 
     def get_sandbox_tablenames(self):
-        pass
+        return [self.sandbox_table_for(table) for table in self.affected_tables]
```
dcarbone, line 270: 👍

data_steward/tools/fitbit/generate_fitbit_dataset.py
```diff
@@ -80,7 +80,7 @@ def cast_to_schema_type(field, schema_type):
     if schema_type.lower() not in bq_int_float:
         col = f'SAFE_CAST({field} AS {schema_type.upper()}) AS {field}'
     else:
-        col = f'SAFE_CAST({field} AS {bq_int_float[schema_type]}) AS {field}'
+        col = f'SAFE_CAST({field} AS {bq_int_float[schema_type.lower()]}) AS {field}'
```
dcarbone, line 83: 👍

docker/develop/Dockerfile
```diff
@@ -66,6 +66,9 @@ ENV CIRCLE_PROJECT_NAME "all-of-us"
 ENV CIRCLE_WORKING_DIRECTORY "${CURATION_HOME}/project/curation"
 ENV CIRCLE_ARTIFACTS "/tmp/circleci-artifacts"
 
+# Enables parallelism in Circle CI but can be 0 if running locally
+ENV CIRCLE_NODE_INDEX "${CIRCLE_NODE_INDEX:-0}"
```
dcarbone, line 70: Setting this envvar in here isn't functionally useful as it's also set in `define_common_envvars`.  We could also use that to our advantage, as described below.

docker/develop/scripts/funcs.sh
```diff
@@ -201,9 +201,12 @@ function define_common_envvars() {
   set_env "GOOGLE_CLOUD_PROJECT_ID" "${APPLICATION_ID}"
   set_env "PROJECT_PREFIX" "${APPLICATION_ID//-/_}"
 
+  # Parallelism env var
+  set_env "CIRCLE_NODE_INDEX" "${CIRCLE_NODE_INDEX}" 0
```
dcarbone, line 205: One thing to consider here is perhaps updating this statement to be this:

`set_env "CIRCLE_NODE_INDEX" "local"`

This would effectively:

1. Ensure that the value is always set to _something_
2. If already defined, leave alone
2. If not already defined, set the value to `"local"`

This could be used to easily determine if a given set of datasets / buckets was created from a ci pipeline run or from being run locally.

tests/run_tests.sh
```diff
@@ -42,21 +42,27 @@ fi
...
+  test_arg=""
 fi
 
 if [[ -z ${substring} ]]
 then
-  cmd="tests/runner.py --test-path ${path} ${sdk_dir} --coverage-file ${coverage_file}"
+  cmd="tests/runner.py --test-dir ${path}  ${coverage_arg} ${test_arg}"
 else
-  cmd="tests/runner.py --test-path ${path} ${sdk_dir} --test-pattern $substring --coverage-file ${coverage_file}"
+  cmd="tests/runner.py --test-dir ${path} --test-pattern $substring ${coverage_arg} ${test_arg}"
```
dcarbone, line 65: One thing I'd like to see here is replacing this with a similar setup to what is done in the `curation.sh` script, namely building an array of positional options to pass to the command, rather than building an elongated string.

There are several reasons for this, not the least of which is it's easier to see the logic and you end up with a cleaner / less error prone output.

It would also allow you to more easily print a debug statement akin to `echo "Executing: tests/runner.py ${runner_args[*]}"`

dcarbone (changes requested):
@nishanthpp93 Overall this is fantastic!  Will have a huge benefit to the team :)

I have a few small changes that I'd like to see, however none of them are critical.  If you aren't able to get them done today, we can simply merge in as-is and you can work with the team to address things later.

docker/develop/scripts/funcs.sh
nishanthpp93: Thanks, this will definitely be a useful feature.

tests/run_tests.sh
nishanthpp93: I did a quick and dirty update for this, but for more enhancements, we might need a different ticket.

dcarbone:
Great work on this, @nishanthpp93!

PR #1046, opened 2021-12-09
[DC-2017] Migrate gcs_utils.get_hpo_bucket(hpo_id) to StorageClient()

lledezma:
This function may best be served as a method in StorageClient().  In every case, it relates to a bucket, and bucket functionality is now in StorageClient().  Having this method defined in that class may prove to increase readability and other human factors.  The change is also subtle.  This task does not include removing gcs_utils.get_hpo_bucket(...) but instead helps the GCS migration.  If accepted, a new ticket will target replacing existing calls with this new method.


MichaelSchmidt82 (approved):
I tested it and it looks good to me.

PR #1047, opened 2021-12-09
[DC-2032] Adding basic function debug command

dcarbone:


nishanthpp93 (approved):
Thanks for demonstrating the example. Tested and works fine.

PR #1050, opened 2021-12-10
[DC-1239] [Build - Validation] Perform DRC Address Comparisons

hiro-mishima:


jp3477 (approved):
This query generation method is ingenious! I wish there was an easier way to do the string replacements, but this does exactly what it's supposed to do.

PR #1051, opened 2021-12-10
[DC-1982] Remove and update calls to test_utils.empty_bucket() in tests

lledezma:
Currently, integration test files achilles_heel_test.py, achilles_test.py, ehr_union_test.py, and main_test.py include calls to test_utils.py to empty buckets or have self/re-defined helpers.  Remove and replace calls using test_util.py and helpers with ones to the new StorageClient() instead.


MichaelSchmidt82 (changes requested):
There are still references to `_empty_bucket(...)` with similar functionality defined and used in bq_utils_test.py and gcs_utils_test.py.

MichaelSchmidt82 (approved):
looks good then!

PR #1052, opened 2021-12-13
[DC-2020] secret manager

lrwb-aou:
* initial commit creates the secret manager extension
* secondary commits use the secret manager to obtain and use the mandrill token

### DIFF
--- a/data_steward/constants/validation/email_notification.py
+++ b/data_steward/constants/validation/email_notification.py
@@ -1,5 +1,5 @@
-MANDRILL_SECRET_ID = 'mandrill_token'
-UNSET_MANDRILL_SECRET_MSG = f"Mandrill secret: `{MANDRILL_SECRET_ID}` is not set in secret manager"
+MANDRILL_TOKEN_SECRET_ID = 'mandrill_token'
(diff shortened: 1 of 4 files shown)


data_steward/constants/validation/email_notification.py
```diff
@@ -1,5 +1,5 @@
-MANDRILL_API_KEY = 'MANDRILL_API_KEY'
-UNSET_MANDRILL_API_KEY_MSG = f"Mandrill API key not set in environment variable {MANDRILL_API_KEY}"
+MANDRILL_TOKEN = 'mandrill_token'
```
dcarbone, line 1: Imo lets rename this to something that indicates its the name of a secret.  Maybe something like `MANDRILL_TOKEN_SECRET_ID` or something.  Not

dcarbone (changes requested):
My only request is that the constant be renamed to be more obvious in-name as to what it's value is, but other than that lgtm.

data_steward/constants/validation/email_notification.py
lrwb-aou: @dcarbone How about now?

dcarbone (approved):
lgtm!

dcarbone:
The codacy warnings are red herrings and can be ignored.

PR #1053, opened 2021-12-13
[DC-2019] Migrate gcs_utils.get_drc_bucket() to StorageClient()

lledezma:
This function may best be served as a method in StorageClient().  In every case, it relates to a bucket, and bucket functionality is now in StorageClient().  Having this method defined in that class may prove to increase readability and other human factors.  The change is also subtle.  This task does not include removing gcs_utils.get_drc_bucket(...) but instead helps the GCS migration.  If accepted, a new ticket will target replacing existing calls with this new method.

### DIFF
--- a/data_steward/gcloud/gcs/__init__.py
+++ b/data_steward/gcloud/gcs/__init__.py
@@ -18,6 +18,10 @@ class StorageClient(Client):
     See https://googleapis.dev/python/storage/latest/client.html
     """
 
+    def get_drc_bucket(self) -> str:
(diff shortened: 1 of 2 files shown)


MichaelSchmidt82 (changes requested):
you likely need to merge in upstream changes as there is a conflict.

MichaelSchmidt82 (approved):
looks good to me.  You don't necessarily have to copy and past a method in -- there's nothing stopping you to use `return os.environ.get('DRC_BUCKET_NAME')` instead of using `result`

PR #1055, opened 2021-12-15
[DC-2043] Ensure envvars always defined

dcarbone:
- Removing "00_init_env.sh" in favor of "define_common_envvars" func
- Adding (hopefully useful) comments to each func in "func.sh"
- Misc cleanup


docker/develop/entrypoint.sh
```diff
@@ -1,7 +1,11 @@
 #!/usr/bin/env bash
 
+# load functions
 source "${CURATION_SCRIPTS_DIR}"/funcs.sh
 
+# DC-2048: Always load common envvars
```
mark-velez, line 6:
```suggestion
# DC-2043: Always load common envvars
```

docker/develop/scripts/funcs.sh
```diff
@@ -51,52 +134,157 @@ function define_username() {
...
     echo "Writing decoded CircleCI GCP key to ${GOOGLE_APPLICATION_CREDENTIALS}"
 
+    # decrypt credentials, then write the decrypted contents to the filepath
+    # specified by $GOOGLE_APPLICATION_CREDENTIALS
     echo "${GCLOUD_CREDENTIALS}" |
       openssl enc -d -aes-256-cbc -base64 -A -md md5 -k "${GCLOUD_CREDENTIALS_KEY}" \
         -out "${GOOGLE_APPLICATION_CREDENTIALS}"
   fi
 
+  # DC-2048: ensure project is set for the runtime config (defaults to "default")
```
mark-velez, line 266:
```suggestion
  # DC-2043: ensure project is set for the runtime config (defaults to "default")
```

mark-velez (approved):
This functionally looks good; the gsutil cmd Works On My Machine as do the other cmds. The only feedback close to "blocking" is the ticket is DC-2043 not DC-2048. I couldn't find a way to change JIRAs issue key to match DC-2048 so I suggested where to reword in the code below.

We can go over this later, but I couldn't figure out how to set the env vars for the `docker run`  command line. I don't think this is related to this PR. Could be I'm just doing something wrong.

:+1: comments are great

PR #1056, opened 2021-12-16
[DRAFT!] [DC-2037] Implement gcs_utils.list_bucket(...) functionality in gcloud.gcs.StorageClient

MichaelSchmidt82:
Implement DC-2036 in gcloud.gcs.StorageClient and write a test.  Name this method get_items_metadata(...) instead.  get_items_metadata(…) should take a google.cloud.storage.bucket.Bucket as one of its arguments.

### DIFF
--- a/data_steward/gcloud/gcs/__init__.py
+++ b/data_steward/gcloud/gcs/__init__.py
@@ -22,7 +22,7 @@ class StorageClient(Client):
     See https://googleapis.dev/python/storage/latest/client.html
     """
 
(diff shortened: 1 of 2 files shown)


MichaelSchmidt82:
Integration test will depend `[DC-2035] Implement gcs_utils.upload_object(...) functionality in gcloud.gcs.StorageClient`

MichaelSchmidt82:
This is not a good solution

PR #1059, opened 2021-12-17
[DC-1947] Add Tablenamer to deid_clean

jp3477:


ksdkalluri (approved):
Looks good! I think we might have to modify create_tier to pass table_namer to rules for it to actually reflect in table names.

PR #1061, opened 2021-12-20
[DC-2049] Nightly delete_stale_test_datasets error "drc_ops is still in use"

hiro-mishima:


data_steward/tools/delete_stale_test_datasets.py
```diff
@@ -97,8 +100,13 @@ def main(first_n):
...
 
     for stale_dataset in datasets_to_delete:
+
         LOGGER.info(f"Running - bq_client.delete_dataset({stale_dataset})")
-        bq_client.delete_dataset(stale_dataset)
+
+        try:
+            bq_client.delete_dataset(stale_dataset)
+        except exceptions.BadRequest as e:
+            LOGGER.warning(f"Failed to delete {stale_dataset}. Message: {e}")
```
nishanthpp93, line 109: I think `e` may not print the message. Maybe `e.message`?
```suggestion
            LOGGER.warning(f"Failed to delete {stale_dataset}. Message: {e.message}")
```

nishanthpp93 (commented):
One small thing to check, but looks good otherwise.

nishanthpp93 (approved):
I just tested it out and `e` seems to work the same as `e.message`, so it may be how google has defined their exception's `__str__`. Sorry for the trouble, but thanks for updating it!

hiro-mishima:
> I just tested it out and `e` seems to work the same as `e.message`, so it may be how google has defined their exception's `__str__`. Sorry for the trouble, but thanks for updating it!

I did not know we can use `e.message` in this context. This way the statement is more explicit and easier to understand. Thanks for the advice! :)

PR #1062, opened 2021-12-20
[DC-2050] Remove EHR slack alerts configuration

lrwb-aou:
* removes the wrapper function from the ehr_union job
* removes the wrapper function import
* removes the wrapper function defintion, as it was only used by the ehr_union job.


nishanthpp93 (approved):
Looks good!

lrwb-aou:
@nishanthpp93 thanks!

PR #1064, opened 2021-12-20
[DC-2051] removing slack messaging for exceptions

lrwb-aou:
* GCP monitoring is configured to monitor for warnings or higher in the application
* should remove redundant messaging from the application now.
* update the unit tests to expect alert removal


lrwb-aou:
This one also failed with what appears to be a keyword restriction on google buckets.  This branch is closed in favor of the pr at https://github.com/all-of-us/curation/pull/1065.

PR #1066, opened 2021-12-20
[DC-2051] removing slack messaging for exceptions

lrwb-aou:
* GCP monitoring is configured to monitor for warnings or higher in the application
* should remove redundant messaging from the application now.
* update the unit tests to expect alert removal

### DIFF
--- a/data_steward/validation/app_errors.py
+++ b/data_steward/validation/app_errors.py
@@ -13,7 +13,6 @@ import flask
 from googleapiclient.errors import HttpError
 
 # Project imports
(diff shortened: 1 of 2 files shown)


MichaelSchmidt82:
- I see that the code which catches the exception and uses post_message(...) is removed.
- The unused import is removed
- The only test which uses a mock_post_message was taken out (and it's usage)

I don't see any reason to request any changes.

MichaelSchmidt82 (approved):

PR #1067, opened 2021-12-21
[DC-1948] Add `table_namer` to `*deid_base` rules

ksdkalluri:


jp3477 (approved):
This looks good! If you want, you can trigger the integration tests again because I think they were skipped. But it should probably be fine.

PR #1068, opened 2021-12-21
[DC-2037] Implement gcs_utils.list_bucket(...) functionality in gcloud.gcs.StorageClient

MichaelSchmidt82:
This PR defines a suitable replacement for gcs_utils.list_bucket(str).  It contains a helper function to return blob metadata.  It also stages (but does not implement) work for three future tickets:

- Replace the return value from a string to a google.cloud.storage.bucket.Bucket as discussed [here](https://precisionmedicineinitiative.atlassian.net/browse/DC-2017?focusedCommentId=104032) and [here](https://precisionmedicineinitiative.atlassian.net/browse/DC-2019?focusedCommentId=104034).
- A future ticket to replace [this code](https://github.com/all-of-us/curation/blob/1863aa81bb3f51d03271eef5a4f12bc120b9475c/data_steward/validation/main.py#L174) with the "helper" function.

To reiterate, this code simply _stages_ this work -- future tickets will implement this work in a controlled manner.

Finally, tests were created for both the helper and list_items_metadata methods -- some test code was consolidated during this development.


data_steward/gcloud/gcs/__init__.py
```diff
@@ -18,6 +20,40 @@ class StorageClient(Client):
...
+        blobs: list = list(self.list_blobs(bucket))
+        metadata: list = [self.get_blob_metadata(blob) for blob in blobs]
+        return metadata
+
+    def get_blob_metadata(self, blob: Blob) -> dict:
+        """
+        Gather and ship a blob's metadata in dictionary form.
+        Note: Date and times are in the python stl format.  They are datetime objects.
+        """
+        metadata: dict = {
```
nishanthpp93, line 40: Do you think this should have a call to the client if it received a blob object that is not instantiated with the HTTP request?
```suggestion
        if blob.id == None:
            blob = self.client.get_bucket(blob.bucket.name).get_blob(blob.name)
        metadata: dict = {
```

tests/integration_tests/data_steward/gcloud/gcs_client_test.py
```diff
@@ -25,36 +25,68 @@ def setUp(self):
         self.prefix: str = 'prefix'
         self.data: bytes = b'bytes'
 
+        # NOTE: this needs to be in sorted order
```
nishanthpp93, line 28: Curious about the ordering requirement.

nishanthpp93 (commented):
Overall looks great! Added a couple questions we can discuss tomorrow over a call.

data_steward/gcloud/gcs/__init__.py
MichaelSchmidt82: Will implement

tests/integration_tests/data_steward/gcloud/gcs_client_test.py
MichaelSchmidt82: found better solution

tests/integration_tests/data_steward/gcloud/gcs_client_test.py
```diff
@@ -25,36 +25,77 @@ def setUp(self):
...
-        self.assertEqual(len(self.sub_prefixes), len(items))
-        for item in items:
-            self.assertIn(item[:-1], self.sub_prefixes)
+        # Check same number of elements
+        self.assertEqual(len(self.sub_prefixes), len(sorted_items))
 
-        self.client.empty_bucket(self.bucket_name)
+        # Check same prefix
+        for index, item in enumerate(sorted_items):
+            self.assertEqual(item[:-1], self.sub_prefixes[index])
```
nishanthpp93, line 91: Do you think this sorting can be done away with using the same `assertCountEqual` check in the other test?

nishanthpp93 (commented):
Looks good, thanks for making the changes! I have one last question. Let me know your thoughts.

tests/integration_tests/data_steward/gcloud/gcs_client_test.py
MichaelSchmidt82: yes.  It can.  Removing and pushing up.  Good catch as usual.

nishanthpp93 (approved):
Looks good!

PR #1069, opened 2021-12-22
[DC-2052] Replace gcs_utils.upload_object() calls with StorageClient.upload_from_file()

lledezma:
As part of the new Google Client migration, the Google client already provides a method to upload files, upload_from_file(). This method provided by the Google client provides the same functionality that  `gcs_utils.upload_object()` provides, and therefore, no new helper method needs to be created. Only all `gcs_utils.upload_object()` calls need to be replaced with StorageClient.upload_from_file().


lledezma:
Commits are no longer needed

PR #1070, opened 2021-12-22
[DC-2053] Remove last remaing empty_bucket() calls found in tests

MichaelSchmidt82:
DC-2053 targets the last remaining "empty_bucket(...)" calls outside of `gcs_utils`.  In fact, before this PR, there were no calls to gcs_utils at all -- the function was already deleted.  However,`test_util.empty_bucket(...)` remained, and some hidden functions still existed.  This PR removes the last of empty_bucket(...) calls outside of `StorageClient.empty_bucket(...)`.  Hooray 🎉.

Diligence is required here because testing their replacements is tricky at best; they were used in tests themselves.  So care was taken to ensure their replacement matched the existing call.  The `StorageClient.empty_bucket(...)` (the replacement) is already used in tests, and has been thoroughly unit- and integration-test tested.

### DIFF
--- a/tests/integration_tests/data_steward/bq_utils_test.py
+++ b/tests/integration_tests/data_steward/bq_utils_test.py
@@ -77,12 +77,8 @@ class BqUtilsTest(unittest.TestCase):
             }
         ]
         self.DT_FORMAT = '%Y-%m-%d %H:%M:%S'
-        self._empty_bucket()
-
-    def _empty_bucket(self):
-        bucket_items = gcs_utils.list_bucket(self.hpo_bucket)
-        for bucket_item in bucket_items:
-            gcs_utils.delete_object(self.hpo_bucket, bucket_item['name'])
+        self.client = StorageClient()
+        self.client.empty_bucket(self.hpo_bucket)
 
     def _drop_tables(self):
         tables = bq_utils.list_tables()
@@ -378,4 +374,4 @@ class BqUtilsTest(unittest.TestCase):
 
     def tearDown(self):
         test_util.delete_all_tables(self.dataset_id)
-        self._empty_bucket()
+        self.client.empty_bucket(self.hpo_bucket)
--- a/tests/integration_tests/data_steward/bq_utils_test.py
+++ b/tests/integration_tests/data_steward/bq_utils_test.py
@@ -103,8 +103,7 @@ class BqUtilsTest(unittest.TestCase):
         csv_file_name = table_name + '.csv'
         local_csv_path = os.path.join(test_util.TEST_DATA_EXPORT_PATH,
                                       csv_file_name)
-        sc = StorageClient()
-        sc_bucket = sc.get_bucket(self.hpo_bucket)
+        sc_bucket = self.client.get_bucket(self.hpo_bucket)
         bucket_blob = sc_bucket.blob(csv_file_name)
         with open(local_csv_path, 'rb') as fp:
             bucket_blob.upload_from_file(fp)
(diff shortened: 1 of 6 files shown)


tests/integration_tests/data_steward/gcs_utils_test.py
```diff
@@ -2,6 +2,7 @@
 from io import open
 
 from googleapiclient.errors import HttpError
+from gcloud.gcs import StorageClient
 
 import gcs_utils
```
lrwb-aou, line 7: Will the additional `gcs_utils` calls be refactored in a future ticket?  Or will this entire file go away when we have fully implemented google capabilities?

tests/integration_tests/data_steward/tools/top_heel_errors_test.py
```diff
@@ -66,7 +66,7 @@ def setUp(self):
         self.bucket: str = gcs_utils.get_drc_bucket()
```
lrwb-aou, line 66: Is changing this to 
```suggestion
        self.bucket: str = self.storage_client.get_drc_bucket()
```
out of scope for this PR?

lrwb-aou (commented):
looks good.  I had some small questions/suggestions.  But overall, very happy to soon see this stuff gone.

tests/integration_tests/data_steward/bq_utils_test.py
```diff
@@ -77,12 +77,8 @@ def setUp(self):
             }
         ]
         self.DT_FORMAT = '%Y-%m-%d %H:%M:%S'
-        self._empty_bucket()
-
-    def _empty_bucket(self):
-        bucket_items = gcs_utils.list_bucket(self.hpo_bucket)
-        for bucket_item in bucket_items:
-            gcs_utils.delete_object(self.hpo_bucket, bucket_item['name'])
+        self.client = StorageClient()
```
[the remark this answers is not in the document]
MichaelSchmidt82: If you're talking about what I think your talking about:

That `self.client = StorageClient()` mean while:

```
sc = StorageClient()
sc_bucket = sc.get_bucket(self.hpo_bucket)
```
later down.   Seems it happens 5ish times... Git blame says Luis in DC-1969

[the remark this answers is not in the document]
lrwb-aou: That is what I  am referencing.  Didn't know if you had plans to clean this up in a later PR or if you might do it as part of this one.

tests/integration_tests/data_steward/gcs_utils_test.py
MichaelSchmidt82: So I have actually kinda struggled with how this might look, because things are getting pretty close to being finished.

I think this is the idea:

- in StorageClient, make get_hpo_bucket and get_drc_bucket return actual buckets and not strings.  gcs_utils continues to return strings.
- the rest of the code at this point still is using gcs_utils as well.
- with StorageClient now returning buckets, we can target files that already contain an instance of StorageClient as these are likely to be easier to transition.
- the rest of these files are then transitioned.
- gcs_utils loses get_*_bucket(...) when all files are using Bucket objects.
- audit gcs_utils, transfer any last functionality which will probably result in removing the file altogether.

MichaelSchmidt82: I think that could be a 13 story pointer easily

tests/integration_tests/data_steward/tools/top_heel_errors_test.py
MichaelSchmidt82: Yes, because ultimately we don't want to use anything in the StorageClient that returns strings because of the plan above.  Return buckets, then migrate cases like these to buckets and not strings.

tests/integration_tests/data_steward/bq_utils_test.py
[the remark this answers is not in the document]
MichaelSchmidt82: I can just get the change in there.  Pretty straightforward.

tests/integration_tests/data_steward/tools/top_heel_errors_test.py
lrwb-aou: 👍

tests/integration_tests/data_steward/bq_utils_test.py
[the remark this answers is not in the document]
MichaelSchmidt82: I made the changes, confirm it with a diff.  - we can let the tests chug away since it'll only take ~25 minutes now.  Or if you feel comfortable, just review and accept the change.

lrwb-aou (approved):
all the `bq_utils` tests passed.  Looks good.

PR #1071, opened 2021-12-22
[DC-2054] remove slack notifications from admin api

lrwb-aou:
* remove ability to send messages from admin api to slack channel
* use f-strings to match rest of code base conventions
* update the unit test to expect log messages instead of slack post messages
* update the test runner script to run a test(s) based on a test filename pattern instead of a filepath.  (adds back a functionality that was removed by a previous commit.)

### DIFF
--- a/tests/run_tests.sh
+++ b/tests/run_tests.sh
@@ -12,7 +12,8 @@ export PYTHONPATH=:"${BASE_DIR}":"${BASE_DIR}/data_steward":"${BASE_DIR}/tests":
 function usage() {
   echo "Usage: run_test.sh " \
     "[unit|integration]" \
-    "[Set env CURATION_TESTS_FILEPATH for specific tests, usage in runner.py]" >&2
+    "[Set env CURATION_TESTS_FILEPATH for specific tests, usage in runner.py]" \
+    "[Set env TEST_PATTERN to run tests whose file names match the pattern, usage in runner.py]" >&2
   exit 1
 }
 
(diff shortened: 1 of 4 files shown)


tests/run_tests.sh
```diff
@@ -79,4 +81,11 @@ if [[ -n "${CURATION_TESTS_FILEPATH}" ]]; then
     "${CURATION_TESTS_FILEPATH}")
 fi
 
+if [[ -n "${TEST_PATTERN}" ]]; then
```
ksdkalluri, line 84: This I don't completely understand. is this for just adding a --test_pattern param?

ksdkalluri (approved):
I think this is good to merge.

Thanks for updating the log strings to use f strings.

tests/run_tests.sh
lrwb-aou: Yea, I like to be able to run a specific unit/integration test from the command line without being required to know the file path.  `TEST_PATTERN` lets me do that.  It's already built into the runner file and was part of the previous bash script.  It just got refactored away.  This functions just like the `CURATION_TESTS_FILEPATH` env variable.  I would prefer just to pass a command line parameter instead of setting an ENV var, but that would have been a more significant change than I wanted to make at this time.

lrwb-aou: I used it to run `TEST_PATTERN=admin_api* ./tests/run_tests.sh unit` so I could just run the one class.

PR #1072, opened 2021-12-23
[DC-2057] Replace gcs_utils.get_metadata() calls with StorageClient.get_blob_metadata()

lledezma:
gcs_utils.get_metadata() call in validation/main.py was replaced with StorageClient.get_blob_metadata(). gcs_utils.get_metadata() is used only once in the curation project.

### DIFF
--- a/data_steward/validation/main.py
+++ b/data_steward/validation/main.py
@@ -696,8 +696,10 @@ def perform_validation_on_file(file_name, found_file_names, hpo_id,
 
 
 def _validation_done(bucket, folder):
-    if gcs_utils.get_metadata(bucket=bucket,
-                              name=folder + common.PROCESSED_TXT) is not None:
+    storage_client = StorageClient()
+    sc_bucket = storage_client.get_bucket(bucket)
+    bucket_blob = sc_bucket.blob(f'{folder}{common.PROCESSED_TXT}')
+    if storage_client.get_blob_metadata(bucket_blob):
         return True
     return False
 
--- a/data_steward/validation/main.py
+++ b/data_steward/validation/main.py
@@ -17,6 +17,7 @@ from io import StringIO, open
 import dateutil
 from flask import Flask
 from googleapiclient.errors import HttpError
+from google.cloud.storage import Blob
 
 # Project imports
 import api_util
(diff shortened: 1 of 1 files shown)


data_steward/validation/main.py
```diff
@@ -696,8 +697,10 @@ def perform_validation_on_file(file_name, found_file_names, hpo_id,
 
 
 def _validation_done(bucket, folder):
-    if gcs_utils.get_metadata(bucket=bucket,
-                              name=folder + common.PROCESSED_TXT) is not None:
+    storage_client = StorageClient()
+    sc_bucket = storage_client.get_bucket(bucket)
```
nishanthpp93, line 701: I just wanted to mention that this is fine in this case, but instantiating a client is resource intensive. Ideally we'd want to pass in already created clients to the function or use their built-ins/added functionalities to achieve this. Maybe in the future a refactor would require the caller of this function to pass the client object which can be used here.

data_steward/validation/main.py
```diff
@@ -696,8 +697,10 @@ def perform_validation_on_file(file_name, found_file_names, hpo_id,
...
+    bucket_blob = sc_bucket.blob(f'{folder}{common.PROCESSED_TXT}')
```
nishanthpp93, line 702: This is fine here as well, but important to note that bucket.blob does not make an HTTP request ([see example here](https://cloud.google.com/storage/docs/viewing-editing-metadata#storage-view-object-metadata-python)). However, since it is only calling get_blob_metadata below (which has an added check to make an HTTP request if not already done), it is handled there. However, I'd suggest taking a step back and figuring out what is being achieved. It is looking for the creation of a specific file. Try writing a simple integration test to see if this really figures out if a blob exists or not. If not, what would be the best way to check if a blob exists or not via [their docs](https://googleapis.dev/python/storage/latest/blobs.html)?

data_steward/validation/main.py
```diff
@@ -696,8 +697,10 @@ def perform_validation_on_file(file_name, found_file_names, hpo_id,
...
+    if storage_client.get_blob_metadata(bucket_blob):
         return True
     return False
```
nishanthpp93, line 705: There is some redundancy here, since you are already checking if something is true or not and just passing the booleans. You could simply return the whole thing instead and achieve the same result. This existed in the old code, so I understand you were just following the existing convention. But I'd say look at the usages of _validation_done and see if it makes sense and replace it.
```suggestion
    return storage_client.get_blob_metadata(bucket_blob)
```

nishanthpp93 (commented):
Looks good but I added a few comments. Let me know your thoughts.

data_steward/validation/main.py
lledezma: Yes. I thought about that and decided to do it like this since it is just one method that is using it but will have only one client for all methods to call as more of them need to be modified in the future

data_steward/validation/main.py
lledezma: Yes. line 702 has to go because we are not creating a blob.

lledezma: Yes. We don't need the metadata because we just have to check that the blob exists.

nishanthpp93 (approved):
Looks good! I see that there's an integration test failing due to rate limits. Kick it off again, hopefully it'll fix itself and then merge.

PR #1073, opened 2021-12-23
[DC-2058] removing slack infrastructure

lrwb-aou:
* remove slack utility
* remove slack logging handler
* remove slack logging handler initialization


MichaelSchmidt82:
[this](https://github.com/all-of-us/curation/blob/e80447e56d269dc2c9c8bc79e78218d4b0dc504c/data_steward/validation/app_errors.py#L63) function still exists.  It's being used 5 times, so I'm assuming this is still relevant?

MichaelSchmidt82:
Two question.

- There only seems to be a few comments left behind (like [here](https://github.com/all-of-us/curation/blob/e80447e56d269dc2c9c8bc79e78218d4b0dc504c/data_steward/gcs_utils.py#L44)).  Do they still provide help?
- Can you comment on the use CircleCI's config which seems to be linked to slack?  Am I reading that correctly?

lrwb-aou:
> [this](https://github.com/all-of-us/curation/blob/e80447e56d269dc2c9c8bc79e78218d4b0dc504c/data_steward/validation/app_errors.py#L63) function still exists. It's being used 5 times, so I'm assuming this is still relevant?

right.  This function is related to app logging, and not to sending messages to slack.  Things may be better now, but there was a time when the logs were not grouped logically in stack driver and we had to set up a logging mechanism to do that.  This function is just handling potential error conditions to make sure it returns a generic view.  If we end up eventually moving the entire thing out of a web app and into cloud functions, these probably won't be necessary.

I do need to update the comment [here](https://github.com/all-of-us/curation/blob/e80447e56d269dc2c9c8bc79e78218d4b0dc504c/data_steward/validation/app_errors.py#L65).

lrwb-aou:
> Two question.
> 
>     * ~There only seems to be a few comments left behind (like [here](https://github.com/all-of-us/curation/blob/e80447e56d269dc2c9c8bc79e78218d4b0dc504c/data_steward/gcs_utils.py#L44)).  Do they still provide help?~  Looks like my comment above answers this.

I do want to update that comment.  By logging the message correctly, GCP's metrics monitoring and alerts will let us know this has happened.

>     * Can you comment on the use of CircleCI's config which seems to be linked to slack?  I'm unfamiliar with reading that config.

CircleCi's config is using the slack orb, provided by circleci.  All we have to do is provide a webhook to send messages from circleci.  We have provided that info via an environment variable in the circleci settings.  There is no GAE application code for this and it is not dependent on a token value to run.  All we use it for is to send a notification if a failure has occurred on the master and/or develop branches.

MichaelSchmidt82:
One last question.  Do the requirements need to be updated to not include slack?

lrwb-aou:
> One last question. Do the requirements.txt need to be updated to not include slack?

Yes, I think you're right.

MichaelSchmidt82:
> > One last question. Do the requirements.txt need to be updated to not include slack?
> 
> Yes, I think you're right.

Looks good. I can't see the requirement change affecting anything.

MichaelSchmidt82 (approved):

lrwb-aou:
> > > One last question. Do the requirements.txt need to be updated to not include slack?
> > 
> > 
> > Yes, I think you're right.
> 
> Looks good. I can't see the requirement change affecting anything.

I have worked on the dependencies a little more.  The slack client installed some dependencies that are unused by other packages.  I've tried to clean those out too.  Please take a look when you have time.

PR #1080, opened 2022-01-07
[DC-1949] Implements the `table_namer` parameter to `C*_deid` data stage cleaning rules.

ksdkalluri:
update *Ct_deid cleaning rules to accept table_namer parameter


jp3477 (changes requested):
Looks good for the most part. I added one request for a `get_sandbox_tablenames()` implementation.

I also think you're missing the `RemoveExtraTables` cleaning rule, so that should be added.

jp3477 (approved):
Thanks for the updates, looks good!

PR #1081, opened 2022-01-07
[DC-1854] Draft - please do not review yet

hiro-mishima:


hiro-mishima:
@ksdkalluri It looks like this issue is already solved by your following PR
https://github.com/all-of-us/curation/pull/1080/files
I think I should just close this PR without merging or anything. What do you think?

PR #1083, opened 2022-01-07
[DC-1935] Add the list of sandbox table names to the `GeneralizeCopeInsuranceAnswers` cleaning rule

hiro-mishima:
In addition to what the title says, I added sandbox creation SQL and test for `GeneralizeCopeInsuranceAnswers`, as discussed in the ticket comments: https://precisionmedicineinitiative.atlassian.net/browse/DC-1935?focusedCommentId=106158

### DIFF
--- a/data_steward/cdr_cleaner/cleaning_rules/deid/generalize_cope_insurance_answers.py
+++ b/data_steward/cdr_cleaner/cleaning_rules/deid/generalize_cope_insurance_answers.py
@@ -19,6 +19,15 @@ from utils import pipeline_logging
 
 LOGGER = logging.getLogger(__name__)
 
+SANDBOX_RECORDS_QUERY = JINJA_ENV.from_string("""
+CREATE OR REPLACE TABLE `{{project_id}}.{{sandbox_id}}.{{sandbox_table}}` AS (
+SELECT * FROM `{{project_id}}.{{dataset_id}}.observation`
+WHERE
+  observation_source_concept_id = 1332737
+  AND value_source_concept_id IN (1332904, 1333140)
+)
+""")
+
 ANSWER_GENERALIZATION_QUERY = JINJA_ENV.from_string("""
 UPDATE
   `{{project_id}}.{{dataset_id}}.observation`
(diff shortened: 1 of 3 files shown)


hiro-mishima:
@MichaelSchmidt82 Now this PR is ready for review! I solved the potential merge conflict.

data_steward/cdr_cleaner/cleaning_rules/deid/generalize_cope_insurance_answers.py
```diff
@@ -92,6 +104,14 @@ def get_query_specs(self, *args, **keyword_args):
...
             an ordering.
         """
+
+        sandbox_records_query = dict()
+        sandbox_records_query[cdr_consts.QUERY] = SANDBOX_RECORDS_QUERY.render(
+            project_id=self.project_id,
+            sandbox_id=self.sandbox_dataset_id,
+            sandbox_table=self.sandbox_table_for(OBSERVATION),
+            dataset_id=self.dataset_id)
+
```
MichaelSchmidt82, line 114: I thought about this section of code for a while.  At first, I had thought maybe something could be improved here, but there are no changes in syntax that will enhance readability.

data_steward/cdr_cleaner/cleaning_rules/deid/generalize_cope_insurance_answers.py
```diff
@@ -81,7 +90,10 @@ def __init__(self,
                          table_namer=table_namer)
 
     def get_sandbox_tablenames(self):
```
MichaelSchmidt82, line 92: For context, can you elaborate on why this function is defined, but appears never to be used in the project?

MichaelSchmidt82 (approved):

data_steward/cdr_cleaner/cleaning_rules/deid/generalize_cope_insurance_answers.py
hiro-mishima: `get_sandbox_tablenames()` was introduced in [DC-734](https://precisionmedicineinitiative.atlassian.net/browse/DC-734) as part of the base cleaning rule created in [DC-712](https://precisionmedicineinitiative.atlassian.net/browse/DC-712). The intention was to have a standardized way of listing the table names in the sandbox. However, in most of the cleaning rules, there are only one or zero sandbox tables. That caused this function to be not utilized in some cleaning rules (however it is defined in each of the cleaning rules because the base class has `get_sandbox_tablenames()` as the abstract method from DC-734). Also, there were already several cleaning rules before the base class was created by DC-712. For some of those rules, `get_sandbox_tablenames()` was added but just left to raise `NotImplementedError`.

hiro-mishima (commented):
Thanks for the review and the comments, @MichaelSchmidt82 ! I have added my response to one of your comments. Please let me know if you have any other questions.

PR #1084, opened 2022-01-07
[DC-1950] Add the optional table namer parameter to Registered Tier Deid cleaning rules

jp3477:


ksdkalluri (approved):
@jp3477 I did verify that both PR's of us doesn't conflict with the cleaning rules in both classes(RT and CT) and all the cleaning rules were covered. And these changes look good!

PR #1087, opened 2022-01-11
[DC-2003] Run participant validation as part of generate_metrics

nishanthpp93:
* [DC-2003] Move relevant scripts and modularize
* [DC-2003] Move .gitignore to root

[DC-2003]: https://precisionmedicineinitiative.atlassian.net/browse/DC-2003?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[DC-2003]: https://precisionmedicineinitiative.atlassian.net/browse/DC-2003?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


ksdkalluri (approved):
@nishanthpp93 This is mostly refactoring, I don't see any issues. I suggest we do a test run before merging this into develop branch.

jp3477 (approved):
@nishanthpp93 To my eye, this looks good to me as well! But I agree with @ksdkalluri that we should do some kind of test runs for the whole process to make sure it's all glued together.

nishanthpp93:
@ksdkalluri @jp3477 I did some more refactoring, please take a look.

data_steward/constants/validation/participants/validate.py
```diff
@@ -232,3 +430,57 @@
...
+            JOIN `{{project_id}}.{{ehr_ops_dataset_id}}.concept` cc
+                ON gender_concept_id = concept_id ) AS ehr_sex
+    ON ehr_sex.person_id = ps.person_id
+WHERE upd.person_id = ps.person_id
+    AND upd._PARTITIONTIME = ps._PARTITIONTIME
+""")
+
+SUMMARY_QUERY = JINJA_ENV.from_string("""
+SELECT COUNT(*) AS row_count
+FROM `{{project_id}}.{{drc_dataset_id}}.{{hpo_id}}`
```
jp3477, line 485: I don't think the variables in this query match with the expectations of the `get_participant_validation_summary_query` function. `drc_dataset_id` here would need to be changed to `dataset_id` and `hpo_id` would need to be changed to `id_match_table`. Am I interpreting this correctly?

jp3477 (dismissed):
Think I found something that can be changed. Will keep looking through.

Also, @nishanthpp93 did you find a good way to execute this in the test environment? I'd love to toy around with running it for validation.

data_steward/constants/validation/participants/validate.py
nishanthpp93: You are right, nice catch. I wasn't using this anywhere since it is too generic, but this needs to be changed in case we do plan to use it.

nishanthpp93:
@jp3477 Unfortunately I haven't. This is mostly refactoring so I'm just relying on the components already being tested.

data_steward/constants/validation/participants/validate.py
```diff
@@ -12,7 +13,204 @@
...
 phonetically similar strings.
+
+
 """
+# Python imports
+
+# Project imports
+from common import JINJA_ENV
+
+RDR_SEX = 'rdr_sex'
```
ksdkalluri, line 24: @nishanthpp93 nit: I think these constants can be moved to the constants directory.

data_steward/validation/main.py
```diff
@@ -364,6 +390,14 @@ def generate_metrics(hpo_id, bucket, folder_prefix, summary):
         report_data[report_consts.COMPLETENESS_REPORT_KEY] = query_rows(
             completeness_query)
 
+        # participant validation metrics
+        logging.info(f"Ensuring participant validation can be run for {hpo_id}")
+        duplicate_tables = check_duplicates_and_validate(hpo_id, report_data)
+        if not duplicate_tables:
```
ksdkalluri, line 396: @nishanthpp93 so the participant validation runs only when there are no duplicates. Can we add messaging about the issue in results.html and is there a decision on how we are going to approach this?

nishanthpp93: Yeah, that makes sense. I didn't want to add the new section in results.html until we know what metrics should go there. I think EHR Ops will come up with metrics to figure out what needs to be shown to sites. Right now, there will be a logging message printed if it was skipped.

data_steward/constants/validation/participants/validate.py
nishanthpp93: I just realized, this is in constants right now

data_steward/validation/main.py
nishanthpp93: I don't think the sites anticipate this yet. @jp3477 Should we start showing warnings in the results.html related to participant validation or hold off on it?

jp3477: @nishanthpp93 , we should hold off. We're still deciding how we would want to display it.

nishanthpp93:
@ksdkalluri @jp3477 I've made the changes and rebased with develop. Let me know if anything else is needed to get this in.

PR #1088, opened 2022-01-11
[DC-2029] Moved ps_api_values schema file

jp3477:
Moved the schema file for the ps_api_values tables from rdr folder to internal folder.


nishanthpp93 (commented):
Looks good. I think the lint check is failing because for single commit PRs, it expects the Jira tag in the commit message rather than the PR title (this is because GitHub switches them while merging). It will switch to checking the PR title if you make another commit.

nishanthpp93 (approved):

PR #1089, opened 2022-01-11
[DC-2016] Draft

hiro-mishima:
This is a draft - please do not review yet.

### DIFF
--- a/data_steward/cdr_cleaner/cleaning_rules/clean_mapping.py
+++ b/data_steward/cdr_cleaner/cleaning_rules/clean_mapping.py
@@ -68,7 +68,11 @@ class CleanMappingExtTables(BaseCleaningRule):
     that exist after the dataset has been fully cleaned.
     """
 
-    def __init__(self, project_id, dataset_id, sandbox_dataset_id):
+    def __init__(self,
+                 project_id,
+                 dataset_id,
+                 sandbox_dataset_id,
+                 table_namer=None):
         """
         Initialize the class with proper info.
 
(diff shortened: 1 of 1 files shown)


jp3477 (changes requested):
This looks good for the most part, but there is one more thing that should be added. The table name that is being sandboxed should take advantaged of the table_namer by wrapping it with the `self.sandbox_table_for()` function. For this cleaning rule, I believe this only involves replacing the `table` variable with `self.sandbox_table_for(table)` on line 186, but want to confirm with Lauren.

@lrwb-aou Is this the only case where this substitution needs to be made (if at all)? Other uses of the `table` variable seem to reference the original mapping tables in the source dataset, not the version in the sandbox dataset.

lrwb-aou:
> This looks good for the most part, but there is one more thing that should be added. The table name that is being sandboxed should take advantage of the table_namer by wrapping it with the `self.sandbox_table_for()` function. For this cleaning rule, I believe this only involves replacing the `table` variable with `self.sandbox_table_for(table)` on line 186, but want to confirm with Lauren.
> 
> @lrwb-aou Is this the only case where this substitution needs to be made (if at all)? Other uses of the `table` variable seem to reference the original mapping tables in the source dataset, not the version in the sandbox dataset.


@jp3477 You are right, `table` does need to be changed on line 186 the way you are describing.  I ran the the cleaning reporter with:
```
 PYTHONPATH=./:$PYTHONPATH python cdr_cleaner/reporter.py -d controlled_tier_deid_clean -f name sql -c -o ./mapping_ext_report.csv
```
Running this command shows that the queries are not using the sandboxed table name or sandbox dataset (it's easier to view the results in something that handles csv files like tables, take your pick).  The destination details are being configured as dictionary parameters on lines 184-187.  I think it will be ok after this change.

hiro-mishima:
Thank you so much, @jp3477 and @lrwb-aou for the comments! I updated the script accordingly.

jp3477 (approved):

jp3477:
Thanks for the changes @hiro-mishima and thanks for the input @lrwb-aou! This is good to go.

PR #1091, opened 2022-01-18
[DC-1908] Modified remove records with wrong date

jp3477:


mark-velez (approved):
Looks right. Thanks for discussing this with me!

PR #1093, opened 2022-01-19
[DC-2077] Modernize integration tests in achilles_test to use GCS

MichaelSchmidt82:
Remove use of strings in favor of buckets.  Empty_bucket() accepts a string or bucket, and the get_bucket() call is no longer needed.


ksdkalluri (approved):
The failing integration test is not related to these changes, I am working to fix that in DC-2125, this can be merged.

PR #1095, opened 2022-01-19
[DC-1802] Update service account key rotation service to exclude keys created by sysadmin

ksdkalluri:


data_steward/admin/key_rotation.py
```diff
@@ -3,13 +3,33 @@
...
+IGNORE_SERVICE_ACCOUNT_EMAILS = "IGNORE_SERVICE_ACCOUNT_EMAILS"
+
+
+def get_ignore_service_account_emails():
+    """
+    Gathers list of emails of service account keys that needs to be ignored in key rotation
+    """
+    try:
+        ignored_email_list = json.loads(
+            os.environ[IGNORE_SERVICE_ACCOUNT_EMAILS])
```
lrwb-aou, line 24: I would like to see this preferably reading from a table,  or less preferably, reading from secret manager so we don't have to redeploy when we want to change this setting.  How would you feel about doing this through secret manager for now (since we are under time constraints) and doing a table implementation in the near future?

ksdkalluri: Sorry, @lrwb-aou I didn't see this message earlier, I was talking to mark while deciding on how I should approach this, we discussed three options, using a table, using secrets manager, and also using an environment variable, we came to an understanding of "preferably this is a one-time thing if there are new accounts to be added to this list then we would move this keys to a table in pipeline_tables.

lrwb-aou: Is this going to load a string or a list?

ksdkalluri: yes, that's correct. To get the list of accounts I used os.environ.get(json.loads()).

lrwb-aou: @ksdkalluri What am I doing wrong?   I run:
```python
>>> import os
>>> import json
>>> #This works as expected
>>> os.environ['APPLICATION_ID']
'xxx-xxx-xxx-xxx'   # value purposely obscured
>>> 
>>> # This fails
>>> json.loads(os.environ['APPLICATION_ID'])
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/usr/local/Cellar/python@3.9/3.9.7_1/Frameworks/Python.framework/Versions/3.9/lib/python3.9/json/__init__.py", line 346, in loads
    return _default_decoder.decode(s)
  File "/usr/local/Cellar/python@3.9/3.9.7_1/Frameworks/Python.framework/Versions/3.9/lib/python3.9/json/decoder.py", line 337, in decode
    obj, end = self.raw_decode(s, idx=_w(s, 0).end())
  File "/usr/local/Cellar/python@3.9/3.9.7_1/Frameworks/Python.framework/Versions/3.9/lib/python3.9/json/decoder.py", line 355, in raw_decode
    raise JSONDecodeError("Expecting value", s, err.value) from None
json.decoder.JSONDecodeError: Expecting value: line 1 column 1 (char 0)
```
Why are we loading it as JSON?  

If line 23 finished successfully, this wouldn't necessarily be a list, which is what the function document string says we are returning.  Perhaps line 28 is supposed to be(?):
```python
return [ignored_email_list]
```

lrwb-aou (changes requested):
Can you please update the comment (and possibly the code) to ensure it is consistently returning a list?

data_steward/admin/key_rotation.py
ksdkalluri: you have to pass the environment variable as a list like I added [here](https://github.com/all-of-us/curation-devops/pull/67/files) in devops PR.

lrwb-aou: still not sure this works.  What am i missing?  I set this in my terminal:
```shell
 export FOO=['bar']
```

Then in the python interpreter, I still get errors.
```python
>>> import os
>>> import json
>>> os.environ['FOO']
'[bar]'
>>> json.loads(os.environ['FOO'])
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/usr/local/Cellar/python@3.9/3.9.7_1/Frameworks/Python.framework/Versions/3.9/lib/python3.9/json/__init__.py", line 346, in loads
    return _default_decoder.decode(s)
  File "/usr/local/Cellar/python@3.9/3.9.7_1/Frameworks/Python.framework/Versions/3.9/lib/python3.9/json/decoder.py", line 337, in decode
    obj, end = self.raw_decode(s, idx=_w(s, 0).end())
  File "/usr/local/Cellar/python@3.9/3.9.7_1/Frameworks/Python.framework/Versions/3.9/lib/python3.9/json/decoder.py", line 355, in raw_decode
    raise JSONDecodeError("Expecting value", s, err.value) from None
json.decoder.JSONDecodeError: Expecting value: line 1 column 2 (char 1)
>>> 
```

ksdkalluri: @lrwb-aou I think the issue is missing "" around the environment variable, when I was testing it I pass the variable like `export FOO="['foo', 'bar']"`

lrwb-aou: Then we probably need to update the PR in devops, [here](https://github.com/all-of-us/curation-devops/pull/67/files).  It is using the string like:
`['foo']`
instead of like `"['foo']"`.

ksdkalluri: @lrwb-aou My bad I didn't realize it was not added there. I'll update that.

lrwb-aou (approved):
Conditionally approved.  Can you add to the function docstring about the decision to investigate better options should we have to start white listing multiple emails?

lrwb-aou:
👍

PR #1097, opened 2022-01-19
[DC-1817] [Build - Pipeline] Leverage existing lookup tables for site validation

hiro-mishima:
**Overview**
See JIRA ticket [DC-1817](https://precisionmedicineinitiative.atlassian.net/browse/DC-1817) for details. [This comment](https://precisionmedicineinitiative.atlassian.net/browse/DC-1817?focusedCommentId=105190) also helps understand the requirement.

--
**Note 1**
For functions that are still using the old GCS client, I added these two `@mock.patch` to the corresponding integration tests. We can safely remove the first patch statement after the GCS client migration is completed.
```
    @mock.patch("gcs_utils.LOOKUP_TABLES_DATASET_ID", dataset_id)
    @mock.patch("gcloud.gcs.LOOKUP_TABLES_DATASET_ID", dataset_id)
```

--

**Note 2**
After this PR is approved, I will update `aou-res-curation-prod.lookup_tables.hpo_id_bucket_name` table. Here is what I will do:
1. Add the new column `service`
2. Populate `default` or `legacy` to each of the values based on [this new CSV file](https://github.com/all-of-us/curation-devops/pull/65/files#diff-8c9e4b07f9bfe67edc668b8de5797a8d26823d87059aa84dc2b9c04669785568).

--

**Note 3**
There is (another PR)[https://github.com/all-of-us/curation-devops/pull/65] for `curation-devops` regarding this ticket.


data_steward/constants/utils/bq.py
```diff
@@ -9,6 +9,12 @@
 HPO_ID_BUCKET_NAME_TABLE_ID = 'hpo_id_bucket_name'
 HPO_ID_CONTACT_LIST_TABLE_ID = 'hpo_id_contact_list'
 
+# Query to select all from a specified table
+GET_BUCKET_QUERY = """
+SELECT bucket_name FROM `{{project_id}}.{{dataset_id}}.{{table_id}}`
+WHERE hpo_id = '{{hpo_id}}'
```
hiro-mishima, line 15: Get everything first, not filter it

data_steward/gcloud/gcs/__init__.py
```diff
@@ -151,5 +154,33 @@ def list_sub_prefixes(self, bucket_name: str, prefix: str) -> list:
...
+
+        Raises:
+            ValueError: [description]
+            BucketDoesNotExistError: [description]
+
+        Returns:
+            str: [description]
+        """
+
+        project_id = os.environ.get('GOOGLE_CLOUD_PROJECT')
```
hiro-mishima, line 170: app_identity should be used instead

data_steward/gcloud/gcs/__init__.py
```diff
@@ -151,5 +154,33 @@ def list_sub_prefixes(self, bucket_name: str, prefix: str) -> list:
...
+
+        hpo_bucket_query = JINJA_ENV.from_string(GET_BUCKET_QUERY).render(
+            project_id=project_id,
+            dataset_id=LOOKUP_TABLES_DATASET_ID,
+            table_id=HPO_ID_BUCKET_NAME_TABLE_ID,
+            hpo_id=hpo_id)
+
+        query_result = list(query(hpo_bucket_query).result())
```
hiro-mishima, line 178: Filter the result here

data_steward/gcloud/gcs/__init__.py
```diff
@@ -151,5 +154,33 @@ def list_sub_prefixes(self, bucket_name: str, prefix: str) -> list:
...
+
+        if len(query_result) != 1:
+            raise ValueError(
+                f'{len(query_result)} buckets are returned for {hpo_id} '
+                f'in {project_id}.{LOOKUP_TABLES_DATASET_ID}.{HPO_ID_BUCKET_NAME_TABLE_ID}.'
+            )
```
hiro-mishima, line 184: try get_bucket and if not found raise bucket not set error

data_steward/gcloud/gcs/__init__.py
hiro-mishima: If the value is empty, raise error here.

hiro-mishima: bucket not set error

data_steward/gcloud/gcs/__init__.py
```diff
@@ -151,5 +154,33 @@ def list_sub_prefixes(self, bucket_name: str, prefix: str) -> list:
...
+        project_id = os.environ.get('GOOGLE_CLOUD_PROJECT')
+
+        hpo_bucket_query = JINJA_ENV.from_string(GET_BUCKET_QUERY).render(
+            project_id=project_id,
+            dataset_id=LOOKUP_TABLES_DATASET_ID,
+            table_id=HPO_ID_BUCKET_NAME_TABLE_ID,
+            hpo_id=hpo_id)
+
+        query_result = list(query(hpo_bucket_query).result())
+
```
hiro-mishima, line 179: Confirm the version is correct (legacy, default, admin)

hiro-mishima (commented):
Leaving comments from the huddle with @nishanthpp93

data_steward/gcloud/gcs/__init__.py
MichaelSchmidt82: I was looking at this and started to wonder.  The `project_id` can get gathered directly through this client (`self.project`), and the `get_application_id()` in app_identity.py uses the project id as well.

tests/integration_tests/data_steward/bq_utils_test.py
```diff
@@ -117,6 +122,8 @@ def test_load_csv(self):
                                         locals())
         self.assertEqual(query_response['kind'], 'bigquery#queryResponse')
 
+    @mock.patch("gcs_utils.LOOKUP_TABLES_DATASET_ID", dataset_id)
+    @mock.patch("gcloud.gcs.LOOKUP_TABLES_DATASET_ID", dataset_id)
```
nishanthpp93, line 126: I think you can mock once in the source and that should automatically propagate to these, since they import from the same source. But let me know if that doesn't work.
```suggestion
    @mock.patch("constants.utils.bq.LOOKUP_TABLES_DATASET_ID", dataset_id)
```

tests/unit_tests/data_steward/gcloud/gcs/gcs_test.py
```diff
@@ -36,22 +36,28 @@ def setUp(self):
...
     @patch('gcloud.gcs.os.environ.get')
-    def test_get_hpo_bucket_not_set(self, mock_environ_get):
+    def test_get_hpo_bucket_not_set(self, mock_environ_get,
+                                    mock_get_hpo_bucket_id):
+
         mock_environ_get.side_effect = [None, '', 'None']
         expected_message = lambda bucket: f"Bucket '{bucket}' for hpo '{self.hpo_id}' is unset/empty"
 
         # run without setting env var (unset env_var)
+        mock_get_hpo_bucket_id.return_value = None
```
nishanthpp93, line 48: You could potentially only replace the mocked function and it should work as before 
```suggestion
        mock_get_hpo_bucket_id.side_effect = [None, '', 'None']
        expected_message = lambda bucket: f"Bucket '{bucket}' for hpo '{self.hpo_id}' is unset/empty"

        # run without setting env var (unset env_var)
```

nishanthpp93 (commented):
Impressive work, Hiro! I think this does the job, and very well. Let's discuss the deployment results later. I added a few comments which are nits

tests/integration_tests/data_steward/bq_utils_test.py
hiro-mishima: Thanks, @nishanthpp93 

I thought that would work too, but it did not :(
After some trials and errors, I learned it did work because `__init__.py` and `gcs_utils.py` know the path to `LOOKUP_TABLES_DATASET_ID` because of this statement and `patch` cannot patch it.
```
from constants.utils.bq import SELECT_ALL_QUERY, LOOKUP_TABLES_DATASET_ID, HPO_ID_BUCKET_NAME_TABLE_ID
```

It would work if we changed it to 
```
from constants.utils import bq
from constants.utils.bq import SELECT_ALL_QUERY, HPO_ID_BUCKET_NAME_TABLE_ID
```
and replace `LOOKUP_TABLES_DATASET_ID ` to `bq.LOOKUP_TABLES_DATASET_ID` in each of the scripts.

I think it is better to leave it as it is now.

hiro-mishima:
@nishanthpp93 Thank you so much for reviewing this!

I updated the scripts. For the second point, I left a follow-up comment in this PR. 

I also deployed the app to the test environment yesterday. One time without `service` specified in `app.yaml` and the other time `service: beta`. I confirmed that `os.environ.get("GAE_SERVICE")` works as expected. Here are screenshots from the experiment. (Each of the screenshots has an error message, but you can ignore that. It is because `aou-res-curation-test:lookup_tables.hpo_id_bucket_name` does not have test data.)

gae_default

gae_beta

data_steward/gcloud/gcs/__init__.py
```diff
@@ -151,5 +156,28 @@ def list_sub_prefixes(self, bucket_name: str, prefix: str) -> list:
...
+            dataset_id=LOOKUP_TABLES_DATASET_ID,
+            table_id=HPO_ID_BUCKET_NAME_TABLE_ID)
+
+        result_df: DataFrame = query(hpo_bucket_query)
+        condition_hpo_id = (result_df['hpo_id'] == hpo_id)
+        condition_service = (result_df['service'] == service)
+        result_filtered = result_df[condition_hpo_id & condition_service]
+
+        if result_filtered['bucket_name'].count() != 1:
+            raise BucketNotSet(
```
hiro-mishima, line 178: Return None, 'None' or something instead. BucketNotset is used in the parent function

data_steward/gcloud/gcs/__init__.py
```diff
@@ -151,5 +156,28 @@ def list_sub_prefixes(self, bucket_name: str, prefix: str) -> list:
...
+        :return: name of the bucket
+        """
+        service = os.environ.get('GAE_SERVICE', 'default')
+
+        hpo_bucket_query = JINJA_ENV.from_string(SELECT_ALL_QUERY).render(
+            project_id=self.project,
+            dataset_id=LOOKUP_TABLES_DATASET_ID,
+            table_id=HPO_ID_BUCKET_NAME_TABLE_ID)
+
+        result_df: DataFrame = query(hpo_bucket_query)
```
hiro-mishima, line 172: bq.query is deprecated. Look up the comments in bq.query and re-write the statement. (But this is optional.)

nishanthpp93 (approved):
Great work on this @hiro-mishima! Looks good. I am approving this since it works as expected. Also, reminder to update the site addition playbook.

tests/integration_tests/data_steward/bq_utils_test.py
nishanthpp93: Yes I see the problem with this.

PR #1100, opened 2022-01-20
[DC-2082] Modernize integration tests in required_labs_tests.py

MichaelSchmidt82:
- Remove uses of gcs_utils
- Restructure elements by topic


nishanthpp93 (approved):
Looks good, one more without gcs_utils!

PR #1101, opened 2022-01-20
[DC-2079] Full Removal of gcs_utils from integration tests in export_test.py

MichaelSchmidt82:
- Full removal of gcs_utils.py dependency
- Variable renaming


hiro-mishima (approved):
It looks good to me. Thanks, Michael!

PR #1102, opened 2022-01-20
[DC-2075] Replace gcs_utils.get_hpo_bucket(hpo_id) calls with StorageClient().get_hpo_bucket(hpo_id)

lledezma:
- Replaces `gcs_utils.get_hpo_bucket(hpo_id)` method calls in ` _upload_achilles_files(…)` and `process_hpo(...)` with `StorageClient().get_hpo_bucket(hpo_id)` in data_steward/validation/main.py.
- Replaces `gcs_utils.upload_object(...)` in `_upload_achilles_files(...)` from data_steward/validation/main.py.'
- Updates` test_unset_bucket(...) `and relocates it from tests/unit_tests/data_steward/validation/main_test.py to tests/unit_tests/data_steward/gcloud/gcs/gcs_test.py.
- Updates `test_validate_all_hpos_exception(...)` and `test_process_hpo_ignore_dirs(...)` in tests/unit_tests/data_steward/validation/main_test.py.
- Adds `self.message = msg` in `class BucketNotSet(RuntimeError)` from data_steward/validation/app_errors.py.

### DIFF
--- a/data_steward/validation/app_errors.py
+++ b/data_steward/validation/app_errors.py
@@ -56,6 +56,7 @@ class BucketNotSet(RuntimeError):
 
     def __init__(self, msg):
         super(BucketNotSet, self).__init__(self, msg)
+        self.message = msg
 
 
 #TODO: refactor code that has BDNE error to NotFound -- all of it w/ticket
--- a/data_steward/validation/main.py
+++ b/data_steward/validation/main.py
@@ -177,8 +177,9 @@ def _upload_achilles_files(hpo_id=None, folder_prefix='', target_bucket=None):
         bucket_file_name = filename.split(resources.resource_files_path +
                                           os.sep)[1].strip().replace('\\', '/')
         with open(filename, 'rb') as fp:
-            upload_result = gcs_utils.upload_object(
-                bucket.name, folder_prefix + bucket_file_name, fp)
+            blob = bucket.blob(f'{folder_prefix}{bucket_file_name}')
+            blob.upload_from_file(fp)
+            upload_result = storage_client.get_blob_metadata(blob)
             results.append(upload_result)
     return results
 
--- a/data_steward/validation/main.py
+++ b/data_steward/validation/main.py
@@ -177,9 +177,9 @@ def _upload_achilles_files(hpo_id=None, folder_prefix='', target_bucket=None):
         bucket_file_name = filename.split(resources.resource_files_path +
                                           os.sep)[1].strip().replace('\\', '/')
         with open(filename, 'rb') as fp:
-            blob = bucket.blob(f'{folder_prefix}{bucket_file_name}')
+            blob: Blob = bucket.blob(f'{folder_prefix}{bucket_file_name}')
             blob.upload_from_file(fp)
-            upload_result = storage_client.get_blob_metadata(blob)
+            upload_result: dict = storage_client.get_blob_metadata(blob)
             results.append(upload_result)
     return results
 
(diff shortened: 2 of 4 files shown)


data_steward/validation/main.py
```diff
@@ -504,37 +510,33 @@ def process_hpo(hpo_id, force_run=False):
     """
     try:
         logging.info(f"Processing hpo_id {hpo_id}")
-        bucket = gcs_utils.get_hpo_bucket(hpo_id)
-        bucket_items = list_bucket(bucket)
-        folder_prefix = _get_submission_folder(bucket, bucket_items, force_run)
+        project_id = app_identity.get_application_id()
```
MichaelSchmidt82, line 513: It would be nice to see these outside the try block since they won't raise the exceptions caught.

tests/unit_tests/data_steward/gcloud/gcs/gcs_test.py
```diff
@@ -32,6 +34,27 @@ def setUp(self):
         self.bucket: str = 'foo_bucket'
         self.prefix: str = 'foo_prefix/'
         self.file_name: str = 'foo_file.csv'
+        self.hpo_id = 'fake_hpo_id'
+
+    def test_unset_bucket(self):
```
MichaelSchmidt82, line 39: The name of the method doesn't follow convention and it could be unclear which method was under test.  I suggest `test_get_hpo_bucket_not_set(...)`

tests/unit_tests/data_steward/gcloud/gcs/gcs_test.py
```diff
@@ -32,6 +34,27 @@ def setUp(self):
...
+        bucket_env_var = f'BUCKET_NAME_{self.hpo_id.upper()}'
+        expected_message = lambda bucket: f"Bucket '{bucket}' for hpo '{self.hpo_id}' is unset/empty"
+
+        # run without setting env var (unset env_var)
+        os.environ.pop(f"BUCKET_NAME_{self.hpo_id.upper()}", None)
```
MichaelSchmidt82, line 44: env var use has the potential to disrupt the namespace, especially if tests are run in parallel.  the dict `os.environ` can be patched:

```
@mock.patch.dict(os.environ, {'mock_key': 'mock_value'})
def test_mock_env(self):
    self.assertEqual(os.environ['mock_key'], 'mock_value')
```
[See here for other ideas.](https://adamj.eu/tech/2020/10/13/how-to-mock-environment-variables-with-pythons-unittest/)

tests/unit_tests/data_steward/validation/main_test.py
```diff
@@ -412,30 +408,26 @@ def test_process_hpo_ignore_dirs(
         main.process_hpo('noob', force_run=True)
```
MichaelSchmidt82, line 408: 'noob' is used 7 times here and it can be difficult to follow exactly what 'noob' is.  I would suggest prefixing the attribute/object with "fake" or "mock" (mock_id, fake_csv.csv, etc.)

MichaelSchmidt82 (changes requested):

data_steward/validation/app_errors.py
```diff
@@ -56,6 +56,7 @@ class BucketNotSet(RuntimeError):
 
     def __init__(self, msg):
         super(BucketNotSet, self).__init__(self, msg)
+        self.message = msg
```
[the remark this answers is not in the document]
nishanthpp93: I think it depends on the exception. Here, we know this exception is expected so we only need the message and don't care about other attributes. Also, it's easier to validate that the message is being printed with the correct params. 
If we run into a non-custom exception, we will need more information so it can help us diagnose the problem, and we'll need more info than just the message so `f'{exc}'` or possibly `f'{repr(exc)}'` should be used.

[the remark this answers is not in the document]
MichaelSchmidt82: I'm not arguing to remove this, but I don't have the context to understand it.

Like when we had the BucketDoesNotExist exception, you attached a bucket attribute.  "It didn't exist--Well, _which_ bucket didn't exist".

but you say, "Also, it's easier to validate that the message is being printed with the correct params." -- so would this be used in testing only?

One might make the case that logging in combination with a stack trace could provide enough information (just log and `raise exc`).  But that's an opinion and style (and verbose logs).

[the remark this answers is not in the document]
nishanthpp93: I agree stack trace and raise are helpful for unexpected exceptions, where we need all that information. But we don't care about that information here, since we are catching a known/expected exception. We just need to catch it, print it and move on. The message attribute helps in the printing. Ideally, all of our custom exceptions should have the message attribute since it is [standard in the docs](https://docs.python.org/3.8/tutorial/errors.html#user-defined-exceptions). The testing is just an added benefit but not the main purpose.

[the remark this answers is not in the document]
MichaelSchmidt82: gold star ⭐️ for the reference.

data_steward/validation/main.py
MichaelSchmidt82: Nope, throws an exc

MichaelSchmidt82:
I will look at this soon.

MichaelSchmidt82 (approved):

PR #1105, opened 2022-01-24
[DC-2084] Refactor test to use buckets directly and phase out gcs_utils

MichaelSchmidt82:
- Previous test method used gcs_utils to get a string and then used the string to create a bucket.  This change uses the string directly to create a bucket, eliminating the intermediate step and removing gcs_utils from the file entirely.

### DIFF
--- a/tests/integration_tests/data_steward/tools/top_heel_errors_test.py
+++ b/tests/integration_tests/data_steward/tools/top_heel_errors_test.py
@@ -7,7 +7,6 @@ import app_identity
 import bq_utils
 import common
 from gcloud.gcs import StorageClient
-import gcs_utils
 import resources
 from tests import test_util
 from tools.top_heel_errors import top_heel_errors, RESULT_LIMIT, FIELD_ANALYSIS_ID, FIELD_RECORD_COUNT, \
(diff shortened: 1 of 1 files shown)


tests/integration_tests/data_steward/tools/top_heel_errors_test.py
```diff
@@ -87,11 +86,10 @@ def load_test_data(self, hpo_id: str = None):
         test_file_path: str = os.path.join(test_util.TEST_DATA_PATH,
                                            test_file_name)
 
-        target_bucket = self.storage_client.get_bucket(self.bucket)
-        test_blob = target_bucket.blob(test_file_name)
+        test_blob = self.drc_bucket.blob(test_file_name)
         test_blob.upload_from_filename(test_file_path)
 
-        gcs_path: str = f'gs://{self.bucket}/{test_file_name}'
+        gcs_path: str = f'gs://{self.drc_bucket.name}/{test_file_name}'
```
lledezma, line 92:
```
     test_blob = self.drc_bucket.blob(f'{table_id}.csv')
     test_file_path: str = os.path.join(test_util.TEST_DATA_PATH,
                                           test_blob.name)
     test_blob.upload_from_filename(test_file_path)
     gcs_path: str = f'gs://{self.drc_bucket.name}/{test_blob.name}'
```

lledezma: I have a suggestion. You could remove and replace test_file_name to use the name attribute from test_blob.

lledezma (changes requested):

tests/integration_tests/data_steward/tools/top_heel_errors_test.py
MichaelSchmidt82: Ah,  I think I see what you're doing here.  Create the blob with the string once then use the name attribute?

lledezma: Yes

lledezma (approved):

PR #1106, opened 2022-01-25
[DST-251] addtional_QCs_for_the_Controlled_Tier

jqian2015:
@mark-velez DST-251 notebooks need to be tested for a further review  and it does not necessarily need to be merged for now. thanks.


data_steward/analytics/cdr_ops/cross_ct_rt/dst_251_addtional_QCs_for_the_Controlled_Tier_part1.py
```diff
@@ -0,0 +1,970 @@
...
+df = pd.DataFrame(columns = ['query', 'result']) 
+
+# # Query1: No dates before birth_date
+#
+
+# +
+# step1 , to get the tables AND columns that have person_id, size >1 AND DATE columns AND save to a data frame
+query = JINJA_ENV.from_string("""
+
+WITH
```
ksdkalluri, line 84: @jqian2015 is there a reason why TIMESTAMP columns are not included in this list of collumns?

data_steward/analytics/cdr_ops/cross_ct_rt/dst_251_addtional_QCs_for_the_Controlled_Tier_part1.py
```diff
@@ -0,0 +1,970 @@
...
+      table_name IN (
+      SELECT
+        DISTINCT table_name
+      FROM
+        table1))
+    AND c.data_type IN ('DATE')
+    AND NOT REGEXP_CONTAINS(column_name, r'(?i)(_PAR)') 
+""")
+
+q = query.render(project_id=project_id, rt_dataset=rt_dataset, ct_dataset=ct_dataset)
```
ksdkalluri, line 123: rt_dataset is not being used in this query.

data_steward/analytics/cdr_ops/cross_ct_rt/dst_251_addtional_QCs_for_the_Controlled_Tier_part1.py
jqian2015: i thought TMESTAMP is same as date column but with extra 'timestamp' and since we only use 'date' , so I did not use it. if this assumption is wrong, we can sure add TIMESTAMP.

data_steward/analytics/cdr_ops/cross_ct_rt/dst_251_addtional_QCs_for_the_Controlled_Tier_part1.py
jqian2015: yes, you may see several places like this, since such 'unused' variable did not have any impact when running the cell, I just went on without concerning. Again, we sure can clean this up.

PR #1108, opened 2022-01-25
[DC-1989] Added ehr_cutoff_date parameter to ehr_union

jp3477:


ksdkalluri (changes requested):
@jp3477 like we discussed on slack. please close this PR. rename your branch and create a new PR with a different branch name.

PR #1109, opened 2022-01-25
[DC-2103] Update the survey version mapping cleaning rule to include Minute survey concepts

ksdkalluri:
* Updated integration test case


jp3477 (approved):
This looks good!

PR #1117, opened 2022-01-31
[DC-1989] added ehr_cutoff_date parameter to ehr_union

jp3477:
Copy of https://github.com/all-of-us/curation/pull/1108 with shorter branch name.


data_steward/validation/ehr_union.py
```diff
@@ -77,6 +77,7 @@
 """
 import argparse
 import logging
+from datetime import datetime
```
ksdkalluri, line 80: @jp3477 changing branch name worked. and the code looks good.

ksdkalluri (approved):

PR #1118, opened 2022-02-01
[DC-2086] Modernize integration tests in main_test to use GCS

MichaelSchmidt82:
- The integration tests in main_test.py use gcs_utils in several places. These tests require updates.
- Sections of these tests (and the code under test) rely on strings of date and times.  These are then converted to `datetime` objects later on.  The new client returns these dates and times as `datetime` objects directly.  However, these objects supply time zone-aware objects (UTC). In the past these `datetime` objects were time zone naive, so keeping compatibility required time zone information stripped out.

### DIFF
--- a/data_steward/bq_utils.py
+++ b/data_steward/bq_utils.py
@@ -858,12 +858,12 @@ def load_table_from_csv(project_id,
                    the fields are stored in a json file in resource_files/fields named table_name.json
     :return: BQ response for the load query
     """
-    if csv_path is None:
+    if not csv_path:
         csv_path = os.path.join(resources.resource_files_path,
                                 table_name + ".csv")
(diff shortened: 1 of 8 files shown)


MichaelSchmidt82:
This is going back to a work in progress due to update on `develop`

MichaelSchmidt82:
Reopening PR after rebasing with `develop` the PR has under gone testing, and the diff show clean updates

hiro-mishima:
Thank you so much for walking me through the codes, and solving the headbutts between my PR and your PR! It looks good to me. For approval, I will defer it to @nishanthpp93 .

MichaelSchmidt82 (commented):
Add comment per Hiro's suggestion

PR #1119, opened 2022-02-01
[DC-1966] Update data_steward.main to use gcloud.gcs.StorageClient()

lledezma:
Removes `list_bucket(...)` function in data_steward/validation/main.py and replaces calls with `StorageClient.get_bucket_items_metadata`.
Adds `mock_google_cloud_error(...)` in tests/test_util.py for `GoogleCloudError` exception unit testing.
Removes `datetime.strptime` from `updated_datetime_object(...)` and `initial_date_time_object(...)` functions in data_steward/validation/main.py.
Removes `datetime.strftime` and `mock_list_bucket` in tests/unit_tests/data_steward/validation/main_test.py.


data_steward/validation/main.py
```diff
@@ -16,7 +16,9 @@
 # Third party imports
 import dateutil
 from flask import Flask
+from google.cloud.exceptions import GoogleCloudError
 from googleapiclient.errors import HttpError
+
```
nishanthpp93, line 21: It would be nice to get rid of all HttpError usages and remove this completely, but that can be in a later ticket. Btw, you don't have to add blank lines within import blocks. Just make sure they are grouped together if coming from the same package/subpackage.

lledezma: I forgot to remove the blank line. I think I left there by mistake when testing different exceptions.

nishanthpp93 (approved):
Overall looks great! Added a small comment

PR #1122, opened 2022-02-04
[DC-2150] Convert only CDM tables to v531

nishanthpp93:


data_steward/tools/migrate_cdm52_to_cdm531.py
```diff
@@ -76,13 +80,16 @@ def get_upgrade_table_query(client, dataset_id, table_id, hpo_id=None):
     try:
```
mark-velez, line 80: nit: consider removing the try catch so it can fail loudly in cases we didn't anticipate

mark-velez (approved):
Thanks @nishanthpp93 . This migration script is only temporary; I think this gets the job done without overthinking it. Added a nitpick which you can ignore if you wish.

data_steward/tools/migrate_cdm52_to_cdm531.py
nishanthpp93: I think this makes sense since we are filtering down the list to only CDM

nishanthpp93: It looks like this is useful for tables we don't have schemas for, like data_browser tables etc., so keeping this.

PR #1131, opened 2022-02-10
[DC-2156] Implement unit test to check for bucket NotFound error

lledezma:
- Adds `test_get_hpo_bucket_not_found` method in tests/unit_tests/data_steward/gcloud/gcs_test.py to test for `NotFound` exceptions.
- Adds `get_bucket_items_metadata` call in `StorageClient.get_hpo_bucket` method to throw `NotFound` exception when a bucket does not exist.

### DIFF
--- a/tests/integration_tests/data_steward/gcloud/gcs_client_test.py
+++ b/tests/integration_tests/data_steward/gcloud/gcs_client_test.py
@@ -35,6 +35,7 @@ class GcsClientTest(unittest.TestCase):
 
     @patch.object(StorageClient, '_get_hpo_bucket_id')
     def test_get_hpo_bucket_not_found(self, mock_get_bucket_id):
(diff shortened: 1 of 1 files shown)


data_steward/gcloud/gcs/__init__.py
```diff
@@ -106,6 +106,7 @@ def get_hpo_bucket(self, hpo_id: str) -> Bucket:
 
         try:
             bucket = self.bucket(bucket_name)
+            self.get_bucket_items_metadata(bucket)
```
nishanthpp93, line 109: Add in a comment here saying this is to verify the bucket can be accessed and we don't care about the result.

nishanthpp93 (commented):
Looks good! I'd say add in an integration test as well and a clarification comment.

nishanthpp93 (approved):
Looks good! Thanks for the changes. Added a nit but not required for merge.

PR #1133, opened 2022-02-10
[DC-2146] remove combined personal family medical health history survey answers

lrwb-aou:
* script to remove the combeind pfmhh survey based on the concept codes from the codebook and the date the answer was submitted
* integration tested
* added to RDR rule prior to the PPI branching logic to drop data before that rule runs.

### DIFF
--- a/data_steward/cdr_cleaner/cleaning_rules/suppress_combined_pfmh_survey.py
+++ b/data_steward/cdr_cleaner/cleaning_rules/suppress_combined_pfmh_survey.py
@@ -1,0 +1,552 @@
+"""
+Removing data associated with the combined Personal Family Health Survey.
+
+This survey is expected to be included in the November CDR and later.  We
+need to be ready to remove this data from the May CDR.
+
+Original Issue:  DC-2146
+"""
+# Python Imports
+import logging
+
+# Third party imports
+from jinja2 import Template
+
+# Project imports
+from common import OBSERVATION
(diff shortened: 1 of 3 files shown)


data_steward/cdr_cleaner/cleaning_rules/suppress_combined_pfmh_survey.py
```diff
@@ -0,0 +1,553 @@
...
+CREATE OR REPLACE TABLE `{{project}}.{{sandbox}}.{{drop_table}}` AS
+SELECT
+  *
+FROM
+  `{{project}}.{{dataset}}.observation`
+WHERE
+  -- separated surveys were removed from participant portal on 2021-11-01 --
+  -- combined survey was launced on 2021-11-01 --
+  observation_date > '2021-10-31'
+  AND LOWER(observation_source_value) IN ('record_id',
```
ksdkalluri, line 35: suggestion: I think it is better to store this list of concepts as a constant or in a separate file, since we will be adding a check in the rdr_qc notebook, we can reuse the list instead of copying this list of 100s of concepts and maintain these in multiple locations.

lrwb-aou: I pulled the list into a separate variable.  I think a notebook should be able to import and use the list.  I was hoping to avoid creating another file since this is likely a one time fix.

ksdkalluri (approved):
Looks Good! I'll be creating a ticket for refactoring the  clean_engine to run validations.

PR #1136, opened 2022-02-14
[DC-2130] Refactor DRC street address comparison

hiro-mishima:


data_steward/resource_files/validation/participants/abbreviation_state.csv
```diff
@@ -0,0 +1,65 @@
+"abbreviated","unabbreviated"
```
nishanthpp93, line 1: This is helpful, but would suggest making it complete using [census reference](https://www.census.gov/library/reference/code-lists/ansi/ansi-codes-for-states.html). Also, I don't think you can have commented lines. A comment where you are loading this data would suffice, it could just reference the source.

data_steward/constants/validation/participants/validate.py
```diff
@@ -229,62 +107,145 @@
...
+
+NORMALIZED_STREET = JINJA_ENV.from_string("""
+    WITH address_abbreviations AS (
+        SELECT *
+        FROM UNNEST(ARRAY<STRUCT<abbreviation STRING, expansion STRING>>[
+            {{abbreviation_tuples}}
+        ])
+    ),
+    remove_commas_and_periods AS (
+        SELECT 'dummy_id' as id, REGEXP_REPLACE(coalesce({{street}}, ''), '[,.]', '') as address
```
nishanthpp93, line 157: I'd suggest using something like row_number over() as the dummy_id.

data_steward/constants/validation/participants/validate.py
```diff
@@ -229,62 +107,145 @@
...
-              END AS ehr_state
-        )
+    WITH normalized_rdr_state AS (
+        SELECT 
+            CASE
+            WHEN REPLACE(LOWER(TRIM(rdr_state)), 'piistate_', '') IN ({{state_abbreviations}})
+            THEN REPLACE(LOWER(TRIM(rdr_state)), 'piistate_', '')
+            WHEN rdr_state IS NULL THEN NULL
+            ELSE ''
+            END AS rdr_state
```
nishanthpp93, line 238: If state_abbreviations can be in a table, you can join it and use COALESCE to simplify this logic

data_steward/validation/participants/validate.py
```diff
@@ -158,10 +104,16 @@ def identify_rdr_ehr_match(client,
...
             missing_rdr=consts.MISSING_RDR,
             missing_ehr=consts.MISSING_EHR,
+            rdr_street=consts.NORMALIZED_STREET.render(
+                abbreviation_tuples=street_tuples_str, street='rdr_street'),
+            ehr_street=consts.NORMALIZED_STREET.render(
+                abbreviation_tuples=street_tuples_str, street='ehr_street'),
+            rdr_city=consts.NORMALIZED_CITY.render(
+                abbreviation_tuples=city_tuples_str, city='rdr_city'),
+            ehr_city=consts.NORMALIZED_CITY.render(
+                abbreviation_tuples=city_tuples_str, city='ehr_city'),
```
nishanthpp93, line 114: I think these could be static tables in drc_ops. They just need the '_' prefix as they are internal tables. We shouldn't need to change them often. Let me know your thoughts.

data_steward/resource_files/validation/participants/abbreviation_street.csv
```diff
@@ -0,0 +1,50 @@
+"abbreviated","unabbreviated"
```
nishanthpp93, line 1: I think `expanded` fits better than `unabbreviated` here and in other places.

data_steward/constants/validation/participants/validate.py
```diff
@@ -229,62 +107,145 @@
...
+        SELECT 'dummy_id' as id, REGEXP_REPLACE(coalesce({{city}}, ''), '[,.]', '') as address
+    ),
+    remove_redundant_whitespaces AS (
+        SELECT id, REGEXP_REPLACE(TRIM(address), ' +', ' ') as address
+        FROM remove_commas_and_periods
+    ),
+    lowercase AS (
+        SELECT id, LOWER(address) as address
+        FROM remove_redundant_whitespaces
+    ),
```
nishanthpp93, line 209: This may be too modular, you can add this to any of the above. Also, all the string manipulation could potentially be in one, and directly placed in the CREATE_CITY_COMPARISON_FUNCTION Jinja template. Similar for street. However, let me know if that's not as doable.

nishanthpp93 (commented):
Thanks @hiro-mishima, overall great improvements. Added a few comments for you to consider. Let me know your thoughts.

data_steward/resource_files/validation/participants/abbreviation_state.csv
hiro-mishima: @nishanthpp93 
I agree with you - I think it is better to have it complete.

I found some discrepancies between our CSV file and the census.

[1] The state abbreviations that exist in our CSV but not in the census reference:
fm
mh
pw
aa
ae
ap

[2] The state abbreviation that exists in the census but not in our CSV:
um

I imagine we should keep the ones in [1], though they are not in the census reference.

nishanthpp93: This is a more complete list from [USPS](https://pe.usps.com/text/pub28/28apb.htm)

data_steward/constants/validation/participants/validate.py
hiro-mishima: I think it is better to keep it this way, because `CompareStreet` (the function that uses this clause) reads only one row at a time, not the whole table. So, `row_number` makes it look more complex but does not improve the logic much.

data_steward/validation/participants/validate.py
```diff
@@ -148,6 +77,23 @@ def identify_rdr_ehr_match(client,
...
+    ])
+
+    city_tuples_str: str = ",\n".join([
+        f"('{abbreviated}','{unabbreviated}')" for abbreviated, unabbreviated in
+        zip(city_df['abbreviated'], city_df['unabbreviated'])
+    ])
+
+    states_str: str = ",\n".join(
+        [f"'{state}'" for state in state_df['abbreviated']])
+
```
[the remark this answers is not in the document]
hiro-mishima: As discussed, I am not implementing internal tables because BigQuery throws the following error if we use it:
`400 Correlated subqueries that reference other tables are not supported unless they can be de-correlated, such as by transforming them into an efficient JOIN.`
If we want to solve this error, we will need to change how we run the updates quite a bit. We will not make that change for now so that we have enough time to implement both `DC-2130` and `DC-2131` before the next CDR run.

hiro-mishima:
@nishanthpp93 
Thank you so much for reviewing this PR, and thanks for your time yesterday for the discussion! Could you please have a look at it again? Please let me know if you have questions or anything. Thanks!

data_steward/constants/validation/participants/validate.py
nishanthpp93: I see, I thought this would be necessary to distinguish different addresses but it looks like this is only processing one address at a time.

nishanthpp93 (commented):
Looks good, thanks for updating it and the test! There is a conflict that needs to be addressed, after resolving it should be good to merge.

hiro-mishima:
@nishanthpp93 Thanks again for the review! I solved the merge conflict. Please let me know if there is anything else!

nishanthpp93 (approved):
Thanks for updating @hiro-mishima, looks good!

data_steward/resource_files/validation/participants/abbreviation_street.csv
```diff
@@ -0,0 +1,50 @@
+"abbreviated","expanded"
```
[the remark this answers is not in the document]
hiro-mishima: Sounds good. I will create a ticket for this. Thanks again for reviewing, Nishanth!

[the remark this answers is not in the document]
hiro-mishima: https://precisionmedicineinitiative.atlassian.net/browse/DC-2177

PR #1144, opened 2022-02-18
[DC-2170] Use StorageClient to create client instances

lledezma:
*Uses `StorageClient` to create new storage client instances in data_steward/tools/load_vacob.py and tests/integration_tests/data_steward/tools/load_vocab_test.py by replacing `storage.Client` calls and function parameters with `StorageClient`

### DIFF
--- a/data_steward/ci/test_buckets.py
+++ b/data_steward/ci/test_buckets.py
@@ -5,7 +5,7 @@ Creates buckets with a 30 day lifecycle.
 Bucket deletion and modification functions (for testing purposes) should be
 added to this module.
 """
-from google.cloud import storage
+from gcloud.gcs import StorageClient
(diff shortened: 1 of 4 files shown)


nishanthpp93 (commented):
Looks good! There are some imports of google.cloud.storage in other scripts (and tests) which may not be necessary. Do you think they can be addressed in a different ticket?

lledezma:
> Looks good! There are some imports of google.cloud.storage in other scripts (and tests) which may not be necessary. Do you think they can be addressed in a different ticket?

I see only an import in data_steward/tools/load_archived_submission.py. It's a minor update. I can take care of it in this PR if that's ok?

nishanthpp93 (approved):
Looks good, thanks for updating @lledezma!

PR #1145, opened 2022-02-18
[DC-2183] Update string match logic to work with UPPER and lower case.

ksdkalluri:


data_steward/gcloud/gcs/__init__.py
```diff
@@ -166,14 +170,17 @@ def _get_hpo_bucket_id(self, hpo_id: str) -> str:
...
             dataset_id=LOOKUP_TABLES_DATASET_ID,
             table_id=HPO_ID_BUCKET_NAME_TABLE_ID)
 
+        LOGGER.info(f"Running: {hpo_bucket_query}")
+
         result_df: DataFrame = query(hpo_bucket_query)
-        condition_hpo_id = (result_df['hpo_id'] == hpo_id)
+        condition_hpo_id = ((result_df['hpo_id'].str.lower()) == hpo_id.lower())
         condition_service = (result_df['service'] == service)
         result_filtered = result_df[condition_hpo_id & condition_service]
```
lrwb-aou, line 185: @ksdkalluri What is the advantage of selecting everything over what we are specifically looking for?  Wouldn't this work:

```suggestion
        # select query is:
        # select * from `{{project_id}}.{{dataset_id}}.{{table_id}}` where lower(hpo_id) = lower('{{hpo_id}}')
        hpo_bucket_query = JINJA_ENV.from_string(SELECT_QUERY).render(
            project_Id=self.project,
            dataset_id=LOOKUP_TABLES_DATASET_ID,
            table_id=HPO_ID_BUCKET_NAME_TABLE_ID,
            hpo_id=hpo_id)

        result_df: DataFrame = query(hpo_bucket_query)

        if result_df['bucket_name'].count() != 1:
            reutnr 'None'

        return result_df['bucket_name'].iloc[0]
```

To me, the code is shorter and easier to understand.  And I've also been in situations where calling "get all" works fine in test but leads to severe performance issues in production.

ksdkalluri: That is a better approach, I will test it.

data_steward/gcloud/gcs/__init__.py
```diff
@@ -166,18 +170,21 @@ def _get_hpo_bucket_id(self, hpo_id: str) -> str:
...
-            table_id=HPO_ID_BUCKET_NAME_TABLE_ID)
+        LOGGER.info(f"Running: {hpo_bucket_query}")
 
         result_df: DataFrame = query(hpo_bucket_query)
-        condition_hpo_id = (result_df['hpo_id'] == hpo_id)
-        condition_service = (result_df['service'] == service)
-        result_filtered = result_df[condition_hpo_id & condition_service]
 
-        if result_filtered['bucket_name'].count() != 1:
+        LOGGER.info(f"hpo_")
```
lrwb-aou, line 187: @ksdkalluri what is this logging?

data_steward/constants/utils/bq.py
```diff
@@ -10,8 +10,14 @@
...
-SELECT_ALL_QUERY = """
-SELECT * FROM `{{project_id}}.{{dataset_id}}.{{table_id}}`
+SELECT_BUCKET_NAME_QUERY = """
+SELECT
+  bucket_name
+FROM
+  `{{project_id}}.{{dataset_id}}.{{table_id}}`
+WHERE
+  LOWER(hpo_id) = LOWER('{{hpo_id}}')
+  AND LOWER(service) = LOWER('{{service}}')
```
lrwb-aou, line 20: 👍

data_steward/gcloud/gcs/__init__.py
ksdkalluri: I am fixing it, didn't realize that was in the code.

lrwb-aou:
Instead of putting `SELECT_ALL_QUERY` back, you could update `https://github.com/all-of-us/curation/blob/develop/data_steward/gcs_utils.py` with the same code.  We know it will be removed when the deprecated module is removed.

lrwb-aou (approved):
This looks good to me.  All tests have passed.  Thank you for the update and only asking for the data you need.

PR #1147, opened 2022-02-21
[DC-1822] zipcode generalization in controlled_tier is not ignoring the `Response removed due to invalid value` values

hiro-mishima:


tests/integration_tests/data_steward/cdr_cleaner/cleaning_rules/generalize_zip_codes_test.py
```diff
@@ -140,7 +147,14 @@ def test_generalize_zip_codes_cleaning(self):
                  'StreetAddress_PIIZIP', 1585250, '723**'),
                 (1011, 1011, 0, datetime.strptime('2020-08-30',
                                                   '%Y-%m-%d').date(), 1011,
-                 'StreetAddress_PIIZIP', 1585247, '37218')
+                 'StreetAddress_PIIZIP', 1585247, '37218'),
+                (1012, 1012, 0, datetime.strptime('2020-08-30',
+                                                  '%Y-%m-%d').date(), 1012,
+                 'StreetAddress_PIIZIP', 1585250, '123*******'),
```
ksdkalluri, line 153: @hiro-mishima, I think we also should make sure this will also show up as 123** instead of 123*******. so that all the zip codes are harmonized in the same format.

ksdkalluri (changes requested):

tests/integration_tests/data_steward/cdr_cleaner/cleaning_rules/generalize_zip_codes_test.py
```diff
@@ -126,9 +129,13 @@ def test_generalize_zip_codes_cleaning(self):
...
                                                   '%Y-%m-%d').date(), 1006,
                  'StreetAddress_PIIZIP', 1585250, '235**'),
+                # 1007: `UpdateInvalidZipCodes` replaces this type of value
+                # with `Response removed due to invalid value` before
+                # `GeneralizeZipCodes` runs. So, '23512-4' here does not
+                # appear in the real world.
                 (1007, 1007, 0, datetime.strptime('2020-08-30',
                                                   '%Y-%m-%d').date(), 1007,
-                 'StreetAddress_PIIZIP', 1585250, '235****'),
+                 'StreetAddress_PIIZIP', 1585250, '23512-4'),
```
ksdkalluri, line 138: @hiro-mishima thanks for the update, this case should also work fine since we have the 5 digits available in the first part of the zip code value we should be generalizing it to 123** format.

ksdkalluri (changes requested):

ksdkalluri (approved):
looks good, thanks for the updates!

PR #1151, opened 2022-02-25
[DC-2119] Verify the covid drug concepts are not suppressed in the May 2022 CDR

hiro-mishima:
#### 1. `data_steward/analytics/cdr_ops/controlled_tier_qc/check_controlled_tier.py`
- Added `display_check_detail_of_rule` for `DC-1496` in `check_controlled_tier.py`. `DC-1496` is already a part of this QC, and the summary shows its result. But the notebook was not showing its detailed result.
#### 2. `data_steward/analytics/cdr_ops/controlled_tier_qc/code/check_concept_suppression.py`
- Added docstrings to the existing functions to make them more descriptive.
- No change in its logic.
#### 3. `data_steward/analytics/cdr_ops/controlled_tier_qc/code/controlled_tier_qc.py`
- Same as 2
#### 4. `data_steward/analytics/cdr_ops/controlled_tier_qc/utils/helpers.py`
- Same as 2

I saved the result of the test runs to the JIRA ticket.


jp3477 (approved):
This looks great! I hadn't seen these additions to the notebooks module yet, so it was very cool to read through. Just to confirm, the registered tier and controlled tier notebooks for the covid_concept check should be identical, correct?

hiro-mishima:
Thanks for reviewing, @jp3477 ! Your understanding is correct - the CT and the RT need the same verification, so I made the notebooks identical.

PR #1153, opened 2022-03-02
[DC-2196] Replace bq.get_client(...) calls in data_steward/retraction

lledezma:
- Implements the new BigQueryClient by replacing bq.get_client(...) calls in data_steward/retraction files.


lrwb-aou (approved):
This looks good!  You've broken down the update task into manageable PRs.  All the tests are passing.

PR #1155, opened 2022-03-04
[DC-2220] Update logic for unioned_ehr_visit_detail to ignore bad data

hiro-mishima:


ksdkalluri (approved):
looks good! Thanks for adding the integration tests.

PR #1158, opened 2022-03-07
[DC-1894] correct cope flu question concepts

lrwb-aou:
* change incorrect concept to correct concept for selected surveys
* add to pipeline at rdr datastage
* integration test


hiro-mishima (approved):
It looks good to me! Thank you for the clean code :)

PR #1163, opened 2022-03-09
[DC-2185] Replace bq.get_client(...) calls in integration tests

lledezma:
- Implements the new BigQueryClient by replacing bq.get_client(...) calls in tests/integration_tests.
- Excludes: 
        tests/integration_tests/data_steward/validation/participants/create_update_drc_id_match_table_test.py.
        tests/integration_tests/data_steward/validation/participants/validate_test.py.
        tests/integration_tests/data_steward/validation/participants/store_participant_summary_results_test.py


MichaelSchmidt82 (changes requested):
There are only two or three cases where `client` is used, and the rest are `bq_client`.  Consider clarifying these to `bq_client` globally.  This avoids ambiguity for current clients (storage_client) and _future_ clients.

MichaelSchmidt82 (approved):
Looks good!

PR #1164, opened 2022-03-10
[DC-2174] Add notebook check that ehr cutoff date is used in ehr observation demographic records

ksdkalluri:


data_steward/analytics/cdr_ops/ehr_snapshot_qc.py
```diff
@@ -89,3 +89,42 @@
...
+  observation_id,
+  person_id,
+  observation_concept_id,
+  observation_date,
+  observation_datetime
+FROM
+  `{PROJECT_ID}.{EHR_SNAPSHOT_DATASET_ID}.unioned_ehr_observation`
+WHERE
+  observation_concept_id IN (4013886,
+    421761,
```
hiro-mishima, line 116: Though the ticket says it is `421761`, I think this needs to be `4271761` as Jason's change for line 576:
https://github.com/all-of-us/curation/pull/1117/files#diff-33687d7723451a4637751a9d35dc8f85aa9ce939865e4cb0770f9b454220ed33L576
Could you please double-check if this is the correct ID?

hiro-mishima (commented):
@ksdkalluri Thanks for working on this!
I left two comments. Could you please check them out?

hiro-mishima (approved):
It looks good to me! Thanks for updating the query :)

PR #1167, opened 2022-03-11
[DC-2197] Replace bq.get_client(...) calls in data_steward/validation

lledezma:
- Implements the new `BigQueryClient` by replacing `bq.get_client(...)` calls in data_steward/validation folder files.


dcarbone (approved):
This lgtm :+1:

PR #1175, opened 2022-03-17
[DC-2253] Add option to display max_rows using notebook_utils/execute()  

ksdkalluri:


hiro-mishima (approved):
This is a very clean solution. Thanks, Krishna!

PR #1176, opened 2022-03-17
[DC-2206] Update `QRIDtoRID` class to use `UPDATE` query instead of `WRITE.TRUNCATE`

hiro-mishima:
- I deleted the unit test, as defined in the scope of the ticket.


nishanthpp93 (changes requested):
Added a suggestion to keep the same logic. I'd also suggest updating the test to make sure rows that are not mapped be set to NULL.

nishanthpp93 (approved):
Looks good! Thanks for updating @hiro-mishima

PR #1178, opened 2022-03-21
[DC-1800] Added a copy of the fitbit dataset to the output-prod script

jp3477:
The `copy_dataset_to_output_prod.py` script now contains a statement to copy fitbit tables into the deid or clean dataset.


nishanthpp93 (approved):
Looks good!

PR #1179, opened 2022-03-21
[DC-2229] Create a Wrapper to get psql client.

ksdkalluri:


dcarbone:
@mark-velez out of curiosity, would it be possible to use [pg8000](https://github.com/tlocke/pg8000) in lieu of `psycopg2`?  pg8000 is a pure-python pgsql driver implementation, and generally works pretty well in my admittedly limited experience with it.

mark-velez:
@dcarbone thanks for the suggestion! @ksdkalluri I say we try pg8000 and see if it meets our current needs.

data_steward/gcloud/gsm/__init__.py
```diff
@@ -23,3 +23,20 @@ def build_secret_full_name(secret_name: str,
         :return: FQN of secret based on provided values
         """
         return f'projects/{project_id}/secrets/{secret_name}/versions/{secret_version}'
+
+    @staticmethod
+    def get_secret_from_secret_manager(secret_name, project_id):
```
lrwb-aou, line 28: For consistency in this module, we should try to use parameter typing here.  

Also, if you allow `project_id` to default the same way it is defaulting in lines 14-15, then we can remove the function definition of `_get_mandrill_secret()` in `email_notification.py` and replace all calls of `_get_mandrill_secret()` with `get_secret_from_secret_manager(consts.MANDRILL_TOKEN_SECRET_ID)`.  The project name would default as expected and your test code should continue to work.

lrwb-aou (commented):
I've left a little feedback. 
It's good to see mark and daniel's feedback.  Could you update the README.md that mark mentioned?
Overall, it looks good.

lrwb-aou:
I did finally get this going.  I was missing the part about starting cloud sql proxy in another window.  What are your thoughts on incorporating the instances as secrets and using subprocess to open the proxy client from the script.  We can add the utility functions to the notebooks.  I think the biggest "gotcha"s would be remembering to have you application default credentials available, cloud sql proxy installed, and remembering to stop the process.  Look over the patch and let me know what looks like something we can use, or throw out, or improve.
[2229_patch.txt](https://github.com/all-of-us/curation/files/8586730/2229_patch.txt)

data_steward/analytics/cdr_ops/cloud_sql_test.py
```diff
@@ -0,0 +1,45 @@
...
+#       jupytext_version: 1.7.1
+#   kernelspec:
+#     display_name: Python 3
+#     language: python
+#     name: python3
+# ---
+
+# + tags=["parameters"]
+project_id = "aou-res-curation-prod"
+run_as = "data-analytics@aou-res-curation-prod.iam.gserviceaccount.com"
```
lrwb-aou, line 17: These should be blank, right?

data_steward/analytics/cdr_ops/notebook_utils.py
```diff
@@ -1,12 +1,82 @@
...
+
+
+def stop_cloud_sql_proxy(process):
+    process.terminate()
+
+
+def start_cloud_sql_proxy(project_id, run_as):
+    """
+    I'd like to consider using secrets to get the instanct information here also.
+    We could get sql proxy for all instances if we wish.  Or the read only replica.
```
lrwb-aou, line 31: This comment needs to be updated or removed.

data_steward/analytics/cdr_ops/notebook_utils.py
```diff
@@ -1,12 +1,82 @@
...
+POSTGRES_USER_NAME = "pdr_psql_username"
+POSTGRES_PASSWORD = "pdr_psql_password"
+POSTGRES_PORT = 7005
+DATABASE_NAME = "drc"
+HOST = "localhost"
+PDR_INSTANCE_NAME = "pdr_cloud_sql_read_only_instance"
+
+
+def stop_cloud_sql_proxy(process):
+    process.terminate()
```
lrwb-aou, line 25: We may want a doc string stating this is specifically killing the subprocess that starts the cloud sql proxy.  Just so we know what the motivation is for this function.

data_steward/validation/email_notification.py
```diff
@@ -211,7 +184,9 @@ def send_email(email_message):
     """
     result = None
     try:
-        api_key = _get_mandrill_secret()
+        smc = SecretManager()
+        api_key = smc.get_secret_from_secret_manager(
+            consts.MANDRILL_TOKEN_SECRET_ID)
```
lrwb-aou, line 189: nice refactor!

lrwb-aou (commented):
This is good.  Just a few doc string and parameters clean ups.

lrwb-aou (approved):
looks good!

PR #1184, opened 2022-03-22
[DC-2282] CR not mapping long to short codes

brendagutman:


mark-velez:
@hiro-mishima I know you're currently looking at the associated data please chime in if you get a minute

data_steward/cdr_cleaner/manual_cleaning_rules/update_questiona_answers_not_mapped_to_omop.py
```diff
@@ -49,7 +49,7 @@
         FIRST_VALUE(standard_c.concept_id) OVER (PARTITION BY source_c.concept_id ORDER BY c_r.relationship_id DESC ) AS new_observation_concept_id
         FROM (
             SELECT
-            SUBSTR(pmi_code,1,50) AS short_pmi_code,
+            pmi_code AS short_pmi_code,
```
hiro-mishima, line 52: Now that we no longer truncate `pmi_code` to 50 char, I think we no longer need to rename it as `short_pmi_code`.

hiro-mishima (changes requested):
Thanks for this PR, @brendagutman ! I left some comments.

hiro-mishima (approved):
Thanks for updating the script, Brenda! I think now it is good to go :sparkles:

PR #1187, opened 2022-03-28
[DC-2291] Add excluded sites check to snapshot_ehr_qc notebook

ksdkalluri:


nishanthpp93 (commented):
Looks good! Added a couple comments about the parametrization.

data_steward/analytics/cdr_ops/ehr_snapshot_qc.py
```diff
@@ -34,6 +37,17 @@
 
 # -
 
+release_tag = '2022q2r1'
```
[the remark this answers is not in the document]
nishanthpp93: This can be removed as well

data_steward/analytics/cdr_ops/ehr_snapshot_qc.py
```diff
@@ -18,12 +19,16 @@
 PROJECT_ID = ""  # identifies the project containing the datasets
 BASELINE_EHR_DATASET_ID = ""  # Identifies the dataset the snapshot was created from
 EHR_SNAPSHOT_DATASET_ID = ""  # Identifies the snapshot dataset
+RELEASE_TAG = ""  # identifies the release tag for the current CDR release
+EXCLUDED_SITES = "default"  # List of excluded sites passed as string: eg. "'hpo_id1', 'hpo_id_2', 'hpo_id3',..."
```
nishanthpp93, line 23: This feels a bit odd. This may be better as a list of hpo_ids than a string.

nishanthpp93, line 23: This feels a bit odd. This may be better as a list of hpo_ids than a string.

nishanthpp93 (commented):
Added a couple more suggestions

data_steward/analytics/cdr_ops/ehr_snapshot_qc.py
ksdkalluri: Papermill takes only strings as input for parameters, that's the reason I have to use this notation. I tried to pass a list but it didn't work.

nishanthpp93: Ah I see, thanks! I see a hardcoded release tag, other than that the ticket looks good.

nishanthpp93 (approved):
Thanks for updating, looks good!

PR #1190, opened 2022-03-30
[DC-2290] Add `ehr_union_qc` notebook

ksdkalluri:


mark-velez:
This looks good @ksdkalluri I would just modify the queries to only show actionable results. You might want to use that heuristic that looks for changes >=25%.

hiro-mishima (approved):
Thanks, Krishna! It looks good to me.

For my future reference, could you please tell me what `Verify Note text data` is checking? What type of results are we expecting for it?

ksdkalluri:
> Thanks, Krishna! It looks good to me.
> 
> For my future reference, could you please tell me what `Verify Note text data` is checking? What type of results are we expecting for it?

Thanks for reviewing this @hiro-mishima, yes for the `Verify Note text data` check we should be expecting an empty string in those columns and that is what we are checking for.

PR #1191, opened 2022-03-30
[DC-2254] Refactor qrid check in RDR export notebook

hiro-mishima:
`data_steward/cdr_cleaner/cleaning_rules/backfill_pmi_skip_codes.py` ... Only the linting for the SQL statement is updated. No change in the logic.


mark-velez (approved):
Please consider refactoring the query to get rid of the `concept_ancestor` references
```
SELECT  
    person_id, 
    STRING_AGG(observation_source_value) AS observation_source_value
FROM
  `aou-res-curation-prod.2022q2r1_rdr.observation`
WHERE 1=1
-- is a survey response
AND observation_type_concept_id = 45905771
AND NOT (observation_id >= 1000000000000 AND value_as_concept_id = 903096)
AND questionnaire_response_id IS NULL
GROUP BY 1
```

PR #1193, opened 2022-03-30
[DC-2261] [Cleaning Rule] Remove participants from fitbit datasets who are not in the combined dataset

hiro-mishima:


nishanthpp93 (commented):
Looks good. I think this still needs an argument in the fitbit cleaning shell script

nishanthpp93 (approved):
Looks good!

hiro-mishima:
@nishanthpp93 Thanks for your time yesterday! For the `__init__() got multiple values for argument` error we faced yesterday, I found out that we needed to add `tablenamer=None` to [ __init__ ](https://github.com/all-of-us/curation/pull/1193/files#diff-ab622bede43dd1eb38c337e589d59ed08839bd0172cb6c5d9d8bf1237fd7d815R40). Otherwise, our additional argument (`reference_dataset_id` for example) are loaded twice: Once as 4th parameter and the second time as `reference_dataset_id=xyz`. The debugging is not showing errors anymore, so it looks good on my end.

Here is the related StackOverflow link: https://stackoverflow.com/questions/21764770/typeerror-got-multiple-values-for-argument

hiro-mishima:
Thanks for checking, @nishanthpp93 ! I updated the unit test accordingly.

nishanthpp93 (approved):
Thanks @hiro-mishima, looks good!

PR #1196, opened 2022-03-31
[DC-2302] change left join to join in visit_detail table for visit_occurrence_id in `combine_ehr_rdr.py`

ksdkalluri:


hiro-mishima (approved):
Thanks for the quick fix, Krishna! It looks good to me.

PR #1199, opened 2022-04-01
[DC-2263] Create script to generate serology data

nishanthpp93:
* [DC-2263] Add script to create serology dataset

[DC-2263]: https://precisionmedicineinitiative.atlassian.net/browse/DC-2263?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/data_steward/tools/snapshot_serology_dataset.py
+++ b/data_steward/tools/snapshot_serology_dataset.py
@@ -125,7 +125,7 @@ def create_serology_snapshot(client: BigQueryClient, release_tag: str,
     """
     dataset_id = f"C{release_tag}_antibody_quest"
     dataset = Dataset(f'{client.project}.{dataset_id}')
-    dataset.description = f'Source dataset: {src_serology_dataset_id}; JIRA issue number: {ISSUE_NUMBER}'
+    dataset.description = f'Source dataset: {src_serology_dataset_id} *_ct views; JIRA issue number: {ISSUE_NUMBER}'
     dataset.labels = {'release_tag': release_tag, 'data_tier': 'controlled'}
     dataset = client.create_dataset(dataset)
     LOGGER.info(f'Successfully created empty dataset {dataset.dataset_id}')
--- a/data_steward/tools/snapshot_serology_dataset.py
+++ b/data_steward/tools/snapshot_serology_dataset.py
@@ -190,7 +190,7 @@ def main():
     bq_client = BigQueryClient(args.project_id, credentials=impersonation_creds)
 
     # Create serology dataset
-    dataset_id = create_serology_snapshot(bq_client, args.snapshot_date,
+    dataset_id = create_serology_snapshot(bq_client, args.release_tag,
                                           args.src_serology_dataset_id)
 
     # Create serology tables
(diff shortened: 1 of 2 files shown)


data_steward/tools/snapshot_serology_dataset.py
```diff
@@ -0,0 +1,202 @@
...
+S_TITER = 'titer'
+S_ROCHE_ORTHO = 'roche_ortho'
+S_TEST = 'test'
+S_RESULT = 'result'
+
+SEROLOGY_TABLES = [PERSON, S_TITER, S_ROCHE_ORTHO, S_TEST, S_RESULT]
+
+PERSON_QUERY = JINJA_ENV.from_string("""
+CREATE TABLE `{{project_id}}.{{dest_dataset_id}}.serology_person`
+LIKE `{{project_id}}.{{source_dataset_id}}.person` AS
```
ksdkalluri, line 33: @nishanthpp93 for these tables aren't we adding the schemas and using them to create the tables? is there any benefit in doing it like this?

ksdkalluri, line 33: @nishanthpp93 for these tables aren't we adding the schemas and using them to create the tables? is there any benefit in doing it like this?

nishanthpp93: Since we don't maintain these tables, this way any changes to the schema in the original dataset made by source devs will automatically be updated in the snapshot as well.

ksdkalluri: That's true, But that also brings in a problem where if the source schema is changed then the queries might fail too right?

nishanthpp93: These queries won't fail because they are schema agnostic. For the person, it is selecting only 4 fields, so if these are changed, it will fail. Any changes to other fields will not cause a failure and it won't affect the snapshot process.

data_steward/tools/snapshot_serology_dataset.py
```diff
@@ -0,0 +1,202 @@
...
+    pipeline_logging.configure(add_console_handler=True)
+
+    # get credentials and create client
+    impersonation_creds = auth.get_impersonation_credentials(
+        args.run_as_email, CDR_SCOPES)
+
+    bq_client = BigQueryClient(args.project_id, credentials=impersonation_creds)
+
+    # Create serology dataset
+    dataset_id = create_serology_snapshot(bq_client, args.snapshot_date,
```
ksdkalluri, line 193: args.snapshopt_date should be released tag.

ksdkalluri (changes requested):

data_steward/tools/snapshot_serology_dataset.py
nishanthpp93: Thanks, updated it

ksdkalluri (approved):
Looks good to me, good to merge once the integration tests pass.

PR #1200, opened 2022-04-04
[DC-2311] Add visit_detail to affected tables list in EnsureDateDateTimeConsistency CR

ksdkalluri:


data_steward/cdr_cleaner/cleaning_rules/ensure_date_datetime_consistency.py
```diff
@@ -57,6 +57,10 @@
     common.VISIT_OCCURRENCE: {
         'visit_start_datetime': 'visit_start_date',
         'visit_end_datetime': 'visit_end_date'
+    },
+    common.VISIT_DETAIL: {
+        'visit_detail_start_date': 'visit_detail_start_datetime',
+        'visit_detail_end_date': 'visit_detail_end_datetime'
```
nishanthpp93, line 63: This may need to be flipped to date time -> date

nishanthpp93 (changes requested):
Added a suggestion. I'd also suggest updating the test with visit_detail

data_steward/cdr_cleaner/cleaning_rules/ensure_date_datetime_consistency.py
nishanthpp93: Yeah If you verified the queries this is fine.

nishanthpp93 (approved):
Looks good!

PR #1201, opened 2022-04-04
[DC-1799] Converted retraction dates to timestamps

jp3477:


data_steward/retraction/retract_deactivated_pids.py
```diff
@@ -48,23 +48,23 @@
 {% endif %}
 
 {% if has_start_date %}
-WHERE (COALESCE({{end_date}}, EXTRACT(DATE FROM {{end_datetime}}),
-    {{start_date}}, EXTRACT(DATE FROM {{start_datetime}})) >= d.deactivated_date
```
hiro-mishima, line 52: Please update the column name from `deactivated_date` to `deactivated_datetime`, as @molly-giddings's DC-1795 will be also updating the column name.

hiro-mishima (changes requested):
Thanks for working on this, @jp3477!
I added one comment for changes. Can you please check it out? The integration test will NOT likely pass still, as this development is dependent on DC-1795. We can work on that once DC-1795 is almost complete.

data_steward/retraction/retract_deactivated_pids.py
jp3477: @hiro-mishima Thanks for cross-referencing these two tickets. Do you think it would be best to hold off on the final work for this until DC-1795 is merged?

hiro-mishima:
Closing this PR as this change was included in DC-1795. Thanks!

PR #1202, opened 2022-04-04
[DC-2217] Replace bq.get_table_schema(...) calls in tests/integration_tests

lledezma:
- Replaces all bq.get_table_schema(...) calls with BigQueryClient.get_table_schema(...) in test/integration_tests folder files.

### DIFF
--- a/tests/integration_tests/data_steward/cdr_cleaner/cleaning_rules/ppi_branching_test.py
+++ b/tests/integration_tests/data_steward/cdr_cleaner/cleaning_rules/ppi_branching_test.py
@@ -7,6 +7,7 @@ from google.cloud import bigquery
 from google.cloud.bigquery import Table, TimePartitioning
 
 import app_identity
+from gcloud.bq import BigQueryClient
 import bq_utils
 from cdr_cleaner.cleaning_rules import ppi_branching
 from cdr_cleaner.cleaning_rules.ppi_branching import PpiBranching
@@ -98,7 +99,9 @@ class Observation(object):
     Helper class to initialize test observation rows
     """
 
-    SCHEMA = bq.get_table_schema('observation')
+    project_id = app_identity.get_application_id()
+    bq_client = BigQueryClient(project_id)
+    SCHEMA = bq_client.get_table_schema('observation')
     """List of schema fields for observation table"""
 
     _FIELD_DEFAULTS = dict(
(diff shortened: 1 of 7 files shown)


tests/integration_tests/data_steward/utils/participant_summary_requests_test.py
```diff
@@ -194,10 +194,12 @@ def test_get_deactivated_participants(self, mock_get_session, mock_token):
...
         fq_table = f'{self.project_id}.{self.dataset_id}.{DIGITAL_HEALTH_SHARING_STATUS}'
         self.fq_table_names.append(fq_table)
-        table = Table(fq_table,
-                      schema=get_table_schema(DIGITAL_HEALTH_SHARING_STATUS))
+        table = Table(
+            fq_table,
+            schema=bq_client.get_table_schema(DIGITAL_HEALTH_SHARING_STATUS))
         table.time_partitioning = TimePartitioning(
             type_=TimePartitioningType.DAY)
         _ = self.client.create_table(table, exists_ok=True)
```
nishanthpp93, line 205: I see `bq_client` and `self.client` being used here. Are they the same?

tests/integration_tests/data_steward/cdr_cleaner/cleaning_rules/ppi_branching_test.py
```diff
@@ -98,7 +99,9 @@ class Observation(object):
     Helper class to initialize test observation rows
     """
 
-    SCHEMA = bq.get_table_schema('observation')
+    project_id = app_identity.get_application_id()
+    bq_client = BigQueryClient(project_id)
+    SCHEMA = bq_client.get_table_schema('observation')
```
nishanthpp93, line 104: This feels odd (code smell?) because `get_table_schema` is a static/non-gcloud function but now a BQ client needs to be instantiated to run it. Not a big deal, so this is fine for now.

nishanthpp93, line 104: This feels odd (code smell?) because `get_table_schema` is a static/non-gcloud function but now a BQ client needs to be instantiated to run it. Not a big deal.

nishanthpp93 (commented):
Looks good, but added a question and a comment.

tests/integration_tests/data_steward/utils/participant_summary_requests_test.py
lledezma: It is. No need to create another instance. Good catch.

nishanthpp93 (approved):
Looks good!

PR #1203, opened 2022-04-07
[DC-2278] Adjust Controlled Tier concept suppression list

ksdkalluri:
* Added a CSV file with missing standard concepts 
* Added a CSV file with replaced concepts
* Added a cleaning rule to suppress missing privacy standard concepts and replaced concepts.


data_steward/cdr_cleaner/cleaning_rules/deid/missing_and_replaced_concept_suppression.py
```diff
@@ -0,0 +1,92 @@
...
+                         table_namer=table_namer)
+
+    def get_suppressed_concept_ids(self):
+        with open(MISSING_PRIVACY_STANDARD_CONCEPTS_PATH) as f:
+            missing_concept_ids_df = read_csv(f, delimiter=',')
+            missing_concept_ids = missing_concept_ids_df['concept_id'].to_list()
+        with open(REPLACED_PRIVACY_CONCEPTS_PATH) as f:
+            replaced_concept_ids_df = read_csv(f, delimiter=',')
+            replaced_concept_ids = replaced_concept_ids_df[
+                'concept_id'].to_list()
```
lrwb-aou, line 63: We are dropping the requirement to provide the Missing Standard Concept IDs.  I've got Brenda checking with privacy about 4 RT replacement concepts.  I'm going to check the list of replacement concepts you've got against what we already know about CT and RT privacy to see if anything else might have slipped through.  But you should be able to make this change about dropping the Standard Concepts csv file.

data_steward/cdr_cleaner/cleaning_rules/deid/missing_and_replaced_concept_suppression.py
```diff
@@ -0,0 +1,92 @@
+"""
+Ensures that all the Missing Standard concepts and the replaced concepts in vocabulary are being suppressed
+ in Controlled tier datasets
```
lrwb-aou, line 3: This comment, the class name, and maybe the module name could/should be updated.

lrwb-aou, line 3: This comment, the class name, and maybe the module name could/should be updated based on the requirements change.

lrwb-aou (changes requested):
Please make the changes based on the requirements change.

data_steward/cdr_cleaner/cleaning_rules/deid/missing_and_replaced_concept_suppression.py
```diff
@@ -0,0 +1,88 @@
...
+import constants.cdr_cleaner.clean_cdr as cdr_consts
+from cdr_cleaner.cleaning_rules.deid.concept_suppression import \
+    AbstractInMemoryLookupTableConceptSuppression
+
+LOGGER = logging.getLogger(__name__)
+ISSUE_NUMBERS = ['DC2278']
+
+
+class ControlledTierReplacedConceptSuppression(
+        AbstractInMemoryLookupTableConceptSuppression):
```
lrwb-aou, line 30: I think this may be ok for now.  I don't think it will hit any quota limits.  But if the list were to expand, it could.  That is why I was suggesting to extend the `AbstractBqLookupTableConceptSuppression` class.  You don't have to create the table via a query. You could load it via the `create_suppression_lookup_table` method since it has access to a client object.  But, this is an optional change.

lrwb-aou (changes requested):
I think you may need to change the line in `data_steward/resources.py`.  The other suggestion is an optional suggestion that would let us keep the concepts we are suppressing with in a sandbox lookup table.

data_steward/cdr_cleaner/cleaning_rules/deid/missing_and_replaced_concept_suppression.py
ksdkalluri: you are right, updated CR to store the concept_ids in the table.

data_steward/cdr_cleaner/cleaning_rules/deid/ct_replaced_concept_suppression.py
```diff
@@ -0,0 +1,125 @@
...
+        df = pd.read_csv(REPLACED_PRIVACY_CONCEPTS_PATH)
+        dataset_ref = bigquery.DatasetReference(self.project_id,
+                                                self.sandbox_dataset_id)
+        table_ref = dataset_ref.table(self.concept_suppression_lookup_table)
+        result = client.load_table_from_dataframe(df, table_ref).result()
+
+        if hasattr(result, 'errors') and result.errors:
+            LOGGER.error(f"Error running job {result.job_id}: {result.errors}")
+            raise GoogleCloudError(
+                f"Error running job {result.job_id}: {result.errors}")
```
lrwb-aou, line 70: 👍 nice error checking!

lrwb-aou (approved):
This looks good to me!

PR #1204, opened 2022-04-07
[DC-2314] [RT] add standard and replacement suppressed concept_ids

brendagutman:


lrwb-aou (commented):
These map to 37 codes.  ✅ 
I did notice that 13 appear to be deprecated.  Did you expect this?  It won't hurt to leave them in, I just want to make sure this is expected.

brendagutman:
Those are most likely the concepts that replaced something, and then were replaced themselves. I can check on those if you want.

lrwb-aou:
Please confirm that.

brendagutman:
These are concepts that are equivalent to a suppressed concept. They are invalid and not standards.

lrwb-aou:
@brendagutman one other question, why are there only 37 added concepts?  I expected more.

brendagutman:
I checked the concepts I had listed against the file for duplicates. I hadn't done that previously.

lrwb-aou (approved):
This looks good!

PR #1205, opened 2022-04-07
[DC-2313] Replace bq.get_table_schema(...) calls in data_steward/validation

lledezma:


ksdkalluri (approved):
Looks good to me and I see the integration tests were passing well.

PR #1206, opened 2022-04-11
[DC-2188] Relocate bq.py helper functions

lledezma:
- Relocates `to_scalar(...)`, `get_table_count(...)`, `list_tables(...)`, and `copy_datasets(...)` helper functions from data_steward/utils/bq.py to be methods of the new `BigQueryClient`.

### DIFF
--- a/data_steward/gcloud/bq/__init__.py
+++ b/data_steward/gcloud/bq/__init__.py
@@ -211,8 +211,59 @@ class BigQueryClient(Client):
         :return: tables contained within the requested dataset
         """
         _MAX_RESULTS_PADDING = 100
-        table_count = super(BigQueryClient,
-                            self).list_tables(dataset=dataset).num_results
+        table_count = self.get_table_count(dataset)
         return super(BigQueryClient, self).list_tables(dataset=dataset,
                                                        max_results=table_count +
                                                        _MAX_RESULTS_PADDING)
+
+    def get_table_count(self, dataset: bigquery.DatasetReference) -> int:
+        """
+        Get the number of tables currently in a specified dataset
+
+        :param client: active bigquery client
+        :param dataset: the dataset
+        :return: number of tables
+        :raises:
+            google.cloud.exceptions.GoogleCloudError:
+                If the job failed.
+            concurrent.futures.TimeoutError:
+                If the job did not complete in the given timeout.
+        """
+
+        TABLE_COUNT_TPL = JINJA_ENV.from_string(
+            "SELECT COUNT(1) table_count FROM `{{dataset.project}}.{{dataset.dataset_id}}.__TABLES__`"
+        )
+        """Query template to retrieve the number of tables in a dataset.
+        Requires parameter `dataset`: :class:`DatasetReference` and
+        yields a scalar result with column `table_count`: :class:`int`."""
+
+        q = TABLE_COUNT_TPL.render(dataset=dataset)
+        return self.to_scalar(self.query(q))
+
+    def to_scalar(
+        self, result: typing.Union[bigquery.table.RowIterator,
+                                   bigquery.QueryJob]
+    ) -> typing.Any:
+        """
+        Get a scalar query result
+
+        :param result: a query job or a resultant :class:`bigquery.table.RowIterator`
+        :return: the singular result value
+        """
+        row_iter = None
+        if isinstance(result, bigquery.table.RowIterator):
+            row_iter = result
+        elif isinstance(result, bigquery.QueryJob):
+            row_iter = result.result()
+        else:
+            raise ValueError(
+                f'Scalar result requires a RowIterator or QueryJob '
+                f'but `{type(result)}` was supplied.')
+        if row_iter.total_rows != 1:
+            raise ValueError(f'Cannot get scalar result from '
+                             f'row iterator with {row_iter.total_rows} rows.')
+        _, row = next(enumerate(row_iter))
+        if len(row_iter.schema) == 1:
+            return row[0]
+
+        return dict(row.items())
--- a/data_steward/gcloud/bq/__init__.py
+++ b/data_steward/gcloud/bq/__init__.py
@@ -220,7 +220,6 @@ class BigQueryClient(Client):
         """
         Get the number of tables currently in a specified dataset
 
-        :param client: active bigquery client
         :param dataset: the dataset
         :return: number of tables
         :raises:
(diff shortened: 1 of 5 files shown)


data_steward/gcloud/bq/__init__.py
```diff
@@ -184,3 +183,86 @@ def define_dataset(self, dataset_id: str, description: str,
         dataset.location = "US"
 
         return dataset
+
+    def copy_datasets(self, input_dataset: str, output_dataset: str):
```
nishanthpp93, line 187: nit: Since only one dataset can be copied at a time
```suggestion
    def copy_dataset(self, input_dataset: str, output_dataset: str):
```

data_steward/gcloud/bq/__init__.py
```diff
@@ -184,3 +183,86 @@ def define_dataset(self, dataset_id: str, description: str,
...
+        Copies tables from source dataset to a destination datasets
+
+        :param input_dataset: name of the input dataset
+        :param output_dataset: name of the output dataset
+        :return:
+        """
+        # Copy input dataset tables to backup and staging datasets
+        tables = super(BigQueryClient, self).list_tables(input_dataset)
+        for table in tables:
+            staging_table = f'{output_dataset}.{table.table_id}'
```
nishanthpp93, line 198: nit: I think this may need to be fully qualified
```suggestion
            staging_table = f'{self.project}.{output_dataset}.{table.table_id}'
```

data_steward/gcloud/bq/__init__.py
```diff
@@ -184,3 +183,86 @@ def define_dataset(self, dataset_id: str, description: str,
...
+
+    def list_tables(
+        self, dataset: bigquery.DatasetReference
+    ) -> typing.Iterator[bigquery.table.TableListItem]:
+        """
+        List all tables in a dataset
+
+        NOTE: Ensures all results are retrieved by first getting total
+        table count and setting max_results in list tables API call
+
```
nishanthpp93, line 209: It think it would be good to add the behavior without _MAX_RESULTS_PADDING as a note, and why it is needed.

data_steward/gcloud/bq/__init__.py
```diff
@@ -184,3 +183,86 @@ def define_dataset(self, dataset_id: str, description: str,
...
+        Requires parameter `dataset`: :class:`DatasetReference` and
+        yields a scalar result with column `table_count`: :class:`int`."""
+
+        q = TABLE_COUNT_TPL.render(dataset=dataset)
+        return self.to_scalar(self.query(q))
+
+    def to_scalar(
+        self, result: typing.Union[bigquery.table.RowIterator,
+                                   bigquery.QueryJob]
+    ) -> typing.Any:
```
nishanthpp93, line 245: This is great!

tests/unit_tests/data_steward/gcloud/bq/bq_test.py
```diff
@@ -40,6 +45,109 @@ def _get_all_field_types(self,) -> typing.FrozenSet[str]:
                             all_field_types.add(field.get('type'))
         return frozenset(all_field_types)
 
+    def list_item_from_table_id(self, table_id: str) -> TableListItem:
+        """
+        Get a table list item as returned by :meth:`bigquery.Client.list_tables` 
+        from a table ID
+        
+        :param table_id: A table ID including project ID, dataset ID, and table ID,
```
nishanthpp93, line 53: nit: For these I'd suggest using `fq_table_id` instead of just `table_id`. This refers to the fully qualified table ID as `project_id.dataset_id.table_id`.
```suggestion
    def list_item_from_table_id(self, fq_table_id: str) -> TableListItem:
        """
        Get a table list item as returned by :meth:`bigquery.Client.list_tables` 
        from a fully qualified table ID
        
        :param fq_table_id: A fully qualified table ID including project ID, dataset ID, and table ID, 
```

tests/unit_tests/data_steward/gcloud/bq/bq_test.py
```diff
@@ -40,6 +45,109 @@ def _get_all_field_types(self,) -> typing.FrozenSet[str]:
...
+        :return: the ordered dict
+        """
+        if len(d) == 1:
+            return OrderedDict(d)
+        else:
+            if key_order is None:
+                raise ValueError(
+                    'Parameter key_order is required in order to convert'
+                    ' a dict with multiple items to OrderedDict')
+        return OrderedDict((key, d[key]) for key in key_order)
```
nishanthpp93, line 107: This seems very specific. If it is not used much (maybe 1/2 instances), I'd suggest removing this function and performing this where it is used. This is not really relevant to the BQ object.

tests/unit_tests/data_steward/gcloud/bq/bq_test.py
```diff
@@ -40,6 +45,109 @@ def _get_all_field_types(self,) -> typing.FrozenSet[str]:
...
+        :param value: value of the field
+        :return: an appropriate schema field object
+        """
+        _TYPE_TO_FIELD_TYPE = {
+            str(int): 'INT64',
+            str(str): 'STRING',
+            str(float): 'FLOAT64',
+            str(datetime.date): 'DATE',
+            str(datetime.datetime): 'TIMESTAMP'
+        }
```
nishanthpp93, line 144: This mapping would be great to have in a constants file. Maybe as Python type to BigQuery type to be more specific?

nishanthpp93 (commented):
Great work on the tests, Luis! Added a few suggestions

data_steward/gcloud/bq/__init__.py
lledezma: So, keep the method as it is, and just remove lines 207 and 208?

nishanthpp93: I'd say keep the method as is, and keep lines 207 and 208, but also add "Without setting max_results, the API has a bug causing it to randomly return 0 tables"

tests/unit_tests/data_steward/gcloud/bq/bq_test.py
lledezma: I thought I would put it inside `def _field_from_key_value` since it's used only there. Maybe relocate it in future if there's more use for it?

nishanthpp93: Yes, that works too. I think there are other usages but they have their own dictionary. Maybe it can be consolidated in a separate ticket.

nishanthpp93 (approved):
Looks good!

nishanthpp93 (approved):

PR #1207, opened 2022-04-11
[DC-1795] Store deactivated_date as a timestamp for deactivated participants

molly-giddings:


hiro-mishima (changes requested):
Thanks for working on this ticket, @molly-giddings!
I ran the integration test for this PR and found the following things we need to update. (The integration test skips everything unless there is a corresponding PR. So, technically the integration test was never run until now.) Can you please check them and update them accordingly? 

1. Now we have `deactivated_datetime`, not `deactivated_date`, please update this insert statement accordingly: https://github.com/all-of-us/curation/blob/develop/tests/integration_tests/data_steward/retraction/retract_deactivated_pids_test.py#L18 and 
2. For the same reason, please update this test function: https://github.com/all-of-us/curation/blob/develop/tests/integration_tests/data_steward/cdr_cleaner/cleaning_rules/remove_participant_data_past_deactivation_date_test.py#L113

Not all the errors might be solved because we have the dependent PR for DC-1799, but these changes will get us closer to what we want.

molly-giddings:
Thank you, @hiro-mishima !  I've made the suggested updates but expect the retract_deactivated_pids cleaning rule to fail until DC-1799 is merged in.

data_steward/retraction/retract_deactivated_pids.py
```diff
@@ -48,23 +48,23 @@
...
-WHERE datetime >= PARSE_DATETIME('%F', CAST(d.deactivated_date as STRING))
+WHERE datetime >= PARSE_DATETIME('%F', CAST(d.deactivated_datetime as STRING))
 {% elif table_ref.table_id in ['drug_era', 'condition_era', 'dose_era', 'payer_plan_period', 'observation_period']  %}
-WHERE COALESCE(DATE({{table_ref.table_id + '_end_date'}}),
-DATE({{table_ref.table_id + '_start_date'}})) >= d.deactivated_date
+WHERE COALESCE(TIMESTAMP({{table_ref.table_id + '_end_date'}}),
+TIMESTAMP({{table_ref.table_id + '_start_date'}})) >= d.deactivated_datetime
 {% else %}
-WHERE COALESCE({{date}}, EXTRACT(DATE FROM {{datetime}})) >= d.deactivated_date
+WHERE COALESCE({{datetime}}, TIMESTAMP({{date}})) >= d.deactivated_datetime
```
molly-giddings, line 67: @hiro-mishima, @jp3477  - For this test to pass, I had to flip the order of the fields in the COALESCE on line 67.  Now that the comparison is based on a TIMESTAMP value, 3 test records should have been excluded as the deactivation time was 1:00:00 UTC.  When the created timestamp from date (TIMESTAMP({{date}})) is used as the primary field in the COALESCE, the comparison time for the records becomes 0:00:00 UTC, which is before the deactivation time and the records will not be excluded.  

It seems likely lines 51, 52, and 58 might need the same treatment, but I would welcome your input.

hiro-mishima: Thanks for checking, @molly-giddings!

1. Casting Date type to DateTime type here may cause an incorrect condition check. I think this line should be updated to something like this:
`WHERE ({{datetime}} IS NOT NULL AND {{datetime}} >= d.deactivated_datetime)`
`OR ({{datetime}} IS NULL AND {{date}} >= DATE(d.deactivated_datetime))`

2. 51, 52, and 58 should be updated similarly.

3. For the same reason, 54, 60, and 64-65 should be Date vs. Date comparisons, not DateTime vs. DateTime.

4. Would it be possible to add more test data to `test_queries_to_retract_from_fake_dataset` to cover each of the patterns?

molly-giddings: @hiro-mishima , I have made changes for all of your suggestions.  I also had to break out the conditions starting on line 66 into two different groups as `drug_era`, `condition_era`, and `dose_era `now have start_date and end_date as TIMESTAMP rather than DATE.

data_steward/retraction/retract_deactivated_pids.py
```diff
@@ -48,23 +48,30 @@
...
+OR (death_datetime IS NULL AND death_date >= DATE(d.deactivated_datetime))
 {% elif table_ref.table_id in ['activity_summary', 'heart_rate_summary'] %}
-WHERE date >= d.deactivated_date
+WHERE date >= DATE(d.deactivated_datetime)
 {% elif table_ref.table_id in ['heart_rate_minute_level', 'steps_intraday']  %}
-WHERE datetime >= PARSE_DATETIME('%F', CAST(d.deactivated_date as STRING))
-{% elif table_ref.table_id in ['drug_era', 'condition_era', 'dose_era', 'payer_plan_period', 'observation_period']  %}
-WHERE COALESCE(DATE({{table_ref.table_id + '_end_date'}}),
-DATE({{table_ref.table_id + '_start_date'}})) >= d.deactivated_date
+WHERE datetime >= PARSE_DATETIME('%F', CAST(d.deactivated_datetime as STRING))
```
hiro-mishima, line 65: I think we no longer need to use `PARSE_DATETIME` here as the column is already in DateTime format.

hiro-mishima (approved):
Thank you so much for working on this, @molly-giddings! It looks great. I left one minor comment. You can either update the script again and merge it to develop or just skip that comment and merge it. It's up to you.

PR #1210, opened 2022-04-14
[DC-823] [DC-1192] Update ID Validation CR to New Base Class

hiro-mishima:
- Refactored the CR to the new base class
- Added the integration test
- Removed the unit test as everything is covered in the integration test and it is obsolete
- Updated the CR logic based on [this comment in the ticket](https://precisionmedicineinitiative.atlassian.net/browse/DC-823?focusedCommentId=116415)


data_steward/cdr_cleaner/cleaning_rules/remove_non_matching_participant.py
```diff
@@ -35,214 +45,275 @@
...
+        :param field_names: a list of field names for counting `missing`s
+        :return: Multiple CAST statements joined by ' + '.
+        """
+        joined_column_expr = ' + '.join([
+            CAST_MISSING_COLUMN.render(column=field_name)
+            for field_name in field_names
+        ])
+        return joined_column_expr
+
+    def get_non_match_participants(self,
```
nishanthpp93, line 165: Couple of changes here: 
1. Remove all instances of bq_utils (being deprecated) and use client as a passed parameter that has these built in, so you can just call them as methods.
2. Call this function from inside setup_rule since that provides you the client.
3. Ideally you don't need the actual pids as a list of numbers. Use a common table expression to do something like 
`WHERE person_id IN (SELECT person_id FROM table WHERE match_criteria)`

data_steward/cdr_cleaner/cleaning_rules/remove_non_matching_participant.py
```diff
@@ -35,214 +45,275 @@
...
+                        f'{hpo_id} submitted the participant_match table, and all data is validated.'
+                    )
+
+            else:
+                LOGGER.info(f'{hpo_id} is missing the participant_match table.')
+                non_matching_person_ids.extend(
+                    self.get_non_match_participants(self.validation_dataset_id,
+                                                    hpo_id))
+
+        if non_matching_person_ids:
```
nishanthpp93, line 307: This looks like logic that is needed for the setup, so would be ideal in setup_rule. I'd say setup_rule should store the results for each hpo in a dictionary, and get_query_specs will use this to determine how many queries to return.

nishanthpp93 (changes requested):
Overall looks great and does the job. Added a few suggestions on how it's done. Ideally setup_rule should handle a bulk of the querying before calling get_query_specs. Let me know if you have any questions and we can set up a call.

data_steward/cdr_cleaner/cleaning_rules/remove_non_matching_participant.py
```diff
@@ -35,214 +45,275 @@
...
+    Removes records with person_ids that are not validated by sites and non-matching.
     """
 
-    # get the the hpo specific <hpo_id>_identity_match
-    identity_match_table = bq_utils.get_table_id(hpo_id, IDENTITY_MATCH)
-    result = []
-    fq_identity_match_table = f'{client.project}.{validation_dataset_id}.{identity_match_table}'
-    if not exist_identity_match(client, fq_identity_match_table):
+    def __init__(self, project_id, dataset_id, sandbox_dataset_id,
+                 ehr_dataset_id, validation_dataset_id):
```
nishanthpp93, line 87: I think `table_namer=None` should be a parameter here passed to super `__init__`.

hiro-mishima:
@nishanthpp93 Thanks for the comments and the follow-up meeting! I updated the scripts accordingly. Could you please have a look? It ended up a significant change from the last review. Thanks again for reviewing!

data_steward/cdr_cleaner/cleaning_rules/remove_non_matching_participant.py
```diff
@@ -31,218 +38,247 @@
...
+        NOTE: self.sandbox_table_for(NOT_MATCH) is not included in this list
+        because it is a temporary lookup table, not a backup table.
+        """
+        return [
+            self.sandbox_table_for(table)
+            for table in remove_pids.get_tables_with_person_id(
+                self.project_id, self.dataset_id)
+        ]
+
+    def exist_table(self, client: BigQueryClient, project_id, dataset_id,
```
hiro-mishima, line 201: We can remove this function when the following JIRA ticket is implemented. Shall we wait for that?
https://precisionmedicineinitiative.atlassian.net/browse/DC-2460

hiro-mishima, line 201: We can remove this function when the following JIRA ticket is implemented. Shall we wait for that?
https://precisionmedicineinitiative.atlassian.net/browse/DC-2460

nishanthpp93: Yeah that's fine.

hiro-mishima: @nishanthpp93 DC-2460 is done and I updated the scripts accordingly.

tests/integration_tests/data_steward/cdr_cleaner/cleaning_rules/remove_non_matching_participant_test.py
```diff
@@ -0,0 +1,324 @@
...
+
+    def tearDown(self):
+        pass
+
+    @classmethod
+    def tearDownClass(cls):
+        test_util.delete_all_tables(cls.dataset_id)
+        test_util.delete_all_tables(cls.ehr_dataset_id)
+        test_util.delete_all_tables(cls.validation_dataset_id)
+        test_util.delete_all_tables(cls.sandbox_id)
```
nishanthpp93, line 324: I hope we can get rid of these and use a custom method added to the client, but that's for later.

tests/integration_tests/data_steward/cdr_cleaner/cleaning_rules/remove_non_matching_participant_test.py
```diff
@@ -0,0 +1,324 @@
...
+from cdr_cleaner.cleaning_rules.remove_non_matching_participant import RemoveNonMatchingParticipant
+from common import JINJA_ENV, IDENTITY_MATCH, OBSERVATION, PARTICIPANT_MATCH, PERSON
+from validation.participants.create_update_drc_id_match_table import create_drc_validation_table
+from tests import test_util
+from tests.integration_tests.data_steward.cdr_cleaner.cleaning_rules.bigquery_tests_base import BaseTest
+import resources
+
+HPO_1, HPO_2, HPO_3, HPO_4 = 'fake', 'pitt', 'nyc', 'chs'
+
+POPULATE_STATEMENTS = {
```
nishanthpp93, line 22: Great work on the tests! Took me a while to go through and they look good.

nishanthpp93 (commented):
Overall looks good! Added one comment. Let me know your thoughts.

data_steward/cdr_cleaner/cleaning_rules/remove_non_matching_participant.py
```diff
@@ -1,249 +1,254 @@
...
+        pass
+
+    def get_sandbox_tablenames(self) -> list:
+        """
+        Return a list table names created to backup deleted data.
+        """
+        return [
+            self.sandbox_table_for(table)
+            for table in remove_pids.get_tables_with_person_id(
+                self.project_id, self.dataset_id)
```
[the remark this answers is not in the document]
hiro-mishima: Thanks, Nishanth! I agree with you - it is better to define affected_tables and follow the standard cleaning rule format.

hiro-mishima:
@nishanthpp93 Thanks again for the review! I updated the script again. Could you please see if it looks good? If not, we can re-connect once you are back.

tests/integration_tests/data_steward/cdr_cleaner/cleaning_rules/remove_non_matching_participant_test.py
```diff
@@ -176,6 +176,8 @@ def setUpClass(cls):
             ehr_dataset_id=cls.ehr_dataset_id,
             validation_dataset_id=cls.validation_dataset_id)
 
+        cls.rule_instance.setup_rule(cls.client)
+
```
nishanthpp93, line 180: You shouldn't have to call this here. It will be called as part of `default_test`. Does it fail without this?

hiro-mishima: Yes, it fails. At line 181, `cls.rule_instance.get_sandbox_tablenames()` is called. And for that, `affected_tables` is needed.

nishanthpp93: I think it may have to do with `self.affected_tables` being empty. Ideally in the `super().__init__()` method for the cleaning rule, `affected_tables=[table for table in CDM_TABLES if any(field['name']='person_id' for field in resources.fields_for(table))]` (pseudo code) can be a parameter.

hiro-mishima: Oh, OK! Instead of calling the client, we can use resources.fields_for(table)). That never occurred to me. Let me try that instead.

tests/integration_tests/data_steward/cdr_cleaner/cleaning_rules/remove_non_matching_participant_test.py
```diff
@@ -0,0 +1,325 @@
...
+                    (404, 0, 1994, 0, 0),
+                ]
+            },
+            {
+                'name':
+                    f'{OBSERVATION}',
+                'fq_table_name':
+                    f'{self.project_id}.{self.dataset_id}.{OBSERVATION}',
+                'fq_sandbox_table_name':
+                    f'{self.project_id}.{self.sandbox_id}.{OBSERVATION}_{"_".join(self.rule_instance.issue_numbers)}',
```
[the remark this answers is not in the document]
hiro-mishima: Thanks for pointing this out! How `sandbox_and_remove_pids` defines sandbox table name was a little off from the standard.

nishanthpp93 (approved):
Looks good! Thanks for updating, @hiro-mishima. I think at some point sandbox_and_remove_pids needs to be converted to a CR and this CR needs to inherit from that, but that should be in a separate ticket.

PR #1211, opened 2022-04-14
[DC-2115] add api logging setup to the admin service

lrwb-aou:
* this should allow logging to occur
* logging will allow logging metrics to happen
* logging metrics will allow GCP notifications


ksdkalluri (approved):
Looks good to me!

PR #1212, opened 2022-04-15
[DC-2315] Replace bq.copy_datasets(...) calls in data_steward/tools

lledezma:
- Replaces all `bq.copy_datasets(...)` calls with `BigQueryClient.copy_dataset(...)` in data_steward/tools folder.
- Updates associated tests.


hiro-mishima (approved):
Thanks, Luis! It looks good to me.

PR #1216, opened 2022-04-18
[DC-2251] end_datetimes earlier than start_datetimes in drug_exposure

jp3477:
Modify datetime consistency cleaning rule to assign end_datetime fields a time component off 11:59:59, in order to avoid falling before the corresponding start_datetime.


ksdkalluri (changes requested):

ksdkalluri (approved):
Looks Good!

PR #1217, opened 2022-04-18
[DC-2324] Remove "reply-to" email address in results.html email from Mandrill

jp3477:


jp3477:
@nishanthpp93, the test should be fixed now.

nishanthpp93 (approved):
Looks good!

PR #1219, opened 2022-04-18
[DC-2365] [DC-2370] Remove visit_detail_id and add gender_concept_id to Demographic columns and add a config file for visit_detail to suppress ICD9 and ICD10 codes

ksdkalluri:


lrwb-aou (approved):
I think this looks good.

PR #1223, opened 2022-04-20
[DC-1832] Accept path to hpo_site_mappings.csv as argument to add_hpo.py script

jp3477:


nishanthpp93 (commented):
Overall looks good! Only suggestion is to update docstrings for all the sub functions which use the new param as well.

jp3477:
@nishanthpp93 Thanks for catching that, I added the docstrings.

nishanthpp93 (approved):

PR #1225, opened 2022-04-20
[DC-2289] Add `trace.append` to CDR_SCOPES

ksdkalluri:


nishanthpp93 (approved):
Worked with the CDR run!

PR #1230, opened 2022-04-21
[DC-2382] visit_detail and note_nlp tables dropped by remove_extra_tables CR

nishanthpp93:
* [DC-2382] Use dynamic list of final tables

[DC-2382]: https://precisionmedicineinitiative.atlassian.net/browse/DC-2382?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/data_steward/cdr_cleaner/cleaning_rules/remove_extra_tables.py
+++ b/data_steward/cdr_cleaner/cleaning_rules/remove_extra_tables.py
@@ -17,79 +17,11 @@ from google.cloud import bigquery
 from cdr_cleaner.cleaning_rules.base_cleaning_rule import BaseCleaningRule
 from constants.cdr_cleaner import clean_cdr as cdr_consts
 from common import JINJA_ENV
+from resources import cdm_schemas, has_domain_table_id
 from utils import pipeline_logging
 from utils.bq import list_tables
 
 LOGGER = logging.getLogger(__name__)
-
-FINAL_TABLES = [
-    '_cdr_metadata',
-    'attribute_definition',
-    'attribute_definition_ext',
-    'care_site',
-    'care_site_ext',
-    'cdm_source',
-    'cohort',
-    'cohort_attribute',
-    'cohort_definition',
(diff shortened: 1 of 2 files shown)


data_steward/cdr_cleaner/cleaning_rules/remove_extra_tables.py
```diff
@@ -123,11 +55,22 @@ def __init__(self,
         this SQL, append them to the list of Jira Issues.
         DO NOT REMOVE ORIGINAL JIRA ISSUE NUMBERS!
         """
-        desc = 'Remove any tables that are not OMOP, OMOP extension, or Vocabulary tables.'
+        desc = 'Remove any tables that are not OMOP, OMOP extension, Vocabulary or AOU custom tables.'
+        # Use custom cdr_metadata instead of metadata
+        cdm_achilles_vocab_tables = list(
+            set(
+                cdm_schemas(include_achilles=True, include_vocabulary=True).
+                keys()) - {'metadata'}) + ['_cdr_metadata']
```
ksdkalluri, line 63: you would need a metadata table as it is one of the tables added in CDM 5.3 https://ohdsi.github.io/CommonDataModel/cdm53.html#Metadata_Tables

ksdkalluri (changes requested):

data_steward/cdr_cleaner/cleaning_rules/remove_extra_tables.py
nishanthpp93: I thought we were using `_cdr_metadata` instead of that. Are we using both?

ksdkalluri:
@nishanthpp93 we are using the _cdr_metadata for info, but we should include all the tables in CDM although they are empty right.

ksdkalluri (approved):
changes look good!

PR #1234, opened 2022-04-22
[DC-2389] Identifying field suppression CR needs to use existing client

nishanthpp93:
* [DC-2389] Identifying field suppression CR needs to use existing client

[DC-2389]: https://precisionmedicineinitiative.atlassian.net/browse/DC-2389?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/data_steward/cdr_cleaner/cleaning_rules/identifying_field_suppression.py
+++ b/data_steward/cdr_cleaner/cleaning_rules/identifying_field_suppression.py
@@ -34,9 +34,9 @@ from constants import bq_utils as bq_consts
 from constants.cdr_cleaner import clean_cdr as cdr_consts
 from common import JINJA_ENV, CDM_TABLES
 from utils import pipeline_logging
-from utils.bq import get_table_schema
 from cdr_cleaner.cleaning_rules.base_cleaning_rule import BaseCleaningRule, query_spec_list
 from cdr_cleaner.cleaning_rules.table_suppression import TableSuppression
+from resources import fields_for
 
 LOGGER = logging.getLogger(__name__)
 
@@ -104,23 +104,23 @@ class IDFieldSuppression(BaseCleaningRule):
         queries = []
         schemas = {}
         for table in self.affected_tables:
-            schemas[table] = get_table_schema(table)
+            schemas[table] = fields_for(table)
         for table in self.affected_tables:
             statements = []
             for item in schemas[table]:
-                if item.name in fields:
-                    if item.mode.lower() == 'nullable':
+                if item.get('name') in fields:
+                    if item.get('mode').lower() == 'nullable':
                         value = 'NULL'
-                    elif item.field_type.lower() == 'integer':
+                    elif item.get('type').lower() == 'integer':
                         value = 0
-                    elif item.field_type.lower() == 'string':
+                    elif item.get('type').lower() == 'string':
                         value = ''
                     else:
                         raise RuntimeError(
-                            f"Required field {item.name} needs to be integer or string type to be replaced"
+                            f"Required field {item.get('name')} needs to be integer or string type to be replaced"
                         )
                     suppression_statement = REPLACE_STRING.render(
-                        suppression_statement=value, field=item.name)
+                        suppression_statement=value, field=item.get('name'))
                     statements.append(suppression_statement)
             if statements:
                 suppression_statement = ', '.join(statements)
--- a/data_steward/cdr_cleaner/cleaning_rules/identifying_field_suppression.py
+++ b/data_steward/cdr_cleaner/cleaning_rules/identifying_field_suppression.py
@@ -102,12 +102,10 @@ class IDFieldSuppression(BaseCleaningRule):
             are optional but the query is required.
         """
         queries = []
-        schemas = {}
         for table in self.affected_tables:
-            schemas[table] = fields_for(table)
-        for table in self.affected_tables:
+            schema = fields_for(table)
             statements = []
-            for item in schemas[table]:
+            for item in schema:
                 if item.get('name') in fields:
                     if item.get('mode').lower() == 'nullable':
                         value = 'NULL'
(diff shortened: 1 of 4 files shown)


data_steward/resources.py
```diff
@@ -174,6 +175,34 @@ def fields_for(table, sub_path=None):
     return fields
 
 
+def get_table_schema(table_name: str, fields=None) -> list:
```
lrwb-aou, line 178: Why is this pulled and copied from https://github.com/all-of-us/curation/blob/develop/data_steward/utils/bq.py#L97.  Why not just use the original code?

lrwb-aou (commented):
I've got some questions and would appreciate some feedback.  I'm glad to be using a single host and getting this fixed, but want to make sure we're describing what we're implementing.

data_steward/cdr_cleaner/cleaning_rules/identifying_field_suppression.py
```diff
@@ -102,11 +103,12 @@ def get_query_specs(self, *args, **keyword_args) -> query_spec_list:
             are optional but the query is required.
         """
         queries = []
+        schemas = {}
+        for table in self.affected_tables:
+            schemas[table] = get_table_schema(table)
         for table in self.affected_tables:
-            bq_client = BigQueryClient(self.project_id)
-            schema = bq_client.get_table_schema(table)
```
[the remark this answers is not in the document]
nishanthpp93: Thanks for bring that up so I can clarify. If you check my previous commit, I explored that option but realized that the list of `self.affected_tables` gets updated in between so some tables lack schema by the time get_query_specs is called. In the current commit, the list of `self.affected_tables` is the final updated list. You are describing my previous commit here which shows a failed test due to this difference in the list of affected tables.

data_steward/resources.py
nishanthpp93: This is because we are deprecating the functions in utils/bq.py. It will be retired and all functions converted to methods in the new BigQueryClient. This function should ideally be outside the client since it doesn't make any API calls but calling it would require a client to be instantiated, which is why I moved it to resources.py. This way all usages can be converted to this function, and it fits well with fields_for etc.

lrwb-aou: This function is currently defined in three places with this patch.  
1.  It's original location, which is being deprecated.
2. Inside `data_steward/gcloud/bq/__init__.py`, where it has been relocated.
3. By this patch in `resources.py`.

 It's starting to feel like a copy/paste bug.  I know the original will eventually be deleted.  I would still suggest this rightfully belongs in the bq package, because it is returning a list of `SchemaField` objects which are specific to BigQuery.  If all you want is a list of fields, we've got the `fields_for` function in `resources.py`.

lrwb-aou (commented):
i accidentally hit start review instead of add comment.

data_steward/resources.py
nishanthpp93: Yeah the entire bq package will be deleted at some point, but I can use that for now. `fields_for` does not fulfill the requirement, so we cannot use that. Since `get_table_schema` is a static function, I was hoping we don't have to instantiate a client to use it and hence relocated it to `resources.py`. It does use the bigquery library but does't make any API calls.

nishanthpp93: @lrwb-aou I've updated it to use the deprecated function for now.

nishanthpp93: @lrwb-aou I've updated it to use the deprecated function for now.

data_steward/cdr_cleaner/cleaning_rules/identifying_field_suppression.py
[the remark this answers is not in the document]
lrwb-aou: I agree that you don't need the `client` object. I don't think you need to use either the deprecated `get_table_schema` or create a new one either.  I would replace this with `fields_for`.  You stated you could not do that because

>  fields_for does not fulfill the requirement

What requirement is not fulfilled?  `fields_for` is used by ` get_table_schema` to access the JSON data.  I ran the following in a python interpreter and it worked as expected for me.
```python
import resources
schema = resources.fields_for('observation')
for item in schema:
    print(f"{item.get('mode')}\t\t{item.get('type')}\t\t{item.get('name')}")
```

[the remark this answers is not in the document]
nishanthpp93: The following logic uses schema as a BigQuery `SchemaField` object list but `fields_for` returns a dictionary which can have idiosyncrasies and possibly change the existing logic. Since the CR works as-is except for the client side-effect, I was looking to avoid logic changes as a de facto requirement. However, the use case here is simple enough that `fields_for` could be used without logic changes, with existing checks for casing, so I'll make that change.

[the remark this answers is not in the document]
nishanthpp93: @lrwb-aou Updated with `fields_for` instead of `get_table_schema`

data_steward/cdr_cleaner/cleaning_rules/identifying_field_suppression.py
```diff
@@ -102,24 +102,25 @@ def get_query_specs(self, *args, **keyword_args) -> query_spec_list:
             are optional but the query is required.
         """
         queries = []
+        schemas = {}
+        for table in self.affected_tables:
+            schemas[table] = fields_for(table)
```
lrwb-aou, line 107: One last suggestion, instead of looping through the list of affected tables twice, you could just do one loop.

I think you've got :
```python
        for table in self.affected_tables:
            schemas[table] = fields_for(table)
        for table in self.affected_tables:
            statements = []
            for item in schemas[table]:
```
And I think it could be simplified to:
```python
        for table in self.affected_tables:
            statements = []
            schema = fields_for(table)
            for item in schema:
```
Or is there a design decision that I am not understanding for looping through this list twice?

nishanthpp93: Thanks for catching this, it is an artifact of moving it from the setup function.

lrwb-aou (approved):
Looks good!

PR #1236, opened 2022-04-26
[DC-2347] Replace all bq.define_dataset(...) calls in curation

lledezma:
- Replaces all `bq.define_dataset(...)` calls in curation by invoking the function using the `BigQueryClient`

### DIFF
--- a/data_steward/ci/test_datasets.py
+++ b/data_steward/ci/test_datasets.py
@@ -6,9 +6,8 @@ Also contains test dataset deletion functions.  Dataset modification functions
 should be added here as well.
 """
 from google.oauth2 import service_account
-from google.cloud import bigquery
 
-from utils import bq
+from gcloud.bq import BigQueryClient
 
 CLIENT = None
 """Bigquery  client object"""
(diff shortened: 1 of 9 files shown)


data_steward/ci/test_datasets.py
```diff
@@ -18,41 +17,40 @@ def get_client(project_id, app_creds):
...
+    :returns: A BigQueryClient object
     """
     global CLIENT
     if not CLIENT:
         credentials = service_account.Credentials.from_service_account_file(
             app_creds)
-        CLIENT = bigquery.Client(project=project_id, credentials=credentials)
+        CLIENT = BigQueryClient(project_id=project_id, credentials=credentials)
 
     return CLIENT
```
nishanthpp93, line 30: Nice work parametrizing client in every function! I hope we can retire this global client.

nishanthpp93 (approved):
Looks good! Added a comment but not necessary for merge.

data_steward/ci/test_datasets.py
lledezma: It is possible to remove the `get_client(...)` function and just create the bq client inside `create_test_datasets(...)` and that still ensures that only one client is created and reused.

PR #1238, opened 2022-04-29
[DC-2403] Replace all bq.upload_csv_data_to_bq_table(...) calls in curation

lledezma:


ksdkalluri (changes requested):
changes looks good but lint checks are failing.

ksdkalluri (approved):
looks good!

PR #1239, opened 2022-05-02
[DC-2072] Add logic to omit datasets that are "in use" from delete_stale_test_datsets

jp3477:
Added checks for routines and models to exclude from stale datasets.


hiro-mishima (approved):
Thanks for working on this, @jp3477! This looks good to me. Let's see how the nightly jobs go once this is merged to develop.

PR #1240, opened 2022-05-02
[DC-2350] Process submissions after fixed time interval

nishanthpp93:
* [DC-2350] Send email for failed submission
* [DC-2350] Refactor functionality

[DC-2350]: https://precisionmedicineinitiative.atlassian.net/browse/DC-2350?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[DC-2350]: https://precisionmedicineinitiative.atlassian.net/browse/DC-2350?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/data_steward/validation/main.py
+++ b/data_steward/validation/main.py
@@ -432,7 +432,8 @@ def get_eastern_time():
         consts.DATETIME_FORMAT)
 
 
-def perform_reporting(hpo_id, report_data, folder_items, bucket, folder_prefix):
+def perform_reporting(hpo_id, report_data, folder_items, bucket, folder_prefix,
+                      failed_submission):
     """
     Generate html report, upload to GCS and send email if possible
 
@@ -441,6 +442,7 @@ def perform_reporting(hpo_id, report_data, folder_items, bucket, folder_prefix):
     :param folder_items: items in the folder without folder prefix
     :param bucket: bucket containing the folder
     :param folder_prefix: submission folder
+    :param failed_submission: Indicates if a submission has failed
     :return:
     """
     processed_time_str = get_eastern_time()
@@ -460,7 +462,8 @@ def perform_reporting(hpo_id, report_data, folder_items, bucket, folder_prefix):
     processed_txt_blob.upload_from_string(processed_time_str)
 
     folder_uri = f"gs://{bucket.name}/{folder_prefix}"
-    if folder_items and is_first_validation_run(folder_items):
+    if (folder_items and
+            is_first_validation_run(folder_items)) or failed_submission:
         logging.info(f"Attempting to send report via email for {hpo_id}")
         email_msg = en.generate_email_message(hpo_id, results_html, folder_uri,
                                               report_data)
@@ -531,11 +534,13 @@ def process_hpo(hpo_id, force_run=False):
                                               folder_prefix)
                 report_data = generate_metrics(hpo_id, bucket, folder_prefix,
                                                summary)
+                failed_submission = False
             else:
                 # do not perform validation
                 report_data = generate_empty_report(hpo_id, folder_prefix)
+                failed_submission = True
             perform_reporting(hpo_id, report_data, folder_items, bucket,
-                              folder_prefix)
+                              folder_prefix, failed_submission)
     except BucketNotSet as exc:
         logging.info(f'{exc}')
     except BucketDoesNotExistError as exc:
--- a/data_steward/validation/main.py
+++ b/data_steward/validation/main.py
@@ -763,13 +763,7 @@ def list_submitted_bucket_items(folder_bucketitems):
         logging.info(f"All required files found, processing.")
         return to_process_items
 
-    # Check submission times and validate if > 3 hrs old
-    logging.info(
-        f"Delaying processing for hpo_id by 3 hrs (to next cron run) "
-        f"since all required files not present. "
-        f"Missing {set(AOU_REQUIRED_FILES) - set(folder_bucketitems_basenames)}"
-    )
-
+    # Check submission times and validate if > 3 hrs old and < 29 days old
     upper_age_threshold = min(item['timeCreated'] +
                               datetime.timedelta(days=object_retention_days) -
                               datetime.timedelta(days=1)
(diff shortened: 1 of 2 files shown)


data_steward/validation/main.py
```diff
@@ -738,49 +743,58 @@ def list_submitted_bucket_items(folder_bucketitems):
...
-            retention_time = datetime.timedelta(days=object_retention_days)
-            retention_start_time = datetime.timedelta(days=1)
-            upper_age_threshold = created_date + retention_time - retention_start_time
-            upper_age_threshold = upper_age_threshold.replace(tzinfo=None)
-
-            if upper_age_threshold > utc_today:
-                files_list.append(item)
-
-            if basename(item) in AOU_REQUIRED_FILES:
-                # restrict processing time for 5 minutes after all required files
```
hiro-mishima, line 771: Do you know why the old logic used to wait for extra 5 minutes even if the files were all present?

hiro-mishima, line 771: Do you know why the old logic used to wait for extra 5 minutes even if the files were all present?

data_steward/validation/main.py
```diff
@@ -738,49 +743,58 @@ def list_submitted_bucket_items(folder_bucketitems):
...
+
+    # Check submission times and validate if > 3 hrs old
+    logging.info(
+        f"Delaying processing for hpo_id by 3 hrs (to next cron run) "
+        f"since all required files not present. "
+        f"Missing {set(AOU_REQUIRED_FILES) - set(folder_bucketitems_basenames)}"
+    )
+
+    upper_age_threshold = min(item['timeCreated'] +
+                              datetime.timedelta(days=object_retention_days) -
```
hiro-mishima, line 774: How does this `object_retention_days` come into play? I think I am having difficulty understanding what `upper_age_threshold` and `lower_age_threshold` each stand for.

hiro-mishima, line 774: How does this `object_retention_days` come into play? I think I am having difficulty understanding what `upper_age_threshold` and `lower_age_threshold` each stand for.

hiro-mishima (commented):
I left a few comments. I think those are coming from my lack of understanding. It would be great if we could have a quick huddle to go over this once you come back!
Thanks, @nishanthpp93!

tests/unit_tests/data_steward/validation/main_test.py
```diff
@@ -57,10 +57,10 @@ def test_retention_checks_list_submitted_bucket_items(self):
             days=25)
         outside_retention = datetime.datetime.now(tz=None) - datetime.timedelta(
```
hiro-mishima, line 58: It looks like `outside_retention` is no longer necessary.

data_steward/validation/main.py
hiro-mishima: 5 min was implemented to make sure the files become downloadable after the upload. However, you can download the files if you can see the files in GCS. So, we confirmed that we do not need this 5 min wait logic anymore.

data_steward/validation/main.py
```diff
@@ -738,49 +743,58 @@ def list_submitted_bucket_items(folder_bucketitems):
...
-                if lower_age_threshold > utc_today:
-                    logging.info(
-                        f"Delaying processing for hpo_id by 3 hrs (to next cron run) "
-                        f"since files are still being uploaded. "
-                        f"Latest file was uploaded less than {object_process_lag_minutes} minutes ago."
-                    )
-                    return []
+            f"since files were recently uploaded. Latest file was uploaded "
+            f"less than {hrs} hours ago.")
+
```
hiro-mishima, line 797: Maybe we can add `else:` statement to clarify what is happening.

data_steward/validation/main.py
hiro-mishima: `upper_age_threshold` ... If now is past `upper_age_threshold`, the files are too old and we should not process them. The files are deleted from the bucket after the retention period(=30 days). If the files are almost reaching the retention period(=less than one day until the retention period), this script omits them from the list of the submitted items.

`lower_age_threshold` ... If now is before `lower_age_threshold`, the files are too new and we wait to process them. We wait because EHR sites might be in the middle of uploading more files and we might expect more files to come. If now is after `lower_age_threshold`, we can safely assume that the EHR sites finished their uploading process whether or not all the expected files are present.

tests/unit_tests/data_steward/validation/main_test.py
```diff
@@ -123,22 +133,26 @@ def test_retention_checks_list_submitted_bucket_items(self):
...
-            'name': '2018-09-01/observation.csv',
-            'timeCreated': within_retention,
-            'updated': before_lag_time
-        })
-
         actual_result = main.list_submitted_bucket_items(bucket_items)
         expected_result = []
         self.assertCountEqual(expected_result, actual_result)
 
+        # If any file is missing but submission is stale, process it
```
hiro-mishima, line 146: If the submission is stale, shouldn't it be ignored?
I think the logic should go into this `else,` and the function should return `files_list,` an empty list.

https://github.com/all-of-us/curation/pull/1240/files#diff-43507efb5f679e0321f8eabb85c7231a109b4000e39babde951c331470dd1593R791

I might be confused again. Let me know if my understanding is correct. Thanks!

hiro-mishima (commented):
Thanks for updating the scripts, @nishanthpp93!
I left one more comment for clarity. Please let me know what you think. Thanks!

tests/unit_tests/data_steward/validation/main_test.py
nishanthpp93: I referred to stale here as past 3 hours since submission, but don't have a word for past retention. Maybe you are referring to stale as past retention?

hiro-mishima: Oh OK, now I get it! Thanks for the clarification :)

hiro-mishima (approved):
Thank you, @nishanthpp93! It looks good to me now 🎉

PR #1245, opened 2022-05-05
[DC-2423] Nightly delete_stale_test_datasets execution fails with "Too long with no output"

hiro-mishima:


nishanthpp93 (approved):
Looks good!

PR #1247, opened 2022-05-06
[DC-2411] Update all bq.build_and_copy_contents(...) calls in curation

lledezma:


data_steward/gcloud/bq/__init__.py
```diff
@@ -405,3 +406,40 @@ def update_labels_and_tags(self,
             return {**existing_labels_or_tags, **updates}
 
         return {**existing_labels_or_tags, **updates}
+
+    def build_and_copy_contents(self, src_dataset: str, dest_dataset: str):
```
ksdkalluri, line 410: @lledezma this function in util/bq is updated recently, it would be great to update this as well.

ksdkalluri (changes requested):

data_steward/gcloud/bq/__init__.py
lledezma: In this ticket or a separate ticket?

ksdkalluri: @lledezma tools/create_tier still uses bq.build_and_copy_contents() please upthat that as well.

ksdkalluri (changes requested):

ksdkalluri (approved):
Thanks for the changes @lledezma!

PR #1251, opened 2022-05-09
[DC-1846] Refactor clean_cdr to make create_tier generate rt_base and rt_clean datasets

ksdkalluri:


data_steward/cdr_cleaner/manual_cleaning_rules/survey_version_info.py
```diff
@@ -88,7 +88,7 @@ def __init__(self,
             description=desc,
             # first seen in the `deid_base` dataset, although
             # not explicitly part of this set of cleaning rules yet
-            affected_datasets=[cdr_consts.DEID_BASE],
+            affected_datasets=[cdr_consts.REGISTERED_TIER_DEID],
```
hiro-mishima, line 91: I think we also need to add `CONTROLLED_TIER_DEID` to this list.

data_steward/cdr_cleaner/cleaning_rules/clean_mapping.py
```diff
@@ -86,7 +86,11 @@ def __init__(self,
         super().__init__(issue_numbers=ISSUE_NUMBERS,
                          description=desc,
                          affected_datasets=[
-                             cdr_consts.RDR, cdr_consts.CONTROLLED_TIER_DEID,
+                             cdr_consts.RDR, cdr_consts.COMBINED,
```
hiro-mishima, line 89: I don't think we need `RDR` here. We need `EHR` and `UNIONED` instead.

hiro-mishima (changes requested):
I left a few comments. Could you please check if those make sense? Thanks, @ksdkalluri!

hiro-mishima:
And it looks like there is a conflict.

ksdkalluri:
> And it looks like there is a conflict.

Resolved them as well

hiro-mishima (approved):
Looks good! Thanks, @ksdkalluri 🎉

PR #1252, opened 2022-05-09
[DC-2436] Update `bq.build_and_copy_contents` to copy tables without pre-defined schema in resource_files/schemas.

ksdkalluri:


hiro-mishima (approved):
It looks good to me! I also confirmed this change works by generating Fitbit CT DEID creation.
Thanks, @ksdkalluri!

data_steward/utils/bq.py
```diff
@@ -676,28 +676,38 @@ def build_and_copy_contents(client, src_dataset, dest_dataset):
 
     for table_item in table_list:
         # create empty schemaed tablle with client object
-        schema_list = get_table_schema(table_item.table_id)
+        try:
+            schema_list = get_table_schema(table_item.table_id)
+        except RuntimeError as re:
+            schema_list = None
```
lrwb-aou, line 682: I would ask for a logging statement added here so we know what is happening.

```suggestion
            schema_list = None
            LOGGER.warning(f"No schema available for {table_item.table_id}.  This function will "
                           f"copy table contents without specifying a schema.")
```

lrwb-aou (changes requested):
I think the PR should work, but would appreciate the logging.  Also, would it be better for this to fail to force the usage of schema files?

data_steward/utils/bq.py
ksdkalluri: @lrwb-aou Added the log.

lrwb-aou (approved):
Thank you!  Looks good.

PR #1253, opened 2022-05-09
[DC-2437] Update `create_tier.py` to use `bq.build_and_copy_contents` instead of `snapshot_by_query. create_schemaed_snapshot_dataset`

ksdkalluri:


hiro-mishima (approved):
It looks good to me! I also confirmed this change works by generating Fitbit CT DEID creation.
Thanks, @ksdkalluri!

PR #1255, opened 2022-05-10
[DC-2438] update `date_shift_cope_responses` CR to use observation_ext table instead of cope_concepts_table.

ksdkalluri:


lrwb-aou (changes requested):
@ksdkalluri I've got a few small questions.  
1.) This removes our dependence on the `cope_concepts` table but places a restraint on the `survey_version_concept_id` column.  If we ever started to map all surveys in this column, then this would no longer work.  Just something to be aware of.  
2.)  Ideally, (and we're not there yet), we'd be able to read all the PPI `concept_codes` from RedCap and just update a reference table with Q/A pairs.  But, we don't have that right now and now is not the time to build it out.
3.)  I would eventually like to switch the select statement to an update statement.  But that can be a future optimization.
4.)  I think we need  to add this issue number to the list of jira issues as it is directly impacting the SQL produced.
5.)  I would also recommend adding a `depends_on` list to the initializer.  This class depends on `COPESurveyVersionTask` to have run before it can execute successfully now.

With some minor bookkeeping changes and team knowledge sharing, this is a good PR.  Think about items 1, 2, and 3 but please implement 4 and 5.

lrwb-aou (approved):
Thanks for the changes!

PR #1256, opened 2022-05-11
[DC-2433] Update all bq_utils.get_table_id calls in ehr_union.py

lledezma:


hiro-mishima (approved):
Looks good! Thanks, @lledezma 🎉

PR #1259, opened 2022-05-16
[DC-2410] Fix the error that prevents generation of a `pipeline_tables.site_masking` record

hiro-mishima:
- I ran the updated script for DC2409 and confirmed that `site_maskings` is updated as expected.
- I added the exceptions for clarity.
- I updated the insert logic since `pipeline_tables_sandbox` does not exist in our dataset.
- I updated the playbook to specify what we need to do in each step.
- Please see the comment section in the JIRA ticket DC2410 for details.

### DIFF
--- a/data_steward/tools/add_hpo.py
+++ b/data_steward/tools/add_hpo.py
@@ -51,7 +51,7 @@ SELECT
 """)
 
 UPDATE_SITE_MASKING_QUERY = JINJA_ENV.from_string("""
-INSERT INTO `{{project_id}}.{{dataset_id}}.{{table_id}}` (hpo_id, src_id)
+INSERT INTO `{{project_id}}.{{pipeline_tables_dataset}}.{{site_maskings_table}}` (hpo_id, src_id)
 WITH available_new_src_ids AS (
   SELECT 
     ROW_NUMBER() OVER(ORDER BY GENERATE_UUID()) AS temp_key,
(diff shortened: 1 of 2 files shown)


data_steward/tools/add_hpo.py
```diff
@@ -23,61 +24,60 @@
...
+""")
 
 UPDATE_SITE_MASKING_QUERY = JINJA_ENV.from_string("""
-INSERT INTO
-   `{{project_id}}.{{dataset_id}}.{{table_id}}` (hpo_id, src_id)
-     WITH random_ids AS
- (
- -- Generates the random EHR site id for the new hpo_site --
+INSERT INTO `{{project_id}}.{{dataset_id}}.{{table_id}}` (hpo_id, src_id)
+WITH available_new_src_ids AS (
```
nishanthpp93, line 55: This is a great breakdown of the logic used! Thanks for updating.

data_steward/tools/add_hpo.py
```diff
@@ -23,61 +24,60 @@
...
-    ROW_NUMBER() OVER(ORDER BY GENERATE_UUID()) AS row_id
-  FROM
-    UNNEST(GENERATE_ARRAY(100, 999)) AS new_id
+    ROW_NUMBER() OVER() AS temp_key,
+    m.hpo_id
+  FROM `{{project_id}}.{{lookup_tables_dataset}}.{{hpo_site_id_mappings_table}}` AS m
+  WHERE m.HPO_ID IS NOT NULL 
+  AND m.HPO_ID != '' 
+  AND LOWER(m.HPO_ID) NOT IN (
+    SELECT LOWER(hpo_id) FROM `{{project_id}}.{{dataset_id}}.{{table_id}}`
```
nishanthpp93, line 74: I'd suggest using `site_maskings_table` instead of `table_id` and `pipeline_tables_dataset` instead of `dataset_id`
```suggestion
    SELECT LOWER(hpo_id) FROM `{{project_id}}.{{pipeline_tables_dataset}}.{{site_maskings_table}}`
```

nishanthpp93 (commented):
Overall looks great! Added a small suggestion to make it easier to read the query

hiro-mishima:
Thanks for the comment, @nishanthpp93! I updated the scripts accordingly. Please let me know if there is anything else. Thanks!!

nishanthpp93 (approved):
Looks good!

PR #1271, opened 2022-05-31
[DC-2260] CR for remove pts<18 should be based on age at consent

hiro-mishima:
Please see the following comment in the JIRA ticket for the age calculation:
https://precisionmedicineinitiative.atlassian.net/browse/DC-2260?focusedCommentId=121588


ksdkalluri (approved):
looks good, thanks for the changes @hiro-mishima.

PR #1272, opened 2022-06-01
[DC-1852] Use notebook_utils.execute instead of pandas.read_gbq (merge conflict resolution)

jp3477:


nishanthpp93 (approved):
Looks good!

PR #1274, opened 2022-06-03
[DC-2474] Update all bq_utils.table_exists calls in data_steward/validation

lledezma:


ksdkalluri (approved):
Looks Good to me! Adding TODO is a good Idea.

PR #1275, opened 2022-06-06
[DC-2446] Fix run_achilles_and_export script to upload achilles index…

molly-giddings:
… files to GCS without error

### DIFF
--- a/data_steward/validation/main.py
+++ b/data_steward/validation/main.py
@@ -159,7 +159,9 @@ def upload_achilles_files(hpo_id):
     return json.dumps(result, sort_keys=True, indent=4, separators=(',', ': '))
 
 
-def _upload_achilles_files(hpo_id: str = None, folder_prefix: str = '', target_bucket: str = None) -> list:
+def _upload_achilles_files(hpo_id: str = None,
+                           folder_prefix: str = '',
+                           target_bucket: str = None) -> list:
     """
     uploads achilles web files to the corresponding hpo bucket
     :hpo_id: which hpo bucket do these files go into
@@ -173,14 +175,16 @@ def _upload_achilles_files(hpo_id: str = None, folder_prefix: str = '', target_b
         if not hpo_id:
             raise RuntimeError(
                 f"Either hpo_id or target_bucket must be specified")
-        target_bucket = storage_client.get_hpo_bucket(hpo_id).name
+        target_bucket = storage_client.get_hpo_bucket(hpo_id)
+    target_bucket = storage_client.bucket(target_bucket)
     logging.info(
-        f"Uploading achilles index files to 'gs://{target_bucket}/{folder_prefix}'"
+        f"Uploading achilles index files to 'gs://{target_bucket.name}/{folder_prefix}'"
     )
 
     for filename in resources.ACHILLES_INDEX_FILES:
         logging.info(
-            f"Uploading achilles file '{filename}' to bucket {target_bucket}")
+            f"Uploading achilles file '{filename}' to bucket {target_bucket.name}"
+        )
         bucket_file_name = filename.split(resources.resource_files_path +
                                           os.sep)[1].strip().replace('\\', '/')
         with open(filename, 'rb') as fp:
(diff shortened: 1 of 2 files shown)


data_steward/validation/main.py
```diff
@@ -175,13 +175,12 @@ def _upload_achilles_files(hpo_id=None, folder_prefix='', target_bucket=None):
                 f"Either hpo_id or target_bucket must be specified")
         target_bucket = storage_client.get_hpo_bucket(hpo_id)
```
hiro-mishima, line 176: We also need to change this line to something like this. Otherwise, there will be a type mismatch:
`target_bucket = storage_client.get_hpo_bucket(hpo_id).name`

hiro-mishima: Also, let's update line 162 to have type hinting. That way, `target_bucket` will be less ambiguous:
`def _upload_achilles_files(hpo_id: str = None, folder_prefix: str = '', target_bucket: str = None) -> list:`

hiro-mishima (changes requested):
Thanks, @molly-giddings. It looks overall good. I added a few comments for a suggestion. Let me know if you have questions. Thanks!

data_steward/validation/main.py
molly-giddings: Thanks, @hiro-mishima !  Resolved both comments.

molly-giddings: Actually, kept the type hinting, but had to turn the bucket name into a bucket object to allow a blob to be created further down the line.  I reverted back to the .name references but added:
    `target_bucket = storage_client.bucket(target_bucket)`
on line 179 after the check for target_bucket

hiro-mishima: Thanks, @molly-giddings!
It looks like the unit test `test_process_hpo_ignore_dirs` is failing. I think it is failing because of the type mismatch:
`target_bucket = storage_client.get_hpo_bucket(hpo_id)` -> `target_bucket` becomes a bucket object
`storage_client.bucket(target_bucket)` -> `target_bucket` is a bucket object if the `if` condition is processed, otherwise it is `str`.

I believe we can fix it by adding `else:` to what you added:
```
    if not target_bucket:
        if not hpo_id:
            raise RuntimeError(
                f"Either hpo_id or target_bucket must be specified")
        target_bucket = storage_client.get_hpo_bucket(hpo_id)
    else:
        target_bucket = storage_client.bucket(target_bucket)
```

data_steward/validation/main.py
```diff
@@ -174,6 +176,8 @@ def _upload_achilles_files(hpo_id=None, folder_prefix='', target_bucket=None):
             raise RuntimeError(
                 f"Either hpo_id or target_bucket must be specified")
         target_bucket = storage_client.get_hpo_bucket(hpo_id)
+    else:
+        target_bucket = storage_client.bucket(target_bucket.name)
```
hiro-mishima, line 180: If we use `target_bucket.name` here, we assume `target_bucket` to be a bucket object. However, we are passing `str` to `target_bucket` when we call this function. So, we need to keep `storage_client.bucket(target_bucket)`. Otherwise, the problem described in `DC-2446` will remain. 

The test assumes the wrong data type, which is causing this error.

https://github.com/all-of-us/curation/blob/develop/tests/integration_tests/data_steward/validation/main_test.py#L261

The test needs `.name` added to represent the real-life use case. 
```
        main._upload_achilles_files(hpo_id=None,
                                    folder_prefix=folder_prefix,
                                    target_bucket=bucket_nyc.name)
```

hiro-mishima (changes requested):
Thanks for updating the script, @molly-giddings! I am leaving one more comment here. I believe this will solve the underlying problems. Let me know if you have any questions!

hiro-mishima (approved):
Thanks, @molly-giddings! Everything looks good now 😃

data_steward/validation/main.py
molly-giddings: Thanks, @hiro-mishima !

PR #1277, opened 2022-06-06
[DC-2486]

lledezma:


data_steward/gcloud/bq/__init__.py
```diff
@@ -23,6 +23,10 @@
 from constants.utils import bq as consts
 from common import JINJA_ENV
 
+tracer_provider = TracerProvider()
```
hiro-mishima, line 26: Can you please explain why we needed to move these lines out of the class?

tests/integration_tests/data_steward/gcs_utils_test.py
```diff
@@ -11,27 +11,29 @@
...
+from gcloud.bq import BigQueryClient
 from tests import test_util
 from tests.test_util import FIVE_PERSONS_PERSON_CSV, FAKE_HPO_ID
 
 
 class GcsUtilsTest(unittest.TestCase):
 
     dataset_id = bq_utils.get_dataset_id()
+    project_id = app_identity.get_application_id()
+    bq_client = BigQueryClient(project_id)
```
hiro-mishima, line 23: I think `bq_client = BigQueryClient(project_id)` can be placed here or within `setUpClass(cls)`. Is there any particular reason you chose here over `setUpClass(cls)`? (I am just curious to know :slightly_smiling_face:)

hiro-mishima (commented):
Thanks, @lledezma! Overall it looks good. I left a few comments for suggestions and clarifications. Please let me know your thoughts!

data_steward/gcloud/bq/__init__.py
lledezma: If tracer provider is not set global, then the DefaultTracerProvider will provide no-op spans that don't do anything. We set it global so we don't fail to collect telemetry.

tests/integration_tests/data_steward/gcs_utils_test.py
lledezma: I remember I was asked to place a `StorageClient` in the same place when working on PR updates a while back. So, I decided to place it there to stay consistent with `dataset_id` and the `StorageClient`. If you think it would be better to place it in `setUpClass(cls)` then I can I do that.🙂

hiro-mishima (approved):
Thanks, @lledezma! It looks all good to me now. I confirmed that the error is no longer showing in the integration tests 🎉

PR #1279, opened 2022-06-09
[DC-2480] [P0] Remove `DropParticipantsWithoutPPI` from the RT and CT cleaning lists

hiro-mishima:


lrwb-aou (approved):
Looks good!

lrwb-aou:
@hiro-mishima We should consider reverting the changes introduced by the PR for DC-2379.

hiro-mishima:
@lrwb-aou Good catch! We should do that. I will create another PR for that under the same JIRA number and ask for your review!

PR #1283, opened 2022-06-10
[DC-2480] [P0] Remove `DropParticipantsWithoutPPI` from the RT and CT cleaning lists #2

hiro-mishima:
Revert changes from DC-2379.


lrwb-aou (approved):
Looks good.

PR #1287, opened 2022-06-14
[DC-2398] Update unmapped foreign keys

jp3477:


hiro-mishima (changes requested):
Thanks, @jp3477. It looks like the integration test failed for this PR. Could you please fix it?

jp3477:
@hiro-mishima I just made a pretty hacky change to exclude checking `visit_detail` and `visit_occurrence` in the failing test. This may not be the best way to fix the test, but the code changes I made are very difficult to incorporate in the existing tests otherwise. Let me know if you want to discuss further.

hiro-mishima:
Hi @jp3477! I was exploring the integration test to figure out how not to exclude `visit_detail` and `visit_occurrence`. And I think I figured it out. The test assumes the order of `JOIN`s follow the order of the columns in the `SELECT` clause. 
I created this PR https://github.com/all-of-us/curation/pull/1297 that shows the proposed update to your PR. Could you please have a look at it?
In that PR, I added an extra test case to make sure those self-reference IDs are actually mapped to the new IDs. That made the test scripts a bit longer.

PR #1288, opened 2022-06-14
[DC-2488] Added check for zero-valued identifiers in rdr

jp3477:


data_steward/analytics/cdr_ops/rdr_export_qc.py
```diff
@@ -669,3 +670,111 @@
...
+SELECT
+  ac.pmi_code, COUNT(*) invalid_id_count
+FROM `{{project_id}}.{{dataset}}.observation` o
+JOIN answer_codes ac
+  ON ac.pmi_code = o.value_source_value
+WHERE (o.observation_source_concept_id = 0
+  OR o.observation_concept_id = 0)
+GROUP BY ac.pmi_code
+ORDER BY 2 DESC
+""")
```
lrwb-aou, line 749: I asked for your update to the code above.  When I run your query with a few extra fields, it is interesting and something we may want to incorporate.  This shows that for some Answers, their questions are not mapped to a non-zero concept_id.  Meaning, this data would be dropped in the clean datasets because both `observation_concept_id` and `observation_source_concept_id` are zero.  But the answer itself is ok.  These will be addressed when the csv file is updated, but I still think this is a useful check.  What do you think about incorporating it?
```sql
WITH answer_codes AS (
  SELECT
    pmi_code
  FROM `{{project_id}}.{{sandbox_dataset}}.old_map_short_codes` 
  WHERE type = 'Answer'
)
SELECT
  ac.pmi_code, o.value_source_value, o.value_source_concept_id, o.value_as_concept_id, o.observation_source_value, o.observation_concept_id, o.observation_source_concept_id,COUNT(*) invalid_id_count
FROM `{{project_id}}.{{dataset}}.observation` o
JOIN answer_codes ac
  ON ac.pmi_code = o.value_source_value
WHERE (o.observation_source_concept_id = 0
  OR o.observation_concept_id = 0)
GROUP BY ac.pmi_code, 2, 3, 4, 5, 6, 7
ORDER BY 8 DESC;
```

lrwb-aou (changes requested):
I like the explicit definition of success and failure.  I've got a couple of suggestions, and noticed one bug that needs to be addressed.  Thanks for picking this up!

data_steward/analytics/cdr_ops/rdr_export_qc.py
```diff
@@ -669,3 +670,111 @@
...
+  FROM `{{project_id}}.{{sandbox_dataset}}.old_map_short_codes` 
+  WHERE type = 'Answer'
+)
+SELECT
+  ac.pmi_code, COUNT(*) invalid_id_count
+FROM `{{project_id}}.{{dataset}}.observation` o
+JOIN answer_codes ac
+  ON ac.pmi_code = o.value_source_value
+WHERE (o.observation_source_concept_id = 0
+  OR o.observation_concept_id = 0)
```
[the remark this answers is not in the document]
jp3477: Got it, will fix this.

jp3477:
@lrwb-aou Thanks for the suggestions, I pushed some changes.

data_steward/analytics/cdr_ops/rdr_export_qc.py
```diff
@@ -669,3 +670,152 @@
                    project_id=project_id,
                    combined_pfmhh_concepts=DROP_PFMHH_CONCEPTS)
 execute(client, query)
+
+
+# # Check that the Question and Answer Concepts in the old_map_short_codes tables are not paired with 0-valued concept_identifiers
+
+# According to this [ticket](https://precisionmedicineinitiative.atlassian.net/browse/DC-2488), Question and Answer concepts that are identified in the `old_map_short_codes` table should not be paired with 0-valued concept_identifiers after the RDR dataset is cleaned. These concept identifiers include the `observation_concept_id` and `observation_source_concept_id` fields.
+
+def render_message(results_df, success_msg=None, failure_msg=None, success_msg_args={}, failure_msg_args={}):
```
lrwb-aou, line 679: Love this!  I can see this being used in multiple places.

data_steward/analytics/cdr_ops/rdr_export_qc.py
```diff
@@ -669,3 +670,152 @@
...
+    <b>{code_count}</b> answer codes have 0-valued IDs. Report failure back to curation team.
+    Bug likely due to failure in the <code>update_questions_answers_not_mapped_to_omop</code> cleaning rule.
+'''
+
+render_message(df, success_msg, failure_msg, failure_msg_args={'code_count': len(df)})
+# -
+
+# ### Question-Answer Codes Combo
+
+# Check the answer codes
```
lrwb-aou, line 789: I would suggest making this description different from the one above.
```suggestion
# Check that mapped answer codes are paired with correctly mapped question codes.  If the question codes are zero valued, the question and answer pair will be dropped from the clean version of the CDR.
```

lrwb-aou (approved):
This is great!  Thank you.  It should help prevent the need for another hotfix like this last one in the future.  I've suggested a more verbose description for the final check.

Looks great!

PR #1293, opened 2022-06-15
[DC-809] \ [DC-1201] conversion to base class and integration test

lrwb-aou:
* converts a legacy module to using the base class method
* moves the instantiation from the combined rules to the rdr rules
* implements an integration test
* removes the legacy module
* sandboxes the data before removing the data (for easier debugging)


ksdkalluri (approved):
Changes look good. I ran integrations locally, works fine. Thanks for the changes, Lauren!

PR #1294, opened 2022-06-16
[DC-2340] Modify the EHR Union process to drop site submitted race/ethnicity/gender observation records

hiro-mishima:


ksdkalluri (approved):
Looks good to me!

PR #1298, opened 2022-06-20
[DC-2538] using the table_namer setter as originally designed

lrwb-aou:
* actually use the table_namer setter
* this causes the default value of table_namer to be used when an explicit value isn't provided
* update tests that are broken because of bug in setter usage


data_steward/cdr_cleaner/cleaning_rules/base_cleaning_rule.py
```diff
@@ -423,13 +424,14 @@ def affected_tables(self, affected_tables):
...
     @table_namer.setter
-    def table_namer(self, table_namer, **kwargs):
+    def table_namer(self, table_namer):
         """
         Set the table_namer for this class instance. If no value is provided, it is set to a default value.
         """
         if not table_namer:
-            for key, value in kwargs.items():
-                self._table_namer = kwargs.get(value)
+            self._table_namer = re.sub('\d\d\d\dq\dr\d', '', self.dataset_id)
```
nishanthpp93, line 432: nit: regex can be contracted 
```suggestion
            self._table_namer = re.sub('\d{4}q\dr\d', '', self.dataset_id)
```

data_steward/cdr_cleaner/cleaning_rules/remove_non_matching_participant.py
```diff
@@ -75,7 +74,7 @@
     """({{column_expr}}) >= {{num_of_missing}}""")
 
 
-class RemoveNonMatchingParticipant(BaseCleaningRule):
+class RemoveNonMatchingParticipant(SandboxAndRemovePids):
```
nishanthpp93, line 77: This is a much needed change!

data_steward/cdr_cleaner/cleaning_rules/sandbox_and_remove_pids.py
```diff
@@ -1,120 +1,148 @@
...
+        """
+        if self.person_table_list:
+            return self.person_table_list
+
+        person_table_query = PERSON_TABLE_QUERY.render(project=self.project_id,
+                                                       dataset=self.dataset_id)
+        person_tables = client.query(person_table_query).result()
+        # utils.bq.query(person_table_query).get(
+        #     TABLE_NAME_COLUMN).to_list()
+        # exclude mapping tables from list, to be removed after all cleaning rules
```
nishanthpp93, line 89: nit: comments can be removed
```suggestion
        """
        Get list of tables that have a person_id column, excluding mapping tables
        """
        if self.person_table_list:
            return self.person_table_list

        person_table_query = PERSON_TABLE_QUERY.render(project=self.project_id,
                                                       dataset=self.dataset_id)
        person_tables = client.query(person_table_query).result()
```

data_steward/cdr_cleaner/cleaning_rules/sandbox_and_remove_pids.py
```diff
@@ -1,120 +1,148 @@
...
+            queries_list.append({
+                cdr_consts.QUERY:
+                    CLEAN_QUERY.render(project=self.project_id,
+                                       dataset=self.dataset_id,
+                                       table=table,
+                                       sandbox_dataset=self.sandbox_dataset_id,
+                                       lookup_table=lookup_table)
+            })
+
+        return queries_list
```
nishanthpp93, line 148: Should this define get_query_specs and other methods?

lrwb-aou: I purposely left this as an abstract class so it cannot be instantiated by iteself.  It must be extended to be useful.  The extending rule should define `get_query_specs` and any other methods that are required to make this class "concrete".

What are your thoughts?

nishanthpp93: Yes, that makes sense. As long as this rule isn't called by itself, it can stay this way.

nishanthpp93 (commented):
There's one last cleaning rule defining `sandbox_table_for` in `drop_rows_for_missing_persons`. Do you think that can be removed?

lrwb-aou:
> There's one last cleaning rule defining `sandbox_table_for` in `drop_rows_for_missing_persons`. Do you think that can be removed?

Unfortunately, this isn't an easy remove.  This affects the unit test of a non-base classed cleaning rule.  But it's a rule that I don't think we currently use.  I can either base class that rule (with potentially other changes because I don't think it would function properly if invoked right now) or remove the thing entirely.

For now, I'm working on the base class and correction of the existing rule.

nishanthpp93:
Yes that is. We could address this CR in a separate PR later, I'm fine with that as well.

data_steward/cdr_cleaner/cleaning_rules/base_cleaning_rule.py
```diff
@@ -423,13 +424,14 @@ def affected_tables(self, affected_tables):
...
+    def table_namer(self, table_namer):
         """
         Set the table_namer for this class instance. If no value is provided, it is set to a default value.
         """
         if not table_namer:
-            for key, value in kwargs.items():
-                self._table_namer = kwargs.get(value)
+            self._table_namer = re.sub('\d{4}q\dr\d', '', self.dataset_id)
+            LOGGER.info(f"'table_namer' was not set.  "
+                        f"Using default value of `{table_namer}`.")
```
lrwb-aou, line 434:
```suggestion
                        f"Using default value of `{self._table_namer}`.")
```

lrwb-aou:
> Yes that is. We could address this CR in a separate PR later, I'm fine with that as well.

I opted to go ahead and base class that rule, create an integration test for it, and remove the unit test.  If we ever want to add it back to the pipeline, it will be better suited than it was.  Up for consideration will be splitting it into two separate rules, one to remove EHR data for folks who have not given EHR consent and another (which we may already have) to drop records for participants who do not exist in the `person` table.

nishanthpp93 (approved):
Looks good! Thank you for the updates, @lrwb-aou!

PR #1301, opened 2022-06-27
[DC-2544] Enhancements to vocabulary_upload_qc

brendagutman:
[DC-2544]  Enhancements to vocabulary_upload_qc

[DC-2544]: https://precisionmedicineinitiative.atlassian.net/browse/DC-2544?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


data_steward/analytics/cdr_ops/vocabulary_upload_qc.py
```diff
@@ -51,17 +59,24 @@
                    vocabulary_dataset_new=vocabulary_dataset_new)
 execute(client, query, max_rows=True)
 
-# # Removed PPI. Where ppi concepts are not present in the new vocabulary
+#
+# # PPI concept comparison, retired or removed concepts 
+#
+# > - The dataframe will contain all PPI concepts that were newly deprecated or removed from Athena. <br>
+# > - Generally none would be deprecated or removed. <br>
+# > - If the dataframe is not empty contact the survey team to inquire about the concepts.<br>
```
lrwb-aou, line 67: I like this being specific to the PPI vocabulary.  This comment is related to a future scope issue we can create.  

I would like to do something similar with non-PPI concepts that are used to generalize or suppress information.  This could help us loop in privacy earlier when a non-PPI concept is deprecated and a potentially new concept is introduced.  Would you like to create a jira issue for this?

data_steward/analytics/cdr_ops/vocabulary_upload_qc.py
```diff
@@ -82,7 +97,12 @@
 # ''')
 # -
 
-# # The difference between the number of rows in each table of the dataset. All 'changes' should increase. https://github.com/OHDSI/Vocabulary-v5.0/releases
+#
+# # Tables row count comparison
+#
+# > - The difference between the number of rows in each table of the datasets.  <br>
+# > - All 'changes' should increase. Expected changes are found in the [release notes](https://github.com/OHDSI/Vocabulary-v5.0/releases).  <br>
```
lrwb-aou, line 104: They are almost always increasing.  We have had at least one instance where a vocabulary got merged into another vocabulary and the vocabulary count dropped from 83 to 82.  Again, this is an edge case and as a general rule we expect row counts to increase as you mention.

data_steward/analytics/cdr_ops/vocabulary_upload_qc.py
```diff
@@ -132,7 +155,24 @@
                    vocabulary_dataset_new=vocabulary_dataset_new)
 execute(client, query, max_rows=True)
 
-# Shows the number of individual concepts added to each vocab type.The 'diff' should always increase.
+# # Confirm presence of vocabularies 
+# > - This table should contain five rows.<br>
+# > - These vocabulary_ids should exist in the new vocabulary <br>
+# > - Investigate if this is not the case. <br>
```
lrwb-aou, line 161: Do you know why this specific check exists and why it only focuses on these 5 vocabularies?  If so, that would be good information to add to this description.

data_steward/analytics/cdr_ops/vocabulary_upload_qc.py
```diff
@@ -156,8 +196,10 @@
 execute(client, query, max_rows=True)
 
 # # Access to the new vocabulary
-# This query shows the users and the roles who have access to the new vocabulary.
-# Confirm that the PDR service account has BigQuery Data Viewer access to it.
+# > - This query shows the users and the roles who have access to the new vocabulary.
+# > - The project_id should be prod here.
+# > - Confirm that the PDR service account has BigQuery Data Viewer access to it.
+# > - If no access, re-run the step in the [vocabulary playbook](https://docs.google.com/document/d/1U8AIunEVRdJOUGgYJASKcTGkzPDg5MVU_hiIKDwmipk/edit#). <br>
```
lrwb-aou, line 202: Nice pointing to new documentation.

lrwb-aou (commented):
These descriptions are much improved.  I did add a few historical knowledge/edge case comments here that you may or may not choose to incorporate.  I do recommend adding the notes about find info in jira for the changes to the AoU_Custom and AoU_General vocabularies.

data_steward/analytics/cdr_ops/vocabulary_upload_qc.py
lrwb-aou: I remember the TypeConcept and Condition Status vocabs causing trouble.  They were removed and added back inadvertently or something similar.  I'm not sure we have to have this particular query anymore since we  do have the above query.  Perhaps you could do a `FULL OUTER JOIN` on the previous query to identify all vocab changes from old to new vocabularies and highlight anything that either didn't match or where a `null` value existed.  If nothing is changed, we don't actually have to see the full 80-ish results list.  Thoughts?

brendagutman: I removed this query, and made a small change to the one above it. I also thought it unnecessary to see the entire vocabulary list.

PR #1305, opened 2022-06-30
[DC-821] [DC-1174] Update Skip Code Backfill CR to New Base Class

hiro-mishima:
- Updated the SQL from the `SELECT` statement with `WRITE_TRUNCATE` to the `INSERT` statement.
- The old SQL was not populating `value_source_value`=`PMI_Skip` properly. With the new SQL, `PMI_Skip` will be populated as designed.
- Confirmed the old and the new SQLs return the same result except `observation_id`(new IDs are added on the fly in the SQL with `+1000000000000`) and `value_source_value`(the reason mentioned above). The SQL I ran for the confirmation is attached to the JIRA ticket.


tests/integration_tests/data_steward/cdr_cleaner/cleaning_rules/backfill_pmi_skip_codes_test.py
```diff
@@ -0,0 +1,245 @@
...
+                     'PMI_Skip', None),
+                    (1000000000015, 4, 1586166,
+                     parser.parse('2020-01-01').date(), None, 45905771, None,
+                     None, 903096, 0, 0, None, None, None,
+                     'ElectronicSmoking_ElectricSmokeParticipant', 1586166,
+                     None, None, 903096, 'PMI_Skip', None),
+                ]
+            },
+            {
+                'fq_table_name':
```
ksdkalluri, line 228: @hiro-mishima the above test was very comprehensive. But this test is not needed as we are not modifying anything in-person table. I mean it is not going to help us test anything and increase test running time. Is there anything I am missing regarding this?

data_steward/cdr_cleaner/cleaning_rules/backfill_pmi_skip_codes.py
```diff
@@ -1,208 +1,195 @@
...
+# Project imports
 import constants.cdr_cleaner.clean_cdr as cdr_consts
+from cdr_cleaner.cleaning_rules.base_cleaning_rule import BaseCleaningRule
+from common import JINJA_ENV, OBSERVATION
+from resources import fields_for
 
 LOGGER = logging.getLogger(__name__)
 
-OBSERVATION_TABLE_NAME = 'observation'
+SKIPPED_QUESTIONS = [
```
ksdkalluri, line 18: good idea to refactor this :+1:

ksdkalluri (approved):
Changes look good Hiro! I added a comment about a test case, If you think we can leave it as is, this is good to merge.

PR #1307, opened 2022-07-06
[DC-2545] Update all bq_utils.create_dataset calls in curation

lledezma:


ksdkalluri (approved):
I don't see any function calls for create_dataset from bq_utils, Looks good to merge!

PR #1308, opened 2022-07-06
[DC-2461] Pass `table_namer` parameter to `DropParticipantsWithoutPPI` superclass instantiation.

ksdkalluri:


lledezma (approved):
Looks good!

PR #1310, opened 2022-07-07
[DC-2491] Update create_tier.py to handle fitbit datasets and sunset clean_fitbit.sh

ksdkalluri:
* Add Fitbit to the deid_stage list in create_tier
* Fix the ordering of the cleaning classes based on pipeline flow.


lrwb-aou (changes requested):
This does not address the acceptance criteria about removing `clean_fitbit.sh`.  I think we may also want to remove the `truncate_fitbit.sh` bash script.  Thoughts?

ksdkalluri:
> 

You are right @lrwb-aou, dropping them now.

lrwb-aou:
> > 
> 
> You are right @lrwb-aou, dropping them now.

@ksdkalluri Did those file drops get pushed?

ksdkalluri:
> > > 
> > 
> > 
> > You are right @lrwb-aou, dropping them now.
> 
> @ksdkalluri Did those file drops get pushed?

@lrwb-aou my bad for got to push to the repo. It is updated now.

lrwb-aou (approved):
Looks good!

PR #1311, opened 2022-07-11
[DC-2367] Update `clean_ppi_numeric_fields` cleaning rule to also update value_source_concept_id

ksdkalluri:


hiro-mishima (approved):
Looks good!

PR #1312, opened 2022-07-11
[DC-2554] Update all bq_utils.has_primary_key calls in curation

lledezma:


hiro-mishima (approved):
Looks good, thanks @lledezma  🎉

PR #1313, opened 2022-07-12
[DC-2372] NOTE_NLP QC check falsely says NOT PASS

hiro-mishima:
- pylint automatically updated the script linting, but the only changes in the logic are the following:
- The SQL for `NOTE_NLP` is updated: https://github.com/all-of-us/curation/pull/1313/files#diff-51278d23d019b0eaecc3ebd75524d0adbd25ffe4121e952739f86cb63356fb13R362
- `import urllib` is removed since it is not used: https://github.com/all-of-us/curation/pull/1313/files#diff-51278d23d019b0eaecc3ebd75524d0adbd25ffe4121e952739f86cb63356fb13L20


lledezma (approved):
Looks good!

PR #1316, opened 2022-07-13
[DC-1157] Remove deprecated bigquery list_tables - Google wrapper

lledezma:


nishanthpp93 (approved):
Looks good! I also wonder if delete_all_tables should be a client method? Can be done in a separate ticket. What are your thoughts?

lledezma:
Looks like the `delete_all_tables` function in test_util.py is for removing all non-vocabulary and non-lookup tables from a dataset. I haven't seen the use for creating a method like that for the client. Maybe in the future as I update more calls.

PR #1317, opened 2022-07-15
[DC-2454] update the unit normalization rule

lrwb-aou:
* solves issues with the 5.3 upgrade by changing the write_truncate query to an update query
* adds a record that should be loaded but not modified by the integration test
* fixes any schema issues caused by this rule's use of write truncate

### DIFF
--- a/tests/integration_tests/data_steward/cdr_cleaner/cleaning_rules/unit_normalization_test.py
+++ b/tests/integration_tests/data_steward/cdr_cleaner/cleaning_rules/unit_normalization_test.py
@@ -141,10 +141,17 @@ class UnitNormalizationTest(BaseTest.CleaningRulesTestBase):
         """
         Tests unit_normalization for the loaded test data
         """
+        sandbox_table_name = ''
+        for table in self.fq_sandbox_table_names:
+            if MEASUREMENT in table:
+                sandbox_table_name = table
+
         # Expected results list
         tables_and_counts = [{
             'fq_table_name':
                 f'{self.project_id}.{self.dataset_id}.{MEASUREMENT}',
+            'fq_sandbox_table_name':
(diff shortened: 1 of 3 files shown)


tests/integration_tests/data_steward/cdr_cleaner/cleaning_rules/unit_normalization_test.py
```diff
@@ -143,7 +145,7 @@ def test_unit_normalization(self):
         tables_and_counts = [{
             'fq_table_name':
                 f'{self.project_id}.{self.dataset_id}.{MEASUREMENT}',
-            'loaded_ids': [1, 2, 3, 4, 5, 6, 7, 8, 9, 10],
+            'loaded_ids': [1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11],
```
hiro-mishima, line 148: Shall we also add the following to `tables_and_counts`? I recently learned that `sandbox_ids` are not tested if 'fq_sandbox_table_name' is not explicitly specified in `tables_and_counts`.

```
            'fq_sandbox_table_name':
                f'{self.project_id}.{self.sandbox_id}.{self.rule_instance.sandbox_table_for(MEASUREMENT)}',
```

hiro-mishima (commented):
@lrwb-aou I left one comment. Please let me know if you have any questions. Thanks!

tests/integration_tests/data_steward/cdr_cleaner/cleaning_rules/unit_normalization_test.py
lrwb-aou: nice catch!  I added that to the integration test.

hiro-mishima (approved):
Looks good!

PR #1319, opened 2022-07-18
[DC-2456] make clean_height_weight OMOP 5.3 compliant

lrwb-aou:
* use select * when appropriate to grab the fields as defined in the table
* use fields_for to determine all expected measurement fields from schema file
* update the integration test to ensure the measurement_time and visit_detail_id fields are being populated and referenced as needed by the cleaning rule


data_steward/cdr_cleaner/cleaning_rules/clean_height_weight.py
```diff
@@ -16,18 +16,24 @@
...
 
 HEIGHT_TABLE = 'height_table'
 WEIGHT_TABLE = 'weight_table'
 NEW_HEIGHT_ROWS = 'new_height_rows'
 NEW_WEIGHT_ROWS = 'new_weight_rows'
 
+WEIGHT_CONCEPT_IDS = [3025315, 3013762, 3023166]
+HEIGHT_CONCEPT_IDS = [3036277, 3023540, 3019171]
+
+MEASUREMENT_FIELDS = [field.get('name') for field in fields_for(MEASUREMENT)]
```
ksdkalluri, line 35: Thanks for making this change, by this we don't have to worry about future schema changes.

ksdkalluri (approved):
changes look good, Thanks for making the changes @lrwb-aou!

PR #1321, opened 2022-07-20
[DC-2273] [P1] `ethnicity_source_concept_id` and `ethnicity_source_value` values are not broken down in controlled tier

hiro-mishima:
On top of adding more test cases to the integration test, I tested the updated logic is working against the DEID data from the past. Please see the JIRA ticket's comment section.


jp3477 (approved):
This looks good!

PR #1322, opened 2022-07-20
[DC-2578] Deprecate bq_utils list_dataset_contents

lledezma:


ksdkalluri (approved):
looks good thanks for the changes!

PR #1326, opened 2022-07-22
[DC-2590] Update maps_to_value_ppi_vocab_update CR to New Base Class

hiro-mishima:


jp3477 (approved):
This looks great! I especially like the use of that new `copy_vocab_tables` method.

PR #1327, opened 2022-07-25
[DC-2551] Move and Update `DropParticipantsWithoutPPI` rule to the RDR data cleaning phase

lledezma:


data_steward/cdr_cleaner/cleaning_rules/drop_participants_without_completed_basics.py
```diff
@@ -54,26 +57,23 @@
...
-    USING (observation_id)
   WHERE concept_class_id = 'Module'
     AND concept_name IN ('The Basics')
-    AND src_hpo_id = 'rdr'
     AND questionnaire_response_id IS NOT NULL)
 """)
 
 
-class DropParticipantsWithoutPPI(DropMissingParticipants):
+class DropParticipantsWithoutCompletedBasics(DropMissingParticipants):
```
lrwb-aou, line 66: The name is much better.  But I might suggest changing it to `DropParticipantsWithoutAnyBasics` because as long as the participant has answered at least one question from The Basics and we have that data, we will not drop that participant.  They don't technically have to complete the entire survey.

data_steward/cdr_cleaner/cleaning_rules/drop_participants_without_completed_basics.py
```diff
@@ -54,26 +57,23 @@
...
+    Drops participants who have not completed any of the "The Basics" survey
     """
 
     def __init__(self,
                  project_id,
                  dataset_id,
                  sandbox_dataset_id,
                  table_namer='stage_less'):
-        desc = (f'Sandbox and remove PIDs with no PPI basics or EHR data.'
+        desc = (f'Sandbox and remove PIDs with no PPI basics.'
```
lrwb-aou, line 76: 👍

data_steward/cdr_cleaner/cleaning_rules/drop_rows_for_missing_persons.py
```diff
@@ -1,7 +1,7 @@
 """
 Deletes records for participants who are missing from the person table.
 
-Developed to assist drop_participants_without_ppi_or_ehr.py in removing records
+Developed to assist drop_participants_without_completed_basics.py in removing records
```
lrwb-aou, line 4: nice comment update.  These are sometimes hard to spot.

tests/integration_tests/data_steward/cdr_cleaner/cleaning_rules/drop_participants_without_completed_basics_test.py
```diff
@@ -33,20 +32,17 @@ def setUpClass(cls):
         cls.dataset_id = os.environ.get('COMBINED_DATASET_ID')
```
lrwb-aou, line 32: While not completely necessary, it would be good to change this to use the RDR dataset since it is closer to how we plan to execute the rule. 

```suggestion
        cls.dataset_id = os.environ.get('RDR_DATASET_ID')
```

data_steward/cdr_cleaner/cleaning_rules/drop_participants_without_completed_basics.py
```diff
@@ -54,26 +57,23 @@
...
                 f'Use drop missing participants CR to remove their records.')
 
         super().__init__(issue_numbers=ISSUE_NUMBERS,
```
lrwb-aou, line 79: The constructor here also needs to be updated.  Because we moved the rule out of the `COMBINED_CLEANING_CLASSES` list and into the `RDR_CLEANING_CLASSES` list, we should also update the `affected_datasets` variable to reflect this same change.  This is used if someone ever wants to run the reporting tool to gain more info about the cleaning rules.

Also, I think we need to drop the dependency on `RemoveParticipantDataPastDeactivationDate` now.  That rule has been updated to use `Timestamp` objects instead of `Date` objects now, so the previous dependency is no longer an issue.  I'd be happy to describe the prior issue to you in a call if you'd like.  This will also cause an update to your comment in `clean_cdr.py`.

lrwb-aou (changes requested):
Overall, a nice patch.  I've requested a few  small changes and would be happy to go over them if you want.  Just let me know.

👍

data_steward/cdr_cleaner/clean_cdr.py
```diff
@@ -185,6 +185,8 @@
     (DropMultipleMeasurements,),
     (CleanByBirthYear,),
     (UpdateInvalidZipCodes,),
+    (DropParticipantsWithoutAnyBasics,
+    ),  # dependent on RemoveParticipantDataPastDeactivationDate
```
lrwb-aou, line 189: This comment is no longer valid.  Remove it.

tests/integration_tests/data_steward/cdr_cleaner/cleaning_rules/drop_participants_without_any_basics_test.py
```diff
@@ -30,23 +28,20 @@ def setUpClass(cls):
         cls.project_id = os.environ.get(PROJECT_ID)
 
         # Set the expected test datasets
-        cls.dataset_id = os.environ.get('COMBINED_DATASET_ID')
+        cls.dataset_id = os.environ.get('RDR_DATASET_ID')
```
lrwb-aou, line 31: 👍

data_steward/cdr_cleaner/cleaning_rules/drop_participants_without_any_basics.py
```diff
@@ -54,26 +57,23 @@
...
     def __init__(self,
                  project_id,
                  dataset_id,
                  sandbox_dataset_id,
                  table_namer='stage_less'):
-        desc = (f'Sandbox and remove PIDs with no PPI basics or EHR data.'
+        desc = (f'Sandbox and remove PIDs with no PPI basics.'
                 f'Use drop missing participants CR to remove their records.')
 
         super().__init__(issue_numbers=ISSUE_NUMBERS,
```
lrwb-aou, line 79: The class initializer needs to be refined.  Line 81 should be updated to 
```python
                         affected_datasets=[cdr_consts.RDR],
```

Lines 86-87 should become
```python
                         table_namer=table_namer)
```

This also means you can drop the import for `RemoveParticipantDataPastDeactivationDate`.

data_steward/cdr_cleaner/cleaning_rules/drop_participants_without_any_basics.py
```diff
@@ -1,16 +1,19 @@
 """
-Run the drop_participants_without_ppi_or_ehr validation clean rule.
+Run the drop_participants_without_any_basics validation clean rule.
```
lrwb-aou, line 2: Nice name change.

lrwb-aou (changes requested):
The class initializer needs to be updated and one more comment needs updating.  Then this can be merged.  👍

lrwb-aou (approved):
This is good!

PR #1330, opened 2022-07-26
[DC-2475] Account for text answers when assigning the invalid code for '1333023'.

hiro-mishima:
-  `cdr_deid_qa_report8_household_state_genera.py` has lots of changes because of `pylint`, but the only notable change is removing [this comment](https://github.com/all-of-us/curation/pull/1330/files#diff-bfc8a31439776b1f1be946f98df34a8a06a6911ecced825491ce2c1029d33b1cL82) that will be no longer relevant after this PR
- `observation_id=122` is added to cover the updated logic in the integration test.

### DIFF
--- a/data_steward/cdr_cleaner/cleaning_rules/clean_ppi_numeric_fields_using_parameters.py
+++ b/data_steward/cdr_cleaner/cleaning_rules/clean_ppi_numeric_fields_using_parameters.py
@@ -2,7 +2,7 @@
 Apply value ranges to ensure that values are reasonable and to minimize the likelihood
 of sensitive information (like phone numbers) within the free text fields.
 
-Original Issues: DC-1058, DC-1061, DC-827, DC-502, DC-487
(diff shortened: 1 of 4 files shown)


data_steward/cdr_cleaner/cleaning_rules/clean_ppi_numeric_fields_using_parameters.py
```diff
@@ -2,7 +2,7 @@
 Apply value ranges to ensure that values are reasonable and to minimize the likelihood
 of sensitive information (like phone numbers) within the free text fields.
 
-Original Issues: DC-1058, DC-1061, DC-827, DC-502, DC-487
+Original Issues: DC-1058, DC-1061, DC-827, DC-502, DC-487, DC-2475
```
lledezma, line 5: Thanks! Adding it in the integration and unit test files would be a good idea.

lledezma (approved):
Thanks, @hiro-mishima! It looks good to me.

PR #1333, opened 2022-07-26
[DC-2584] Modify WHERE condition for DATETIME data type

hiro-mishima:
- `retract_deactived_pids.py` which is mentioned in the original ticket is refactored into `remove_participant_data_past_deactivation_date.py`


nishanthpp93 (approved):
Looks good!

PR #1335, opened 2022-07-29
[DC-2596] Clean up environment after main_test finishes

ksdkalluri:


tests/integration_tests/data_steward/validation/main_test.py
```diff
@@ -58,6 +57,7 @@ def setUp(self):
         self.storage_client.empty_bucket(self.drc_bucket)
         self.storage_client.empty_bucket(self.hpo_bucket)
 
+        test_util.delete_all_tables(self.bq_client, self.dataset_id)
```
hiro-mishima, line 60: This line is identical to the line below. If my understanding is correct, we want it for `self.rdr_dataset_id`.

hiro-mishima (commented):
Thanks for working on it, @ksdkalluri! I left one comment. Could you please have a look? Thank you:)

hiro-mishima (approved):
Looks good!

PR #1336, opened 2022-08-01
[DC-2458] Update gcloud/bq/__init__.py copy_dataset to expect project as part of input_dataset and output_dataset parameters

lledezma:
This was the branch that updated the `bq.copy_dataset` calls: #1212

### DIFF
--- a/data_steward/gcloud/bq/__init__.py
+++ b/data_steward/gcloud/bq/__init__.py
@@ -194,8 +194,8 @@ class BigQueryClient(Client):
         """
         Copies tables from source dataset to a destination datasets
 
-        :param input_dataset: name of the input dataset
-        :param output_dataset: name of the output dataset
+        :param input_dataset: name of the input(source) dataset
+        :param output_dataset: name of the output(destination) dataset
         :return:
         """
         # Copy input dataset tables to backup and staging datasets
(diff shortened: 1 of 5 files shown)


data_steward/gcloud/bq/__init__.py
```diff
@@ -194,14 +194,14 @@ def copy_dataset(self, input_dataset: str, output_dataset: str):
...
-        :param output_dataset: name of the output dataset
+        :param input_dataset: name of the input(source) dataset
+        :param output_dataset: name of the output(destination) dataset
         :return:
         """
         # Copy input dataset tables to backup and staging datasets
         tables = super(BigQueryClient, self).list_tables(input_dataset)
         for table in tables:
-            staging_table = f'{self.project}.{output_dataset}.{table.table_id}'
+            staging_table = f'{output_dataset}.{table.table_id}'
```
ksdkalluri, line 204: I am confused, from what I understand the ticket says we need to add project_id for both input and output dataset parms please update accordingly.

ksdkalluri (changes requested):

data_steward/gcloud/bq/__init__.py
lledezma: **Scope**
- Fix gcloud/bq/__init__.py copy_dataset function so that project is not appended to the output_dataset parameter.

The project_id is added as part of the argument, meaning the method is called with fully qualified dataset names.

ksdkalluri: can you update the docstring to say that the dataset name should be fully qualified?

PR #1337, opened 2022-08-02
[DC-2132] added sleep tables to cleaning rule

jp3477:


molly-giddings:
@lledezma - Would you mind resolving the conflicts in @jp3477 's PR while he is out?  I believe these are the same files that you are looking at in DC-2134.

molly-giddings (approved):
Thank you, @jp3477 and @lledezma !  These changes look good and also correspond with the work on `DC-2134`.

PR #1341, opened 2022-08-05
[DC-2552] [P1] failed participant match should remove EHR data, not all participant data

hiro-mishima:


tests/integration_tests/data_steward/cdr_cleaner/cleaning_rules/remove_non_matching_participant_test.py
```diff
@@ -47,22 +46,76 @@
         INSERT INTO `{{fq_table_name}}` 
         (observation_id, person_id, observation_concept_id, observation_date, observation_type_concept_id)
         VALUES
```
lrwb-aou, line 48: @hiro-mishima I spent some time yesterday looking at this integration test.  It is duplicating a lot of work that the base class does for free.  I'm adding a patch file for your review with this test.  We can talk about it whenever you would like.
[remove_non_matching_participant_test_patch.txt](https://github.com/all-of-us/curation/files/9376283/remove_non_matching_participant_test_patch.txt)

Let me know if you need help applying the patch file.

lrwb-aou (changes requested):
Thank you for working on this.  I think you can simplify some of the SQL statements by making use of the jinja2 templates.  This will make the code more maintainable.

I also added a suggested patch for the integration test.  It takes advantage of the base test class.

data_steward/cdr_cleaner/cleaning_rules/sandbox_and_remove_pids.py
```diff
@@ -70,76 +80,98 @@ def __init__(self,
...
-        for table in self.person_table_list:
+        for table in self.affected_tables:
+
+            if ehr_only:
+                ehr_only_join = SANDBOX_EHR_ONLY_JOIN.render(
+                    project=self.project_id,
+                    dataset=self.dataset_id,
+                    table=table)
+            else:
+                ehr_only_join = ''
```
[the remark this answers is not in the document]
hiro-mishima: This is really cool, Lauren! Thanks for sharing. The code is way cleaner now.

tests/integration_tests/data_steward/cdr_cleaner/cleaning_rules/remove_non_matching_participant_test.py
hiro-mishima: Thanks for the patch file! The patch makes sense to me. Now the test looks cleaner 🎉

hiro-mishima: This addition to the `data_steward/gcloud/bq/__init__.py` will be quite helpful. Thanks.

data_steward/gcloud/bq/__init__.py
```diff
@@ -73,7 +73,12 @@ def get_table_schema(self, table_name: str, fields=None) -> list:
         if fields:
             fields = fields
         else:
-            fields = fields_for(table_name)
+            if table_name.endswith('identity_match'):
+                fields = fields_for('identity_match')
+            elif table_name.endswith('participant_match'):
+                fields = fields_for('participant_match')
```
lrwb-aou, line 79: I know this is directly from the patch file I gave you, but I was using the quick and dirty method for the patch.  These strings should be replaced by values defined in common.py.

lrwb-aou (approved):
This is a good patch.   I did leave one comment about some of the patch file code I created that should be cleaner than it is.  Don't hesitate to update patch file code when you see it can be better.  :)  Enjoy your time off.

PR #1343, opened 2022-08-11
[DC-2608] Update the survey version and dates check in rdr qc nb

brendagutman:


molly-giddings (approved):
@brendagutman - This looks great!  Your description of the check is very clear and I especially like the documentation link for the implementation ranges.  I ran the updated query on the rdr20220815 data and verified vaccine_4.  Go ahead and merge your PR when ready.

PR #1344, opened 2022-08-12
[DC-2134] Verify if `RemoveParticipantDataPastDeactivationDate` does or does not require an update

lledezma:


molly-giddings:
@lledezma - This looks good to me, but could you resolve the conflict?  Once that's taken care of, I will approve.  Thanks!

molly-giddings (approved):
Thanks, @lledezma !  Adding `sleep_level` table test data is a good addition for testing the addition of both sleep_level and sleep_daily_summary tables to the fitbit dataset.  I like that you chose it rather than sleep_daily_summary, as it has the datetime field comparison as well as the date comparison and will test the more complicated branch of the rule.

PR #1347, opened 2022-08-15
[DC-2643] Move Basic questions check to rdr_export_qc notebook.

ksdkalluri:


data_steward/analytics/cdr_ops/rdr_export_qc.py
```diff
@@ -439,6 +439,54 @@
...
+  FROM `{{DATASET_ID}}.observation` o
+  JOIN basics_concept c
+   ON o.observation_concept_id = c.concept_id
+  WHERE 1=1
+  GROUP BY 1)
+
+ -- list all pids for whom no basics questions are found
+SELECT * 
+FROM pid_basics
+WHERE ARRAY_LENGTH(basics_codes) = 0
```
lrwb-aou, line 484: This does not work.  This check always passes because nothing is aggregated when a string doesn't exist.  A better check is:
```suggestion
FROM `{{DATASET_ID}}.person`
WHERE person_id not in (select person_id from pid_basics)
```
You can run both queries against a raw rdr dataset and see the difference.

data_steward/analytics/cdr_ops/rdr_export_qc.py
```diff
@@ -439,6 +439,54 @@
...
+ basics_concept AS
+ (SELECT
+   c.concept_id
+  ,c.concept_name
+  ,c.concept_code
+  FROM `{{DATASET_ID}}.concept_ancestor` ca
+  JOIN `{{DATASET_ID}}.concept` c
+   ON ca.descendant_concept_id = c.concept_id
+  WHERE 1=1
+    AND ancestor_concept_id={BASICS_MODULE_CONCEPT_ID}
```
lrwb-aou, line 466: I think this needs to be double bracketed and passed in as a parameter.

lrwb-aou (changes requested):
Thanks for moving this to earlier in our pipeline.  Please look at the requested changes.

lrwb-aou (approved):
Looks good!

PR #1349, opened 2022-08-16
[DC-2135] Update `RemoveFitbitDataIfMaxAgeExceeded` class (Registered Tier)

lledezma:


lrwb-aou (approved):
Looks good!  I like it.

PR #1352, opened 2022-08-17
[DC-2609] Update the survey_version_info with New Year Minute

brendagutman:


molly-giddings (approved):
Thank you, @brendagutman !  It looks great.

PR #1354, opened 2022-08-17
[DC-2683] Add insurance_lookup creation to MapHealthInsuranceResponses CR setup

ksdkalluri:


data_steward/cdr_cleaner/cleaning_rules/map_health_insurance_responses.py
```diff
@@ -191,7 +193,14 @@ def setup_rule(self, client, *args, **keyword_args):
         the logic to get the row counts of the tables prior to applying cleaning rule
 
         """
-        pass
+        client.create_tables(
+            [f'{self.project_id}.{self.dataset_id}.{INSURANCE_LOOKUP}'], False,
```
lrwb-aou, line 197: Should this be the sandbox dataset id?

ksdkalluri: you are right, I will update it.

MichaelSchmidt82 (approved):
Works for me

PR #1356, opened 2022-08-18
[DC-2136] Remap `person_id` to correct `research_id` for sleep tables

lledezma:


molly-giddings (commented):
Thanks, @lledezma - this is really nice.  I like your clean coding style, replacing concatenation with f strings and eliminating unnecessary variables.

It looks like `pid_rid_map.py` snuck in due to some spacing in comments - can you remove it from the PR?  I'd rather not update it unless a real change was made.

molly-giddings (approved):

PR #1360, opened 2022-08-19
[DC-2585] Modify PS script to allow running on full participant summary data or specified hpo_id

nishanthpp93:
* [DC-2585] Use string all_hpo
* [DC-2585] Add pagination

[DC-2585]: https://precisionmedicineinitiative.atlassian.net/browse/DC-2585?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[DC-2585]: https://precisionmedicineinitiative.atlassian.net/browse/DC-2585?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/data_steward/utils/participant_summary_requests.py
+++ b/data_steward/utils/participant_summary_requests.py
@@ -149,6 +149,60 @@ def get_participant_data(api_project_id: str,
     return participant_data
 
 
+def get_paginated_participant_data(api_project_id: str,
+                                   params: Dict,
+                                   url: str = None,
+                                   session: Session = None,
+                                   expected_fields: List[str] = None) -> Dict:
+    """
+    Fetches participant data via ParticipantSummary API with pagination
+
+    :param api_project_id: RDR project id when PS API rests
+    :param params: the fields and their values
+    :param url: if paginating, use url with token to continue
+    :param session: persisting http session
+    :param expected_fields: filter participants not containing any of the fields.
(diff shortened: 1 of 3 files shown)


molly-giddings (commented):
This is great - thank you, @nishanthpp93 !  I have a few questions just to clarify:

- Since `get_org_participant_information`, called when getting data for a specific hpo, will still call `get_participant_data` where all the results are processed to dataframe and added to BigQuery at one time, could this function also benefit from the new `get_paginated_participant_data`?  Or is the data volume in this case small enough to not cause a problem as with getting all hpo data?
- With the addition of the `append` parameter to `store_participant_data` function with default value of `False`, is no-append the desired behavior when `fetch_and_store_ps_hpo_data` is run?  See lines 126-130 in `store_participant_summary_results.py`.

nishanthpp93:
@molly-giddings Thanks! To address your comments:
* Yes, get paginated data is not necessary in those cases, where the data is ~50x smaller.
* Ideally, we want fresh data each time so no append is the default, and only used when loading paginated data.

molly-giddings (approved):
Excellent - thanks again, @nishanthpp93

PR #1366, opened 2022-08-24
[DC-2626] Update common.py to include the survey_conduct table

MichaelSchmidt82:
Add variables (and list variable) needed to include the survey_conduct CATI table.


molly-giddings (approved):
This looks good, @MichaelSchmidt82.  Thanks!

PR #1371, opened 2022-08-31
[DC-2624] Add `questionnaire_response_additional_info` table schema to rdr schema files

MichaelSchmidt82:
This adds an additional RDR schema, questionnaire_additional_info

### DIFF
--- a/data_steward/resource_files/schemas/rdr/questionnaire_response_additional_info.json
+++ b/data_steward/resource_files/schemas/rdr/questionnaire_response_additional_info.json
@@ -15,6 +15,6 @@
     "type": "string",
     "name": "value",
     "mode": "nullable",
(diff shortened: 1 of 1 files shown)


ksdkalluri (commented):
This looks good, you can merge it in when Molly gets a chance to review and approve this.

molly-giddings (changes requested):
@MichaelSchmidt82, @ksdkalluri - This looks good and matches the data we received from RDR.  Could we amend the description of the value field to something like:
`Values in { LANGUAGE: [en, es], NON_PARTICIPANT_AUTHOR_INDICATOR: [CATI]}, CODE: [survey taken]`

I'm thinking right now there are 91 coming through the data, but that won't always be the case.

MichaelSchmidt82:
@molly-giddings , didn't catch that.  Updated

molly-giddings (approved):

PR #1373, opened 2022-08-31
[DC-2695] Add UnityPoint Health to pipeline

hiro-mishima:
- `COOK_COUNTY` is missing, so I included it as well. I double-checked there is no other HPO site missing from the list.


nishanthpp93 (approved):
Looks good!

PR #1379, opened 2022-08-31
[DC-2631] update remove_participant_data_past_deactivation_date tou uses resources.CDM_TABLES instead of common.CDM_TABLES to include survey_conduct table

ksdkalluri:


molly-giddings (changes requested):
Thanks for this, @ksdkalluri - it looks good!  Did you run `reporter.py` to verify that survey_conduct would be included in the SQL?  If so, would you please attach a screenshot to the ticket?  Alternatively, you could add survey_conduct to the integration test `remove_participant_data_past_deactivation_date_test.py`, whichever you'd prefer.

ksdkalluri:
> Thanks for this, @ksdkalluri - it looks good! Did you run `reporter.py` to verify that survey_conduct would be included in the SQL? If so, would you please attach a screenshot to the ticket? Alternatively, you could add survey_conduct to the integration test `remove_participant_data_past_deactivation_date_test.py`, whichever you'd prefer.

@molly-giddings Added integration test for survey_conduct_table.

molly-giddings (approved):
Thank you, @ksdkalluri !

PR #1380, opened 2022-08-31
[DC-2634] Update temporal_consistency cleaning rule to include the survey_conduct table

ksdkalluri:


nishanthpp93 (approved):
Looks good!

PR #1382, opened 2022-09-02
[DC-2627] Create a cleaning rule that will populate the `survey_conduct_ext` table

hiro-mishima:
- See the comment in the JIRA ticket for DC-2627. This new cleaning rule assumes `survey_conduct_ext` and `questionnaire_response_additional_info` can be joined on `survey_conduct_id = questionnaire_response_id`.

### DIFF
--- a/data_steward/cdr_cleaner/cleaning_rules/populate_survey_conduct_ext.py
+++ b/data_steward/cdr_cleaner/cleaning_rules/populate_survey_conduct_ext.py
@@ -22,14 +22,16 @@ AS
     SELECT sce.* FROM `{{project_id}}.{{dataset_id}}.survey_conduct_ext` AS sce
     JOIN `{{project_id}}.{{dataset_id}}.questionnaire_response_additional_info` AS qrai
     ON sce.survey_conduct_id = qrai.questionnaire_response_id
-    WHERE sce.type IS NULL OR sce.value IS NULL OR sce.type != qrai.type OR sce.value != qrai.value
+    WHERE (sce.language IS NULL OR sce.language != qrai.value)
+    AND qrai.type = 'LANGUAGE'
 """)
 
 UPDATE_SURVEY_CONDUCT_EXT_QUERY = JINJA_ENV.from_string("""
 UPDATE `{{project_id}}.{{dataset_id}}.survey_conduct_ext` AS sce
-SET type = qrai.type, value = qrai.value
+SET language = qrai.value
 FROM `{{project_id}}.{{dataset_id}}.questionnaire_response_additional_info` AS qrai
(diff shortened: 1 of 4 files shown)


data_steward/cdr_cleaner/cleaning_rules/populate_survey_conduct_ext.py
```diff
@@ -0,0 +1,128 @@
...
+LOGGER = logging.getLogger(__name__)
+
+SANDBOX_SURVEY_CONDUCT_EXT_QUERY = JINJA_ENV.from_string("""
+CREATE TABLE `{{project_id}}.{{sandbox_dataset_id}}.{{sandbox_table}}`
+AS
+    SELECT sce.* FROM `{{project_id}}.{{dataset_id}}.survey_conduct_ext` AS sce
+    JOIN `{{project_id}}.{{dataset_id}}.questionnaire_response_additional_info` AS qrai
+    ON sce.survey_conduct_id = qrai.questionnaire_response_id
+    WHERE (sce.language IS NULL OR sce.language != qrai.value)
+    AND qrai.type = 'LANGUAGE'
```
molly-giddings, line 26: Because these queries are doing a string comparison, I would suggest modifying lines 26 and 34 to something like:
`AND UPPER(qrai.type) = 'LANGUAGE'`

to preserve the logic if what we receive changes case or has inconsistent case.

tests/integration_tests/data_steward/cdr_cleaner/cleaning_rules/populate_survey_conduct_ext_test.py
```diff
@@ -0,0 +1,111 @@
...
+
+        # set the test project identifier
+        project_id = os.environ.get(PROJECT_ID)
+        cls.project_id = project_id
+
+        # set the expected test datasets
+        dataset_id = os.environ.get('RDR_DATASET_ID')
+        cls.dataset_id = dataset_id
+        sandbox_id = dataset_id + '_sandbox'
+        cls.sandbox_id = sandbox_id
```
molly-giddings, line 57: I've reviewed some other PRs recently that have collapsed lines 50 and 51 into one like so:
`cls.project_id = os.environ.get(PROJECT_ID)`

Same for lines 54 and 55:
`cls.dataset_id = os.environ.get('RDR_DATASET_ID')`

And also removed string concatenation in favor of f-strings on line 56.

For consistency, could you make those few changes?

molly-giddings (changes requested):
Thank you so much, @hiro-mishima  - very easy to follow code and I really appreciate your flexibility with changing the schema of the `questionnaire_response_additional_info` table while this work was in progress.  I have a few minor suggestions for you and it should be good to go.

PR #1386, opened 2022-09-06
[DC-2705] Modify `survey_conduct_ext` extension table schema in extension_tables schema files

MichaelSchmidt82:
Remove two schema fields (`value`, `type`) and add `language`


molly-giddings (approved):
Thanks, @MichaelSchmidt82 !

PR #1387, opened 2022-09-06
[DC-2637] Update questionnaire_response_id_map cleaning rule to include the survey_conduct table

hiro-mishima:
- Updated docstrings to better represent what this CR does.
- Refactored the mapping query to handle both `observation` and `survey_conduct`.
- Added sandbox and delete queries for `survey_conduct`.
- Added the test case for `survey_conduct` in the integration test.
- Updated the test case for `observation` so it is easier to understand.


molly-giddings (approved):
Thanks, @hiro-mishima!

ksdkalluri (approved):

PR #1389, opened 2022-09-07
[DC-2651] Add cleaning rule to copy lookup table and use values to update `measurement.value_as_concept_id` rows

ksdkalluri:


mark-velez:
* Remove the ')' in the UPDATE query
 * Reference the lookup table from the sandbox dataset rather than pipeline tables
 * Set the standard concept field in the lookup table (in BQ, not in this PR)
 * Add comment to integration test with expectations

data_steward/cdr_cleaner/cleaning_rules/dedup_measurement_value_as_concept_id.py
```diff
@@ -0,0 +1,168 @@
...
+            )
+        except NotFound:
+            job = client.copy_table(
+                f'{self.project_id}.{PIPELINE_TABLES}.{IDENTICAL_LABS_LOOKUP_TABLE}',
+                f'{self.project_id}.{self.sandbox_dataset_id}.{IDENTICAL_LABS_LOOKUP_TABLE}'
+            )
+            job.result()
+            LOGGER.info(
+                f'Copied {PIPELINE_TABLES}.{IDENTICAL_LABS_LOOKUP_TABLE} to '
+                f'{self.sandbox_dataset_id}.{IDENTICAL_LABS_LOOKUP_TABLE}')
```
mark-velez, line 139: Is there a reason not to do a create/replace of the lookup table here?
```python
client.copy_table(
                f'{self.project_id}.{PIPELINE_TABLES}.{IDENTICAL_LABS_LOOKUP_TABLE}',
                f'{self.project_id}.{self.sandbox_dataset_id}.{IDENTICAL_LABS_LOOKUP_TABLE}',
                job_config=CopyJobConfig(write_disposition=google.cloud.bigquery.WriteDisposition.WRITE_TRUNCATE)
            )
```
It seems strange that the primary scenario (i.e. table not loaded yet) throws/catches exception. Also if table does exist at destination I think it is more likely that we will want to overwrite it e.g. if the pipeline table is more up to date. Let me know what you think.

ksdkalluri: Thanks for the suggestion Mark.

mark-velez (approved):
Thanks @ksdkalluri . I also learned UPDATE syntax with this review. 😃

PR #1390, opened 2022-09-08
[DC-2639] Remap `person_id` to correct `research_id` for survey_conduct table

jp3477:


data_steward/cdr_cleaner/clean_cdr.py
```diff
@@ -278,6 +278,7 @@
 ]
 
 REGISTERED_TIER_DEID_CLEAN_CLEANING_CLASSES = [
+    (RtCtPIDtoRID,),
```
nishanthpp93, line 281: Could this apply pid-rid mapping twice? I think we have some code to map pid to rid in the deid package.

nishanthpp93 (commented):
I think this achieves the objective but would like some more clarity on the application of pid-rid mapping.

data_steward/cdr_cleaner/clean_cdr.py
jp3477: Yeah, the ticket says to remove the functionality from the deid module. I'm sorting through it to figure out where that is.

nishanthpp93: Thanks, I think @hiro-mishima just looked at the module so you may get more insight from him.

hiro-mishima: `run_deid.py` runs `aou.py`, which takes care of PID->RID mapping for the registered tier for all the tables. `run_deid.py` runs before `clean_cdr.py`. I think it is simpler for now to keep `run_deid.py` and `aou.py` do the PID->RID mapping for the registered tier and keep this CR's focus on the controlled tier only. 
Please see the comment in the JIRA ticket DC-2638 for some additional info.

jp3477: Thanks for the response @hiro-mishima ! I believe this ticket specifically wants the registered tier pid_rid mapping functionality to move from the legacy deid code to a cleaning rule. Would you suggest it be a Won't Fix then?

hiro-mishima: I think we can make this ticket `Won't Fix`. I would love to double-check with @molly-giddings to see if we are missing anything we should cover in `DC-2638` and `DC-2639`.
@molly-giddings any thoughts?

nishanthpp93: Also @molly-giddings I just merged #1393 but I think @hiro-mishima's ticket already covers it. I could just remove it from the list of RT rules.

molly-giddings: Hi @hiro-mishima, @jp3477, @nishanthpp93 - we would eventually like to move off of the legacy deid code, and pulling these pieces off would be a big step towards that.  I would like to keep this cleaning rule and have the functionality commented out/removed in the the legacy code.  I think if you comment out line 199 in `deid/cofig/ids/config.json`, that will remove the pid-rid portion of the query that is built.  If that works, we can also remove lines 269-299 of `deid/aou.py` but will have to modify the function return value.  And finally, that allows us to refactor `tools/run_deid.py` to remove the age-limit parameter and all of the functionality for loading the deid map table.  If anyone would like a meeting to discuss, let me know and we can set something up.

nishanthpp93: Thanks @molly-giddings, I'd appreciate a meeting tomorrow for this. My morning is free.

hiro-mishima: I will be free in the morning too

molly-giddings: Yes, @nishanthpp93 - you are correct about #1393 being covered, but your work on it will be the model for date-shifting all of the OMOP tables in the near future as we move further off the legacy code.  Lauren and I had a long conversation about the date shifting, and I had some confusion about `date_shift_cope_responses.py` as it is actually providing an UNshifting of the dates that were already shifted specific to COPE surveys.  I will be making a new ticket to modify `date_shift_cope_responses.py` to provide unshifting for survey_conduct, and as part of that we will want to comment out (not remove) the cleaning rule SurveyConductDateShift in `clean_cdr.py`.  I would also request that the cleaning rule name is modified to clarify that it is unshifting, whereas the FitbitDateshift rule and SurveyConductDateshift rule are shifting.

I will also make a new ticket for making sure the survey_conduct table data is removed if the maximum age is exceeded, as that will be needed due to the refactoring of `run_deid.py`.

We can discuss all this and more fun tomorrow morning - I'll post on slack for whoever wants/is able to join.

PR #1391, opened 2022-09-08
[DC-2635] Update `ensure_date_datetime_consistency` cleaning rule to include the survey_conduct table

ksdkalluri:


nishanthpp93 (changes requested):
Looks good! Only change is to add the ticket number too.

nishanthpp93 (approved):

PR #1397, opened 2022-09-12
[DC-2606] Add a check to the Fitbit QA notebook for data in the sleep_level table with invalid level

lledezma:


data_steward/analytics/cdr_ops/rt_cdr_qc/cdr_deid_qa_report6_fitdata.py
```diff
@@ -549,6 +552,53 @@
         ignore_index=True)
 df1.T
 
+# # Verify that all records with invalid level values are being dropped from sleep_level table
+#
+# DC-2606
+#
+# Queries the sandbox table for the corresponding cleaning rule (created in DC-2605) and outputs any records that are being dropped due to invalid level values. It also outputs any records in the sleep_level table that should have been dropped but were not.
```
lrwb-aou, line 559: This line has 258 characters.  How about breaking it somewhere between 100 and 110 characters in length.

data_steward/analytics/cdr_ops/rt_cdr_qc/cdr_deid_qa_report6_fitdata.py
```diff
@@ -549,6 +552,53 @@
...
+
+df2 AS (
+SELECT 
+*,'not dropped' AS dropped_status
+FROM 
+`{project_id}.{fitbit_dataset}.sleep_level`
+WHERE 
+(
+    LOWER(level) NOT IN 
+         ('awake','light','asleep','deep','restless','wake','rem','unknown') OR level IS NULL))
```
lrwb-aou, line 577: I would suggest this might be more meaningful if the data is condensed to a different format rather than a "get all" format.  Perhaps
```python
WITH df1 AS (
SELECT 
level
, 'dropped' AS dropped_status,
, count(*) as n_violations
FROM 
`{project_id}.{fitbit_sandbox_dataset}.{sleep_level_sandbox_table}`
group by 1, 2),
df2 AS (
SELECT 
level
, 'not dropped' AS dropped_status,
, count(*) as n_violations
FROM 
`{project_id}.{fitbit_dataset}.sleep_level`
WHERE 
(
    LOWER(level) NOT IN 
         ('awake','light','asleep','deep','restless','wake','rem','unknown') OR level IS NULL)
)
```

@molly-giddings thoughts on this change?

data_steward/analytics/cdr_ops/rt_cdr_qc/cdr_deid_qa_report6_fitdata.py
```diff
@@ -549,6 +552,53 @@
...
+            'result': 'PASS'
+        },
+        ignore_index=True)
+else:
+    df = df.append(
+        {
+            'query': 'Query6 no invalid level records in sleep_level table',
+            'result': ''
+        },
+        ignore_index=True)
```
lrwb-aou, line 599: This doesn't seem to include a failure or notification if we have a `dropped` status for something in that field.  I think it should have something if we are getting records we are not expecting and dropping them.  Maybe it's not a failure status but a "Follow up" status.

data_steward/analytics/cdr_ops/rt_cdr_qc/cdr_deid_qa_report6_fitdata.py
```diff
@@ -549,6 +552,53 @@
...
+df2
```
lrwb-aou, line 600: Is this line necessary since you are using the `execute` function?

lrwb-aou (commented):
I've added some suggestions for you, @lledezma , and Molly to review.  Let me know if anything seems odd or if I'm not understanding some intention.

data_steward/analytics/cdr_ops/rt_cdr_qc/cdr_deid_qa_report6_fitdata.py
molly-giddings: @lledezma, @lrwb-aou - I love this change.  This should give us the counts of the dropped data by distinct level which may be helpful in correcting it, and also cut down on the number of records output.  Maybe also throw in an `order by n_violations` in df1.

lrwb-aou: @lledezma let me know when you've had a chance to review the comments and I'll take another look at the PR.

data_steward/analytics/cdr_ops/rt_cdr_qc/cdr_deid_qa_report6_fitdata.py
lledezma: This is the code that prints out a list that includes the status for each QA check in the notebook. The status for Query6 will be in that list

data_steward/analytics/cdr_ops/rt_cdr_qc/cdr_deid_qa_report6_fitdata.py
```diff
@@ -549,6 +552,61 @@
...
+
 # # Summary_fitdata
 
 # if not pass, will be highlighted in red
```
lledezma, line 612: This is the code that prints out the list that includes the status for each QA check in the notebook. The status for Query6 will be in that list.

data_steward/analytics/cdr_ops/rt_cdr_qc/cdr_deid_qa_report6_fitdata.py
lledezma: `if not pass, will be highlighted in red
df = df.mask(df.isin(['Null', '']))
df.style.highlight_null(null_color='red').set_properties(
    **{'text-align': 'left'})`

Lines 613-615

This is the code that prints out the list that includes the status for each QA check in the notebook. The status for Query6 will be in that list.

lrwb-aou (approved):
I like the changes.  I approve if @molly-giddings agrees also.

molly-giddings (approved):
@lledezma, @lrwb-aou - It looks good to me.  Thanks for the updates, Luis!

PR #1398, opened 2022-09-13
[DC-2638] Add survey_conduct to deid_tables list

hiro-mishima:
- See the comment in the JIRA ticket for more details.


molly-giddings (approved):
Thanks, @hiro-mishima !

PR #1399, opened 2022-09-13
[DC-806] Clean smoking ppi to base class and integration test

brendagutman:


data_steward/cdr_cleaner/cleaning_rules/clean_smoking_ppi.py
```diff
@@ -0,0 +1,362 @@
...
+                f'{self.__class__.__name__} unexpectedly changed table counts.\n'
+                f'initial count is: {self.init_counts.get("total_rows")}\n'
+                f'clean count is: {clean_counts.get("total_rows")}')
+
+        if clean_counts.get('child_count') or clean_counts.get('parent_count'):
+            raise RuntimeError(
+                f'{self.__class__.__name__} did not clean as expected.\n'
+                f'Found data in: {clean_counts}')
+
+    def _get_counts(self, client):
```
lrwb-aou, line 318: Nice usage of the leading underscore.  Consider adding the usage definitions/hints to all method parameters.

data_steward/cdr_cleaner/cleaning_rules/clean_smoking_ppi.py
```diff
@@ -0,0 +1,362 @@
...
+        """
+        Counts query.
+        Used for job validation.
```
lrwb-aou, line 321: Add a blank line after the initial summary line.  This is useful for automatic document generation, if we ever start using it. 

https://peps.python.org/pep-0257/#multi-line-docstrings

data_steward/cdr_cleaner/clean_cdr.py
```diff
@@ -173,8 +173,7 @@
     # won't work with mocked strings.  should use base class
     # setup_query_execution function to load dependencies before query execution
```
lrwb-aou, line 174: this comment should now be removed.  It was specific to the old style smoking rule that you are updating.

tests/integration_tests/data_steward/cdr_cleaner/cleaning_rules/clean_smoking_test.py
```diff
@@ -0,0 +1,110 @@
...
+    def setUpClass(cls):
+        print('**************************************************************')
+        print(cls.__name__)
+        print('**************************************************************')
+
+        super().initialize_class_vars()
+
+        cls.project_id = os.environ.get(PROJECT_ID)
+        cls.dataset_id = os.environ.get('RDR_DATASET_ID')
+        cls.sandbox_id = cls.dataset_id + '_sandbox'
```
lrwb-aou, line 49: This could be updated to an f-string representation.

tests/integration_tests/data_steward/cdr_cleaner/cleaning_rules/clean_smoking_test.py
```diff
@@ -0,0 +1,110 @@
...
+
+        cls.rule_instance = CleanSmokingPpi(cls.project_id, cls.dataset_id,
+                                            cls.sandbox_id)
+
+        cls.fq_table_names = [
+            f'{cls.project_id}.{cls.dataset_id}.{OBSERVATION}'
+        ]
+
+        cls.fq_sandbox_table_names = [
+            f'{cls.project_id}.{cls.sandbox_id}.{cls.rule_instance.sandbox_table_for(OBSERVATION)}'
```
lrwb-aou, line 59: You've got a lookup table loaded in the sandbox too, right?  Add that name here and it will be deleted by the base integration test class when the class cleans up after itself.

The integration test should be non-side effecting.  Nothing should be left in the environment that was created by the class being tested.

lrwb-aou (changes requested):
This is really nice.  I've got some suggestions for following some coding conventions.  The integration test looks good.  I've you've got some questions about any of my comments, let me know.

nishanthpp93:
@brendagutman The linting check I added long ago fails for anything not matching [DC-123] so I updated the PR title

lrwb-aou:
> @brendagutman The linting check I added long ago fails for anything not matching [DC-XYZ] so I updated the PR title

It isn't the title of the PR on github that causes the failure.  The PR title should contain the Jira issue number to link it back to the correct Jira issue.  
It is the first line of your commit message that is causing the linting failure.  That must adhere to the `[DC-###] extra info here if you want it ` pattern.

data_steward/cdr_cleaner/clean_cdr.py
```diff
@@ -184,6 +183,7 @@
     (RoundPpiValuesToNearestInteger,),
     (UpdateFamilyHistoryCodes,),
     (ConvertPrePostCoordinatedConcepts,),
+    (CleanSmokingPpi,),
```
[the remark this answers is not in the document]
brendagutman: These concepts are not post-coordinated at this time, but they are on the list to be updated by Odysseus because they have incorrect information in Athena. These might be made to use post-coordination at that time. I believe there is something more going on with these concepts than the comments on the CR suggest and that the mapping we have defined here should not be changed with post-coordination CR if the concepts are eventually updated.

data_steward/cdr_cleaner/cleaning_rules/clean_smoking_ppi.py
```diff
@@ -0,0 +1,362 @@
...
+    "description": ""
+}, {
+    "type": "integer",
+    "name": "new_value_source_concept_id",
+    "mode": "nullable",
+    "description": ""
+}]
+
+SANDBOX_CREATE_QUERY = JINJA_ENV.from_string("""
+CREATE TABLE `{project_id}.{sandbox_dataset_id}.{new_smoking_rows}`
```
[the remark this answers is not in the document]
brendagutman: You are correct! Expect these changes.

data_steward/cdr_cleaner/clean_cdr.py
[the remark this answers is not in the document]
lrwb-aou: This info sounds like something we should note in Jira, in the class, and potentially here as well.  This doesn't need a long comment, but maybe something hinting that Post Coordination may have unanticipated impacts and we may need to revisit this rule once that is complete.

data_steward/cdr_cleaner/cleaning_rules/clean_smoking_ppi.py
```diff
@@ -0,0 +1,362 @@
...
+            try:
+                load_job = client.load_table_from_file(
+                    csv_file, table_id, job_config=job_config,
+                    job_id=job_id)  # Make an API request.
+
+                load_job.result()  # Waits for the job to complete.
+            except (ValueError, TypeError) as exc:
+                LOGGER.info(
+                    'Something went wrong and the table did not load correctly')
+                raise exc
```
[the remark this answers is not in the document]
brendagutman: Haha thanks! I got it from you!

tests/integration_tests/data_steward/cdr_cleaner/cleaning_rules/clean_smoking_test.py
```diff
@@ -0,0 +1,113 @@
...
+from tests.integration_tests.data_steward.cdr_cleaner.cleaning_rules.bigquery_tests_base import BaseTest
+
+LOAD_QUERY = JINJA_ENV.from_string("""
+    INSERT INTO `{{project_id}}.{{dataset_id}}.observation` (
+        observation_id,
+        person_id,
+        observation_concept_id, 
+        observation_source_concept_id,
+        value_as_concept_id,
+        value_source_concept_id
```
lrwb-aou, line 22: You're going to have to load some test data for `observation_date` and `observation_type_concept_id`.  These fields are required in the OMOP schema and the test table schemas are built from our json files.  You can fill them with dummy data, they just have to be in the loaded data.  

You are not required to test their cleaned versions.  Lines 96 - 110 do not have to change.

tests/integration_tests/data_steward/cdr_cleaner/cleaning_rules/clean_smoking_test.py
```diff
@@ -0,0 +1,111 @@
...
+        cls.rule_instance = CleanSmokingPpi(cls.project_id, cls.dataset_id,
+                                            cls.sandbox_id)
+
+        cls.fq_table_names = [
+            f'{cls.project_id}.{cls.dataset_id}.{OBSERVATION}'
+        ]
+
+        cls.fq_sandbox_table_names = [
+            f'{cls.project_id}.{cls.sandbox_id}.{cls.rule_instance.sandbox_table_for(OBSERVATION)}',
+            f'{cls.project_id}.{cls.sandbox_id}.{cls.lookup_table}'
```
lrwb-aou, line 64: This needs to be set up to delete the `NEW_SMOKING_ROWS` table in the sandbox also.

tests/integration_tests/data_steward/cdr_cleaner/cleaning_rules/clean_smoking_test.py
```diff
@@ -0,0 +1,111 @@
...
+
+        cls.rule_instance = CleanSmokingPpi(cls.project_id, cls.dataset_id,
+                                            cls.sandbox_id)
+
+        cls.fq_table_names = [
+            f'{cls.project_id}.{cls.dataset_id}.{OBSERVATION}'
+        ]
+
+        cls.fq_sandbox_table_names = [
+            f'{cls.project_id}.{cls.sandbox_id}.{cls.rule_instance.sandbox_table_for(OBSERVATION)}',
```
lrwb-aou, line 63: This table doesn't look like it is even created by the cleaning rule.  This line can be removed.

tests/integration_tests/data_steward/cdr_cleaner/cleaning_rules/clean_smoking_test.py
```diff
@@ -0,0 +1,111 @@
...
+    def test_clean_smoking_ppi(self):
+        """
+        Tests that the specifications for the queries perform as designed.
+        """
+
+        tables_and_counts = [{
+            'fq_table_name':
+                f'{self.project_id}.{self.dataset_id}.{OBSERVATION}',
+            'fq_sandbox_table_name':
+                self.fq_sandbox_table_names[0],
```
lrwb-aou, line 95: As of now, the rule is not sandboxing the original data.  It is instead creating a table with the corrected data in the sandbox.  You can either:
1. Point this test line at the `NEW_SMOKING_ROWS` table.
2. OR Start sandboxing the data before it is deleted.  Then point the test line at the sandboxed original data.
3. OR Do both 1 and 2.
The first item is the smallest change to the existing code.  It has no impacts on the generated SQL. The implementation of the second item will introduce a little more SQL and change the test code.  The third implementation is the most exhaustive, but may be harder to implement the test.  Implementing the first item is the fastest and most important fix.  The other two options are "nice to haves".

data_steward/cdr_cleaner/cleaning_rules/clean_smoking_ppi.py
```diff
@@ -24,56 +33,8 @@
...
-}, {
-    "type": "integer",
-    "name": "new_value_source_concept_id",
-    "mode": "nullable",
-    "description": ""
-}]
-
 SANDBOX_CREATE_QUERY = JINJA_ENV.from_string("""
-CREATE TABLE `{project_id}.{sandbox_dataset_id}.{new_smoking_rows}`
+CREATE TABLE `{{project_id}}.{{sandbox_dataset_id}}.{{new_smoking_rows}}`
```
lrwb-aou, line 38: Any reason to prefer `CREATE` over `CREATE OR REPLACE`?

lrwb-aou (changes requested):
There are some minor issues with the integration test.  I've left comments there.  I did get it running locally with the following patch.
[smoking_patch.txt](https://github.com/all-of-us/curation/files/9600535/smoking_patch.txt)

You need to be able to run the integration test twice without a failure or code change to help ensure you are cleaning up the environment behind yourself.
 
```shell
TEST_PATTERN=clean_smoking_* RDR_DATASET_ID=<your_identifier> ./tests/run_tests.sh integration
```

tests/integration_tests/data_steward/cdr_cleaner/cleaning_rules/clean_smoking_test.py
brendagutman: At this point, I'll stick with the first method so as to not confuse myself more.

data_steward/cdr_cleaner/clean_cdr.py
brendagutman: Sorry, I thought I had. It will be fixed with the next commit.

lrwb-aou (approved):
I think this looks good.  Thanks for working with my feedback.

PR #1402, opened 2022-09-15
[DC-2721] Update `validation/ehr_union` to ignore `survey_conduct` table

ksdkalluri:


hiro-mishima (approved):
Thanks, @ksdkalluri! This is good to go.

PR #1404, opened 2022-09-20
[DC-2711] Suppress Monkeypox Data from Registered Tier Only for One-Year

hiro-mishima:
- I ended up building from `BaseCleaningRule` instead of extending `AbstractBqLookupTableConceptSuppression` since the one-year holdout condition would make the script more complicated with `AbstractBqLookupTableConceptSuppression`.


jp3477 (approved):
This looks great, approved!

PR #1406, opened 2022-09-23
[DC-838] Update Gender/Sex Mismatch CR to New Base Class & Rules Dir

lledezma:


data_steward/cdr_cleaner/clean_cdr.py
```diff
@@ -263,6 +264,7 @@
     (
         GeneralizeStateByPopulation,),
     (GeneralizeCopeInsuranceAnswers,),
+    (GeneralizeSexGenderConcepts,),
```
lrwb-aou, line 267: I'm not sure we need to include this in the Clean cdr yet.  This is currently done by the the legacy Registered Tier code and we don't want to try to do this twice.  i would suggest commenting this out for now.

data_steward/cdr_cleaner/cleaning_rules/generalize_sex_gender_concepts.py
```diff
@@ -1,103 +1,206 @@
...
-    query = dict()
-    query[cdr_consts.
-          QUERY] = parse_query_for_updating_man_to_generalized_concept_id(
-              project_id, dataset_id)
-    query[cdr_consts.BATCH] = True
-    queries.append(query)
-
-    return queries
+SANDBOX_CONCEPT_ID_QUERY_TEMPLATE = JINJA_ENV.from_string("""
+INSERT INTO
```
lrwb-aou, line 39: This will need to be a `CREATE OR REPLACE TABLE` statement.  The `INSERT INTO` statement assumes a table already exists.  Most cleaning rules assume the table does not exist when the rule begins.

lrwb-aou (changes requested):
This is good.  We do need to make a few changes for the upcoming CDR.

molly-giddings (commented):
@lledezma - I agree with @lrwb-aou that it looks good.  If you address the few things she pointed out, I'm happy to approve.

data_steward/cdr_cleaner/cleaning_rules/generalize_sex_gender_concepts.py
lledezma: Since the query will be called twice, I decided to create the sandbox table in the `setup_rule` method to not overwrite any sandbox records from the first run.

lrwb-aou (changes requested):

data_steward/cdr_cleaner/cleaning_rules/generalize_sex_gender_concepts.py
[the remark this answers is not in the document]
lrwb-aou: not necessary since the `setup_rule` method will create the table.

lrwb-aou, line 39: One of the curation "unspoken" conventions is we usually use `CREATE OR REPLACE` table statements.  You don't need to do that because you are creating the table as part of the rule setup.  But, since this is a little different than normal curation processes, I would suggest including a jinja comment here.  Something like
```python
-- Using INSERT query because table creation is handled by rule_setup function --
```
Jinja comments do require the opening and closing `--`.

lrwb-aou (approved):
I would appreciate the comment in the SQL.

Also, once merged, I would rebase your integration test PR on the develop branch so you can just pull in the changes for the integration test.

PR #1407, opened 2022-09-23
[DC-2729] Added functionality to exclude `survey_conduct` table primary key id to increment at combined stage

ksdkalluri:


data_steward/constants/tools/create_combined_backup_dataset.py
```diff
@@ -118,6 +118,21 @@
 {% endif %}
 """)
 
+SURVEY_CONDUCT_MAPPING_QUERY = JINJA_ENV.from_string("""
```
lrwb-aou, line 121: You can do this with this separate string.  I achieved something similar by modifying the `MAPPING_QUERY` with an `if` to ignore the constant when working on the `survey_conduct` table.
```python
{% if domain_table != 'survey_conduct' %}
 {{domain_table}}_id + {{mapping_constant}}  AS {{domain_table}}_id,
{% else %}
{{domain_table}}_id AS {{domain_table}}_id,
{% endif %}
```

I had also thought about adding a constant of zero for the `survey_conduct` table when rendering this query instead of using the jinja template.

Which do you think is easier to maintain?

lrwb-aou (commented):
I left a comment.  I can think of three ways (including your implementation), to alter the process and achieve the same results.  I would like to pick the most maintainable way.  Which way do you want to go?

lrwb-aou (approved):
Looks good.

data_steward/constants/tools/create_combined_backup_dataset.py
ksdkalluri: I think, using the if here is easier to maintain than adding 0 to the survey_conduct_id.

PR #1412, opened 2022-09-27
[DC-2731] Fix `truncate_fitbit_data` cleaning rule to include sleep tables in list of tables to be cleaned.

ksdkalluri:


MichaelSchmidt82 (approved):
A simple constant.

PR #1414, opened 2022-09-27
[DC-2734] Fix main() function call in `ehr_union.py` to include `hpo_id_ex` list

ksdkalluri:


MichaelSchmidt82 (approved):
Just adding a parameter.  Thanks for updating the docstring

PR #1416, opened 2022-09-28
[DC-817] [DC-1212] Update scripts

hiro-mishima:


data_steward/cdr_cleaner/clean_cdr.py
```diff
@@ -220,11 +215,7 @@
     (RemoveEhrDataWithoutConsent,),
     (DedupMeasurementValueAsConceptId,),
     (drug_refills_supply.get_days_supply_refills_queries,),
-    # trying to load a table while creating query strings,
-    # won't work with mocked strings.  should use base class
-    # setup_query_execution function to load dependencies before query execution
-    (
-        populate_routes.get_route_mapping_queries,),
+    (PopulateRouteIds,),
```
molly-giddings, line 218: Is `COMBINED_CLEANING_CLASSES` the right location for this CR?  The ticket says `Needs to be moved to between base and clean when added to the pipeline` and it should be removed from both UNIONED_EHR and COMBINED.

molly-giddings (changes requested):
Hi, @hiro-mishima!  Overall, this looks really good.  I have one minor suggestion and a clarifying question.  There is also a merge conflict with clean_cdr.py that needs to be resolved.  Thanks!

data_steward/cdr_cleaner/clean_cdr.py
hiro-mishima: I believe `COMBINED_CLEANING_CLASSES` is the right location for this cleaning rule. The ticket was written with old terminology, and I think what it meant is we need to run this CR only once in the whole pipeline, and it needs to run where it can clean all the incoming data (=combined).

@molly-giddings @lrwb-aou Please let me know if my understanding is correct or not.

lrwb-aou: @hiro-mishima I think you are right.  This rule is meant to be run once instead of twice.  Adding it to the combined cleaning classes lists makes sense.

molly-giddings: That works for me - thanks, @hiro-mishima and @lrwb-aou!

molly-giddings (approved):

PR #1418, opened 2022-09-29
[DC-2737] Update truncate fitbit cleaning rule test to accommodate new list of sandbox tables

hiro-mishima:
- [DC-2731](https://github.com/all-of-us/curation/pull/1412/files) updated the values in `FITBIT_DATE_TABLES`. This change added a few new sandbox tables to the cleaning rule.
- The integration test failed because it used positional arguments for specifying `fq_sandbox_table_name`. The newly added sandbox tables changed the positions in the list.
- This PR updates the logic so that the test uses the correct `fq_sandbox_table_name` regardless of the position of the sandbox table in the list.


ksdkalluri (approved):
Thanks for the changes @hiro-mishima!

PR #1421, opened 2022-10-03
[DC-849]  Update Height/Weight/BMI CR to New Base Class & Correct Pipeline Step

jp3477:


jp3477:
Replicated in #1426 with shorter branch name.

PR #1423, opened 2022-10-04
[DC-845] Update CPT Modifiers CR to New Base Class & Correct Pipeline Step

lledezma:


data_steward/cdr_cleaner/clean_cdr.py
```diff
@@ -133,6 +132,7 @@
 
 EHR_CLEANING_CLASSES = [
     (CleanMappingExtTables,),  # should be one of the last cleaning rules run
+    (RemoveInvalidProcedureSourceRecords,),
```
hiro-mishima, line 135: Is this the right place to put this cleaning rule? The ticket mentions `it needs to run at the clean step,` and I think `the clean step` with the old terminology can mean the combined stage. Let me know if you have already had some conversations about it.

data_steward/cdr_cleaner/cleaning_rules/remove_invalid_procedure_source_records.py
```diff
@@ -46,61 +47,113 @@
...
+WHERE EXISTS 
+(
+  SELECT 
+    p1.procedure_occurrence_id 
+  FROM 
+    `{{project}}.{{sandbox_dataset}}.{{sandbox_table}}` AS p1
+  WHERE 
+    p0.procedure_occurrence_id = p1.procedure_occurrence_id
+)
+""")
```
hiro-mishima, line 69: This query's WHERE clause can be simplified as something like this:

```
WHERE procedure_occurrence_id IN (
  SELECT DISTINCT procedure_occurrence_id
  FROM `{{project}}.{{sandbox_dataset}}.{{sandbox_table}}`
)
```

This way we also do not need to assign the alias `p0` and `p1` to each of the tables.

hiro-mishima (changes requested):
Thanks for your great work, @lledezma!
It looks good overall. I have a question and a suggestion. Please let me know if you have any questions. Thanks again :)

data_steward/cdr_cleaner/clean_cdr.py
lledezma: @lrwb-aou Should this be in the unioned cleaning or combined cleaning?

lrwb-aou: @lledezma and @hiro-mishima After reviewing the comments in DC-583, I think this rule is expected to run at `UNIONED_EHR_CLEANING_CLASSES`.  

The cleaning engine was a completely different piece of code when this rule was originally written and being updated when the ticket to update the rule was written.

PR #1425, opened 2022-10-06
[DC-814] [DC-1173] Update Concept Table CR to New Base Class

hiro-mishima:
- The cleaning rule is updated to use the base class method.
- `_logging_domain_alignment` is renamed to `lookup_domain_alignment`
- All `SELECT` statements are re-written to `CREATE TABLE`/`INSERT`/`DELETE` depending on the purpose of the statement
  - As a result, all the SQL statements got significant refactoring. 
  - `_logging_domain_alignment` contained the IDs not affected by this CR. `lookup_domain_alignment` only includes the moved or dropped IDs. It makes the rest of the process in this CR more straightforward than it used to be.
- The integration test became more detailed to handle the complexity of this CR.


data_steward/cdr_cleaner/cleaning_rules/domain_mapping.py
```diff
@@ -52,8 +52,8 @@ def get_rerouting_criteria(src_table, dest_table):
     """
     for t in table_mappings_csv:
         if t[SRC_TABLE] == src_table and t[DEST_TABLE] == dest_table:
-            return re.sub('^\\s+$', EMPTY_STRING, t[REROUTING_CRITERIA])
-    return EMPTY_STRING
+            return re.sub('^\\s+$', '', t[REROUTING_CRITERIA])
+    return ''
```
MichaelSchmidt82, line 56: I like this change.

MichaelSchmidt82 (approved):
Good work!

MichaelSchmidt82 (approved):
Good job porting this cleaning rule to a solid base-class CR.
Good documentation too

hiro-mishima:
Thanks for reviewing, @MichaelSchmidt82!

@lrwb-aou There is one thing I want to double-check with you before merging this branch to `develop`.
I kept the logic as-is that adds `1970-01-01` to the newly mapped rows with mandatory date/datetime columns when the source rows have a NULL value for that date/datetime column.
https://github.com/all-of-us/curation/blob/hm/dc814/data_steward/cdr_cleaner/cleaning_rules/domain_alignment.py#L146
This inevitably causes the inconsistency between the date column and the datetime column. I kept the logic as-is because I did not want to add extra complexity to this CR. But I am not sure adding a wrong date value/ datetime value as default to the mandatory columns is a good idea. Should I update the logic for this part? I would love your thoughts on this. Thanks :)

lrwb-aou:
> @lrwb-aou There is one thing I want to double-check with you before merging this branch to `develop`. I kept the logic as-is that adds `1970-01-01` to the newly mapped rows with mandatory date/datetime columns when the source rows have a NULL value for that date/datetime column. https://github.com/all-of-us/curation/blob/hm/dc814/data_steward/cdr_cleaner/cleaning_rules/domain_alignment.py#L146 This inevitably causes the inconsistency between the date column and the datetime column. I kept the logic as-is because I did not want to add extra complexity to this CR. But I am not sure adding a wrong date value/ datetime value as default to the mandatory columns is a good idea. Should I update the logic for this part? I would love your thoughts on this. Thanks :)
> 
> For example, this test case shows that `procedure_date` is date`2015-07-15` but timestamp `1970-01-01` is assigned to `procedure_datetime ` after this CR because the source record in `observation` does not have the timestamp column populated. https://github.com/all-of-us/curation/blob/hm/dc814/tests/integration_tests/data_steward/cdr_cleaner/cleaning_rules/domain_alignment_test.py#L371

@hiro-mishima I think it is ok to keep the logic "as is" for the purposes of this update.  Yes, this will create the problem your example illustrates.  A little later on in the combined cleaning rule list we have `TemporalConsistency` followed by `EnsureDateDatetimeConsistency`.  I believe `EnsureDateDatetimeConsistency` will fix the issue you are concerned about.  This is an example of the SQL that class builds:  

```sql
-- for a table with start and end date and datetime fields
SELECT drug_exposure_id, person_id, drug_concept_id, drug_exposure_start_date, 
CASE
WHEN drug_exposure_start_datetime IS NULL
THEN CAST(DATETIME(drug_exposure_start_date, TIME(00,00,00)) AS TIMESTAMP)
WHEN EXTRACT(DATE FROM drug_exposure_start_datetime) = drug_exposure_start_date
THEN drug_exposure_start_datetime
-- this else will catch the inconsistency for the start date
ELSE CAST(DATETIME(drug_exposure_start_date, EXTRACT(TIME FROM drug_exposure_start_datetime)) AS TIMESTAMP)
END AS drug_exposure_start_datetime, drug_exposure_end_date, 
CASE
WHEN drug_exposure_end_datetime IS NULL
THEN CAST(DATETIME(drug_exposure_end_date, TIME(11,59,59)) AS TIMESTAMP)
WHEN EXTRACT(DATE FROM drug_exposure_end_datetime) = drug_exposure_end_date
THEN drug_exposure_end_datetime
-- this else will catch the inconsistency for the end date
ELSE CAST(DATETIME(drug_exposure_end_date, EXTRACT(TIME FROM drug_exposure_end_datetime)) AS TIMESTAMP)
END AS drug_exposure_end_datetime, verbatim_end_date, drug_type_concept_id, stop_reason, refills, quantity, days_supply, sig, route_concept_id, lot_number, provider_id, visit_occurrence_id, visit_detail_id, drug_source_value, drug_source_concept_id, route_source_value, dose_unit_source_value
FROM `tester.foo.drug_exposure`

-- for a table with only date and datetime fields
SELECT procedure_occurrence_id, person_id, procedure_concept_id, procedure_date, 
CASE
WHEN procedure_datetime IS NULL
THEN CAST(DATETIME(procedure_date, TIME(00,00,00)) AS TIMESTAMP)
WHEN EXTRACT(DATE FROM procedure_datetime) = procedure_date
THEN procedure_datetime
-- this ELSE would catch the date inconsistency
ELSE CAST(DATETIME(procedure_date, EXTRACT(TIME FROM procedure_datetime)) AS TIMESTAMP)
END AS procedure_datetime, procedure_type_concept_id, modifier_concept_id, quantity, provider_id, visit_occurrence_id, visit_detail_id, procedure_source_value, procedure_source_concept_id, modifier_source_value
FROM `tester.foo.procedure_occurrence`
```

hiro-mishima:
Thanks for the confirmation, @lrwb-aou! Great, I will keep this logic as it is. I will close this PR after making an update for this comment -> https://github.com/all-of-us/curation/pull/1425#discussion_r1011972548

PR #1430, opened 2022-10-14
[DC-2727] Bug fix preceding 0s in update_invalid_zip_codes CR

brendagutman:


data_steward/cdr_cleaner/cleaning_rules/update_invalid_zip_codes.py
```diff
@@ -78,27 +79,30 @@ class UpdateInvalidZipCodes(BaseCleaningRule):
     Any invalid zip code will be sandboxed and updated.
     """
 
-    def __init__(self, project_id, dataset_id, sandbox_dataset_id):
+    def __init__(self,
+                 project_id,
+                 dataset_id,
+                 sandbox_dataset_id,
+                 table_namer=None):
```
hiro-mishima, line 86: Thanks for adding `table_namer`!

tests/integration_tests/data_steward/cdr_cleaner/cleaning_rules/update_invalid_zip_codes_test.py
```diff
@@ -112,31 +107,35 @@ def test_update_invalid_zip_codes(self):
...
                 -- invalid zip codes, excess whitespace, will be sandboxed and updated in observation table --
                 -- once cleaned will be valid --
                 (14, 14, 0, date('2019-03-03'), 0, 1585250, null, '     34567', null, null), 
                 -- once cleaned will be invalid --
-                (15, 15, 0, date('2019-03-03'), 0, 1585250, null, '56789     ', null, null) 
+                (15, 15, 0, date('2019-03-03'), 0, 1585250, null, '56789     ', null, null), 
+
+                -- valid zip codes with preceding 0s will not be affected --
+                (16, 16, 0, date('2019-03-03'), 0, 1585250, null, '02035', null, null),
+                (17, 17, 0, date('2019-03-03'), 0, 1585250, null, '00601', null, null)
```
hiro-mishima, line 138: I understand that `12345` is a valid zip code, and `12345-6789` is also valid. Is `123456789`(zip code without hyphen) valid or invalid? I saw 9-digit codes in the discussion in the comments from `DC-2739`(the investigation ticket). I think it is worth adding that test case to this test to clarify the expected behavior.

data_steward/cdr_cleaner/cleaning_rules/update_invalid_zip_codes.py
```diff
@@ -12,6 +10,7 @@
         value_as_string and value_source_value = 'Response removed due to invalid value'
         value_as_number = 0
         value_source_concept_id = 2000000010
+zip3_lookup is a view that uses "Census Bureau US Boundaries" public dataset available in BigQuery.
```
hiro-mishima, line 13: I found this Lauren's comment in `DC-2739`(the investigation ticket). Should we update the logic as suggested? Let me know if I am missing any discussions in the background.

```
- #####
- #########
- #####-####

~ the work for DC-2727 be changed to ignore that zip code table entirely and focus on meeting a regular expression criteria.
```

hiro-mishima (commented):
@brendagutman Thanks for your great work! I left a couple of comments. Let me know what you think! Thanks :)

data_steward/cdr_cleaner/cleaning_rules/update_invalid_zip_codes.py
brendagutman: I left a comment in DC-2739 on this subject. I'll start work on this again when I have a reply. I believe the subject is being discussed.

brendagutman: The question of whether to use a zip code lookup table or otherwise will be discussed in DC-2802. For this CR we will continue using zip3_lookup.

hiro-mishima (approved):
Thanks for the updates and also for creating the followup JIRA ticket. It looks great :tada:

PR #1437, opened 2022-10-25
[DC-2701] Write a prod stored SQL script that identifies pre-deid person_id values in the non-prod environments

hiro-mishima:


jp3477 (approved):
Very cool dynamic SQL. I ran the first part of your scheduled query to create the temp tables, and it ran well.

hiro-mishima:
Thanks for the review, @jp3477! I added the details of how I set up the scheduled SQL in the comment section of the JIRA DC-2701. Hope it helps when you work on DC-2702 :)

PR #1438, opened 2022-10-26
[DC-2755] Check that mapping, extension, and OMOP tables contain the same number of records

jp3477:


data_steward/analytics/cdr_ops/combined.py
```diff
@@ -285,6 +290,197 @@
 '''
 execute(client, query)
 
+# ## OMOP, Mapping, and Extension Tables Contain Same Record Count
```
brendagutman, line 293: Perhaps add instructions for the check. Mainly mentioning that the results for both conditions are consolidated  in the output of the 'results' cell.

If it is possible at this time, add actions for the types of failure. Perhaps giving hints to the teams or process' that are responsible for creating these tables.

brendagutman (commented):
I learned a lot while working on this review. I really like how you went about creating this check!

data_steward/analytics/cdr_ops/combined.py
```diff
@@ -285,6 +290,197 @@
...
+            CASE
+              WHEN ext.{mapped_table}_id IS NOT NULL THEN 'BC'
+              ELSE 'B'
+            END
+          ELSE 'C'
+        END cat
+      FROM `aou-res-curation-prod.jp_test_dataset.{mapped_table}` m
+      FULL JOIN `aou-res-curation-prod.jp_test_dataset._mapping_{mapped_table}` mm
+        ON mm.{mapped_table}_id = m.{mapped_table}_id
+      FULL JOIN `aou-res-curation-prod.jp_test_dataset.{mapped_table}_ext` ext
```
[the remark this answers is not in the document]
jp3477: Thanks for catching this.

data_steward/analytics/cdr_ops/combined.py
jp3477: Nice suggestions, I added some instructions. As for actions to take for different failure types, I'm not sure which processes/cleaning rules would cause the most issues.

@lrwb-aou Are there any specific cleaning rules are or other processes that would cause these failures to occur? If so, we can point notebook users in those directions.

jp3477:
@brendagutman I rebased on develop, the tests pass now.

PR #1439, opened 2022-10-28
Same code and QCs, but updated format

kouamea:
Updated the format a bit - added PASS if QC passes. I also hard coded rt_dataset as this parameter does not change

### DIFF
--- a/data_steward/analytics/cdr_ops/qc_serology_ct_dataset.py
+++ b/data_steward/analytics/cdr_ops/qc_serology_ct_dataset.py
@@ -13,9 +13,8 @@
 #     name: python3
 # ---
 # + tags=["parameters"]
-rt_dataset = ""
 ct_ser_dataset = ""
-new_ct_dataset = ""
+new_ser_ct_dataset = ""
 cur_project = ""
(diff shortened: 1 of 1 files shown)


kouamea:
@ksdkalluri when you have time, please approve this. I just updated the format but it is the same code.

data_steward/analytics/cdr_ops/qc_serology_ct_dataset.py
```diff
@@ -38,15 +37,17 @@
...
+    q = query.render(cols = cols, project_id=project, dataset=dataset)
+    df =execute(client, q)
     return df
 
 
-def get_data_input(rt_dataset, ct_ser_dataset, new_ct_dataset, cur_out_project,
-                   cur_project):
+def get_data_input(ct_ser_dataset, new_ser_ct_dataset, cur_out_project, cur_project):
+    
+    rt_dataset='R2020q4r1_antibody_quest'
```
ksdkalluri, line 50: @kouamea although the dataset name doesn’t change we do not want our prod dataset names to be publicly available. can you revert this back to a parameter.

PR #1440, opened 2022-10-28
[DC-2674] Fix Serology QC notebook 

MichaelSchmidt82:
- Modified the get_table function query to correctly template variables.
- Functions require the client parameter (DI).
- Removed unused `pandas` import
- Some lines contained trailing white space.  I confirmed the removal has no side effect.


hiro-mishima (approved):
Thanks, @MichaelSchmidt82 it looks good!
I ran this notebook for the non-AIAN dataset, and it worked fine without errors.

PR #1442, opened 2022-11-02
[DC-2670] Fix fill source value text fields CR to ignore `validated_survey_source_value`

ksdkalluri:


MichaelSchmidt82 (approved):
A simple check.

PR #1444, opened 2022-11-02
[DC-2702] Update/Write a prod stored SQL script to identify post-deid person_id values in the non-prod curation environments

jp3477:


hiro-mishima (approved):
Thanks, @jp3477! It looks great :)

Would you mind adding some comments in the docstrings that this SQL checks both `person_id` and `research_id`?
https://github.com/all-of-us/curation/blob/jp/check_research_ids/data_steward/resource_files/scheduled_queries/monitor_nonprod_person_id_violation.txt#L2

Once updating the comments, please merge this PR and update the job in our prod environment. 

Thanks again! 🚀

PR #1446, opened 2022-11-03
[DC-2754] Update the combined notebook to check for null or orphaned `survey_conduct_id`s

hiro-mishima:
- Moved `render_message()` from `rdr_export_qc.py` to `notebook_utils.py` as it's useful for other QC notebooks.


jp3477 (approved):
Looks great! Good idea moving render_message to a more accessible location.

PR #1448, opened 2022-11-08
[DC-2806] Create a retraction script that covers destination dataset creation, sandboxing, etc.

hiro-mishima:


data_steward/retraction/run_retraction.py
```diff
@@ -0,0 +1,370 @@
...
+    lookup_table_name = regex_created_table.group(2)
+
+    LOGGER.info(
+        f"Created the lookup table {lookup_table_name} in the sandbox dataset.")
+    return lookup_table_name
+
+
+def parse_args(raw_args=None):
+    parser = ArgumentParser(description='abc')
+
```
lledezma, line 198: Is this the actual description?

PR #1450, opened 2022-11-16
[DC-2841] Create a validation notebook to check retraction status

ksdkalluri:


hiro-mishima (approved):
This looks good, and I confirmed that it works as expected. Thanks for your work, @ksdkalluri!

PR #1451, opened 2022-11-19
[DC-2775] Create cleaning rule to drop data related to unverified surveys

brendagutman:


jp3477 (approved):
Thanks for the clarifications, it's good to go.

PR #1452, opened 2022-11-21
[DC-2786] Make sure *_era tables are empty in Ct_deid dataset

lledezma:


data_steward/analytics/cdr_ops/controlled_tier_qc/check_controlled_tier_part2.py
```diff
@@ -1157,6 +1157,67 @@ def my_sql(table_name, column_name):
...
+# +
+era_tables = ['dose_era', 'drug_era', 'condition_era']
+
+
+def my_query(table_era):
+
+    query = JINJA_ENV.from_string("""
+      WITH df1 AS (
+        SELECT 
+          person_id
```
ksdkalluri, line 1171: @lledezma suggestion, instead of using person_id in the query, you can use {{table_era}}_id. Since we have a primary_key for these tables it is better to use those keys instead of person_id.

ksdkalluri (approved):
@lledezma This is good to merge. I added a small suggestion for the query in the validation notebook, if you choose to use it make the change and wait for the checks to pass and merge it.

PR #1456, opened 2022-12-05
[DC-2889] Adds QC notebook for EHR only retraction validation

ksdkalluri:


data_steward/analytics/cdr_ops/ehr_retraction_validation.py
```diff
@@ -0,0 +1,216 @@
...
+row_counts_union_all_query = '\nUNION ALL\n'.join(row_counts_queries_list)
+
+retraction_table_count_query = (f'{rids_query}\n{row_counts_union_all_query}'
+                                if is_deidentified == 'true' else
+                                f'{pids_query}\n{row_counts_union_all_query}')
+execute(client, retraction_table_count_query)
+# -
+
+# ## 3. Verify Death table retraction.
+
```
lrwb-aou, line 170: It would be good to explain why this table is checked separately from other tables with `person_id`.  I think it is because of the mapping table issue.  I think we will be revisiting this in the future.

data_steward/analytics/cdr_ops/ehr_retraction_validation.py
```diff
@@ -0,0 +1,216 @@
...
+                                       table_name=table)
+
+retraction_status_query = (f'{rids_query}\n{death_query}'
+                           if is_deidentified == 'true' else
+                           f'{pids_query}\n{death_query}')
+execute(client, retraction_status_query)
+# -
+
+# ## 4. Verify if mapping/ext tables are cleaned after retraction
+
```
lrwb-aou, line 195: It is good to explain what this check is actually doing and the problem it indicates if a 'PROBLEM' is found.  Assume the person running this has never run this notebook before.

data_steward/analytics/cdr_ops/ehr_retraction_validation.py
```diff
@@ -0,0 +1,216 @@
...
+# +
+mapping_ext_check_query = JINJA_ENV.from_string('''
+select \'{{mapping_table}}\' as table_name,
```
lrwb-aou, line 198: Can this check be made to resemble the other checks with a status column?  You could just add a case statement to this query.

lrwb-aou (commented):
I think the notebook is functionally sound.  If we need to commit it for use, we will.  But there are multiple conventions I would like to make.  If we commit this PR as is, we need a new issue created to come back and clean it up for conventionality.

data_steward/analytics/cdr_ops/retraction_validation.py
```diff
@@ -97,19 +97,19 @@
 table_row_counts_query = JINJA_ENV.from_string('''
 SELECT
 \'{{table_name}}\' as table_name,
-q4r3_minus_aian_row_count,
-q4r4_row_count,
-case when q4r3_minus_aian_row_count = q4r4_row_count then 'OK'
-  when q4r3_minus_aian_row_count is null AND q4r4_row_count is null then 'OK'
-  when q4r3_minus_aian_row_count is NOT NULL and q4r4_row_count is null then 'PROBLEM'
-  when q4r3_minus_aian_row_count is NULL and q4r4_row_count is not NULL then 'PROBLEM'
+newr3_minus_aian_row_count,
```
lrwb-aou, line 100: would this be better as `old_minus_aian_row_count` so we don't have `r3` and `r4` still mixed into the naming scheme?

data_steward/analytics/cdr_ops/retraction_validation.py
```diff
@@ -144,10 +144,8 @@
 RIGHT JOIN pids p2
 on fr.fact_id_2 = p2.person_id
 where domain_concept_id_1 = 56
-''').render(project=project_id, dataset=q4_dataset)
-retraction_status_query = (f'{rids_query}\n{fact_relationship_retraction}'
-                           if is_deidentified == 'true' else
-                           f'{pids_query}\n{fact_relationship_retraction}')
+''').render(project=project_id, dataset=new_dataset)
+retraction_status_query = f'{pids_query}\n{fact_relationship_retraction}'
```
lrwb-aou, line 148: so, we don't need the `is_deidentified` check in the `retraction_status_query` anymore?

lrwb-aou (commented):
My only functional concern with the new commit is the removal of the de-identified check for the fact_relationship table.  We did have a ticket out there that Jason has worked on to try to get proper de-identification of this table working so we don't have to drop all that data anymore.

Can we leave the is_deidentified check in for it?

data_steward/analytics/cdr_ops/retraction_validation.py
```diff
@@ -37,11 +37,25 @@
...
+    """
+    Returns a mapping table for a domain table.
+    
+    :param table: identifies domain table name
+    :param is_deidentified: identifies if a the dataset is de-identified choose b/e true/false.
+    """
+    if is_deidentified.lower() == 'false':
+        return f'_mapping_{table}'
+    else:
+        return f'{table}_ext'
```
lrwb-aou, line 56: Does it make sense to move this code snipped to the https://github.com/all-of-us/curation/blob/develop/data_steward/analytics/cdr_ops/notebook_utils.py module?  You've used it in two different files now.  It could be a useful helper function in other notebooks also.

lrwb-aou (approved):
This looks good.  I left some feedback for minor typos.

But, I did want to ask what you thought about moving that one function into a utility module and importing it instead of defining the same function in two different modules.

data_steward/analytics/cdr_ops/ehr_retraction_validation.py
```diff
@@ -0,0 +1,268 @@
...
+    """
+    if is_deidentified.lower() == 'false':
+        return f'_mapping_{table}'
+    else:
+        return f'{table}_ext'
+
+
+# ## 1. Verify ehr data for participants listed to be dropped in the lookup table are dropped from the pid_tables
+#
+# Here we are checking for no EHR data in a given dataset. We are joining the domain tables onto the provenence tables to identify if a particulary record is from EHR submission.
```
lrwb-aou, line 89: nit:
```suggestion
# Here we are checking for no EHR data in a given dataset. We are joining the domain tables onto the provenence tables to identify if a particular record is from an EHR submission.
```

data_steward/analytics/cdr_ops/ehr_retraction_validation.py
```diff
@@ -0,0 +1,268 @@
...
+
+retraction_status_query = (f'{rids_query}\n{union_all_query}'
+                           if is_deidentified == 'true' else
+                           f'{pids_query}\n{union_all_query}')
+execute(client, retraction_status_query)
+# -
+
+# ## 2. Row counts of participants in source dataset minus the ehr data is equal to the row counts for the participants in new dataset.
+#
+# We expect PPI/PM data to exist for the listed participants even after the retraction. So here we are checking the record count of the source data minus the EHR records is equal to the recvord count post retraction.
```
lrwb-aou, line 137: nit:
```suggestion
# We expect PPI/PM data to exist for the listed participants even after the retraction. So here we are checking the record count of the source data minus the EHR records is equal to the record count post retraction.
```

data_steward/analytics/cdr_ops/ehr_retraction_validation.py
```diff
@@ -0,0 +1,268 @@
...
+
+retraction_table_count_query = (f'{rids_query}\n{row_counts_union_all_query}'
+                                if is_deidentified == 'true' else
+                                f'{pids_query}\n{row_counts_union_all_query}')
+execute(client, retraction_table_count_query)
+# -
+
+# ## 3. Verify Death table retraction.
+#
+# Death table is copied as-is from EHR dataset as we do not recieve data for death from RDR export. Death table is missing the provenence table as it doesn't have a domain_id column so we will use person_id column to identify the records if any.
```
lrwb-aou, line 206: nit:
```suggestion
# Death table is copied as-is from EHR dataset as we do not receive death data via the RDR export yet. Death table is missing the provenence table as it doesn't have a domain_id column so we will use person_id column to identify any records.
```

lrwb-aou (approved):
Changes are still approved, with some minor typos.  I just wanted to review one more time.

PR #1457, opened 2022-12-07
[DC-2964] load ad hoc file from RDR bucket

lrwb-aou:
* ability to load an ad hoc file from the RDR bucket
* Will require the user to impersonate a service account with the access
* will require a json schema file
* should be able to load any file from that bucket with no (or minimal) changes needed, even if it is the first time we have seen the file

### DIFF
--- a/data_steward/tools/load_rdr_file.py
+++ b/data_steward/tools/load_rdr_file.py
@@ -72,7 +72,8 @@ def parse_args(raw_args=None) -> Namespace:
     secure bucket.
     """
     parser = ArgumentParser(
-        description='Arguments pertaining to a file in the RDR bucket')
+        description='Arguments pertaining to a file in the RDR bucket',
+        allow_abbrev=False)
 
     parser.add_argument('--bucket_filepath',
                         action='store',
--- a/data_steward/tools/load_rdr_file.py
+++ b/data_steward/tools/load_rdr_file.py
@@ -1,4 +1,4 @@
-from argparse import ArgumentParser
+from argparse import ArgumentParser, Namespace
 from datetime import datetime
 import json
 import logging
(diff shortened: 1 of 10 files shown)


data_steward/tools/load_rdr_file.py
```diff
@@ -0,0 +1,147 @@
...
+                                        validate_file_exists,
+                                        validate_email_address,
+                                        validate_bucket_filepath)
+from gcloud.bq import BigQueryClient
+from common import CDR_SCOPES
+
+LOGGER = logging.getLogger(__name__)
+
+
+def load_rdr_table(client: BigQueryClient, src_uri: str, table_id: str,
```
ksdkalluri, line 19: @lrwb-aou I understand this is to load a single file, do you think If we can expand this script to load multiple files from a directory?  If that will be too much of a scope creep we can create a new script to add that functionality.

lrwb-aou: @ksdkalluri I think we can expand it to do that, but would suggest making it a new PR (not necessarily a new script).  Right now, we try to load all the files they send us that we know about with the import rdr omop script.  This script allows our teams to use the bucket to pass data from their prod env to our prod env without downloading the data.  Hopefully, we don't have this happen too often. 🤞 

I think the directory import may be useful, but would like to do that as an additional PR (primarily because I want to enforce useful schemas with valid column descriptions).  Thoughts?

ksdkalluri: My only concern was, if RDR wants to send us multiple files in the same directory that are not related to the rdr submission white-listed files, we would have to run this script for each file in the directory which will become a manual repetitive process. Since we are not expecting such transfers from RDR anytime soon we can currently go ahead with this scope. We can expand on this to load multiple files if we expect more such transfers in future.

ksdkalluri (approved):
This is good to merge, thanks for adding all the parameter validators, it'll be helpful for us to add new ones to the validators file in the future.  All the tests were helpful in testing and understanding the code.

lrwb-aou:
Hopefully, if they are going to send us multiple files and it's not an ad hoc situation, we will have a json schema file built for that.  🤞 

If sending data that we don't have a schema for becomes the norm, we've got an SOP problem.

data_steward/tools/load_rdr_file.py
lrwb-aou: Hopefully, if they are going to send us multiple files and it's not an ad hoc situation, we will have a json schema file built for that. 🤞

If sending data that we don't have a schema for becomes the norm, we've got an SOP problem.

PR #1458, opened 2022-12-08
[DC-2704] Set up a GCP alert for the curation slack channel if any disallowed values are found

hiro-mishima:
- After this branch is merged to `develop` and deployed, I will create GCP alerting policies to catch [these three log messages](https://github.com/all-of-us/curation/pull/1458/files#diff-55244244d8407609fe3cb9c4721d9290c6531afe54aa8d0fe67558a627c13858R26).
- [Scheduled query was updated to loop with a smaller size,](data_steward/resource_files/scheduled_queries/monitor_nonprod_person_id_violation.txt) to avoid BigQuery's resource limit.

### DIFF
--- a/data_steward/admin/admin_api.py
+++ b/data_steward/admin/admin_api.py
@@ -1,10 +1,11 @@
 """
 A module for the basic `admin` endpoint.
 
-Its solely responsible for automating service key deletions and notifications.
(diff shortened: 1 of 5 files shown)


MichaelSchmidt82 (approved):
Solid work 😎

PR #1459, opened 2022-12-12
[DC-2769] Update geolocation suppression rule

brendagutman:


jp3477 (approved):
This looks good! And good catch with the ethnicity string

PR #1464, opened 2022-12-28
[DC-2839] Refactor queries to use research_ids and update test.

brendagutman:


nishanthpp93 (approved):
Looks good! Thanks for updating. Added a couple comments just as an FYI.

PR #1466, opened 2023-01-06
[DC-2788] no_data_30_days_after_death cleaning rule fails to identify Greatest when there is null in date/datetime fields

lledezma:


tests/integration_tests/data_steward/cdr_cleaner/cleaning_rules/no_data_30_days_after_death_test.py
```diff
@@ -98,22 +118,20 @@ def setUp(self):
...
+            PERSON_DATA_TEMPLATE, DEATH_DATA_TEMPLATE,
+            VISIT_OCCURRENCE_DATA_TEMPLATE, OBSERVATION_DATA_TEMPLATE,
+            CONDITION_OCCURRENCE_DATA_TEMPLATE, DEVICE_EXPOSURE_DATA_TEMPLATE
+        ]
+
+        test_queries = []
+        for template in templates:
+            test_queries.append(
+                template.render(project_id=self.project_id,
+                                dataset_id=self.dataset_id))
```
nishanthpp93, line 131: Nice use of looping!

nishanthpp93 (approved):
Looks good! I saw a warning about table namer while running the test so even though this is unrelated but very minor, I would suggest specifying the table_namer=None parameter in the cleaning rule args.

PR #1468, opened 2023-01-09
[DC-2749] Add COPESurveyVersionTask to combined cleaning classes to run at the end of all other cleaning classes.

jp3477:
Note: I also moved the currently commented `SurveyConductDateShiftRule` rule from the original data stages to the combined date stage. If this needs to remain in the original stages, request a change.

### DIFF
--- a/data_steward/cdr_cleaner/clean_cdr.py
+++ b/data_steward/cdr_cleaner/clean_cdr.py
@@ -224,10 +224,7 @@ COMBINED_CLEANING_CLASSES = [
     (GenerateExtTables,),
     (COPESurveyVersionTask,
     ),  # Should run after GenerateExtTables and before CleanMappingExtTables
-    # TODO: Uncomment rule after date-shift removed from deid module
-    # (SurveyConductDateShiftRule,),
-    (
-        PopulateSurveyConductExt,),
+    (PopulateSurveyConductExt,),
     (CleanMappingExtTables,),  # should be one of the last cleaning rules run
 ]
 
@@ -242,6 +239,8 @@ FITBIT_CLEANING_CLASSES = [
 REGISTERED_TIER_DEID_CLEANING_CLASSES = [
     # Data mappings/re-mappings
     ####################################
+    # TODO: Uncomment rule after date-shift removed from deid module
+    # (SurveyConductDateShiftRule,),
     (
         QRIDtoRID,),  # Should run before any row suppression rules
 
--- a/data_steward/cdr_cleaner/clean_cdr.py
+++ b/data_steward/cdr_cleaner/clean_cdr.py
@@ -221,6 +221,13 @@ COMBINED_CLEANING_CLASSES = [
     (RemoveNonMatchingParticipant,),
     (DropOrphanedSurveyConductIds,),
     (DropOrphanedPIDS,),
+    (GenerateExtTables,),
+    (COPESurveyVersionTask,
+    ),  # Should run after GenerateExtTables and before CleanMappingExtTables
+    # TODO: Uncomment rule after date-shift removed from deid module
+    # (SurveyConductDateShiftRule,),
+    (
+        PopulateSurveyConductExt,),
     (CleanMappingExtTables,),  # should be one of the last cleaning rules run
 ]
 
@@ -236,14 +243,7 @@ REGISTERED_TIER_DEID_CLEANING_CLASSES = [
     # Data mappings/re-mappings
     ####################################
     (
-        GenerateExtTables,),
-    (COPESurveyVersionTask,
-    ),  # Should run after GenerateExtTables and before CleanMappingExtTables
-    # TODO: Uncomment rule after date-shift removed from deid module
-    # (SurveyConductDateShiftRule,),
-    (
-        PopulateSurveyConductExt,),
-    (QRIDtoRID,),  # Should run before any row suppression rules
+        QRIDtoRID,),  # Should run before any row suppression rules
 
     # Data generalizations
     ####################################
@@ -297,10 +297,6 @@ REGISTERED_TIER_FITBIT_CLEANING_CLASSES = [
 
 CONTROLLED_TIER_DEID_CLEANING_CLASSES = [
     (RtCtPIDtoRID,),
-    (GenerateExtTables,),
-    (COPESurveyVersionTask,
-    ),  # Should run after GenerateExtTables and before CleanMappingExtTables
-    (PopulateSurveyConductExt,),
     (QRIDtoRID,),  # Should run before any row suppression rules
     (TruncateEraTables,),
     (NullPersonBirthdate,),
--- a/data_steward/cdr_cleaner/cleaning_rules/populate_survey_conduct_ext.py
+++ b/data_steward/cdr_cleaner/cleaning_rules/populate_survey_conduct_ext.py
@@ -19,7 +19,7 @@ LOGGER = logging.getLogger(__name__)
 UPDATE_SURVEY_CONDUCT_EXT_QUERY = JINJA_ENV.from_string("""
 UPDATE `{{project_id}}.{{dataset_id}}.survey_conduct_ext` AS sce
 SET language = qrai.value
-FROM `{{project_id}}.{{cope_lookup_dataset_id}}.questionnaire_response_additional_info` AS qrai
+FROM `{{project_id}}.{{clean_survey_dataset_id}}.questionnaire_response_additional_info` AS qrai
 WHERE sce.survey_conduct_id = qrai.questionnaire_response_id
 AND UPPER(qrai.type) = 'LANGUAGE'
 """)
@@ -34,7 +34,7 @@ class PopulateSurveyConductExt(BaseCleaningRule):
                  project_id,
                  dataset_id,
                  sandbox_dataset_id,
-                 cope_lookup_dataset_id=None,
+                 clean_survey_dataset_id=None,
                  table_namer=None):
         """
         Initialize the class with proper information.
@@ -57,10 +57,10 @@ class PopulateSurveyConductExt(BaseCleaningRule):
                          depends_on=[GenerateExtTables, COPESurveyVersionTask],
                          table_namer=table_namer)
 
-        if not cope_lookup_dataset_id:
-            raise RuntimeError("'cope_lookup_dataset_id' must be set")
+        if not clean_survey_dataset_id:
+            raise RuntimeError("'clean_survey_dataset_id' must be set")
 
-        self.cope_lookup_dataset_id = cope_lookup_dataset_id
+        self.clean_survey_dataset_id = clean_survey_dataset_id
 
     def get_query_specs(self):
         """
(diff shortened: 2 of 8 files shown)


data_steward/cdr_cleaner/clean_cdr.py
```diff
@@ -221,6 +221,13 @@
     (RemoveNonMatchingParticipant,),
     (DropOrphanedSurveyConductIds,),
     (DropOrphanedPIDS,),
+    (GenerateExtTables,),
+    (COPESurveyVersionTask,
+    ),  # Should run after GenerateExtTables and before CleanMappingExtTables
+    # TODO: Uncomment rule after date-shift removed from deid module
+    # (SurveyConductDateShiftRule,),
```
lrwb-aou, line 228: This doesn't need to be here.

data_steward/cdr_cleaner/clean_cdr.py
```diff
@@ -236,14 +243,7 @@
     # Data mappings/re-mappings
     ####################################
     (
-        GenerateExtTables,),
-    (COPESurveyVersionTask,
-    ),  # Should run after GenerateExtTables and before CleanMappingExtTables
-    # TODO: Uncomment rule after date-shift removed from deid module
-    # (SurveyConductDateShiftRule,),
```
lrwb-aou, line 243: This does need to be here.

data_steward/cdr_cleaner/clean_cdr.py
```diff
@@ -297,10 +297,6 @@
 
 CONTROLLED_TIER_DEID_CLEANING_CLASSES = [
     (RtCtPIDtoRID,),
-    (GenerateExtTables,),
-    (COPESurveyVersionTask,
-    ),  # Should run after GenerateExtTables and before CleanMappingExtTables
-    (PopulateSurveyConductExt,),
```
lrwb-aou, line 303: Both of these rules reference the survey data set but with different names.  It's better to use the same name so we only have to pass the value once instead of twice.  Please update this as well.

data_steward/cdr_cleaner/cleaning_rules/populate_survey_conduct_ext.py
```diff
@@ -19,7 +19,7 @@
 UPDATE_SURVEY_CONDUCT_EXT_QUERY = JINJA_ENV.from_string("""
 UPDATE `{{project_id}}.{{dataset_id}}.survey_conduct_ext` AS sce
 SET language = qrai.value
-FROM `{{project_id}}.{{additional_info_dataset}}.questionnaire_response_additional_info` AS qrai
+FROM `{{project_id}}.{{cope_lookup_dataset_id}}.questionnaire_response_additional_info` AS qrai
```
lrwb-aou, line 22: I would prefer for this to have a more meaningful name.  `cope_lookup_dataset_id` doesn't have much to do with the `survey_conduct_ext` table specifically.  A possible solution is to change the variable name to something meaningful for both `survey_conduct_ext` and `observation_ext` tables.  One suggestion is to use `clean_survey_dataset_id` since it should denote a cleaned version of the dataset in both rules.
```suggestion
FROM `{{project_id}}.{{clean_survey_dataset_id}}.questionnaire_response_additional_info` AS qrai
```
I'm ok with `raw_survey_dataset_id`, `survey_dataset_id`, or something else that is descriptive but generic and applicable to both rules.

jp3477: Sure, I'll use `clean_survey_dataset_id`

jp3477: Hmm, my last commit seems to have caused one test to fail. I can't see why at all. Can you tell?

lrwb-aou: I just saw this.  Sorry for the delay.  Yes, I think I know what is happening.

These are the tests that ran before the failing integration test.
```text
tests/integration_tests/data_steward/validation/ehr_union_test.py
tests/integration_tests/data_steward/validation/achilles_heel_test.py
tests/integration_tests/data_steward/cdr_cleaner/cleaning_rules/clean_height_weight_test.py                       ----- NOT THIS ONE
tests/integration_tests/data_steward/cdr_cleaner/cleaning_rules/replace_standard_id_in_domain_tables_test.py     ------ NOT THIS ONE
tests/integration_tests/data_steward/cdr_cleaner/cleaning_rules/deid/remove_fitbit_data_if_max_age_exceeded_test.py
tests/integration_tests/data_steward/cdr_cleaner/cleaning_rules/set_unmapped_question_answer_survey_concepts_test.py
tests/integration_tests/data_steward/cdr_cleaner/cleaning_rules/ppi_branching_test.py
tests/integration_tests/data_steward/cdr_cleaner/cleaning_rules/monkeypox_concept_suppression_test.py
tests/integration_tests/data_steward/cdr_cleaner/cleaning_rules/populate_route_ids_test.py
tests/integration_tests/data_steward/cdr_cleaner/cleaning_rules/remove_ehr_data_without_consent_test.py
tests/integration_tests/data_steward/cdr_cleaner/cleaning_rules/null_invalid_foreign_keys_test.py
tests/integration_tests/data_steward/cdr_cleaner/cleaning_rules/deid/suppress_year_of_birth_records_test.py    -----NOT THIS ONE
tests/integration_tests/data_steward/cdr_cleaner/cleaning_rules/deid/questionnaire_response_id_map_test.py
tests/integration_tests/data_steward/cdr_cleaner/cleaning_rules/clean_smoking_test.py                     -----NOT THIS ONE
tests/integration_tests/data_steward/cdr_cleaner/cleaning_rules/remove_operational_pii_fields_test.py     -----NOT THIS ONE
tests/integration_tests/data_steward/cdr_cleaner/cleaning_rules/id_de_duplicate_test.py
tests/integration_tests/data_steward/cdr_cleaner/cleaning_rules/drop_unverified_survey_data_test.py
tests/integration_tests/data_steward/cdr_cleaner/cleaning_rules/missing_concept_record_suppression_test.py                      -- FAILURE
```
I checked these against the last passing integration test you ran and already marked those that don't seem to impact the pass/failure of this rule (i.e. they ran before the failing test in both pass and fail scenarios).

When I look at your test dataset, I see that the vocabulary tables (except for `concept`) still exist in the combined dataset.  `concept` does not exist because this failing rule accounted for the table creation and deletion.  The test is failing to remove the data for `observation_type_concept_id = 19`.  This is a valid `concept_id` in a full `concept` table.  The test was expecting to create a `concept` table with 3 records where `concept_id = 19` does not exist.

One of the rules running before this test copied the vocabulary tables into the dataset and didn't delete them.  This rule then appends three more records to the existing table.  The rule does not expect a `concept_id = 19` to exist in the `concept` table.  Because a preceding test didn't clean up after itself, it does.  Therefore, this test is failing.  

You're code didn't cause this failure.  It is the fact that test timings are used to determine which tests are batched together.  A rule that leaves a dirty environment is now running before the failing test.  You'll need to find which one of the rules running before this one is copying in the vocabulary tables and not adding them to the list of `fq_table_names`.  If they had been added to the list, they would have been deleted.

lrwb-aou: @jp3477 Did you see the comment above?  Does it make sense?

jp3477: Hi @lrwb-aou, sorry forgot to respond. Yes this makes sense. Haven't been able to figure out which one it is yet. Tried to see which rules had the string "19" in them, but couldn't find one with the issue. Will keep trying.

jp3477: Oh I see it now. Wasn't a manual definition of the concept table as I originally thought. It was a use of the `copy_vocab_tables` method. I'm thinking I can add a  `truncate_vocab_tables` method or something like that in the base class for cleanup.

jp3477: @lrwb-aou It looks like the `PopulateRouteIds` rule was the culprit. I started running into another random test failure, but I think it should be resolved once this workflow finishes running.

jp3477: @lrwb-aou Sorry this is taking so long, running into another weird test failure.

tests/integration_tests/data_steward/cdr_cleaner/cleaning_rules/bigquery_tests_base.py
```diff
@@ -354,18 +350,6 @@ def copy_vocab_tables(self, vocabulary_id):
...
-            self.vocabulary_loaded = True
-
-        def tearDown(self):
-            super().tearDown()
-
-            # Remove all vocabulary tables if created
-            if self.vocabulary_loaded:
-                for table in self.vocabulary_tables:
-                    self.client.delete_table(table, not_found_ok=True)
-
```
lrwb-aou: @jp3477 It's good to revert this commit, but I like your thought process.  If we are going to add the vocabulary tables with the base class, we can add them to the "automatic cleanup" at this time as well.  But I think instead of all of this, you could say:
```python
self.client.copy_table(src_table, dst_table)    # this currently exists in copy_vocab_tables
self.fq_table_names.append(destination)     # this could be added
```
This should add them to the list of names that will be deleted by either the class or test `tearDown` functions.  Both say `not_found_ok=True` so if something is added twice, it isn't a failure.

data_steward/cdr_cleaner/cleaning_rules/populate_survey_conduct_ext.py
lrwb-aou: yeah, it's weird to have that many random failures all at once.  I'm going to see if someone has updated the test base recently.

lrwb-aou: There must have been a system hiccup last week that prevented your datasets from being created.  The tests are passing now.

tests/integration_tests/data_steward/cdr_cleaner/cleaning_rules/bigquery_tests_base.py
lrwb-aou: @jp3477 Are you interested in adding this line to the `copy_vocab_tables` method?

lrwb-aou (approved):
@jp3477 Thanks for seeing this through.

data_steward/cdr_cleaner/cleaning_rules/populate_survey_conduct_ext.py
jp3477: @lrwb-aou Awesome, yeah I was running out of ideas haha

tests/integration_tests/data_steward/cdr_cleaner/cleaning_rules/bigquery_tests_base.py
jp3477: @lrwb-aou Sure, I'll add this in. Think it'll be useful

PR #1471, opened 2023-01-11
[DC-2799] Update cleaning rule for case-insensitive free text suppression

ksdkalluri:
* Update FreeTextSurveyResponseSuppression to suppress case-insensitive free text concept codes
* Add FreeTextSurveyResponseSuppression to registered tier cleaning classes
* Update integration tests to cover case-insensitive concept codes


nishanthpp93 (commented):
Looks good! Some tests are failing but maybe rebase to develop might fix it?

PR #1476, opened 2023-01-11
[DC-2981] CircleCI Security Alert: update gcloud_cred and slack_webhook

ratuagga:
CircleCI config update with new slack-orb version.


MichaelSchmidt82 (approved):
solid work

PR #1478, opened 2023-01-19
[DC-3007] Update clean_mapping cleaning rule to correctly handle EHR datasets

hiro-mishima:
- For EHR datasets, the `table_id` was overwritten to a wrong value.
- I fixed the logic. I added test cases to ensure the update is working as expected.
- I updated `fields_for()` in `resources.py` so it can work for `unioned_ehr_xyz` tables in EHR datasets.


lledezma (approved):
Looks good. Thank you Hiro

PR #1483, opened 2023-01-26
[DC-2704] Bug fix

hiro-mishima:
- Updated the logic to find any violation of the day, even when there is one record that says `No violation found`.
- Updated the description of the arg parser.


MichaelSchmidt82 (approved):
the good ol' `any()` to the rescue

PR #1485, opened 2023-01-31
[DC-3019] Hot-fix scripts for "missing basics"

hiro-mishima:


ratuagga (approved):
Changes look good to me. But it would be great if 1 more developer can review this PR before moving forward.
Thanks

ksdkalluri (dismissed):
@hiro-mishima Thanks for adding the script, I verified that

1. We were able to sandbox the bad basics data
2. add the remidiated basics back to observation table with new observation_ids
3. add extra basics records to observation table
4. Fix survey_conduct table records as well.
5. Update state_of_residence fields in person_ext table in _deid datasets.
6. Integration tests were added as well.

The only test that is left for me to do is to run run_basics_remediation.py script to see if it runs without errors in prod environment. 🤞🏽

hiro-mishima:
For your question, person_id 9 is a participant that we must not include to the output.
(e.g. AIAN participants are included in incremental_dataset while  the final output must not include AIAN participants for releases W/O AIAN)
I added the comment above to the test cases too for clarity. Thanks!!

ksdkalluri (approved):
Thanks for adding the comments Hiro!

data_steward/resources.py
```diff
@@ -575,9 +576,43 @@ def get_git_tag():
...
+    Based on the old dataset name and the new release tag, creates a new
+    dataset name. This function assumes both the dataset name and the release
+    tag follow the Curation team's naming convention. It may return an
+    incorrect dataset name if those are irregularly named.
+    Args:
+        src_dataset_name: Name of the source dataset.
+        release_tag: Release tag for the new datasets.
+    Returns: Name of the new dataset.
+    """
+    return re.sub(r'\d{4}q\dr\d', release_tag, src_dataset_name)
```
molly-giddings, line 618: Hi, @hiro-mishima - might we suggest the following?
`r'\d{4}q[1-4]r\d{1,3}'`
This will:

- ensure quarter is one digit only, 1-4
- allow 1, 2, or 3 digit revision numbers

hiro-mishima: Thanks for your suggestion, @molly-giddings! I asked @MichaelSchmidt82 to include this change since he is working on a relevant ticket right now: https://github.com/all-of-us/curation/pull/1491#issuecomment-1428748870

data_steward/tools/run_basics_remediation.py
```diff
@@ -0,0 +1,255 @@
...
+    ask_if_continue()
+    create_sandbox_dataset(client, tag)
+    LOGGER.info(
+        f"Completed [3/5] Completed an empty sandbox dataset if not exists.\n")
+
+    LOGGER.info(
+        f"Starting [4/5] Run remediation and cleaning rules on the new dataset."
+    )
+    ask_if_continue()
+    if not is_deid_dataset(new_dataset) and not is_combined_dataset(
```
molly-giddings, line 223: @hiro-mishima, This logic check is fine for now, but will need to be removed when we remediate the lower priority datasets.  Just putting the comment here as a placeholder.

data_steward/cdr_cleaner/manual_cleaning_rules/remediate_basics.py
```diff
@@ -0,0 +1,395 @@
...
+    dataset_id STRING NOT NULL OPTIONS(description="Dataset ID of the destination table."),
+    observation_id INT64 NOT NULL OPTIONS(description="Newly assigned observation_id for the destination dataset.")
+)
+OPTIONS
+(
+    description="Lookup table for observation_ids. New observation_ids are assigned when loaded so no observation_id will be a duplicate."
+)
+""")
+
+# NOTE In case of re-running remediation for the same dataset, you need to run
```
molly-giddings, line 42: We need the generated IDs to be stable across ALL of the pipeline datasets for release tags 9 and 10.  The ID generation uses `ROWNUMBER() OVER(ORDER BY observation_id)` which will provide different results in tables with differing numbers of rows. Because of this, 

1. Logic needs to be added to only generate this lookup table for the largest number of observations dataset (which is `2022q4r7_combined_release with AIAN`)
2. Subsequent runs of this cleaning rule on other datasets should not delete or recreate this table.
3. The order of running the phases should start with `2022q4r7_combined_release`
@hiro-mishima

data_steward/cdr_cleaner/manual_cleaning_rules/remediate_basics.py
```diff
@@ -0,0 +1,395 @@
...
+        SELECT MAX(observation_id) 
+        FROM `{{project}}.{{dataset}}.{{table}}`
+        )
+FROM `{{project}}.{{incremental_dataset}}.{{table}}` inc_o
+WHERE person_id IN (
+    SELECT person_id FROM `{{project}}.{{dataset}}.{{table}}` 
+)
+""")
+
+SANDBOX_OBS = JINJA_ENV.from_string("""
```
molly-giddings, line 64: Imagined scenario:
Person A answers parent question and 1 or more child questions in bad dataset.
Person A has data in incremental dataset and answers parent question differently, such that they should not have answered the child questions from their previous parent answer.
Because the sandboxing is done on `person_id` and `observation_source_concept_id`, the old child data will not be sandboxed.  

Is this a risk we are comfortable with or should we modify the sandboxing to use the `concept_id`list (to be generated to the best of our abilities knowing there is an issue with health insurance related questions)?

@mandyroberts

data_steward/resources.py
hiro-mishima: I ended up adding this suggestion to the open PR for DC-3079. Thanks for the suggestion!
https://github.com/all-of-us/curation/pull/1489/commits/4b12430e03556fc106b1435f871ffd2f94aaa35d#diff-6ecbf63270b1fe0210a08915e522c72d8d164fdb10460211699a92511fb6e929R618

data_steward/tools/run_basics_remediation.py
hiro-mishima: Thanks! I removed this check at the PR for DC-3079: https://github.com/all-of-us/curation/pull/1489/commits/4b12430e03556fc106b1435f871ffd2f94aaa35d#diff-86e18cafe179274064a9a88590e2ce09cf6204860c5ad9d0b4db4ff2135ad5c0L223

data_steward/tools/run_basics_remediation.py
```diff
@@ -0,0 +1,255 @@
...
+    LOGGER.info(f"Running remediation for {new_dataset}...")
+    clean_dataset(project_id,
+                  new_dataset,
+                  f"{tag}_sandbox", [(RemediateBasics,)],
+                  incremental_dataset_id=args.incremental_dataset_id)
+
+    LOGGER.info(f"Completed running remediation for {new_dataset}...")
+
+    LOGGER.info(
+        f"[4/5] Run remediation and cleaning rules on the new datasets.\n")
```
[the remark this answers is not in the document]
hiro-mishima: I added this additional message and pause here so we can be sure we stop here and do validation first: https://github.com/all-of-us/curation/pull/1489/commits/4b12430e03556fc106b1435f871ffd2f94aaa35d#diff-86e18cafe179274064a9a88590e2ce09cf6204860c5ad9d0b4db4ff2135ad5c0R247

data_steward/cdr_cleaner/manual_cleaning_rules/remediate_basics.py
hiro-mishima: Thanks! Your point makes sense. I added the parameter `dataset_with_largest_observation_id ` so we can specify the dataset you mentioned. 
https://github.com/all-of-us/curation/pull/1489/commits/4b12430e03556fc106b1435f871ffd2f94aaa35d#diff-638190b97f9a117d7f9c2a0fc16996cb1608d05f353b383919366a8b7aec007fR50

I also added this `if` statement in `setup_rule` so this table creation and population occurs only once:
https://github.com/all-of-us/curation/pull/1489/commits/4b12430e03556fc106b1435f871ffd2f94aaa35d#diff-638190b97f9a117d7f9c2a0fc16996cb1608d05f353b383919366a8b7aec007fR345

data_steward/cdr_cleaner/manual_cleaning_rules/remediate_basics.py
molly-giddings: @lrwb-aou & @molly-giddings will look at how many person_ids might have such a scenario in the data.

molly-giddings: @lrwb-aou put the SQL in this ticket that identifies data of this scenario.  The results show consistent multi-select question value pairs for all data.  Therefore, there does not seem to be a risk of creating non-sandboxed data for participants with existing real data.

data_steward/cdr_cleaner/manual_cleaning_rules/remediate_basics.py
molly-giddings: Thanks again, @hiro-mishima for your quick responses on this.  The `dataset_with_largest_observation_id` parameter and `if` statement are good, but I think the sandbox parameter value will change based on the `tag` for the run.  So, a different lookup table would be created for revision 9 and revision 10, when they should really use the same table.

hiro-mishima: @molly-giddings @lrwb-aou 
Can r9 and r10 have different generated observation_ids? The current logic creates `_observation_id_map` for r9 and r10 individually. The reason behind it is to have `_observation_id_map` has only the necessary records for each release. If r9 and r10's observation_ids must be consistent, I will need to update both the script and the QC notebook. It shouldn't be too difficult. Let me know if I should go ahead and update the logic.

hiro-mishima: Krishna helped me fix a few more lines: https://github.com/all-of-us/curation/commit/92f4c7bd84d03faa5d0e273723076ba51025c891

molly-giddings: @lrwb-aou and I reviewed and it looks good.  Thanks, @hiro-mishima & @ksdkalluri!

PR #1486, opened 2023-02-06
[DC-3058] Deprecate bq.py and bq_utils.py helper functions with a warning

lledezma:


nishanthpp93 (approved):
Looks good! I would also prefer something like `use xyz instead`, but this is also fine.

lledezma:
I thought about doing what you suggested, but I couldn't find suitable replacements within project.

nishanthpp93:
> I thought about doing what you suggested, but I couldn't find suitable replacements within project.

Aren't these moved to common? You can just say use the constants in common.py. For the other functions, are they not replaced by something else?

lledezma:
They were not moved to common and replacements were not created because they are not used at all in curation.

PR #1491, opened 2023-02-10
[DC-3065] Update scripts to handle two-digit release (2022q4r10) correctly

MichaelSchmidt82:
I searched our code base for any `release_tag` modifiers such as regex substitution, slicing, indexing, or modifiers such as upper() that would cause an error.  Fortunately, Lauren's [commit](https://github.com/all-of-us/curation/pull/1457/commits/415d719c3735ba5139c430119390eea599fff09f) already handled most of the cases, so the only contribution I could make was updating a single case check.

Other than that, I made a nitpick change to two unused variables.  By convention, `_` is often used.


hiro-mishima:
Hi @MichaelSchmidt82, can you please add this change to this PR? `DC-3019` is already closed and it makes sense to include it here I think: https://github.com/all-of-us/curation/pull/1485#discussion_r1105052001

PR #1494, opened 2023-02-14
[DC-3079] table_namer for hotfix CR

hiro-mishima:


ksdkalluri (approved):
This will help since we are going to just use one sandbox for each release version.

PR #1497, opened 2023-02-15
[DC-3063] Update insert statements/ Update QC notebook

hiro-mishima:
I am fixing the scripts based on the execution from 2/14.
- Updated `insert` statements in the main hotfix script so it will also insert new records. (It was inserting only records that had corresponding records in the source dataset.) 
- Introduced `exclude_lookup_table` and `exclude_lookup_dataset` to achieve the above.
- Added a new test case, `test_no_exclusion,` to test whether the above is working as expected.
- Added a new test case (`person_id==3`) to test whether new participants are correctly inserted into the output.
- Deleted `QC 2-2` from the QC notebook as we saw yesterday the check was meaningless.

NOTE: `person_ext` tables' state-related columns can be `NULL` if the participants are new since the incremental dataset.
https://github.com/all-of-us/curation/pull/1497/files#diff-638190b97f9a117d7f9c2a0fc16996cb1608d05f353b383919366a8b7aec007fR198

### DIFF
--- a/data_steward/cdr_cleaner/manual_cleaning_rules/remediate_basics.py
+++ b/data_steward/cdr_cleaner/manual_cleaning_rules/remediate_basics.py
@@ -458,7 +458,7 @@ if __name__ == '__main__':
         action='store',
         dest='exclude_lookup_dataset',
         help=
-        (f'Dataset that has the PID/RID table for exlusion (e.g. AIAN participants for W/O AIAN dataset).'
+        (f'Dataset that has the PID/RID table for exclusion (e.g. AIAN participants for W/O AIAN dataset).'
         ),
         required=False)
     parser.add_argument(
@@ -466,7 +466,7 @@ if __name__ == '__main__':
         action='store',
         dest='exclude_lookup_table',
         help=
-        (f'PID/RID table for exlusion (e.g. AIAN participants for W/O AIAN dataset).'
+        (f'PID/RID table for exclusion (e.g. AIAN participants for W/O AIAN dataset).'
         ),
         required=False)
 
--- a/data_steward/cdr_cleaner/manual_cleaning_rules/remediate_basics.py
+++ b/data_steward/cdr_cleaner/manual_cleaning_rules/remediate_basics.py
@@ -130,8 +130,12 @@ SELECT
 FROM `{{project}}.{{incremental_dataset}}.{{table}}` inc_o
 JOIN `{{project}}.{{obs_id_lookup_dataset}}.{{new_obs_id_lookup}}` new_id
 ON inc_o.observation_id = new_id.source_observation_id
+WHERE inc_o.person_id IN ( -- Include only existing participants --
+    SELECT DISTINCT person_id
+    FROM `{{project}}.{{dataset}}.person`
+)
 {% if exclude_lookup_dataset and exclude_lookup_table %}
-WHERE inc_o.person_id NOT IN (
+AND inc_o.person_id NOT IN (
     SELECT 
         {% if is_deid %} research_id {% else %} person_id {% endif %}
     FROM `{{project}}.{{exclude_lookup_dataset}}.{{exclude_lookup_table}}` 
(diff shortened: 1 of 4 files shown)


data_steward/cdr_cleaner/manual_cleaning_rules/remediate_basics.py
```diff
@@ -129,9 +130,13 @@
 FROM `{{project}}.{{incremental_dataset}}.{{table}}` inc_o
 JOIN `{{project}}.{{obs_id_lookup_dataset}}.{{new_obs_id_lookup}}` new_id
 ON inc_o.observation_id = new_id.source_observation_id
-WHERE inc_o.person_id IN (
-    SELECT person_id FROM `{{project}}.{{dataset}}.{{table}}` 
+{% if exclude_lookup_dataset and exclude_lookup_table %}
```
ksdkalluri, line 133: @hiro-mishima am I assuming right that `exclude_lookup_dataset` and `exclude_lookup_table` corresponds to the names where AIAN participants list is available?

hiro-mishima: Yes, that is correct, @ksdkalluri

data_steward/cdr_cleaner/manual_cleaning_rules/remediate_basics.py
```diff
@@ -146,12 +151,16 @@
 FROM `{{project}}.{{incremental_dataset}}.{{table}}` i
 JOIN `{{project}}.{{obs_id_lookup_dataset}}.{{new_obs_id_lookup}}` oim
 ON i.observation_id = oim.source_observation_id
+{% if exclude_lookup_dataset and exclude_lookup_table %}
 WHERE i.observation_id IN (
```
molly-giddings, line 155: @hiro-mishima - Do we need the WHERE clause on 155-156?  The inner WHERE clause on 157-160 should suffice if not AIAN, unless I'm missing something (which is very possible)?

data_steward/cdr_cleaner/manual_cleaning_rules/remediate_basics.py
```diff
@@ -182,22 +195,48 @@
...
+    FROM `{{project}}.{{exclude_lookup_dataset}}.{{exclude_lookup_table}}` 
+)
+{% endif %}
+""")
+
+GENERIC_INSERT_MAP_EXT = JINJA_ENV.from_string("""
+INSERT INTO `{{project}}.{{dataset}}.{{table}}`
+SELECT * FROM `{{project}}.{{incremental_dataset}}.{{table}}`
+{% if exclude_lookup_dataset and exclude_lookup_table %}
 WHERE {{domain}}_id IN (
```
molly-giddings, line 231: Hi, @hiro-mishima - similar to above question, did you have a specific scenario in mind where this could break?  Otherwise, perhaps we could remove the outer WHERE clause on 231-232.

data_steward/cdr_cleaner/manual_cleaning_rules/remediate_basics.py
hiro-mishima: @molly-giddings We need these two WHERE clauses because this JINJA template is for observation mapping table. Observation mapping table does not have `person_id` column. So we need to reference observation table for `person_id`.

data_steward/cdr_cleaner/manual_cleaning_rules/remediate_basics.py
hiro-mishima: Same as above, mapping tables and ext tables do not have `person_id` columns so these two where clauses are necessary.

data_steward/cdr_cleaner/manual_cleaning_rules/remediate_basics.py
```diff
@@ -182,22 +195,48 @@
...
-JOIN `{{project}}.{{sandbox_dataset}}.{{sandbox_table}}` sb
+LEFT JOIN `{{project}}.{{sandbox_dataset}}.{{sandbox_table}}` sb
 ON i.person_id = sb.person_id
+{% if exclude_lookup_dataset and exclude_lookup_table %}
+WHERE i.person_id NOT IN (
+    SELECT 
+        {% if is_deid %} research_id {% else %} person_id {% endif %}
+    FROM `{{project}}.{{exclude_lookup_dataset}}.{{exclude_lookup_table}}` 
+)
+{% endif %}
```
lrwb-aou, line 212: @hiro-mishima and @ksdkalluri to make sure we are grabbing valid data you might try adding:
```python
AND i.person_id in (
    SELECT distinct person_id
    FROM `{{project}}.{{dataset}}.{{table}}
)
```

molly-giddings:
Thanks, @hiro-mishima!  Pending the integration test passing, this looks good to @lrwb-aou and me.

ksdkalluri (approved):
@hiro-mishima Looks good to me. I think we got all the conditions covered. :crossed_fingers:

PR #1498, opened 2023-02-16
[DC-1862] Refactor uses of `bq_utils.py` `get_table_info(...)` function

lledezma:


data_steward/bq_utils.py
```diff
@@ -158,6 +158,10 @@ def get_table_id(hpo_id, table_name):
     return hpo_id + '_' + table_name
 
 
+@deprecated(
+    reason=
+    'see: https://cloud.google.com/python/docs/reference/bigquery/latest/google.cloud.bigquery.client.Client#google_cloud_bigquery_client_Client_get_table'
+)
```
hiro-mishima, line 164: I see `get_table_info()` is called by `bq_utils.create_snapshot_dataset()`.
https://github.com/all-of-us/curation/blob/develop/data_steward/bq_utils.py#L969

You left it as it is because `bq_utils.create_snapshot_dataset()` is already flagged as `deprecated`, correct? 
I think we should still update it so these deprecated functions have less dependency on each other.
What do you think?

hiro-mishima (commented):
Hi @lledezma, thanks for working on it! I left two comments. Please take a look. Thanks!

data_steward/bq_utils.py
lledezma: `bq_utils.create_snapshot_dataset()` is not used at all, and that's why I didn't update it.

hiro-mishima: Oh, you're right - It is not used anywhere. Let's remove `bq_utils.create_snapshot_dataset()` altogether if that's the case.

hiro-mishima (approved):
Thanks for the updates! It looks good!

data_steward/bq_utils.py
lledezma: I don't plan on removing anything in bq_utils.py, only deprecate. Once everything has been deprecated, bq_utils.py will be deleted.

PR #1502, opened 2023-02-21
[DC-3044] Update cron jobs for retraction to retract in place

nishanthpp93:
* [DC-3044] Update retraction to skip based on type
* [DC-3044] Add env vars for datasets table
* [DC-3044] Update retraction cron to also clean

[DC-3044]: https://precisionmedicineinitiative.atlassian.net/browse/DC-3044?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[DC-3044]: https://precisionmedicineinitiative.atlassian.net/browse/DC-3044?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[DC-3044]: https://precisionmedicineinitiative.atlassian.net/browse/DC-3044?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/data_steward/retraction/retract_data_bq.py
+++ b/data_steward/retraction/retract_data_bq.py
@@ -449,6 +449,14 @@ def run_bq_retraction(project_id,
     :param bq_client: BigQuery client. Reuse the client if one already exists. If not, a new one will be created.
     :return:
     """
+
+    # Skip retraction if type is incorrect
+    if retraction_type not in [RETRACTION_ONLY_EHR, RETRACTION_RDR_EHR]:
+        LOGGER.info(
+            f"{retraction_type} is not {RETRACTION_ONLY_EHR} or {RETRACTION_RDR_EHR}. Skipping BQ retraction."
+        )
+        return
+
     client = bq_client if bq_client else BigQueryClient(project_id)
 
     # NOTE get_datasets_list() excludes sandbox datasets and datasets that are type=OTHER.
--- a/data_steward/validation/main.py
+++ b/data_steward/validation/main.py
@@ -29,7 +29,8 @@ import cdm
 import common
 from cdr_cleaner import clean_cdr
 from cdr_cleaner.args_parser import add_kwargs_to_args
-from constants.cdr_cleaner.clean_cdr import DATA_CONSISTENCY
+from constants.cdr_cleaner.clean_cdr import CRON_RETRACTION
+import constants.global_variables
 from gcloud.bq import BigQueryClient
 from gcloud.gcs import StorageClient
 import resources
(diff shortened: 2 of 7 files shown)


data_steward/retraction/retract_data_bq.py
```diff
@@ -449,6 +449,14 @@ def run_bq_retraction(project_id,
     :param bq_client: BigQuery client. Reuse the client if one already exists. If not, a new one will be created.
     :return:
     """
+
+    # Skip retraction if type is incorrect
```
hiro-mishima, line 453: I love this check! Thanks for adding it.

data_steward/validation/main.py
```diff
@@ -1037,34 +1040,70 @@ def union_ehr():
...
-                                        sandbox_dataset_id,
-                                        pid_table_id,
-                                        hpo_id,
-                                        folder,
-                                        force_flag=True)
-    logging.info(f"Completed retraction from internal bucket folders")
+    # Run memory changes
+    for dataset_id in dataset_ids:
+        logging.info(f"Running CRs for {dataset_id}...")
+        cleaning_args = [
```
hiro-mishima, line 1085: `DATA_CONSISTENCY` has these 3 cleaning rules, and `1` and `2` fail with some old datasets:
1. `DropOrphanedSurveyConductIds`
2. `DropOrphanedPIDS`
3. `CleanMappingExtTables`

`survey_conduct` exits only after `2022q4`. So, `1` needs to run on datasets from `2022q4` and newer.
`2` looks at `visit_detail`, and `visit_detail` exists only  after `2022q2`. So, `2` needs to run on datasets from `2022q2` and newer.
`3` should run with no problem with any datasets.

https://github.com/all-of-us/curation/blob/develop/data_steward/cdr_cleaner/clean_cdr.py#L349

So, instead of running `DATA_CONSISTENCY` for everything, we might need to run individual cleaning rules depending on which dataset it is.

hiro-mishima (commented):
Hi @nishanthpp93, thanks for the update! It looks great. There is one thing we need to consider regarding `DATA_CONSISTENCY`. Please take a look at my comment.

data_steward/validation/main.py
nishanthpp93: Thanks, I added a `RETRACTION_CRON` cleaning stage with only the ext/mapping cleaning rule.

hiro-mishima (approved):
Thanks for all the updates, @nishanthpp93! This is good to go 🚀

PR #1503, opened 2023-02-21
[DC-3014] Use `bq.copy_dataset` in `create_rdr_snapshot` instead of `copy_raw_rdr_dataset`

MichaelSchmidt82:
`data_steward/tools/create_rdr_snapshot.py` is not taking full advantage of the tools at its disposal via the `gcloud/bq/__init__.py` functionality.  Instead, this script defines `copy_raw_rdr_tables` to copy the contents of one dataset to another.   This PR deletes the `copy_raw_rdr_tables(...)` function and replaces the call with our implementation in the BigQueryClient, removing code duplication.

However, when `copy_raw_rdr_tables` created each table, it used a job_config (`CopyJobConfig`) to make each table with a specified write disposition which was later passed to the core copy_table method.  An optional parameter now facilitates this functionality.  If none is provided, a default is created.

This PR adds a feature to add labels `table_id`, `copy_from`, and `copy_to` to the copied tables using this new job_config.

### DIFF
--- a/data_steward/gcloud/bq/__init__.py
+++ b/data_steward/gcloud/bq/__init__.py
@@ -241,13 +241,14 @@ class BigQueryClient(Client):
         :param job_config: An optional google.cloud.bigquery.job.CopyJobConfig
         :return: incomplete jobs
         """
-        job_config = job_config if job_config else CopyJobConfig(write_disposition=WriteDisposition.WRITE_EMPTY)
+        job_config = job_config if job_config else CopyJobConfig(
+            write_disposition=WriteDisposition.WRITE_EMPTY)
         # Copy input dataset tables to backup and staging datasets
         tables = super(BigQueryClient, self).list_tables(input_dataset)
         job_list = []
         for table in tables:
             staging_table = f'{output_dataset}.{table.table_id}'
-            job_config.labels.update( {
+            job_config.labels.update({
                 'table_name': table.table_id,
                 'copy_from': input_dataset,
                 'copy_to': output_dataset
--- a/data_steward/gcloud/bq/__init__.py
+++ b/data_steward/gcloud/bq/__init__.py
@@ -242,6 +242,9 @@ class BigQueryClient(Client):
         :return: incomplete jobs
         """
 
+        if not job_config:
+            job_config = CopyJobConfig()  # create an empty default
+
         # Copy input dataset tables to backup and staging datasets
         tables = super(BigQueryClient, self).list_tables(input_dataset)
         job_list = []
--- a/data_steward/gcloud/bq/__init__.py
+++ b/data_steward/gcloud/bq/__init__.py
@@ -249,9 +249,9 @@ class BigQueryClient(Client):
         for table in tables:
             staging_table = f'{output_dataset}.{table.table_id}'
             job_config.labels.update({
-                'table_name': table.table_id,
-                'copy_from': input_dataset,
-                'copy_to': output_dataset
+                'table_name': table.table_id.lower(),
+                'copy_from': input_dataset.lower(),
+                'copy_to': output_dataset.lower()
             })
             job = self.copy_table(table, staging_table, job_config=job_config)
             job_list.append(job.job_id)
--- a/data_steward/gcloud/bq/__init__.py
+++ b/data_steward/gcloud/bq/__init__.py
@@ -250,11 +250,11 @@ class BigQueryClient(Client):
         job_list = []
         for table in tables:
             staging_table = f'{output_dataset}.{table.table_id}'
-            job_config.labels = {
+            job_config.labels.update( {
                 'table_name': table.table_id,
                 'copy_from': input_dataset,
                 'copy_to': output_dataset
-            }
+            })
             job = self.copy_table(table, staging_table, job_config=job_config)
             job_list.append(job.job_id)
         self.wait_on_jobs(job_list)
--- a/data_steward/gcloud/bq/__init__.py
+++ b/data_steward/gcloud/bq/__init__.py
@@ -12,6 +12,8 @@ from time import sleep
 from google.api_core import retry
 from google.cloud import bigquery
 from google.cloud.bigquery import Client
+from google.cloud.bigquery.job import CopyJobConfig, WriteDisposition
+
 from google.auth import default
 from google.api_core.exceptions import GoogleAPIError, BadRequest
 from google.cloud.exceptions import NotFound
@@ -227,7 +229,10 @@ class BigQueryClient(Client):
 
         return dataset
 
-    def copy_dataset(self, input_dataset: str, output_dataset: str) -> list:
+    def copy_dataset(self,
+                     input_dataset: str,
+                     output_dataset: str,
+                     job_config: CopyJobConfig = None) -> list:
         """
         Copies tables from source dataset to a destination datasets
 
@@ -240,7 +245,7 @@ class BigQueryClient(Client):
         job_list = []
         for table in tables:
             staging_table = f'{output_dataset}.{table.table_id}'
-            job = self.copy_table(table, staging_table)
+            job = self.copy_table(table, staging_table, job_config=job_config)
             job_list.append(job.job_id)
         self.wait_on_jobs(job_list)
         return job_list
--- a/data_steward/gcloud/bq/__init__.py
+++ b/data_steward/gcloud/bq/__init__.py
@@ -232,7 +232,8 @@ class BigQueryClient(Client):
     def copy_dataset(self,
                      input_dataset: str,
                      output_dataset: str,
-                     job_config: CopyJobConfig = None) -> list:
+                     job_config: CopyJobConfig = None,
+                     include_metadata: bool = False) -> list:
         """
         Copies tables from source dataset to a destination datasets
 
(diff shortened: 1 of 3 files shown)


data_steward/gcloud/bq/__init__.py
```diff
@@ -227,20 +229,33 @@ def define_dataset(self, dataset_id: str, description: str,
...
         """
         Copies tables from source dataset to a destination datasets
 
         :param input_dataset: fully qualified name of the input(source) dataset
         :param output_dataset: fully qualified name of the output(destination) dataset
         :return: incomplete jobs
         """
+        if include_metadata and not job_config:
+            job_config = CopyJobConfig()
+
```
lrwb-aou, line 246: I would suggest reverting this change too.

data_steward/gcloud/bq/__init__.py
```diff
@@ -227,20 +229,33 @@ def define_dataset(self, dataset_id: str, description: str,
...
         for table in tables:
             staging_table = f'{output_dataset}.{table.table_id}'
-            job = self.copy_table(table, staging_table)
+            if include_metadata:
+                job_config.labels = {
+                    'table_name': table.table_id,
+                    'copy_from': input_dataset,
+                    'copy_to': output_dataset
+                }
+            job = self.copy_table(table, staging_table, job_config=job_config)
```
lrwb-aou, line 258: If we're going to add labels, lets add them for all copy jobs.  This would become
```python
            job_config = CopyJobConfig()
            job_config.labels = {
                'table_name': table.table_id,
                'copy_from': input_dataset,
                'copy_to': output_dataset
            }
            job = self.copy_table(table, staging_table, job_config=job_config)
```

data_steward/tools/create_rdr_snapshot.py
```diff
@@ -10,7 +10,7 @@
 from cdr_cleaner import clean_cdr
 from cdr_cleaner.args_parser import add_kwargs_to_args
 from utils import auth
-from gcloud.bq import BigQueryClient
+from gcloud.bq import BigQueryClient, CopyJobConfig, WriteDisposition
 from utils import pipeline_logging
```
lrwb-aou, line 14: Then you can remove this.

data_steward/tools/create_rdr_snapshot.py
```diff
@@ -99,9 +99,17 @@ def main(raw_args=None):
 
     # create staging, sandbox, and clean datasets with descriptions and labels
     datasets = create_datasets(bq_client, args.rdr_dataset, args.release_tag)
+    
+    job_config = CopyJobConfig(
+        write_disposition=bigquery.job.WriteDisposition.WRITE_EMPTY)
```
lrwb-aou, line 104: you could remove this.

data_steward/tools/create_rdr_snapshot.py
```diff
@@ -99,9 +99,17 @@ def main(raw_args=None):
...
 
-    # copy raw data into staging dataset
-    copy_raw_rdr_tables(bq_client, args.rdr_dataset, datasets.get('staging'))
+    bq_client.copy_dataset(input_dataset=args.rdr_dataset,
+                           output_dataset=datasets.get('staging'),
+                           job_config=job_config,
+                           include_metadata=True)
```
lrwb-aou, line 109: you could also remove these extra lines.

data_steward/gcloud/bq/__init__.py
```diff
@@ -227,20 +229,33 @@ def define_dataset(self, dataset_id: str, description: str,
...
 
         :param input_dataset: fully qualified name of the input(source) dataset
         :param output_dataset: fully qualified name of the output(destination) dataset
+        :param job_config: An optional google.cloud.bigquery.job.CopyJobConfig
         :return: incomplete jobs
         """
+
+        if not job_config:
+            job_config = CopyJobConfig()  # create an empty default
+
```
lrwb-aou, line 247: This could be a ternary operator if it is truly an empty default.
```suggestion
        job_config = job_config if job_config else CopyJobConfig()
```

I personally prefer to specify a `WRITE_EMPTY` default value if we are copying
```suggestion
        job_config = job_config if job_config else CopyJobConfig(write_disposition=WriteDisposition.WRITE_EMPTY)
```

data_steward/gcloud/bq/__init__.py
```diff
@@ -227,20 +229,33 @@ def define_dataset(self, dataset_id: str, description: str,
...
         tables = super(BigQueryClient, self).list_tables(input_dataset)
         job_list = []
         for table in tables:
             staging_table = f'{output_dataset}.{table.table_id}'
-            job = self.copy_table(table, staging_table)
+            job_config.labels = {
+                'table_name': table.table_id,
+                'copy_from': input_dataset,
+                'copy_to': output_dataset
+            }
```
lrwb-aou, line 257: This would be better as:
```suggestion
            job_config.labels.update( {
                'table_name': table.table_id,
                'copy_from': input_dataset,
                'copy_to': output_dataset
            })
```
Above, you are allowing someone to specify a `CopyJobConfig`, but this line reassigns the `labels` dictionary, potentially wiping out any labels that the calling process has specified.  Using the `update` dictionary feature will allow the job to keep any pre-existing labels and append the extra key/value pairs.  The only catch is if the previously defined labels happen to use one of the three labels defined here.  In that case, those values would be overwritten.

MichaelSchmidt82: This is a great catch and definitely would introduce a bug.  I do know that @brendagutman is working on adding ownership labels.

>  The only catch is if the previously defined labels happen to use one of the three labels defined here. In that case, those values would be overwritten.

This is true too.  I'm not sure how that would happen, though.

data_steward/tools/create_rdr_snapshot.py
```diff
@@ -5,12 +5,10 @@
 from datetime import datetime
```
lrwb-aou, line 5: check on your side, but I think this import can be removed entirely.

data_steward/tools/create_rdr_snapshot.py
```diff
@@ -5,12 +5,10 @@
...
 import logging
 
-from google.cloud import bigquery
-
 from cdr_cleaner import clean_cdr
 from cdr_cleaner.args_parser import add_kwargs_to_args
 from utils import auth
-from gcloud.bq import BigQueryClient
+from gcloud.bq import BigQueryClient, CopyJobConfig, WriteDisposition
```
lrwb-aou, line 11: This works but seems a little odd to me.  `gcloud.bq` isn't defining or extending `CopyJobConfig` or `WriteDisposition`, so why are we importing them from there?  If someone were to remove their imports in `gcloud.bq` in the future, then this would break.  I would recommend importing them directly here as a Third party import (where you removed the `bigquery` import on line 8).
```python
from google.cloud.bigquery.job import CopyJobConfig, WriteDisposition
```

data_steward/tools/create_rdr_snapshot.py
MichaelSchmidt82: I found an removed an additional unused import

data_steward/tools/create_rdr_snapshot.py
MichaelSchmidt82: Corrected

data_steward/tools/create_rdr_snapshot.py
```diff
@@ -1,16 +1,18 @@
...
-from argparse import ArgumentParser
-from datetime import datetime
+# Python standard imports
 import logging
+from argparse import ArgumentParser
 
-from google.cloud import bigquery
+# Third-party imports
+from gcloud.bq import BigQueryClient
+from google.cloud.bigquery.job import CopyJobConfig, WriteDisposition
```
lrwb-aou, line 10: Technically
```python
from gcloud.bq import BigQueryClient     # this is a project level import.  this is referencing our code.
from google.cloud.bigquery.job import CopyJobConfig, WriteDisposition    # this one is the third party import
```

tests/unit_tests/data_steward/gcloud/bq/bq_test.py
```diff
@@ -8,7 +8,7 @@
 # Third party imports
 from unittest.mock import ANY
 
-from google.api_core import retry
+
```
lrwb-aou, line 11: I think linting is complaining about this extra blank line now.

data_steward/tools/create_rdr_snapshot.py
lrwb-aou: was the change pushed?

data_steward/tools/create_rdr_snapshot.py
lrwb-aou: was the change pushed?

data_steward/gcloud/bq/__init__.py
```diff
@@ -227,20 +229,31 @@ def define_dataset(self, dataset_id: str, description: str,
...
         # Copy input dataset tables to backup and staging datasets
         tables = super(BigQueryClient, self).list_tables(input_dataset)
         job_list = []
         for table in tables:
             staging_table = f'{output_dataset}.{table.table_id}'
-            job = self.copy_table(table, staging_table)
+            job_config.labels.update({
+                'table_name': table.table_id,
+                'copy_from': input_dataset,
+                'copy_to': output_dataset
```
lrwb-aou, line 254: @MichaelSchmidt82 I just realized, since you are leaving these here you must ensure the label values are always [lower case](https://cloud.google.com/resource-manager/docs/creating-managing-labels#requirements).  You should wrap these as:
```suggestion
                'table_name': table.table_id.lower(),
                'copy_from': input_dataset.lower(),
                'copy_to': output_dataset.lower()
```

data_steward/tools/create_rdr_snapshot.py
MichaelSchmidt82: Yes

data_steward/tools/create_rdr_snapshot.py
```diff
@@ -1,16 +1,17 @@
...
+from argparse import ArgumentParser
 
-from google.cloud import bigquery
+# Third-party imports
+from gcloud.bq import BigQueryClient
 
+# Project level imports
+from utils import auth
 from cdr_cleaner import clean_cdr
 from cdr_cleaner.args_parser import add_kwargs_to_args
```
lrwb-aou, line 14: You're missing your third party imports and have a project import listed with the set of third party imports.  I would expect this to look like:
```suggestion
# Third-party imports
from google.cloud.bigquery.job import CopyJobConfig, WriteDisposition

# Project level imports
from cdr_cleaner import clean_cdr
from cdr_cleaner.args_parser import add_kwargs_to_args
from gcloud.bq import BigQueryClient
from utils import auth
```

This explains the recent linting error.

lrwb-aou (approved):
@MichaelSchmidt82 All tests are passing.  This looks good!

PR #1505, opened 2023-02-22
[DC-3052] Account for Skips and whitespace in clean_ppi_numeric_fields_using_parameters.py

brendagutman:


tests/integration_tests/data_steward/cdr_cleaner/cleaning_rules/clean_ppi_numeric_fields_using_parameters_test.py
```diff
@@ -206,7 +225,16 @@ def test_invalid_values_fields_cleaning(self):
                  self.invalid_values_value_as_concept_id),
                 (120, 1333023, 1333023, None, None,
                  self.six_plus_value_as_concept_id,
-                 self.six_plus_value_as_concept_id)
+                 self.six_plus_value_as_concept_id),
+                # Test skips
```
hiro-mishima, line 229: Thanks for adding these test cases! It helped me understand the queries this CR runs. It reminded me that there is logic specifically for `1333023` to null out `value_as_string`. (Though I do not understand why this null-out logic is only for `1333023`... Let me know if you know the background.)

hiro-mishima: Your comment in the JIRA ticket was helpful, too!

hiro-mishima (approved):
@brendagutman This looks great! Thanks for the update :tada:

tests/integration_tests/data_steward/cdr_cleaner/cleaning_rules/clean_ppi_numeric_fields_using_parameters_test.py
brendagutman: Hey @hiro-mishima , thanks for bringing this up. It looks like the line that nulls value_as_string for `1333023 ` should only affect questions that were not skipped. That portion of the CR was to account for an actual response string in value_as_string field and should not have affected the skips.

I made a change to the cleaning query, sandbox query, and the integration test expectations. I also added the 'PMI Skip' test cases. Those are the changes since your last review.

hiro-mishima (approved):
Thanks for the additional update, @brendagutman! I am glad we could catch that 👏 I re-approve this PR here.
It looks like the CircleCi testing had failed, but it will succeed by simply re-running it. (It is the BigQuery resource limit issue we sometimes see.)

PR #1507, opened 2023-02-22
[DC-3060] Remove unused variables required by `deid_runner.sh`

MichaelSchmidt82:
This PR removed two unused variables in the deid_runner.sh shell script.  It involved three fundamental changes:

1) The removal of three command line parameters in both deid_runner.sh and cdm.py scripts
2) Refactoring `data_steward/cdm.py`
- Create a [factory function](https://github.com/all-of-us/curation/blob/79c5160495046aa7e117a752c29f11ee3c7a5d53/data_steward/cdm.py#L16) to return an argument parser.  This parser now takes advantage of a [mutual exclusion clause](https://github.com/all-of-us/curation/blob/79c5160495046aa7e117a752c29f11ee3c7a5d53/data_steward/cdm.py#L25).
- Create a [main()](https://github.com/all-of-us/curation/blob/79c5160495046aa7e117a752c29f11ee3c7a5d53/data_steward/cdm.py#L91) function with simplified branching logic.
3) Create a unit test where one didn't exist before.
- Four test cases for [valid argument](https://github.com/all-of-us/curation/blob/79c5160495046aa7e117a752c29f11ee3c7a5d53/tests/unit_tests/data_steward/cdm_test.py#L68) and function calls.
- Five test cases with [bad arguments](https://github.com/all-of-us/curation/blob/79c5160495046aa7e117a752c29f11ee3c7a5d53/tests/unit_tests/data_steward/cdm_test.py#L26), including a NotYetImpleted ([here](https://github.com/all-of-us/curation/blob/883df1bf9f6dc5377dd5948c72af015d089f131f/data_steward/cdm.py#L102)) error ([here](https://github.com/all-of-us/curation/blob/883df1bf9f6dc5377dd5948c72af015d089f131f/tests/unit_tests/data_steward/cdm_test.py#L62)).

One critical piece to review is the branching logic near [line 96](https://github.com/all-of-us/curation/blob/883df1bf9f6dc5377dd5948c72af015d089f131f/data_steward/cdm.py#L96) and the mutex clause on [line 25](https://github.com/all-of-us/curation/blob/883df1bf9f6dc5377dd5948c72af015d089f131f/data_steward/cdm.py#L25), as the original logic had a dead end/do nothing branch. Was this the original intention? There is a pass statement on [line 105](https://github.com/all-of-us/curation/blob/883df1bf9f6dc5377dd5948c72af015d089f131f/data_steward/cdm.py#L105) to retain the original logic.


data_steward/cdm.py
```diff
@@ -13,6 +13,27 @@
 logger = logging.getLogger(__name__)
 
 
+def create_parser():
```
lrwb-aou, line 16: This change and the unit test feels like scope creep.  @ksdkalluri thoughts?

ksdkalluri: @MichaelSchmidt82 Like Lauren said, these changes are good but out of scope. Lets focus on just the scope in 3060 for now, Reason being we may be completely getting rid of this script when we start refactoring registered tier.

MichaelSchmidt82: The function was added for unit testing.  See [here](https://stackoverflow.com/questions/18160078/how-do-you-write-tests-for-the-argparse-portion-of-a-python-module) specifically.  There are other cases where this is done also within the tool directory.  see [here](https://github.com/all-of-us/curation/blob/9741f6364aaca96c87e2d0cf05ce776b710de908/data_steward/tools/delete_stale_test_buckets.py#L85) 

I spoke with @nishanthpp93 about this script, and he suggested a unit test.

nishanthpp93: @MichaelSchmidt82 A unit test is great for testing parsers, so thank you for creating it. However, we don't need it for this particular use case right now, especially for this ticket. Such changes would be good in a separate ticket created for that purpose.

MichaelSchmidt82: Cool.  I'll back it out.

MichaelSchmidt82:
This notebook appears to have been created some time ago.  YAPF did not like it's format whatsoever.

ksdkalluri (approved):
Changes look good!

PR #1509, opened 2023-02-24
[DC-3064] Add column names for partitioned tables

hiro-mishima:
- The RT deid dataset has partitioned tables. With partitioned tables, we need to specify column names for DMLs, because those have a `_PARTITIONTIME` column, and `SELECT *` is too ambiguous in such a case. 
- The solution is to specify columns in the DDLs. This change will not break the DDLs for non-partitioned tables.


ksdkalluri (approved):
Looks good, thanks for the changes.

PR #1516, opened 2023-02-28
[DC-3068] Update `cdr_cleaner/cleaning_rules/update_ppi_negative_pain_level.py` to set `value_as_string` similarly to RDR settings for skipped responses

brendagutman:


hiro-mishima (approved):
Thanks, @brendagutman! It looks good.
There is one comment that needs an update accordingly. Please update it before merging. Thanks 🚀 
https://github.com/all-of-us/curation/blob/develop/data_steward/cdr_cleaner/cleaning_rules/update_ppi_negative_pain_level.py#L52

PR #1518, opened 2023-02-28
[DC-3042] ` _get_kwargs` needs to be refactored to accept `=` key/value pairs in `clean_cdr.py`

MichaelSchmidt82:
In data_steward/cdr_cleaner/clean_cdr.py, the function _get_kwargs will fail with an error when it encounters an = operator with key value usage.  This PR fixes the issue and updates a unit test.

### DIFF
--- a/data_steward/cdr_cleaner/clean_cdr.py
+++ b/data_steward/cdr_cleaner/clean_cdr.py
@@ -6,6 +6,7 @@ to the query engine.
 """
 # Python imports
 import logging
+import typing
 
 # Project imports
 import cdr_cleaner.clean_cdr_engine as clean_engine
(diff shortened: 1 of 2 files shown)


lrwb-aou (changes requested):
[equals_parsing.txt](https://github.com/all-of-us/curation/files/10865996/equals_parsing.txt)

I'm providing a patch file for the `clean_cdr.py` file only.  The unit tests look fine (the patch file doesn't make any changes to the tests) and the patch tests pass.

I think we should talk, because the current patch is over complicating the fix.

data_steward/analytics/cdr_ops/clean_rdr_export_qc.py
```diff
@@ -231,7 +231,7 @@
    -- ,questionnaire_response_id --
```
lrwb-aou, line 231: @MichaelSchmidt82 why is this file in this PR?

lrwb-aou (approved):
@MichaelSchmidt82 This looks good and should be ready to merge!  Woohoo!

PR #1519, opened 2023-03-01
[DC-3094] Birth suppression rule should also be applied to the RT datasets

MichaelSchmidt82:
This PR adds `BirthInformationSuppression` to the  `REGISTERED_TIER_DEID_CLEANING_CLASSES` list.  It should be applied after `VehicularAccidentConceptSuppression`.  It is also added to the `affected_datasets` variable within the rule as well.


lrwb-aou (approved):
This looks good.

lrwb-aou:
> This looks good.

@MichaelSchmidt82 👆

PR #1500, opened 2023-02-20
[DC-3013] Create CR to clean the concept_id fields in survey_conduct table

brendagutman:

PR #1520, opened 2023-03-03
[DC-3082] Create `survey_conduct` table CR to fix COPE data prior to setting `survey_source_concept_id`

brendagutman:


data_steward/cdr_cleaner/cleaning_rules/clean_survey_conduct_recurring_surveys.py
```diff
@@ -0,0 +1,176 @@
+"""
+Inserts data representing the version of a survey into survey_conduct table for surveys with custom ids. Ex COPE/Minute
```
hiro-mishima, line 2: nit: I think `update` or `correct` is more accurate for this CR instead of `insert`.

data_steward/cdr_cleaner/cleaning_rules/clean_survey_conduct_recurring_surveys.py
```diff
@@ -0,0 +1,176 @@
...
+JIRA_ISSUE_NUMBERS = ['DC-3082']
+
+DOMAIN_TABLES = [SURVEY_CONDUCT]
+REFERENCE_TABLES = [COPE_SURVEY_MAP]
+
+SANDBOX_SURVEY_CONDUCT = JINJA_ENV.from_string("""
+CREATE OR REPLACE TABLE `{{project_id}}.{{sandbox_dataset_id}}.{{sandbox_table_id}}` AS (
+SELECT * FROM `{{project_id}}.{{dataset_id}}.survey_conduct`
+WHERE survey_conduct_id IN (SELECT questionnaire_response_id FROM `{{project_id}}.{{dataset_id}}.{{cope_survey_map}}`)
+AND (survey_concept_id NOT IN (2100000006, 2100000007, 2100000002, 2100000005, 2100000004, 2100000003, 905047, 905055, 765936, 1741006)
```
hiro-mishima, line 36: Will this list be expanded in the future? (e.g., We get cope surveys with new `cope_month` that is not listed here)
If yes, this CR will unintentionally sandbox the new surveys. `CLEAN_SURVEY_CONDUCT` will not update such records (those will fall into `ELSE`), so it will not be a problem after all.

So, it is not a problem, but I wanted to mention it here for clarity.

data_steward/cdr_cleaner/cleaning_rules/clean_survey_conduct_recurring_surveys.py
```diff
@@ -0,0 +1,176 @@
...
+    WHEN cope_month = 'nov' THEN 'AoUDRC_SurveyVersion_CopeNovember2020'
+    WHEN cope_month = 'july' THEN 'AoUDRC_SurveyVersion_CopeJuly2020'
+    WHEN cope_month = 'june' THEN 'AoUDRC_SurveyVersion_CopeJune2020'
+    WHEN cope_month = 'vaccine1' THEN 'cope_vaccine1'
+    WHEN cope_month = 'vaccine2' THEN 'cope_vaccine2'
+    WHEN cope_month = 'vaccine3' THEN 'cope_vaccine3'
+    WHEN cope_month = 'vaccine4' THEN 'cope_vaccine4'
+    ELSE sc.survey_source_value
+    END
+FROM (SELECT *
```
hiro-mishima, line 71: nit: You can simplify this SQL using `JOIN` instead of `LEFT JOIN`. Something like this:
```
FROM (SELECT *
      FROM `{{project_id}}.{{dataset_id}}.survey_conduct` sc
      JOIN `{{project_id}}.{{dataset_id}}.{{cope_survey_map}}` m
      ON m.questionnaire_response_id = sc.survey_conduct_id
      ) sub
```

hiro-mishima (approved):
Thanks for this PR, @brendagutman! It looks great.
I left some comments and suggestions. All are not critical. Please update the scripts if necessary before merging this branch to `develop`.

Thanks 🎉

data_steward/cdr_cleaner/cleaning_rules/clean_survey_conduct_recurring_surveys.py
```diff
@@ -0,0 +1,176 @@
...
+
+Recurring versions of surveys such as COPE had all of their questions and answers mapped to a single Module in the PPI
+vocabulary. The same was the case for the Minute surveys. This will make it seem like there is only one version of each
+of these surveys. Researchers need to know the version of the survey the participant took. This CR will use
+`cope_survey_semantic_version_map` which is provided to curation from the rdr team, to update the survey version in
+survey_source_concept_id and survey_source_value for each observation.
+
+There are multiple cleaning rules that clean the survey conduct table.
+This CR will clean survey_concept_id and survey_source_value for the surveys with custom concept ids.
+Another cleans survey_source_concept_id for all valid records.
```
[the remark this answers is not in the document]
brendagutman: @hiro-mishima That is a CR that I almost have the PR out for [DC-3013](https://github.com/all-of-us/curation/pull/1500). This PR blocks its completion, but it will be ready for review soon.

data_steward/cdr_cleaner/cleaning_rules/clean_survey_conduct_recurring_surveys.py
```diff
@@ -0,0 +1,176 @@
...
+Then observations that do not have a valid survey will be dropped in another CR.
```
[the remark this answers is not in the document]
brendagutman: @hiro-mishima Yes, do you think I should add the CR titles in the description?

data_steward/cdr_cleaner/cleaning_rules/clean_survey_conduct_recurring_surveys.py
brendagutman: That I am not sure of. I believe we will not get anymore COPE or Minute surveys but it is very possible that we get more recurring surveys in the future. If other recurring surveys do get added and they do map to the same Module code then I don't know if a new mapping table would be created or the one used here would be expanded. 
I think the plan is to map  any new recurring surveys to different module codes so this **shouldn't** be an issue, but I plan to look into the raw_rdr_export qc notebook. If a check does not exist for new surveys or differences in cope_month, etc. then I'll make a ticket and get it added. Listing the CRs that will need an update if it/they fail.

data_steward/cdr_cleaner/cleaning_rules/clean_survey_conduct_recurring_surveys.py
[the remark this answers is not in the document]
hiro-mishima: Sure, I think it will help the readers' comprehention. Thanks!

PR #1522, opened 2023-03-06
[DC-3096] [DC-3097] Abstract class for backfill/ Backfill CR for "The Basics"

hiro-mishima:
## This PR's scope
- Creating the abstract class `AbstractBackfillSurveyRecords` for the backfill.
- `BackfillTheBasics` and its test are also included in this PR as the pilot case for using the abstract class.
- `BackfillTheBasics` does not use the arg `additional_backfill_conditions`. It will be used when implementing the backfill for the Overall Health survey (`DC-3099`)

## Change in the `observation_id` assignment logic
- The original cleaning rule used to add `1000000000000` for the newly assigned `observation_ids`. ([BackfillPmiSkipCodes](https://github.com/all-of-us/curation/blob/develop/data_steward/cdr_cleaner/cleaning_rules/backfill_pmi_skip_codes.py#L84)). With `AbstractBackfillSurveyRecords`, new `observation_ids` are created based on the max `observation_id` in `observation`. ([link](https://github.com/all-of-us/curation/pull/1522/files#diff-181ea7b99537ef8a6765d1e5fb6ceeec741e2ea8724d4893a4c65ffc367078f2R94))
- The QC [clean_rdr_export_qc.py](https://github.com/all-of-us/curation/blob/develop/data_steward/analytics/cdr_ops/clean_rdr_export_qc.py#L663) assumes the backfilled records to be `observation_id >= 1000000000000`. This QC needs an update. **I will create a new ticket to address this issue once this PR is approved.**

### DIFF
--- a/data_steward/cdr_cleaner/cleaning_rules/backfill_survey_records.py
+++ b/data_steward/cdr_cleaner/cleaning_rules/backfill_survey_records.py
@@ -1,0 +1,188 @@
+"""
+An abstract survey backfill class that expects an instantiating class to
+provide a list of observation_source_concept_ids to use for backfilling.
+If a participant has answered at least one of these questions, but has not
+answered them all, this rule is responsible for filling in the missing
+responses with skip codes.  
+
+This abstract class does most of the backfill work with customization provided
+by extending classes. It expects extending classes to provide the list of
+concepts that will be used to determine backfill eligibility.  
+It also allows extending classes to define criteria that may be specific to a
+sub-group (e.g. we only backfill menstruation questions for female participants).
+
+Original issue: DC-3096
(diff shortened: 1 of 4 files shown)


data_steward/cdr_cleaner/cleaning_rules/backfill_survey_records.py
```diff
@@ -0,0 +1,188 @@
+"""
```
MichaelSchmidt82, line 1: Great docstring:  Concise and informative

MichaelSchmidt82 (approved):
I've broken down the bulk SQL into segments, and they make sense to me.

PR #1523, opened 2023-03-06
[DC-3117] Automate csv file update for hpo_id_bucket_name

ratuagga:
automate bucket_name csv update.

### DIFF
--- a/data_steward/tools/add_hpo.py
+++ b/data_steward/tools/add_hpo.py
@@ -429,23 +429,23 @@ def main(project_id, hpo_id, org_id, hpo_name, bucket_name, display_order,
     if addition_type == "update_config":
         add_hpo_site_to_csv_files(hpo_id, hpo_name, org_id, bucket_name,
                                   hpo_site_csv_path, display_order)
-    # elif addition_type == "update_lookup_tables":
-    #     if bucket_access_configured(gcs_client, bucket_name):
-    #         LOGGER.info(f'Accessing bucket {bucket_name} successful. '
-    #                     f'Proceeding to add site.')
-    #         add_lookups(bq_client, hpo_id, hpo_name, org_id, bucket_name,
-    #                     display_order)
-    #
-    #         LOGGER.info(
-    #             f'hpo_site_id_mappings table successfully updated. Updating `{bq_consts.HPO_SITE_ID_MAPPINGS_TABLE_ID}` '
-    #             f'table')
-    #         update_site_masking_table(bq_client, hpo_id, us_state,
-    #                                   value_source_concept_id)
-    #
-    #     else:
-    #         raise RuntimeError(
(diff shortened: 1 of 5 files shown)


data_steward/tools/add_hpo.py
```diff
@@ -140,6 +179,60 @@ def add_hpo_site_mappings_csv(hpo_id,
...
+    :param hpo_name: name of the hpo
+    :param org_id: hpo organization identifier
+    :param bucket_name: GCS bucket name of the site
+    :param hpo_site_csv_path: path to csv file containing hpo site information
+    :param display_order: index number in which hpo should be added in table
+    :return:
+    """
+    # Check if csv files exist in given path.
+    hpo_site_mappings_path = hpo_site_csv_path + '/hpo_site_mappings.csv'
+    hpo_id_bucket_name_path = hpo_site_csv_path + '/hpo_id_bucket_name.csv'
```
jp3477, line 217: These two lines will cause a bad path if `hpo_site_csv_path` already has a trailing slash (e.g. combined path can become "hpo_site_csv_path//hpo_id_bucket_name.csv". I recommend using pathlib to form the paths.
```suggestion
    hpo_site_mappings_path = Path(hpo_site_csv_path) / 'hpo_site_mappings.csv'
    hpo_id_bucket_name_path = Path(hpo_site_csv_path) / 'hpo_id_bucket_name.csv'
```

tests/unit_tests/data_steward/tools/add_hpo_test.py
```diff
@@ -122,3 +127,70 @@ def test_check_state_code_format(self):
...
+            'bucket_name': 'fake_bucket'
+        }
+        hpo_site_csv_path = 'hpo_site_csv_files'
+
+        # This test will pass because by default, 'hpo_site_csv_files' folder is not present
+        # in Curation repo.
+        self.assertRaises(RuntimeError, add_hpo.add_hpo_site_to_csv_files,
+                          new_site['hpo_id'], new_site['hpo_name'],
+                          new_site['org_id'], new_site['bucket_name'],
+                          hpo_site_csv_path, new_site['display_order'])
```
jp3477, line 146: This is up to you: This test passing by default due to the lack of a folder may not be the most informative. For example, it would not have been able to find an error like the one mentioned above. I recommend adding a failure case of some kind as well.

jp3477 (changes requested):
This looks great, just have a couple suggestions.

data_steward/tools/add_hpo.py
ratuagga: sure.

tests/unit_tests/data_steward/tools/add_hpo_test.py
ratuagga: ok, sure.

jp3477 (approved):
This looks good, thanks for adding those changes.

PR #1526, opened 2023-03-07
[DC-3118] Update `copy_dataset()` not to use dots for labels

hiro-mishima:


brendagutman (approved):
Thanks for getting this updated!

PR #1529, opened 2023-03-09
[DC-3009] Update scheduled query for prod ID prevention

hiro-mishima:
I updated the scheduled query. The previous version's loop logic was incomplete.


MichaelSchmidt82 (approved):
Solid work as usual

PR #1540, opened 2023-03-14
[DC-3061] Add copy ext tables process to registered tier deid step

ksdkalluri:


data_steward/tools/run_deid.py
```diff
@@ -291,6 +291,34 @@ def load_deid_map_table(client, deid_map_dataset_name, age_limit):
...
+    :param input_dataset: Name of the input dataset
+    :param output_dataset: Name of the output dataset
+    :return: job_list: list of job_ids
+    """
+    source_tables = bq_client.list_tables(
+        f'{bq_client.project}.{input_dataset}')
+    job_config = CopyJobConfig(write_disposition=WriteDisposition.WRITE_EMPTY)
+    job_list = []
+    for table in source_tables:
+        if table.table_id.endswith('_ext'):
```
lrwb-aou, line 308: you could import and use `EXT_SUFFIX` here.

data_steward/tools/run_deid.py
```diff
@@ -291,6 +291,34 @@ def load_deid_map_table(client, deid_map_dataset_name, age_limit):
...
+            destination_table = f'{output_dataset}.{table.table_id}'
+            job_config.labels.update({
+                'table_name': table.table_id.lower(),
+                'copy_from': input_dataset.lower(),
+                'copy_to': output_dataset.lower()
+            })
```
lrwb-aou, line 314: This is the second or third time I've seen these copy job labels.  Does it make sense to move this to a utility somewhere?  What are your recommendations?

data_steward/tools/run_deid.py
```diff
@@ -291,6 +291,34 @@ def load_deid_map_table(client, deid_map_dataset_name, age_limit):
         )
 
 
+def copy_ext_tables(bq_client, input_dataset: str, output_dataset: str):
```
lrwb-aou, line 294: Thanks for the typing hints!  Please add a typing hint for the return type.

lrwb-aou (commented):
I've got two requests and two thoughts I'd like your feedback on.  Thank you for doing this here rather than in bash.

data_steward/tools/run_deid.py
```diff
@@ -353,6 +381,12 @@ def main(raw_args=None):
 
     copy_suppressed_table_schemas(known_tables, args.odataset)
 
+    logging.info(
+        f"Copying ext tables from {args.input_dataset} dataset to {args.odataset} dataset..."
+    )
+    copy_ext_tables(bq_client, args.input_dataset, args.odataset)
```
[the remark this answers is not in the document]
ksdkalluri: Makes sense.

data_steward/tools/run_deid.py
ksdkalluri: There is a ticket thay i am working on which fixes the illegal characters in the lables which touches both the places you mentioned, do you think, it will be good ideat to refactor there instead of here?

lrwb-aou: That does sound like a better place to make this change.

PR #1541, opened 2023-03-14
[DC-3119] Update BQ client functions not to use forbidden letters for labels

ksdkalluri:


hiro-mishima (approved):
It looks great. Thanks, @ksdkalluri!
There is a merge conflict. Please solve it before proceeding.

PR #1542, opened 2023-03-14
[DC-3121] Update clean_rdr_export_qc.py to correctly handle the updated backfill logic

hiro-mishima:
- I also removed the misguiding comment for the parameter `raw_rdr` in the notebook.


brendagutman (approved):
Looks good. Thanks for fixing the parameter while you were at it!

PR #1549, opened 2023-03-22
[DC-3172] Remove logging to App Engine drive for Clean CDR

nishanthpp93:
* [DC-3172] Disable log write to app engine

### DIFF
--- a/data_steward/validation/main.py
+++ b/data_steward/validation/main.py
@@ -1096,7 +1096,7 @@ def run_retraction_cron():
         cleaning_args = [
             '-p', project_id, '-d', dataset_id, '-b', sandbox_dataset_id,
             '--data_stage', CRON_RETRACTION, '--run_as',
-            f'{project_id}@appspot.gserviceaccount.com', '-s'
+            f'{project_id}@appspot.gserviceaccount.com'
         ]
         all_cleaning_args = add_kwargs_to_args(cleaning_args, None)
         clean_cdr.main(args=all_cleaning_args)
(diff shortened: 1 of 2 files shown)


data_steward/validation/main.py
```diff
@@ -1096,7 +1096,7 @@ def run_retraction_cron():
         cleaning_args = [
             '-p', project_id, '-d', dataset_id, '-b', sandbox_dataset_id,
             '--data_stage', CRON_RETRACTION, '--run_as',
-            f'{project_id}@appspot.gserviceaccount.com', '-s'
+            f'{project_id}@appspot.gserviceaccount.com'
```
hiro-mishima, line 1099: Do we need to remove `'-s'` here? I think it is still helpful to have the console log, if this `'-s'` is irrelevant from the local log file creation.

hiro-mishima (commented):
Thanks for the update, @nishanthpp93! I have one small question to confirm. Thanks!!

data_steward/validation/main.py
nishanthpp93: Yeah, the log file was my concern too. Should I keep it? I just don't want us to deploy again.

hiro-mishima (approved):
Thanks, @nishanthpp93! I think it is safe to go without `'-s'` here. I think our priority here is to finish the retraction asap. If anything happens, we can do without that small portion of the log.

Thanks again!

nishanthpp93:
Yeah this was my reasoning as well. Only the CR is affected.

PR #1550, opened 2023-03-23
[DC-3076] Create a script to restore a provided list of datasets with _restore suffix

MichaelSchmidt82:
While performing in-place cleaning or retractions, restoring a dataset to a previous timestamp may become necessary if data is wrongly modified/retracted.  This ticket/PR adds a method named `restore_from_time(...)`, which accepts a list of datasets and restores them (I.e, every table within the dataset) to a previous day within a 7-day window as allowed by BigQuery.

The parameter `days_ago` is used to signify how many days ago (of type `int` limited by `<7`)
An optional (but nearly required) parameter, `job_config`, is used to retain dataset labels.

[See the ticket](https://precisionmedicineinitiative.atlassian.net/browse/DC-3076)


data_steward/gcloud/bq/__init__.py
```diff
@@ -577,3 +577,63 @@ def wait_on_jobs(self,
...
+            queries = []
+            for table in org_tables:
+                queries.append(
+                    time_travel_q.render(project_id=self.project,
+                                         dataset_id=rest_dset.dataset_id,
+                                         table_id=table.table_id,
+                                         days_ago=days_ago))
+
+            # Execute each query
+            for query in queries:
```
ksdkalluri, line 638: @MichaelSchmidt82 This method does what is expected to be done. My only suggestion is let us add all the queries job_ids to a list and wait till all the queries are finished before we move on to next step when this method is called.
```suggestion
            for query in queries:
                job = self.query(query, job_config=job_config)
                job_list.append(job.job_id)
      self.wait_on_jobs(job_list)
      return job_list
```

ksdkalluri (changes requested):

data_steward/gcloud/bq/__init__.py
MichaelSchmidt82: I did not know one could do this.  This is a great way to prevent blocking! thanks.

data_steward/gcloud/bq/__init__.py
```diff
@@ -577,3 +577,68 @@ def wait_on_jobs(self,
...
+                    f'The dataset {org_dset.dataset_id}_restore already exists.  '
+                    f'Delete the dataset first to prevent an accidental overwrite.'
+                ))
+            # add the label to include 'temp'
+            rest_dset.labels.update({'temp': ''})
+            self.update_dataset(rest_dset, ['labels'])
+
+            # Generate a list of queries for each table from a template
+            queries = []
+            job_list = []
```
ksdkalluri, line 630: This job_list should be outside of this for loop

ksdkalluri:
@MichaelSchmidt82 is this ready for review again?

PR #1553, opened 2023-04-03
[DC-3048] `site_maskings` is not available at expected location for CT notebook, `analytics/cdr_ops/controlled_tier_qc/check_controlled_tier.py`

ratuagga:
updated site_masking table to be used from pipeline tables.

### DIFF
--- a/data_steward/analytics/cdr_ops/controlled_tier_qc/sql/query_templates.py
+++ b/data_steward/analytics/cdr_ops/controlled_tier_qc/sql/query_templates.py
@@ -114,7 +114,11 @@ WHERE {{ column_name }}
 {% endif %}
  NOT IN (
     SELECT {{ new_id }}
+    {% if mapping_table == 'site_maskings' %}
+    FROM `{{ project_id }}.{{ pipeline_dataset }}.{{ mapping_table }}`
+    {% else %}
     FROM `{{ project_id }}.{{ mapping_dataset }}.{{ mapping_table }}`
+    {% endif %}
 )
 """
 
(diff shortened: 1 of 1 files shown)


data_steward/analytics/cdr_ops/controlled_tier_qc/sql/query_templates.py
```diff
@@ -149,7 +149,7 @@
     post_deid.{{ column_name }} AS output_pid
 FROM `{{ project_id }}.{{ post_deid_dataset }}.{{ table_name }}` post_deid
 LEFT JOIN `{{ project_id }}.{{ pre_deid_dataset }}._mapping_{{ table_name|replace("_ext","") }}` pre_deid USING({{ primary_key }})
-LEFT JOIN `{{ project_id }}.{{ mapping_dataset }}.{{ mapping_table }}` map ON pre_deid.src_hpo_id = map.hpo_id
+LEFT JOIN `{{ project_id }}.pipeline_tables.{{ mapping_table }}` map ON pre_deid.src_hpo_id = map.hpo_id
 )
```
lledezma, line 153:
```suggestion
LEFT JOIN `{{ project_id }}.{{ pipeline_dataset }}.{{ mapping_table }}` map ON pre_deid.src_hpo_id = map.hpo_id
)
```

lledezma (commented):
Your update works but the dataset name is hard coded. The ticket says to reference the pipeline_tables dataset as a parameterized SQL value.

data_steward/analytics/cdr_ops/controlled_tier_qc/sql/query_templates.py
ratuagga: oh.. during the discussion about the ticket there were 2 ways discussed and I interpreted it as 'if going with using pipeline_tables, parameterization is not required.'.. let me make that change.

lledezma (approved):

PR #1554, opened 2023-04-04
[DC-3189] Temporary fix to remove  ValidateAllHpoFiles cron job warning

lledezma:


hiro-mishima (approved):
Thank you so much, Luis! That was quick ⚡

PR #1557, opened 2023-04-06
[DC-3131] Add aou_death to `common.py`

hiro-mishima:


nishanthpp93 (approved):
Looks good! I was expecting the schema file as well, but it can be in another PR. Also, the constant can be AOU_DEATH, and ALL_DEATH may be confusing to devs as to the actual name. Just things to consider.

hiro-mishima:
Thanks for the review! DC-3130 added the schema file. And to your point on `AOU_DEATH` vs `ALL_DEATH`, I was torn between these two. Let me change it to `AOU_DEATH`.

PR #1558, opened 2023-04-07
[DC-3195] Update `retract_data_bq.py` to handle "sandbox" and "other" datasets

hiro-mishima:


data_steward/retraction/retract_data_bq.py
```diff
@@ -389,24 +408,40 @@ def skip_dataset_retraction(dataset_id, retraction_type) -> bool:
...
+        Returns '' if the domain table cannot be identified, the domain table
+        does not contain person_id, or the domain table does not contain EHR data.
+    """
+    domain_tables = set(AOU_REQUIRED) - set(NON_PID_TABLES)
+
+    for domain in domain_tables:
+        if domain in table:
+            if domain == DEATH:
+                return f"{PERSON}_id"
+            return f"{domain}_id"
```
nishanthpp93, line 441: This is tricky, you may get observation_id for observation_period. I would suggest using 'endswith' and making sure the domain_id exists in the table, along with person_id.

nishanthpp93 (commented):
Impressive work testing this @hiro-mishima! This is definitely one of the more challenging tasks. I've suggested one small change, let me know your thoughts.

data_steward/retraction/retract_data_bq.py
hiro-mishima: I assumed we do not have to retract `observation_period` for `only_ehr` since it is not `AOU_REQUIRED`. But I was wrong. Some HPO sites send `observation_period` records. Other non-`AOU_REQUIRED` tables, such as `drug_era`, also have EHR data. Let me update the script.
I will also add a check if the domain_id exists in the table.

Thanks for pointing it out, @nishanthpp93!

hiro-mishima:
I updated the script and the test accordingly. Let me know if it looks good. Thanks, @nishanthpp93!

data_steward/retraction/retract_data_bq.py
```diff
@@ -417,28 +423,44 @@ def skip_dataset_retraction(dataset_id, retraction_type) -> bool:
...
-        if domain in table:
+        if table.endswith(domain):
+
             if domain == DEATH:
-                return f"{PERSON}_id"
-            return f"{domain}_id"
+                domain_id = f"{PERSON}_id"
+
+            elif domain == OBSERVATION and OBSERVATION_PERIOD in table:
+                domain_id = f"{OBSERVATION_PERIOD}_id"
```
nishanthpp93, line 455: Since we are using .endswith, this situation may not arise. Also, there must be some other place in the codebase (maybe in unioned) where there is an hpo prefix added to omop table names and we need to determine the omop table. You can try to replicate that.

nishanthpp93 (commented):
Thanks for updating @hiro-mishima! I would also suggest a small test for the get_domain_id loop, if possible within the test framework.

nishanthpp93 (commented):
Nice work on the tests @hiro-mishima! Let's meet and run this tomorrow and then it should be good to go.

hiro-mishima:
Thanks for reviewing, @nishanthpp93! I have attached the exec logs to the JIRA ticket's comment section. The logs looked good!

nishanthpp93 (approved):
Looks good! Thanks Hiro!

PR #1560, opened 2023-04-10
[DC-3193] Stop DEATH creation in Unioned EHR

hiro-mishima:
- `AOU_DEATH` creation will be handled in `DC-3134.`
- This PR is to suppress the `DEATH` table.


data_steward/validation/ehr_union.py
brendagutman, line 1: Suggestion: Add a short description defining the DEATH table oddities. If you aren't already planning to in the following ticket.

brendagutman (approved):
The changes look great, thanks @hiro-mishima!

PR #1561, opened 2023-04-10
[DC-3134] Update `ehr_union.py` to accommodate `aou_death`

hiro-mishima:
- `death` table is already suppressed with https://github.com/all-of-us/curation/pull/1560
- Updated `import common` and `import resources` for better readability


brendagutman (approved):
Thanks for adding helpful descriptions and I really liked the way you tested the primary death record. I don't have any suggestions but I do have some questions. I'll message you in Slack.
Thanks Hiro!

PR #1576, opened 2023-04-21
[DC-3186] Update `ValidDeathDates` to include `aou_death`

hiro-mishima:


brendagutman (approved):
Looks good Hiro. Thanks

PR #1579, opened 2023-04-25
[DC-3133] Update `NegativeAges` to include `aou_death`

hiro-mishima:


MichaelSchmidt82 (approved):
I looked over these files in their entirety, and the modifications you have made, and these changes are appropriate to what is being asked.  The string literal to constant conversions is a welcomed addition.

PR #1583, opened 2023-04-27
[DC-3204] Add aou_death

hiro-mishima:


MichaelSchmidt82 (approved):
I followed the integration test, and as with others, I didn't see anything in particular just like past tests.

PR #1584, opened 2023-04-28
[DC-3205] Update `YearOfBirthRecordsSuppression` to include `aou_death`

hiro-mishima:


MichaelSchmidt82 (approved):
These AOU_DEATH rule updates follow a common theme: outstanding work.

PR #1562, opened 2023-04-13
[DC-3175] Update the vocabulary QC notebook

brendagutman:

PR #1585, opened 2023-04-28
[DC-3175] Add checks to vocabulary_upload_qc.py

brendagutman:
Replaces PR  #1562


nishanthpp93 (commented):
Looks good! Thanks for updating @brendagutman. I would suggest re-running the circle tests since it shouldn't be affected by these changes.

nishanthpp93 (approved):
Looks good!

PR #1590, opened 2023-05-03
[DC-3161] Create device.json

brendagutman:


hiro-mishima (approved):
Looks good! We can open another PR if we need any other changes depending on the answers we receive.
Thanks!

PR #1591, opened 2023-05-08
[DC-3162] Update generate_fitbit_dataset.py with device and to create additional fields

brendagutman:


MichaelSchmidt82 (approved):
shockingly simple

PR #1594, opened 2023-05-09
[DC-3163] Update truncate_fitbit_data.py to include the device table

brendagutman:


data_steward/cdr_cleaner/cleaning_rules/truncate_fitbit_data.py
```diff
@@ -4,51 +4,44 @@
...
-SANDBOX_QUERY = common.JINJA_ENV.from_string("""
-CREATE OR REPLACE TABLE `{{project}}.{{sandbox}}.{{intermediary_table}}` AS (
+SANDBOX_QUERY = JINJA_ENV.from_string("""
+CREATE OR REPLACE TABLE `{{project_id}}.{{sandbox_id}}.{{intermediary_table}}` AS (
 SELECT * 
-FROM `{{project}}.{{dataset}}.{{table_name}}`
-WHERE {{date_field}} > {{cutoff_date}})""")
+FROM `{{project_id}}.{{dataset_id}}.{{fitbit_table}}`
+{% if datetime_fields and date_fields %}
+  WHERE (GREATEST({{date_fields}}) > DATE("{{truncation_date}}"))
```
hiro-mishima, line 29: I think the `COALESCE()` logic should be in this JINJA template rather than in `get_query_specs()` so we can see the whole query structure in one place. 

Can we change this 
`  WHERE (GREATEST({{date_fields}}) > DATE("{{truncation_date}}"))`
to be something like this?
```
# date_fields is the list of field names with datatype==date
# do the same for datetime_fields

  WHERE (GREATEST(
    {% for field in date_fields %}
    COALESCE({{field}}, DATE("1900-01-01"))
    {% endfor %}{% if not loop.last -%}, {% endif %}
  ) > DATE("{{truncation_date}}"))`
```
Or if `GREATEST` can have both date and datetime values at once, we can do it like this and we can remove the `if/elif` conditions in the JINJA template:
```
  WHERE (GREATEST(
    {% for field in date_fields + datetime_fields %}
    COALESCE({{field}}, DATE("1900-01-01"))
    {% endfor %}{% if not loop.last -%}, {% endif %}
  ) > DATE("{{truncation_date}}"))`
```
^ It's mock - some syntax might be wrong.

data_steward/cdr_cleaner/cleaning_rules/truncate_fitbit_data.py
```diff
@@ -4,51 +4,44 @@
...
 
 LOGGER = logging.getLogger(__name__)
 
-FITBIT_DATE_TABLES = [
-    common.ACTIVITY_SUMMARY, common.HEART_RATE_SUMMARY, common.SLEEP_LEVEL,
-    common.SLEEP_DAILY_SUMMARY
-]
-FITBIT_DATETIME_TABLES = [common.HEART_RATE_MINUTE_LEVEL, common.STEPS_INTRADAY]
-
-FITBIT_TABLES_DATE_FIELDS = {
```
hiro-mishima, line 29: I like this idea - removing hard-coding columns and tables for the CR.
Great refactoring!

data_steward/cdr_cleaner/cleaning_rules/truncate_fitbit_data.py
```diff
@@ -72,19 +65,18 @@ def __init__(self,
         """
         try:
             # set to provided date string if the date string is valid
-            datetime.strptime(truncation_date, '%Y-%m-%d')
-            self.truncation_date = truncation_date
+            self.truncation_date = validate_date_string(truncation_date)
```
hiro-mishima, line 68: 👍

data_steward/cdr_cleaner/cleaning_rules/truncate_fitbit_data.py
hiro-mishima: Great idea to use `GREATEST()` here, by the way!

data_steward/cdr_cleaner/cleaning_rules/truncate_fitbit_data.py
```diff
@@ -186,28 +179,26 @@ def validate_rule(self, client):
...
-                            dest='truncation_date',
-                            help='CDR truncation date',
-                            required=True)
+    ext_parser.add_argument(
+        '-t',
+        '--truncation_date',
+        action='store',
+        dest='truncation_date',
+        help=
+        ('Cutoff date for data based on <table_name>_date and <table_name>_datetime fields.  '
```
hiro-mishima, line 199: nit: This help needs an update - Column names are not `<table_name>_date`/`<table_name>_datetime` for FitBit.

hiro-mishima (changes requested):
Great work!! I added some suggestions for changes. Let me know what you think. And also, let me know if you have any questions. Thanks!!

data_steward/cdr_cleaner/cleaning_rules/truncate_fitbit_data.py
brendagutman: I like your idea of combining the fields!!! 

I went ahead and combined them in the `get_query_specs()` to simplify things. After doing that I realized we might not need a loop in the query at all which I find more readable, but this would require the COALESCE statement to stay in the `get_query_specs()`.  As readability is in the eye of the beholder, what do you think is best?

hiro-mishima: Your latest update is the most straightforward and most readable. You are correct that doing everything in the JINJA template here could make it look messier. Thanks for the update!

hiro-mishima (approved):
Great work :white_check_mark:

PR #1598, opened 2023-05-10
[DC-3203] Add aou_death

hiro-mishima:


MichaelSchmidt82 (approved):
Thanks for the descriptive variable names, string/variable interpolation, and explicit table lists.

PR #1602, opened 2023-05-15
[DC-3164] Apply RemoveParticipantDataPastDeactivationDate to the new table 'device'

brendagutman:


MichaelSchmidt82 (approved):
Thanks for your explanation!

PR #1607, opened 2023-05-22
[DC-3201] Add aou_death

hiro-mishima:


brendagutman (approved):
Thanks for the helpful PR comment and the additional resources in the ticket, Hiro!

PR #1612, opened 2023-05-30
[DC-3214] Add QC for aou_death

hiro-mishima:


hiro-mishima:
@MichaelSchmidt82 I realized some existing QCs are looking at `death`, so I updated it just now to consider `aou_death` as well. FYI!

MichaelSchmidt82 (approved):
I looked deeply into these changes and generated my own queries, and I have not found an issue.  Good work

PR #1613, opened 2023-05-30
[DC-3237] Add a check to the raw RDR export QA notebook to verify data origin flag

lledezma:


ksdkalluri (approved):
Tested with new rdr import. works as expected.

PR #1614, opened 2023-05-31
[DC-3245] Generate mapping tables when creating rdr snapshot dataset

lledezma:


data_steward/tools/create_rdr_snapshot.py
```diff
@@ -179,5 +197,64 @@ def create_datasets(client, rdr_dataset, release_tag):
...
+    :return:
+    """
+    query = mapping_query(domain_table, input_dataset_id, client.project)
+    mapping_table = mapping_table_for(domain_table)
+    logging.info(f'Query for {mapping_table} is {query}')
+    query_job = client.query(query)
+    query_job.result()
+
+
+def mapping_query(table_name, dataset_id=None, project_id=None):
```
ksdkalluri, line 217: We do not have to create mapping tables for all the tables that RDR gives us, we should only create _mpping_tables for only the tables that we are creating earlier. we have MAPPED_CLINICAL_DATA_TABLES + PERSON + DEATH. for all the other tables like fact_relationship and metadata, we can ignore creating mapping tables.

ksdkalluri (changes requested):

data_steward/tools/create_rdr_snapshot.py
lledezma: I believe we are not creating a mapping table for DEATH table.

PR #1618, opened 2023-06-05
[DC-3213] Update QC `ehr_union_qc.py` for `aou_death`

hiro-mishima:


brendagutman (approved):
Looks great as always! Thank you for the documentation, comments, and especially the failure messages. They would be a great help if we were to need them.
Thanks for adding the DEATH table query in addition! Great checks to have.

PR #1621, opened 2023-06-12
[DC-3278] Update PROD ID violation check SQL to handle partitioned tables without error

hiro-mishima:
- I already updated the scheduled query in our BigQuery. This PR is to update the reference SQL file to the latest


brendagutman (approved):
Thanks for the tips! Your explanation and screenshots helped enormously.

PR #1627, opened 2023-06-14
[DC-3258] Update create_combined_backup_dataset.py to include rdr origin flags in mapping tables

ksdkalluri:
Update combined dataset generation script to add data origins for PPI data in _mapping tables.


nishanthpp93 (approved):
Overall looks good! Actual change is minor and make sense and all tests pass.

PR #1628, opened 2023-06-15
[DC-3255] Modify the fitbit device 'date' field name to be 'device_date'

brendagutman:


MichaelSchmidt82 (approved):
A good ol' find & replace.

PR #1632, opened 2023-06-23
[DC-3312] Update historical retraction QC notebook to handle datasets with many tables

hiro-mishima:


brendagutman (approved):
Nice update, thanks Hiro!

PR #1633, opened 2023-06-23
[DC-3309] Update create_load_aou_death to use `CREATE OR REPLACE TABLE` instead of `CREATE TABLE` in ehr_union.py

hiro-mishima:


brendagutman (approved):
Looks good!

PR #1637, opened 2023-06-29
[DC-3317] Update aou_death.json to accept NULL death_date values

hiro-mishima:


brendagutman (approved):
Thanks for updating the descriptions as well, Hiro.

PR #1638, opened 2023-06-30
[DC-3261] Update affected_datasets

hiro-mishima:


brendagutman (approved):
Thanks for catching this Hiro!

PR #1640, opened 2023-07-03
[DC-3275] Ignore DailyParticipants folder in validation cron

ksdkalluri:


brendagutman (approved):
Thanks for the update Krishna!

PR #1642, opened 2023-07-03
[DC-3260] Update PERSON_EXT_TABLE_QUERY in create_person_ext_table.py to populate state_of_residence and sex_at_birth columns without recreating person_ext table.

ksdkalluri:


data_steward/cdr_cleaner/cleaning_rules/create_person_ext_table.py
```diff
@@ -9,7 +9,6 @@
 deid to be correctly de-identified.
 This cleaning rule will populate the person_ext table
 The following fields will need to be copied from the observation table:
-src_id (from observation_ext, should all be “PPI/PM”)
 state_of_residence_concept_id: the value_source_concept_id field in the OBSERVATION table row where
```
lledezma, line 12: This same comment is present in the integration and unit test. I would also remove it from there.

lledezma (changes requested):
Updates look good! I would also add the ticket number to the `issue_numbers` list and as one the the original issues in the docstrings.

ksdkalluri:
> Updates look good! I would also add the ticket number to the `issue_numbers` list and as one of the original issues in the docstrings.

Updated the docstring and added the `issue_number`

lledezma (changes requested):
Thanks for the updates! I would also update the docstring in the unit test.

lledezma (approved):

PR #1650, opened 2023-07-11
[DC-3304] comment out rule and it's usage

MichaelSchmidt82:
Digital health sharing status was first implemented in DC-1836 to check against a participant’s autonomy on whether they would like to share their Fitbit data. However, this will no longer be used to filter Fitbit data.


nishanthpp93 (changes requested):
Thanks @MichaelSchmidt82, looks good but the lint check is failing so you'll need to do a yapf format

nishanthpp93 (approved):
Looks good!

PR #1653, opened 2023-07-13
[DC-3266] Remove PPI/PM src_id in PersonIdValidation cleaning rule

lledezma:


data_steward/cdr_cleaner/cleaning_rules/person_id_validator.py
```diff
@@ -51,10 +51,17 @@
...
     -- should expect mapping tables --
-     and lower(moe.src_hpo_id) <> 'rdr'
+     and lower(moe.src_hpo_id) NOT IN (
+        SELECT 
+            hpo_id
+        FROM
+            `{{project_id}}.pipeline_tables.site_maskings`
+        WHERE NOT
+            REGEXP_CONTAINS(src_id, r'(?i)(PPI/PM)|(EHR site)')
+     )
```
brendagutman, line 64: You might write this differently to avoid the double negative confusion. Suggestions seen below.

```
and lower(moe.src_hpo_id) IN (
        SELECT 
            hpo_id
        FROM
            `{{project_id}}.pipeline_tables.site_maskings`
        WHERE
            REGEXP_CONTAINS(src_id, r'(?i)(EHR site)')
```

Below is another simplification. But either would be fine.

```
and  REGEXP_CONTAINS(moe.src_hpo_id, r'(?i)(EHR site)'
```

tests/integration_tests/data_steward/cdr_cleaner/cleaning_rules/person_id_validator_test.py
```diff
@@ -78,14 +76,14 @@ def setUp(self):
...
+                (101, 'Portal 2', null),
                 (102, 'site bar', null),
-                (201, 'PPI/PM', null),
-                (202, 'PPI/PM', null),
+                (201, 'Portal 3', null),
+                (202, 'Portal 4', null),
                 (203, 'site baz', null),
                 (301, 'site raz', null),
-                (302, 'PPI/PM', null)
+                (302, 'Portal 5', null)
```
brendagutman, line 86: Does the test pick up the sites properly? They might need to be renamed to `EHR site ...`

brendagutman (commented):
Let me know what you think about the comments. Thanks

data_steward/cdr_cleaner/cleaning_rules/person_id_validator.py
brendagutman: Thanks for meeting with me Luis. I see I had a misunderstanding when making the suggestions. Your changes should work as expected I agree.

tests/integration_tests/data_steward/cdr_cleaner/cleaning_rules/person_id_validator_test.py
brendagutman: Thanks for explaining that the query will pick up only things that are not ~'portal'. The other values will be picked up automatically and don't need to be edited for the test.

brendagutman (approved):
Looks good Luis! Thanks for meeting with me!

PR #1654, opened 2023-07-13
[DC-3328] Write the integration test for store_new_measurement...

brendagutman:
...concept_ids

### DIFF
--- a/data_steward/cdr_cleaner/cleaning_rules/store_new_duplicate_measurement_concept_ids.py
+++ b/data_steward/cdr_cleaner/cleaning_rules/store_new_duplicate_measurement_concept_ids.py
@@ -140,12 +140,16 @@ INSERT INTO
     vac_name,
     vac_vocab,
     aou_standard_vac,
+    n_measurement_concept_id,
+    n_measurement,
     date_added)
 SELECT
   value_as_concept_id,
   vac_name,
   vac_vocab,
   aou_standard_vac,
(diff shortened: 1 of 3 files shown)


data_steward/cdr_cleaner/cleaning_rules/store_new_duplicate_measurement_concept_ids.py
```diff
@@ -102,7 +102,7 @@
       COALESCE(CASE
           WHEN dupes.concept_vocab_in_table NOT IN ('LOINC', 'NAACCR') THEN
            ( SELECT MAX(value_as_concept_id) FROM and_cn_dupes AS t2 WHERE 
-           t2.concept_name_in_table = dupes.concept_name_in_table AND t2.concept_vocab_in_table = 'LOINC' )
+           lower(t2.concept_name_in_table) = lower(dupes.concept_name_in_table) AND t2.concept_vocab_in_table = 'LOINC' )
```
brendagutman, line 105: @ksdkalluri I had to lower these as well to have the test pass.

Without this change 100003 was inserted into the table but no change to 'aou_standard_vac'. 
Result was (100003, 'negative', 'SNOMED', 100003) whereas it should be (100003, 'negative', 'SNOMED', 100001)

 Does this change make sense to you?

ksdkalluri (approved):
Thanks for covering all the cases Brenda!

PR #1657, opened 2023-07-17
[DC-3338] Update the 'not mapping to concept_id` check 

brendagutman:
In raw_rdr_export_qc.py


ksdkalluri (approved):
Thanks for updating the existing check as well Brenda!

PR #1661, opened 2023-07-18
[DC-3352] Fix bug EhrSubmissionDataCutoff.affected_tables is unset

mark-velez:
Note: Uninitialized affected_tables causes reporter to fail with error getting sandbox table names [DC-3351]

### DIFF
--- a/data_steward/cdr_cleaner/cleaning_rules/ehr_submission_data_cutoff.py
+++ b/data_steward/cdr_cleaner/cleaning_rules/ehr_submission_data_cutoff.py
@@ -39,6 +39,28 @@ SELECT * FROM `{{project_id}}.{{sandbox_id}}.{{intermediary_table}}`
 """)
 
 
+def get_affected_tables():
+    """
+    This method gets all the tables that are affected by this cleaning rule which are all the CDM tables
+        except for the person table. The birth date field in the person table will be cleaned in another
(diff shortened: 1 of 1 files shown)


data_steward/cdr_cleaner/cleaning_rules/ehr_submission_data_cutoff.py
```diff
@@ -74,28 +96,8 @@ def __init__(self,
...
-            # skips the person table
-            if table == 'person':
-                continue
-
-            # appends all CDM tables except for the person table
-            else:
-                tables.append(table)
-        return tables
+                         sandbox_dataset_id=sandbox_dataset_id,
+                         affected_tables=get_affected_tables())
```
hiro-mishima, line 100: This works. We can also change this line to 
`affected_tables=[table for table in AOU_REQUIRED if table != 'person']`
and remove `get_affected_tables()` altogether.

If we do that, patching for the test will be similar to this one: https://github.com/all-of-us/curation/blob/develop/tests/integration_tests/data_steward/cdr_cleaner/cleaning_rules/remove_records_with_wrong_date_test.py#L56

hiro-mishima (approved):
Thanks, @mark-velez! It looks good. I left another solution in the comment. Both ways exist in our code base, so this will be just a matter of preference. This PR is good to close.
Thanks :tada:

PR #1663, opened 2023-07-18
[DC-3324] Update `NegativeAges` to handle aou_death's NULL death_date

hiro-mishima:
- The cleaning rule can handle the null `death_date` without an update.
- I updated the integration test to ensure that.


lledezma (approved):
Looks good!

PR #1666, opened 2023-07-19
[DC-3318] Update `CalculatePrimaryDeathRecord` to avoid NULL death_date records to be "primary"

hiro-mishima:


MichaelSchmidt82 (approved):
The test looks good.

PR #1676, opened 2023-07-24
[DC-3340] Add QC checks for WEAR study table

brendagutman:


data_steward/analytics/cdr_ops/controlled_tier_qc/check_controlled_tier_part2.py
```diff
@@ -1225,6 +1225,105 @@ def query_template(table_era):
...
+UNION ALL
+
+SELECT
+  'incorrect_start_date' as issue,
+  COUNT(research_id) as bad_rows
+FROM `{{project_id}}.{{rt_cdr_deid}}.wear_study` ws
+LEFT JOIN ( -- all wear study consent records from observation -- 
+  SELECT * 
+  FROM `{{project_id}}.{{rt_cdr_deid}}.observation`
+  WHERE LOWER(observation_source_value) = 'resultsconsent_wear') o
```
nishanthpp93, line 1303: Not a concern, but a comment: This data may not be available if we run the WEAR suppression before this check, so we need to ensure it runs in order.

nishanthpp93 (commented):
Added some comments, let me know your thoughts

data_steward/analytics/cdr_ops/controlled_tier_qc/check_controlled_tier_part2.py
brendagutman: Is it not a concern? I see we have the suppression occurring at the end of the rt and ct deid cleaning classes. Like you said the data should not be in observation at the time this runs.

Should the suppression occur in the base cleaning classes? Or if the suppression location is fine I could join to the combined dataset through the pid rid table instead?

I also see that the suppression CR was placed into the ct fitbit cleaning classes. Most likely on accident. I'll need to remove that.

data_steward/analytics/cdr_ops/controlled_tier_qc/check_controlled_tier_part2.py
```diff
@@ -1225,6 +1225,105 @@ def query_template(table_era):
...
+
+# # Q17 Wear study table
+#
+# DC-3340
+#
+# This check confirms that the wear_s.tudy table contains:
+# 1. Only one row per participant
+# 2. Wear study participants are also found in the CDR person table.
+# 3. Wear study participants have primary consent records in observation.
+# 4. Wear study start dates coincide with records in observation. Date shifted in RT.
```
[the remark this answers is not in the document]
brendagutman: I like the idea of moving this check. It would be cleaner for sure. It would not rely on joining back to the obs table where the data might not exist.

data_steward/analytics/cdr_ops/controlled_tier_qc/check_controlled_tier_part2.py
```diff
@@ -1225,6 +1225,89 @@ def query_template(table_era):
...
+            'result': 'PASS'
+        },
+        ignore_index=True)
+else:
+    df = df.append(
+        {
+            'query': 'Query17 wear_study table is not as expected. See notes in the description.',
+            'result': 'Failure'
+        },
+        ignore_index=True)
```
brendagutman, line 1310: From this file I removed the date shift query, exchanged research_id with person_id, edited to use ct_dataset.

data_steward/analytics/cdr_ops/controlled_tier_qc/check_controlled_tier_part2.py
nishanthpp93: Thanks, I see you moved this check

nishanthpp93 (approved):
Looks good!

PR #1677, opened 2023-07-25
[DC-3239] Add CR to calculate BMI

hiro-mishima:
NOTE:
- Queenie confirmed `measurement_type_concept_id` must be `32865`. See the ticket's comment section (DC2791)

### DIFF
--- a/data_steward/cdr_cleaner/cleaning_rules/calculate_bmi.py
+++ b/data_steward/cdr_cleaner/cleaning_rules/calculate_bmi.py
@@ -1,0 +1,187 @@
+"""
+Some participants self-report their height and weight. This CR calculates the BMI for
+those self-reported measurements. (Note BMIs are already calculated if the data is from HPO sites.)
+
+Each participant can self-report their height and weight multiple times. BMIs are calculated
+for each self-reported hight/weight pair. BMIs are not calculated if either height or weight is missing.
+
+Original Issue: DC-3239
+"""
+# Python imports
+import logging
+
+# Project imports
+from cdr_cleaner.cleaning_rules.base_cleaning_rule import BaseCleaningRule
+from cdr_cleaner.cleaning_rules.clean_height_weight import CleanHeightAndWeight
+from common import JINJA_ENV, MEASUREMENT
+from constants.cdr_cleaner.clean_cdr import (CONTROLLED_TIER_DEID_CLEAN, QUERY,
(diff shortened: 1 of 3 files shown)


data_steward/cdr_cleaner/cleaning_rules/calculate_bmi.py
```diff
@@ -0,0 +1,187 @@
...
+    NULL AS provider_id,
+    h.visit_occurrence_id,
+    NULL AS visit_detail_id,
+    CAST(NULL AS STRING) AS measurement_source_value,
+    NULL AS measurement_source_concept_id,
+    'kg/m2' AS unit_source_value,
+    CAST(NULL AS STRING) AS value_source_value
+FROM self_reported_height h
+JOIN self_reported_weight w
+ON h.person_id = w.person_id AND h.visit_occurrence_id = w.visit_occurrence_id
```
ksdkalluri, line 79: @hiro-mishima why don't you add probably measurment_date as a condition to match here? It is not required when we are matching on visit_occurrence_id but I believe having measurement_date will add an extra layer of conformance in case there is any issue with the visit_occurrence_ids that may be duplicated or what not.

hiro-mishima: Thanks! I added `measurement_date` as a JOIN condition. I also added `src_id` to the condition just to be safer.

tests/integration_tests/data_steward/cdr_cleaner/cleaning_rules/calculate_bmi_test.py
```diff
@@ -0,0 +1,190 @@
...
+                (32, 'Participant Portal: PTSC'),
+                (33, 'Participant Portal: CE'),
+                (34, 'Participant Portal: CE'),
+                (39, 'Participant Portal: PTSC'),
+                (41, 'Participant Portal: PTSC'),
+                (42, 'Participant Portal: PTSC'),
+                (43, 'Participant Portal: CE'),
+                (44, 'Participant Portal: CE'),
+                (49, 'Participant Portal: PTSC'),
+                (50, 'aou_drc'),
```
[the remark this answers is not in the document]
hiro-mishima: Updated!

ksdkalluri (approved):
Looks Good. Thanks!

PR #1679, opened 2023-07-26
[DC-3356] Add a check to QC `combined.py` to verify src_ids in the ext tables

lledezma:


hiro-mishima (approved):
This QC looks great, thanks! 🎉

PR #1686, opened 2023-07-31
[DC-3349] Update the key rotation service to rotate keys more frequently than 180 days

ratuagga:
Updated expiration and alert days for service account keys.


nishanthpp93 (changes requested):
Looks good, but the ticket doesn't mention changing the alert days from 7 to 16, so I would suggest keeping it the same.

nishanthpp93 (approved):
Looks good!

PR #1688, opened 2023-08-02
[DC-3373] Add the new consent_validation check

brendagutman:


data_steward/analytics/cdr_ops/raw_rdr_export_qc.py
```diff
@@ -955,5 +955,57 @@
...
+# Count of participants with multiple validation status' for their ehr consent records.
+query = JINJA_ENV.from_string("""
+
+WITH cte AS (
+SELECT 
+"Have multiple status. These are not duplicates" as issue,
+COUNT(*) AS n
+FROM (SELECT DISTINCT * FROM `{{project_id}}.{{new_rdr}}.consent` )
+GROUP BY person_id
+HAVING COUNT(*)>1
```
nishanthpp93, line 978:
```suggestion
HAVING n>1
```

data_steward/analytics/cdr_ops/raw_rdr_export_qc.py
```diff
@@ -955,5 +955,57 @@
...
+
+# +
+# Count of participants with multiple validation status' for their ehr consent records.
+query = JINJA_ENV.from_string("""
+
+WITH cte AS (
+SELECT 
+"Have multiple status. These are not duplicates" as issue,
+COUNT(*) AS n
+FROM (SELECT DISTINCT * FROM `{{project_id}}.{{new_rdr}}.consent` )
```
nishanthpp93, line 976: Couple changes:
1. consent_validation will be the table name
2. Distinct * will include timestamp as well, and I think there is merit in separating them here (maybe another check?)
```suggestion
FROM (SELECT DISTINCT * EXCEPT (consent_for_electronic_health_records) FROM `{{project_id}}.{{new_rdr}}.consent_validation` )
```

nishanthpp93 (commented):
Overall looks good! Added some comments, let me know your thoughts

data_steward/analytics/cdr_ops/raw_rdr_export_qc.py
brendagutman: I agree, thanks for thinking of that!

data_steward/analytics/cdr_ops/raw_rdr_export_qc.py
```diff
@@ -955,5 +955,58 @@
...
+FROM (SELECT DISTINCT * EXCEPT (consent_for_electronic_health_records_authored) FROM `{{project_id}}.{{new_rdr}}.consent_validation` )
+GROUP BY person_id
+HAVING n>1
+
+UNION ALL 
+
+SELECT 
+"Have multiple status. These are duplicates" AS issue,
+COUNT(*) AS n
+FROM `{{project_id}}.{{new_rdr}}.consent_validation`
```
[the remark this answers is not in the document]
brendagutman: I split the query to give a view of 
  1. the duplicates that would most likely cause participant data to be dropped vs 
  2. those that are duplicates but they shouldn't change the participants status and affect their data being dropped.
 
Because of the way the suppression CR works, I don't think that the second occurrence of duplicates are as important as the first and would not necessarily need to be a priority for rdr. The same would be true for those who have multiples of the same status but with different dates.

In my opinion I don't see those with different times as true duplicates and would be wary to have rdr drop them not knowing more about how they were created.  If anything I could add another issue type and include these. Either way, I'm not sure that they would need to go back to rdr though as they should not affect participant data. I'd almost lean toward removing the second part of the query since the issue isn't really posing a real problem for us and this table isn't going to be seen by the researchers.

Let me know what you think.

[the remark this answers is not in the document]
nishanthpp93: This would be a great opportunity to flag it to RDR, not necessarily asking them to drop rows. The requirements state that we are supposed to get one row for each participant. It is primarily for awareness.
However, the requirements make sense only if a participant has a single consent response, which is not always the case.
In case of multiple consent responses, we need to know which responses were validated and which ones were not. I would also suggest comparing these validation rows to actual consent responses in observation using timestamps to ensure there is a 1-1 mapping. If not, such validation records for non-existent responses should be flagged. 
If we get multiple responses and all of them are validated, we need to know which one to follow. There may be cases where the responses match, and cases where they don't. Typically we use the latest response.
Now, duplicates are a concern at this stage. If the same consent response has different validation records, it would be seen in the 1-1 mapping above as non-conformant. These need to be highlighted to them regardless of the response value itself.
These can further be stratified based on data impact like you mentioned, but not necessary. This is just the validation mechanism we are addressing and shouldn't depend on the value of the response itself. Any fix should address it regardless of the response value.

nishanthpp93 (approved):
Looks good!

PR #1690, opened 2023-08-02
[DC-3375] Update check in QC `fitbit_qc.py` that verifies src_ids in Fitbit tables

lledezma:


brendagutman (approved):
Looks good Luis. Thanks for fixing this!

PR #1692, opened 2023-08-03
[DC-3277] Update site_masking query to handle HPO sites with NULL state

hiro-mishima:


data_steward/tools/add_hpo.py
```diff
@@ -53,13 +53,13 @@
    SELECT 
      "{{hpo_id}}" AS hpo_id,
      CONCAT('EHR site ', new_id) AS src_id,
-     "{{us_state}}" AS state,
-     {{value_source_concept_id}} AS value_source_concept_id
+     {% if us_state %} "{{us_state}}" {% else %} CAST(NULL AS STRING) {% endif %} AS state,
```
nishanthpp93, line 56: Thoughts on empty string field?
```suggestion
     {% if us_state %} "{{us_state}}" {% else %} '' {% endif %} AS state,
```

nishanthpp93 (commented):
Looks good, but added a small comment. Let me know your thoughts!

data_steward/tools/add_hpo.py
hiro-mishima: @nishanthpp93 

Thanks for the comment!

`''` can work. I chose `CAST(NULL AS STRING)` here because the existing records in `{{site_maskings_table}}` have NULL values for that column. (See the values for RDR or VA site in the table, for example)

If we choose `''` here as the solution, I will also manually update those records in the table from null to empty strings. As far as I see, nothing will break even if we change the table's null values to empty strings. I do not have a strong preference for null vs empty strings here. Do you have a preference?

nishanthpp93: No, either is fine, but be sure to check that casting works as intended, sometimes it can fail.

nishanthpp93 (approved):
Looks good!

PR #1698, opened 2023-08-11
[DC-3380] Update `DropMultipleMeasurements` to handle self-reported PM

hiro-mishima:
- Deleted the unit test since the integration test can cover it.
- Renamed the integration test to be more appropriate
- Cleaned up the description and the numbering of the integration test/ test cases.


brendagutman (approved):
The changes look good, thanks Hiro!

PR #1701, opened 2023-08-15
[DC-3379] Update `DropExtremeMeasurements` to handle self-reported PM

hiro-mishima:
- Updated the CR's SQL to process in-person vs self-reported PM.
- Cleaned up the main script for readability
- Reorganized the test cases to represent self-report vs in-person PM


brendagutman (approved):
Thanks for going above and beyond by refactoring this CR and test. They are now very straight forward, and much easier to understand!

PR #1705, opened 2023-08-21
[DC-2707] Correct typo (`cocnept` -> `concept`)

hiro-mishima:


nishanthpp93 (approved):
Looks good! Straightforward and I only found 6 instances

PR #1706, opened 2023-08-21
[DC-3235] Add `survey_conduct` to `NoDataAfterDeath` cleaning rule

hiro-mishima:
- Also updated `table_namer` so it can be referenced correctly
- Also updated `depends_on` to the right value
- Technically, `survey_end_date` is `nullable`, but the column is never `NULL` as defined in the PRD for `survey_conduct`. https://github.com/all-of-us/curation/blob/develop/data_steward/resource_files/schemas/cdm/clinical/survey_conduct.json#L36


lledezma (approved):
Looks good!

PR #1707, opened 2023-08-22
[DC-3230] Avoid WRITE_TRUNCATE

hiro-mishima:


data_steward/cdr_cleaner/cleaning_rules/null_invalid_foreign_keys.py
```diff
@@ -14,43 +14,48 @@
...
+            )
+            AND {{key}} IS NOT NULL
+        ){% if not loop.last -%} OR {% endif %}
+    {% endfor %}
+)""")
+
+DELETE_QUERY = JINJA_ENV.from_string("""
+DELETE FROM `{{project_id}}.{{dataset_id}}.{{table_name}}`
+WHERE {{key}} NOT IN (
+    SELECT {{key}} FROM `{{project_id}}.{{dataset_id}}.{{key[:-3]}}`
```
brendagutman, line 50: I see we are only sandboxing where the key is not null. This would delete NULL rows as well. Is this expected?

brendagutman (approved):
Thanks for upgrading this CR and test @hiro-mishima!

data_steward/cdr_cleaner/cleaning_rules/null_invalid_foreign_keys.py
hiro-mishima: Thanks for reviewing it!

This `DELETE` statement runs against `required` foreign key columns (=`person_id` for all and `visit_occurrence_id` for `visit_detail`). Since they are not nullable, we cannot update the invalid values to `NULL`, so we delete the whole row here. For `nullable` foreign key columns, we update the invalid values to `NULL`.

I added additional comments in the code. Let me know if it makes sense!

brendagutman: Yes, I see what we are doing now, thanks for pointing that out and adding the comments! 

Very clear explanation as well. Thank you

brendagutman (approved):
Thanks Hiro!

PR #1708, opened 2023-08-22
[DC-3408] Add column names

hiro-mishima:


lledezma (changes requested):
Your update will fix the error.
https://stackoverflow.com/questions/62616853/unable-to-load-data-into-ingestion-time-partitioned-table
I would just update `issue_numbers` list to include the ticket number and `Original Issues` in the docstring.

hiro-mishima:
Thanks for reviewing! I updated `issue_numbers`. I did not update `Original Issues` because the origin of the issue is DC-3239 and I do not think it is necessary to keep updating that part.

lledezma (approved):
Looks good!

PR #1712, opened 2023-08-25
[DC-3383] Update `import_rdr_dataset.py` to import death data into `aou_death`

hiro-mishima:
- While testing, I encountered the error that says the label can be only up to 63 letters. I updated this line as a result: https://github.com/all-of-us/curation/pull/1712/files#diff-6ecbf63270b1fe0210a08915e522c72d8d164fdb10460211699a92511fb6e929R812


ratuagga (approved):

ratuagga:
Looks good

PR #1713, opened 2023-08-25
[DC-3413] Update `raw_rdr_export_qc.py` and `clean_rdr_export_qc.py` to run QC on `aou_death`, not `death`

hiro-mishima:
- Pylint also updated some formats where it is not directly related.


ratuagga (approved):
Looks Good

PR #1714, opened 2023-08-25
[DC-3414] Add `CalculatePrimaryDeathRecord` to RDR data stage

hiro-mishima:


ratuagga:
Change looks ok to me. @ksdkalluri might know best where to put the rule in the run sequence.

ksdkalluri:
> Change looks ok to me. @ksdkalluri might know best where to put the rule in the run sequence.

This CR will not have dependency I believe in this stage it can be placed anywhere before clean_mapping. If other changes look okay this can be merged.

ratuagga (approved):

PR #1715, opened 2023-08-28
[DC-3381] Update `create_rdr_snapshot.py` to create an empty death table in RDR clean

hiro-mishima:


brendagutman (approved):
Thanks for the changes Hiro!

PR #1716, opened 2023-08-28
[DC-3412] Update `create_combined_backup_dataset.py` to expect RDR's `aou_death` not `death`

hiro-mishima:


brendagutman (approved):
Thanks for the quick turnaround on this!

PR #1718, opened 2023-08-29
[DC-3371] Prep the RDR export

ratuagga:
Updates in Raw RDR stage.


data_steward/tools/create_rdr_snapshot.py
```diff
@@ -125,8 +125,10 @@ def main(raw_args=None):
     ]
     skip_tables = [
         AOU_DEATH, COPE_SURVEY_MAP, PID_RID_MAPPING,
-        QUESTIONNAIRE_RESPONSE_ADDITIONAL_INFO
-    ]
+        QUESTIONNAIRE_RESPONSE_ADDITIONAL_INFO, 'consent_validation',
+        'wear_consent'
```
brendagutman, line 129: Suggestion. Add `wear_consent` to common.py (line 125?) and use common.EHR_CONSENT_VALIDATION for `consent_validation`

data_steward/resource_files/schemas/rdr_src_id/tables/rdr_specimen.json
```diff
@@ -0,0 +1,98 @@
...
+    "mode": "nullable",
+    "description": "The information about the disease status as detailed in the source."
+  },
+  {
+    "type": "string",
+    "name": "src_id",
+    "mode": "required",
+    "description": "The source of the record."
+  }
+]
```
brendagutman, line 98: I know we have requested these tables from RDR, is there any reason to believe we would not receive these tables or that they won't contain src_id?

ratuagga: If we do not receive these tables, they have to be created by our Import RDR script. And at raw rdar stage, I believe it should have src_id.

data_steward/tools/create_rdr_snapshot.py
ratuagga: sure

data_steward/tools/import_rdr_dataset.py
```diff
@@ -165,7 +168,12 @@ def create_rdr_tables(client, destination_dataset, rdr_project,
             fields_name_str = ',\n'.join(sc_list)
 
             # copy contents from source dataset to destination dataset
-            sql = (f'SELECT {fields_name_str} ' f'FROM `{source_table_id}`')
+            if table == 'cope_survey_semantic_version_map':
+                sql = (f'SELECT {fields_name_str} '
+                       f'FROM `{source_table_id}` '
+                       f'where participant_id is not Null')
```
brendagutman, line 174: Suggestion: WHERE participant_id IS NOT NULL

ksdkalluri (approved):
Looks good to merge!

PR #1719, opened 2023-08-29
[DC-3401] Use `BigQueryClient.create_tables()` instead of `CREATE TABLE` for `aou_death` creation

hiro-mishima:
I used different `create_table()` for EHR union and Combined.
- For EHR union, I used `bq_utils.create_standard_table()`, following other CDM tables creation in EHR union. I had to use it because (1) the destination table name needs the prefix `unioned_ehr_`, and (2) this process runs every night and we need the `drop_existing=True` capability when it runs. `bq_utils.create_standard_table()` creates a partitioned table as a result, and it is not a problem since all the other CDM tables are partitioned at this stage. 
- For Combined, I used `client.create_tables()` as it does not require special consideration like EHR union.

For testing, the existing integration tests can cover these updates. No update was needed.

### DIFF
--- a/data_steward/validation/ehr_union.py
+++ b/data_steward/validation/ehr_union.py
@@ -888,10 +888,16 @@ def create_load_aou_death(bq_client, project_id, input_dataset_id,
         `CalculatePrimaryDeathRecord` updates the table at the end of the
         Unioned EHR data tier creation.
     """
-    bq_utils.create_standard_table(AOU_DEATH,
-                                   f'{UNIONED_EHR}_{AOU_DEATH}',
-                                   drop_existing=True,
-                                   dataset_id=output_dataset_id)
+    # EHR Union runs every night so the table needs to be deleted first if exists.
+    bq_client.delete_table(f'{output_dataset_id}.{UNIONED_EHR}_{AOU_DEATH}',
+                           not_found_ok=True)
+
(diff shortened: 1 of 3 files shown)


data_steward/validation/ehr_union.py
```diff
@@ -877,6 +888,11 @@ def create_load_aou_death(bq_client, project_id, input_dataset_id,
         `CalculatePrimaryDeathRecord` updates the table at the end of the
         Unioned EHR data tier creation.
     """
+    bq_utils.create_standard_table(AOU_DEATH,
+                                   f'{UNIONED_EHR}_{AOU_DEATH}',
+                                   drop_existing=True,
+                                   dataset_id=output_dataset_id)
+
```
lledezma, line 895: bq_utils.py is in the process of being deprecated, so `create_standard_table(...)` function will be deprecated eventually. I left a suggestion to see what you think. 

```suggestion
table_name = f'{bq_client.project_id}.{output_dataset_id}.{UNIONED_EHR}_{AOU_DEATH}'
schema_list = bq_client.get_table_schema(AOU_DEATH)
table_obj = bq.Table(table_name, schema=schema_list)
table_obj.clustering_fields = 'person_id'
table_obj.time_partitioning = bq.table.TimePartitioning(type_='DAY')
bq_client.create_table(table_obj, exists_ok=True)
```

lledezma (changes requested):

lledezma (approved):
Looks good. Thanks @hiro-mishima!

PR #1720, opened 2023-08-29
[DC-3407] Update value_as_string check in clean_rdr notebooks to remove false positives

brendagutman:


hiro-mishima (commented):
Thanks, @brendagutman. I have two questions/comments:
1. In what situation will we need to update the list `expected_strings`? Do we need to update it for each CDR release? It might be helpful to add a note that when we should revisit the list.
2. Will it be helpful if the list is in alphabetical order, or some other order? The list will be more consistent/comprehensive if we add `order by` in the SQL statement in the JIRA ticket DC-3407.

brendagutman:
> Thanks, @brendagutman. I have two questions/comments:
> 
> 1. In what situation will we need to update the list `expected_strings`? Do we need to update it for each CDR release? It might be helpful to add a note that when we should revisit the list.
> 2. Will it be helpful if the list is in alphabetical order, or some other order? The list will be more consistent/comprehensive if we add `order by` in the SQL statement in the JIRA ticket DC-3407.

Hey thanks for the review @hiro-mishima. 
1.  The list helps narrow down the list of possible issues and would not necessarily need to be updated on a schedule. The list is created using the current surveys and updating the list would depend on the types/format of questions in future surveys. If a new survey is implemented and there are no text answers resulting in a number in 'value_as_string' this check would not have any false results due to expected text answers. I think that it is okay to wait for 'issues' to arise, investigate them and if they are expected then to add them to the list. Again, this depends on future surveys. I tried to hint at the possible need to update the list in the check description.  Let me know if that needs to be more clear or if you think something should be added. 

2. Good point, I think ordering by 'form_name' and 'question_code' would be the best. This way any concepts added to the list would most likely still be ordered.

hiro-mishima (approved):
Thanks, @brendagutman! It looks good 🎉

PR #1721, opened 2023-08-30
[DC-3397] Refactor uses of `bq_utils.py` `get_dataset_id(...)` function in data_steward/validation

lledezma:


hiro-mishima (approved):
Looks great! Thanks :)

PR #1722, opened 2023-08-30
[DC-3426] Refactor uses of `bq_utils.py` `get_rdr_project_id` function

lledezma:


hiro-mishima (approved):
This one looks good too!

PR #1728, opened 2023-09-13
[DC-3409] Update site_masking CSV generation

hiro-mishima:


data_steward/tools/add_hpo.py
```diff
@@ -244,8 +246,9 @@ def add_src_hpos_allowed_state_csv(hpo_id, us_state, value_source_concept_id,
     hpo_file_df = add_src_hpos_allowed_state_file_df(
         hpo_id, us_state, value_source_concept_id, src_hpos_allowed_state_path)
     hpo_file_df.to_csv(src_hpos_allowed_state_path,
-                       quoting=csv.QUOTE_ALL,
-                       index=False)
+                       quoting=csv.QUOTE_NONE,
+                       index=False,
+                       float_format=lambda x: '%d' % x)
```
nishanthpp93, line 251: Not sure why we are seeing floats in this file.

nishanthpp93 (commented):
I need some more context on this. We should not be seeing floats here. Also, I would suggest a PR in devops as well which updates the file itself with this new format.

data_steward/tools/add_hpo.py
hiro-mishima: This function calls `bq_utils.get_hpo_site_state_info()` and the [line 980](https://github.com/all-of-us/curation/blob/develop/data_steward/bq_utils.py#L980) loads `value_source_concept_id` as numeric value. And it looks like `to_csv` adds `.0` to the numeric output even when the value is integer.

The whole logic here was implemented by Ratul 5 months ago. And these "care evolution xyz" are the first time we run `add_hpo.py` after it was implemented.

nishanthpp93: Ah okay, thanks for the clarification!

PR #1729, opened 2023-09-18
[DC-3438] Add a check in controlled_tier_qc to verify the race/ethnicity subcategory expansion

MichaelSchmidt82:
## Description

As we remove the suppression of race/ethnicity subcategories, we need to add a check in our QC notebooks to verify if the subcategories are adequately represented in the observation table.

## Scope:

- Add a check in controlled_tier_qc  to verify that race/ethnicity subcategories exist in the observation table.
- The added check should summarise the existence of the subcategories, not display everything.

We can use the logic in the [cleaning rule](https://precisionmedicineinitiative.atlassian.net/browse/DC-3436) that will be dropped to identify the race/ethnicity records.

Acceptance Criteria:

A check is added to controlled_tier_qc


ksdkalluri (approved):
Looks good, can be merged once the integration test passes.

PR #1731, opened 2023-09-20
[DC-3458] Update `AbstractBackfillSurveyRecords` to create mapping records for the new records

brendagutman:


data_steward/cdr_cleaner/cleaning_rules/backfill_survey_records.py
```diff
@@ -170,19 +190,54 @@ def setup_rule(self, client):
...
+            QUERY:
+                SANDBOX_QUERY.render(
+                    project=self.project_id,
+                    dataset=self.dataset_id,
+                    sandbox_dataset=self.sandbox_dataset_id,
+                    sandbox_table=self.sandbox_table_for(OBSERVATION),
+                    obs=OBSERVATION,
+                    mapping_obs=MAPPING_PREFIX + OBSERVATION,
+                    pers=PERSON,
+                    rdr_obs_fields=', '.join(
```
hiro-mishima, line 211: It looks like `rdr_obs_fields` is not referenced so we can remove it.

hiro-mishima (approved):
Thanks for the updates! I like how the sandbox is used to create new records in both observation and mapping tables. I left one minor suggestion.

I did not move the ticket to "in review" in case @nishanthpp93 should review it.

PR #1732, opened 2023-09-20
[DC-3445] Add `raw_fitbit_qc.py` for `steps_intaday`

hiro-mishima:


data_steward/analytics/cdr_ops/raw_fitbit_qc.py
```diff
@@ -0,0 +1,66 @@
...
+impersonation_creds = auth.get_impersonation_credentials(
+    run_as, target_scopes=IMPERSONATION_SCOPES)
+
+client = BigQueryClient(project_id, credentials=impersonation_creds)
+
+# # STEPS_INTRADAY table
+
+# Validation criteria for steps_intraday is the following:
+# - Field names and data types match the approved schema's
+# - The table includes both PTSC and CE data per the src_id field
```
lledezma, line 44: I don't think the first check applies anymore.

lledezma (commented):
The query works and I like the check status message, but check does not display any results or status.

hiro-mishima:
Thanks for reviewing, @lledezma! I just pushed the updated version. Let me know if it suffices.

lledezma (approved):
Thanks for the updates Hiro! You could provide an alias for the src_id count if you think I would help the viewer, but you don't have to.

PR #1737, opened 2023-09-25
[DC-3459] Create `heart_rate_intraday.json` following `heart_rate_minute_level.json`

hiro-mishima:


data_steward/resource_files/schemas/wearables/fitbit/heart_rate_intraday.json
```diff
@@ -0,0 +1,26 @@
...
+    "type": "datetime",
+    "name": "datetime",
+    "mode": "nullable",
+    "description": "Timestamp for the intraday heart rate data."
+  },
+  {
+    "type": "integer",
+    "name": "heart_rate_value",
+    "mode": "nullable",
+    "description": "Heart rate value on a minute level."
```
brendagutman, line 12: This would probably need to change to somthing like 
'Heart rate value at minute level for src_id = PTSC and second level for src_id = CE'. I'll pose the question in the [doc](https://docs.google.com/spreadsheets/d/1IS4WgSZy8VeJjshNCYy9GpXxXI59pdVBI6V47zEmHEA/edit#gid=0).

hiro-mishima: You are right, @brendagutman. I see Hiral mentioned that in the DC-3359's comment section. Let me update the description.

brendagutman: There is still some discussion on this. We remap src_ids so we need to put the remapped src_id in the description. This will likely mean other descriptions that mention ptsc/vibrent/ce would need to be updated in another ticket.

hiro-mishima: I updated the description so that we do not have to write the `src_id` values explicitly. Let me know if it suffices.

brendagutman (approved):
Looks good thanks Hiro!

PR #1749, opened 2023-09-27
[DC-3442] Add cleaning rule to remove participants using a list by extending sandbox_and_remove_pids

lledezma:


nishanthpp93 (approved):
Looks good!

PR #1753, opened 2023-10-03
[DC-3455] Update Unioned EHR process to have mapping tables as expected

hiro-mishima:
Dear reviewer,
- See the comment section of the JIRA ticket DC-3455 for findings from the investigation.
- `generate_unioned_ehr_dataset.sh` ... I removed `--sync false` from the command that copies mapping tables. `--sync false` is still there in the commands before it, but having it in sync mode here will give the preceding commands ample time to finish copying. Making it in sync mode will make the execution longer but the impact is acceptable (~10 minutes).


lledezma (approved):
Looks good!

PR #15, opened 2017-09-22
Upload Instructions Update

cmcphilamy:
In order to meet the Oct 5th milestone the directions have been updated with naming convention information and better upload instructions.

PR #1755, opened 2023-10-03
[DC-3492] Modify RT deid notebook checks for ease of use

brendagutman:


lledezma (changes requested):
1. `cdr_deid_qa_report1_generalization_rule.py`
- There are 2 QA checks with the same number.   
   `# # 3 GR_03 Gender Generalization Rule # QUERY NUMBER CONFLICT`
   `# # 3 Biological Sex Generalization Rule # QUERY NUMBER CONFLICT`

2. `cdr_deid_qa_report2_row_suppression.py`
- What is the purpose of the second query in QA checks 5 and 12? Should the results be          appended to the final list of results? 

3. `cdr_deid_qa_report4_dateshift.py`
- Looks like the QA check numbers need to be updated, because there is no QA check #2.
- QA check #15 results table has multiple rows, so line 14 needs to be updated. I would  suggest `if df1['n_row_not_pass'].sum() == 0:`

data_steward/analytics/cdr_ops/rt_cdr_qc/cdr_deid_qa_report2_row_suppression.py
```diff
@@ -76,8 +76,8 @@
 df1
 
 # # 2 Verify all fields identified for suppression in the OBSERVATION table have been removed from the table in the deid dataset.
-
-# ## error in new cdr
+#
+# If there are issues view them in more detail using query 2.1
```
brendagutman, line 80: Updated these descriptions. The second query in several checks would be helpful for any failure investigations.

data_steward/analytics/cdr_ops/rt_cdr_qc/cdr_deid_qa_report4_dateshift.py
```diff
@@ -597,23 +597,21 @@
...
+UNION ALL SELECT * FROM df8 
 
 """)
 q = query.render(project_id=project_id,
                  pipeline=pipeline,
                  com_cdr=com_cdr,
                  deid_cdr=deid_cdr)
 df1 = execute(client, q)
-if df1.eq(0).any().any():
+if df1['n_row_not_pass'].sum() == 0:
```
brendagutman, line 614: Thanks for the suggestion here! Good catch!

As for the query updates. I tend toward not wanting to update them.
To keep the query numbers consistent with previous notebook results. Also it could be a kind of paper trail for removed queries. Because there is no query2 we can assume it was removed. 
Do you agree or think it better have a more clean notebook?

lledezma: It would be helpful to keep them as they are, so we can compare with previous notebooks.

data_steward/analytics/cdr_ops/rt_cdr_qc/cdr_deid_qa_report1_generalization_rule.py
```diff
@@ -306,7 +306,7 @@
         ignore_index=True)
 df1
 
-# # 3 GR_03 Gender Generalization Rule
+# # 3.1 GR_03 Gender Generalization Rule
 #
 # Verify that the field identified for de-identification action in OBSERVATION table follow the Gender Generalization Rule  for the de-id table.
```
lledezma, line 311: Thanks! I would update the dictionary entry in lines 370 and 379.

data_steward/analytics/cdr_ops/rt_cdr_qc/cdr_deid_qa_report1_generalization_rule.py
```diff
@@ -383,7 +383,7 @@
         ignore_index=True)
 df1
 
-# # 3 Biological Sex Generalization Rule
+# # 3.2 Biological Sex Generalization Rule
 #
```
lledezma, line 387: Same here. Lines 431 and 438.

lledezma (changes requested):

lledezma (approved):
Thanks Brenda!

PR #1756, opened 2023-10-03
[DC-3491] Modify fitbit qc notebook checks

brendagutman:


lledezma (approved):
Assuming that the check `Verify all participants have digital health sharing consent` is the only check that is no longer needed.

PR #1758, opened 2023-10-04
[DC-3435] Add check in combined.py to verify no participant in the pdr_ehr_dup_report list have EHR data

MichaelSchmidt82:
## Scope

In the combined.py notebook, it is necessary to add a verification step to ensure that all EHR records are removed for duplicated accounts with conflicting statuses. This check should analyze the data from all CDM/clinical tables. A failure message with a list of these items should be added if any PIDs still exist.

### DIFF
--- a/data_steward/analytics/cdr_ops/combined.py
+++ b/data_steward/analytics/cdr_ops/combined.py
@@ -862,12 +862,20 @@ execute(client, results)
 # ## Verify no participant in the pdr_ehr_dup_report list has EHR data.
 
 unconsented_records_tpl = JINJA_ENV.from_string("""
-SELECT
-  person_id
+SELECT *
 FROM
-  `{{project_id}}.{{dataset_id}}.person`
-WHERE EXISTS (
-    SELECT person_id FROM `{{project_id}}.{{dataset_id}}_sandbox.{{unconsented}}`
+    `{{project}}.{{dataset}}.{{domain_table}}` d
+  JOIN
+    `{{project}}.{{dataset}}.{{mapping_domain_table}}` md
+  USING
+    ({{domain_table}}_id)
+  WHERE
+    person_id IN (
+    SELECT
+      person_id
+    FROM
+      `{{project}}.{{sandbox_dataset}}.{{unconsented_lookup}}`)
+    AND src_dataset_id LIKE '%ehr%'
   )
 """)
 query = unconsented_records_tpl.render(project_id=PROJECT_ID,
--- a/data_steward/analytics/cdr_ops/combined.py
+++ b/data_steward/analytics/cdr_ops/combined.py
@@ -878,12 +878,15 @@ FROM
     AND src_dataset_id LIKE '%ehr%'
   )
 """)
-query = unconsented_records_tpl.render(project=PROJECT_ID,
-                                       dataset=DATASET_ID,
-                                       domain_table='wip',
-                                       mapping_domain_table='wip',
-                                       sandbox_dataset='wip',
-                                       unconsented_lookup=UNCONSENTED)
+
+for table in MAPPED_CLINICAL_DATA_TABLES:
+    query = unconsented_records_tpl.render(
+        project=PROJECT_ID,
+        dataset=DATASET_ID,
+        domain_table=table,
+        mapping_domain_table=f'_mapping_{table}',
+        sandbox_dataset='wip',
+        unconsented_lookup=UNCONSENTED)
 result = execute(client, query)
 print(f"found {len(result)} PIDs'")
 result
--- a/data_steward/analytics/cdr_ops/combined.py
+++ b/data_steward/analytics/cdr_ops/combined.py
@@ -878,6 +878,10 @@ FROM
     AND src_dataset_id LIKE '%ehr%'
 """)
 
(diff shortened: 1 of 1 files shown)


data_steward/analytics/cdr_ops/combined.py
```diff
@@ -857,3 +858,21 @@ def verify_dataset_labels(dataset):
     result.append(query)
 results = '\nUNION ALL\n'.join(result)
 execute(client, results)
+
+# ## Verify no participant in the pdr_ehr_dup_report list has EHR data.
```
ksdkalluri, line 862: To check for EHR data we need to verify that no person in the ehr_duplicate_table has ehr data in any of the clinical tables for eg.. measurement, visit_occurrence, procedure_occurrence etc. You can identify if any ehr records exist by joining a domain table and the mapping/ext table and filtering like below

```
select *
FROM
    `{{project}}.{{dataset}}.{{domain_table}}` d
  JOIN
    `{{project}}.{{dataset}}.{{mapping_domain_table}}` md
  USING
    ({{domain_table}}_id)
  WHERE
    person_id IN (
    SELECT
      person_id
    FROM
      `{{project}}.{{sandbox_dataset}}.{{unconsented_lookup}}`)
    AND src_dataset_id LIKE '%ehr%' 
  )
```

ksdkalluri (changes requested):

data_steward/analytics/cdr_ops/combined.py
```diff
@@ -857,3 +858,36 @@ def verify_dataset_labels(dataset):
...
+        dataset=DATASET_ID,
+        domain_table=table,
+        mapping_domain_table=f'_mapping_{table}',
+        sandbox_dataset=f'{DATASET_ID}_sandbox',
+        unconsented_lookup=UNCONSENTED)
+
+    result = execute(client, query)
+    if result.any():
+        break
+result
```
lrwb-aou, line 893: @MichaelSchmidt82 This fails when you actually run the notebook.  The error is:
```bash
ValueError: The truth value of a Series is ambiguous. Use a.empty, a.bool(), a.item(), a.any() or a.all().
```

Have you looked through the notebook at how to display query results?  There are examples in here of how to indicate success and failure.  See the `render_message()` function and usage examples.

lrwb-aou (changes requested):
See in line comments.  Meet the ticket's acceptance criteria.

data_steward/analytics/cdr_ops/combined.py
```diff
@@ -857,3 +858,42 @@ def verify_dataset_labels(dataset):
...
+    SELECT
+      person_id
+    FROM
+      `{{project}}.{{sandbox_dataset}}.{{unconsented_lookup}}`)
+    AND src_dataset_id LIKE '%ehr%'
+""")
+
+# +
+success_msg_if_empty = "All PID's are consenting PID's"
+failure_msg_if_empty = "Unconsented PID's where FOUND"
```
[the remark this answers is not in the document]
MichaelSchmidt82: I will update this and provide a full description

data_steward/analytics/cdr_ops/combined.py
```diff
@@ -857,3 +858,44 @@ def verify_dataset_labels(dataset):
...
+    query = unconsented_records_tpl.render(
+        project=PROJECT_ID,
+        dataset=DATASET_ID,
+        domain_table=table,
+        mapping_domain_table=f'_mapping_{table}',
+        sandbox_dataset=f'{DATASET_ID}_sandbox',
+        unconsented_lookup=UNCONSENTED)
+
+    result = execute(client, query)
+    totals.append(result)
```
lrwb-aou, line 896: @MichaelSchmidt82 This is much better.  If a failure happens, the error data is readable.  Whether the test passes or fails is easily discernible from the status check.  The one thing I am going to point out is the execution cycle. 

 As written, this will send, execute, and wait for the results for one query on every loop iteration.  This forces you to aggregate the results in a data frame for eventual usage in the `render_message` function.  I mentioned in our slack conversation that you could do this entire check with one SQL query.  It would take a minor tweak of this code to create a single query instead of multiple queries.  Try to make this improvement.

data_steward/analytics/cdr_ops/combined.py
```diff
@@ -857,3 +858,44 @@ def verify_dataset_labels(dataset):
     result.append(query)
 results = '\nUNION ALL\n'.join(result)
 execute(client, results)
+
+# ## Verify no participant in the pdr_ehr_dup_report list has EHR data.
```
lrwb-aou, line 862: @MichaelSchmidt82 One last request, add more to this check description.  When someone reads this, they should know how to identify if the check passes or fails and what to do/check in case of a failure.  This is just updating verbage, but it's important to whoever runs the notebook because they very likely didn't create this check.  You can reference jira issues, lookup tables, etc.  When writing it, think of the information you would want to have if this failed and you didn't write the check.  Add anything useful.

MichaelSchmidt82: @lrwb-aou , this description has been updated.

ksdkalluri (approved):

PR #1761, opened 2023-10-05
[DC-3454] Add a QC check for expected tables/ mapping tables in clean_rdr and ehr_union

hiro-mishima:


nishanthpp93 (approved):
Looks good!

PR #1765, opened 2023-10-09
[DC-3505] Update the checks that check for participants with basics data

brendagutman:


data_steward/analytics/cdr_ops/controlled_tier_qc/check_controlled_tier_part2.py
```diff
@@ -469,48 +471,60 @@ def my_sql(table_name, column_name):
         },
         ignore_index=True)
 
-# # query 7  All participants have basics,done
+# # query 7  All participants have basics data in the rdr dataset
```
hiro-mishima, line 474: I have a question about the purpose of this QC.

Is this check for making sure:
a. all participants in CT having the basics records in the `CT` observation table, or
b. all participants in CT having the basics records in the `RDR` observation table?

Your update is checking `b`. But I think checking `a` is more appropriate since the basics records might be dropped along the way and they can exist in RDR but not in CT for some reason.

Let me know what you think - I might be missing some context, so I would appreciate your thoughts on this!

hiro-mishima (commented):
Thanks @brendagutman! Your comments in the JIRA ticket and the note from the eng-sync made the review smoother. I have a question on the update for CT QC notebook. Let me know what you think. Thanks!

data_steward/analytics/cdr_ops/controlled_tier_qc/check_controlled_tier_part2.py
brendagutman: That is a good question. I should have put that in the ticket. I'll do that now.
I was working off of the logic described in 'drop_participants_without_any_basics'.

```
...

As part of DC-696, several participants were found with no basics survey still persisting in the CDR
These records were removed, and DC-706 was created to remove such participants in the future. DC-2551
was created to update DropParticipantsWithoutPPI to DropParticipantsWithoutAnyBasics to
reflect what the rule is actually doing, dropping participants who have not completed any question in The Basics.
It also moves the cleaning rule to RDR_CLEANING_CLASSES list to allow curation to deliver a list of
participants expected to exist in the CT dataset when the RDR dataset is cleaned.

...

```

I saw that this CR was moved in order to enable creation of the CT expected list. To me this means that the participants who have the Basics at the RDR stage should be expected in the CDR.

Do you agree with my thought process?

hiro-mishima: Hi Brenda, thanks for the background. I understand the context now. I have two follow-up comments.

1. Your rebase might have wiped out the changes to the raw RDR QC and the clean RDR QC. Please check the files changed in this PR again to see if it's intended.
2. Regarding my previous question, if the intention is to check `b`, I think we do not need this check in the CT QC notebook. All the participants are already confirmed to have the basics in the RDR data stage, and we are not adding any new participants in the downstream data stages. So, the updated query in the CT QC notebook will always return PASS because all the participants are guaranteed to have the basics in the RDR data stage. In the meanwhile, if the intention is to check `a`, it can return `FAILED` because the basics record might be dropped along the way.

brendagutman: 1. The expectations for the raw notebook are changed. The check was removed in another PR. I'll update the ticket soon.  I did make updates to both the clean rdr notebook and the CT notebook with the last commit. Thank you for the extra questions and reviews.
2. After our team discussion (see eng sync notes 'Participants with/without the Basics') we will keep the check but add to the description. I'll also make note of this in the ticket.

data_steward/analytics/cdr_ops/clean_rdr_export_qc.py
```diff
@@ -668,54 +668,34 @@
...
+  WHERE ancestor_concept_id = 1586134
+)
+
+SELECT 
+'persons_without_basics_in_rdr' as issue,
+COUNT(DISTINCT person_id) as n
+FROM `{{project_id}}.{{new_rdr}}.person` 
+WHERE person_id NOT IN (SELECT person_id FROM pids_with_basics)
+GROUP BY person_id
+
```
brendagutman, line 695: Updated to use the same logic as the CR.

data_steward/analytics/cdr_ops/controlled_tier_qc/check_controlled_tier_part2.py
```diff
@@ -1471,10 +1478,12 @@ def query_template(table_era):
 result
 
 
+# # Results Summary
+
```
brendagutman, line 1482: Adding a markdown header for the summary so it shows up in the notebook table of contents.

data_steward/analytics/cdr_ops/clean_rdr_export_qc.py
```diff
@@ -1382,21 +1342,11 @@
...
-
-# +
-# Check the expectations of survey_conduct - survey list
-
-Confirm that all expected surveys have records in survey_conduct. Check ignores snap surveys because these surveys are not expected in any release.
-
-Generally the list of surveys should increase from one export to the next.
-
-Investigate any surveys that were available in the previous export but not in the current export. 
-Also make sure that any new expected surveys are listed in the current rdr.
```
brendagutman, line 1394: This is no longer applicable to the survey_conduct check. Let me know if you would rather this be in a separate PR.

hiro-mishima (approved):
Thanks for all the updates, Brenda! It looks great!!

PR #1771, opened 2023-10-11
[DST-1423] Update rt_cdr_qc_report3 notebook

jqian2015:


data_steward/analytics/cdr_ops/rt_cdr_qc/cdr_deid_qa_report3_col_suppression.py
```diff
@@ -178,85 +155,62 @@
...
 #
-# <font color='red'> no visit_detail_id in old CDR?
-#
-#  Uncomment below when visit_detail_id is available in 2022 CDR
+# <font color='red'> DST-1423,updates, only provide_id is null.
 
 query = JINJA_ENV.from_string("""
 SELECT
 SUM(CASE WHEN provider_id IS NOT NULL THEN 1 ELSE 0 END) AS n_provider_id_not_null,
 SUM(CASE WHEN visit_detail_id IS NOT NULL THEN 1 ELSE 0 END) AS n_visit_detail_id_not_null
```
brendagutman, line 202: This line was to be removed correct? Or is the field to be suppressed?

data_steward/analytics/cdr_ops/rt_cdr_qc/cdr_deid_qa_report3_col_suppression.py
```diff
@@ -178,85 +155,62 @@
...
 
 # +
 # query6 has error
 # no this column called visit_detail_id?
-
+    
 query = JINJA_ENV.from_string("""
 SELECT
 SUM(CASE WHEN provider_id IS NOT NULL THEN 1 ELSE 0 END) AS n_provider_id_not_null,
 SUM(CASE WHEN visit_detail_id IS NOT NULL THEN 1 ELSE 0 END) AS n_visit_detail_id_not_null
```
brendagutman, line 179: Same as below

data_steward/analytics/cdr_ops/rt_cdr_qc/cdr_deid_qa_report3_col_suppression.py
jqian2015: which line? 235-237? if so, they were already removed in my updated one. should be like this, 

# # 6 Verify the provider_id in the DEVICE_EXPOSURE table is null and visit_detail_id is not all null:
#
# <font color='red'> DST-1423,updates, only provide_id is null.

# +
# query6 has error
# no this column called visit_detail_id?
    
query = JINJA_ENV.from_string("""
SELECT
SUM(CASE WHEN provider_id IS NOT NULL THEN 1 ELSE 0 END) AS n_provider_id_not_null,
SUM(CASE WHEN visit_detail_id IS NOT NULL THEN 1 ELSE 0 END) AS n_visit_detail_id_not_null

jqian2015: here, my updated one, https://github.com/jqian2015/curation/blob/dst-1423/data_steward/analytics/cdr_ops/rt_cdr_qc/cdr_deid_qa_report3_col_suppression.py

brendagutman: No line 242. I believe this ticket is to remove the part that checks for visit_detail_id to be NULL. We looked in the privacy rules and believe this field should be kept. 

Same for visit_detail_id in DRUG_EXPOSURE as well.

data_steward/analytics/cdr_ops/rt_cdr_qc/cdr_deid_qa_report3_col_suppression.py
jqian2015: if you are talking about line 203/204, then, yes, they can be removed. would you want me to do that?

jqian2015: due to they are uncommented in the cell, didn't pay attention.

data_steward/analytics/cdr_ops/rt_cdr_qc/cdr_deid_qa_report3_col_suppression.py
jqian2015: ok, there might be misunderstanding here, as seen in the DST-1423, from Krishna, "So, here we are checking for whole column suppression for both the provide_id and visit_detail_id. Even though we have a lot of nulls is fine for visit_detail as that is from a lack of corresponding visit_detail records. The check should only fail if all the visit_detail_ids are null in the table." what i did was to check both columns and fail only provider_id is not null or if all the visit_detail_ids are null. But if we only need to check one column provider_id, we can do that too. But I mentioned that too and then Krishna responded that way. so could you please look through those conversions in the ticket and let me know the decision whether we should check one column or two columns?

data_steward/analytics/cdr_ops/rt_cdr_qc/cdr_deid_qa_report3_col_suppression.py
```diff
@@ -130,46 +118,35 @@
...
-        'result': 'Failure'
-    },
-                   ignore_index=True)
-df1.T
+ df = df.append({'query' : 'Query3 Measurement', 'result' :  'Failure'},  
+                ignore_index = True) 
+df1.T  
 
-# # 4 Verify the following columns in the AOU_DEATH table have been set to null:
+# # 4 Verify the following columns in the DEATH table have been set to null:
```
brendagutman, line 132: I don't think changes to other queries are needed. This seems to be an old version of the notebook.

data_steward/analytics/cdr_ops/rt_cdr_qc/cdr_deid_qa_report3_col_suppression.py
```diff
@@ -178,85 +155,62 @@
...
-df1
+ df = df.append({'query' : 'Query5 Condition provider_id', 'result' :  'Failure'},  
+                ignore_index = True) 
+df1        
 
-# # 6 Verify the following columns in the DEVICE_EXPOSURE table have been set to null:
+# # 6 Verify the provider_id in the DEVICE_EXPOSURE table is null and visit_detail_id is not all null:
 #
-# <font color='red'> Uncomment below when visit_detail_id is available
+# <font color='red'> DST-1423,updates, only provide_id is null.
```
brendagutman, line 170: Suggested change. Would you change this comment to something like:
```
<font color='red'> 'DST-1423, updates. Verify that provider_id is suppressed(null) and that visit_detail_id is not suppressed(not always null) in `device_exposure`'. <br>
    
`n_provider_id_not_null` should return 0 <br>
`n_visit_detail_id_not_null` should return > 0  <br>
```

When we run these notebooks and we see results with a count it usually means the check is failing. We don't always go down to the summary to figure out if the check is failing. I think this comment will clear up any possible confusion.

data_steward/analytics/cdr_ops/rt_cdr_qc/cdr_deid_qa_report3_col_suppression.py
```diff
@@ -178,85 +155,62 @@
...
 # -
 
-# # 7 Verify the following columns in the DRUG_EXPOSURE table have been set to null:
-# provider_id
+# # 7 Verify the provider_id in the DRUG_EXPOSURE table is null and visit_detail_id is not all null:
 #
-# <font color='red'> no visit_detail_id in old CDR?
-#
-#  Uncomment below when visit_detail_id is available in 2022 CDR
+# <font color='red'> DST-1423,updates, only provide_id is null.
```
brendagutman, line 197: A similar comment as the one mentioned above would make the results easier to understand.

data_steward/analytics/cdr_ops/rt_cdr_qc/cdr_deid_qa_report3_col_suppression.py
brendagutman: Thanks I see those conversations now. I misunderstood the ticket. Thanks!

data_steward/analytics/cdr_ops/rt_cdr_qc/cdr_deid_qa_report3_col_suppression.py
brendagutman: No longer an issue.

brendagutman (approved):
Looks good. Thanks Jun!

PR #1774, opened 2023-10-12
[DC-3515] Create check for duplicate observation_ids per person_id in raw_rdr_export_qc.py

lledezma:


data_steward/analytics/cdr_ops/raw_rdr_export_qc.py
```diff
@@ -1293,8 +1289,20 @@
...
-            new_rdr=new_rdr)
+""").render(project_id=project_id, new_rdr=new_rdr)
 execute(client, query)
 
+# # Check to catch duplicate observation_ids
 
+tpl = JINJA_ENV.from_string('''
+SELECT
+  observation_id,
+  COUNT(DISTINCT person_id) AS n
```
nishanthpp93, line 1300: Regardless of person_id, we should not see any duplicates in observation_id
```suggestion
  COUNT(DISTINCT observation_id) AS n
```

nishanthpp93 (changes requested):
Looks good but added a small change. I would also suggest not running yapf on this file since it is excluded for linting tests, that should clear up the unnecessary diffs.

data_steward/analytics/cdr_ops/raw_rdr_export_qc.py
```diff
@@ -1297,4 +1297,18 @@
             new_rdr=new_rdr)
 execute(client, query)
 
+# # Check to catch duplicate observation_ids
+
+tpl = JINJA_ENV.from_string('''
+SELECT
+  observation_id,
+  COUNT(DISTINCT observation_id) AS n
```
nishanthpp93, line 1305: You can drop the distinct in the count too, sorry should've caught this earlier.

nishanthpp93 (approved):
Looks good!

PR #1779, opened 2023-10-18
[DC-3539] Create required Schemas for rdr stage

ratuagga:
adding missing schemas in rdr stage.


nishanthpp93 (commented):
Small changes but looks good

PR #1780, opened 2023-10-18
[DC-3544] Updates to include and exclude new mapping tables at Raw RDR stage

ratuagga:
create mapping tables during snapshot.


nishanthpp93 (commented):
Looks good! We can address the fact_relationship issue later since mapping is required in both Union and RDR.

PR #1784, opened 2023-10-23
[DC-3509] Suppression of organ transplant rows

lledezma:
Suppression of organ transplant rows is no longer applied, so DC-1527: [CT - Analytics] Verify suppression of organ transplant rows QA check is not needed.


hiro-mishima (approved):
Looks good! Thanks for the comments in the PR and the JIRA ticket. Those helped me understand the context.

hiro-mishima (approved):
Thanks for the additional changes!

PR #1785, opened 2023-10-23
[DC-3402] Create CR for AIAN lookup table

hiro-mishima:


ksdkalluri (approved):
Looks good, thanks for the thorough test cases.

PR #1786, opened 2023-10-24
[DC-3548] Update checks and descriptions in clean_rdr_export_qc.py

brendagutman:


data_steward/analytics/cdr_ops/clean_rdr_export_qc.py
```diff
@@ -1062,89 +1090,7 @@
 
 # -
 
-# # COPE survey mapping
```
ksdkalluri, line 1065: @brendagutman why is this check being removed? can you provide some context on this please.

data_steward/analytics/cdr_ops/clean_rdr_export_qc.py
```diff
@@ -680,10 +693,23 @@
 WITH pids_with_basics AS ( -- pids with basics data --
 SELECT
 person_id
-  FROM `{{project_id}}.{{new_rdr}}.concept_ancestor`
-  INNER JOIN `{{project_id}}.{{new_rdr}}.observation` o ON observation_concept_id = descendant_concept_id
-  INNER JOIN `{{project_id}}.{{new_rdr}}.concept` d ON d.concept_id = descendant_concept_id
-  WHERE ancestor_concept_id = 1586134
+FROM `{{project_id}}.{{new_rdr}}.concept_ancestor`
```
ksdkalluri, line 696: @brendagutman can you also update the description of this check by adding context on why the query is extended.

data_steward/analytics/cdr_ops/clean_rdr_export_qc.py
brendagutman: Yes. This check was added when we expected that the cope concepts would be remapped to individual module concepts in Athena. We decided not to go on this route. Instead using custom concepts, the cope_survey_semantic_version_map, and then the survey_conduct table to determine the cope version. 

This check will always fail(cause confusion) and the check  `Survey version and dates` verifies our current cope mapping strategy.

data_steward/analytics/cdr_ops/clean_rdr_export_qc.py
brendagutman: @ksdkalluri I extended this check to align with logic used in the cleaning rule.  This way the check does not fail due to differing criteria. The CR should have the most accurate/current criteria to define a person with the Basics.

 I'm not sure why the CR was written this way but I can look into it and put more in the description if you'd like.

data_steward/analytics/cdr_ops/clean_rdr_export_qc.py
ksdkalluri: Thanks for the context.

data_steward/analytics/cdr_ops/clean_rdr_export_qc.py
ksdkalluri: Now I remember this makes sense.

ksdkalluri (approved):
Looks good!

PR #1789, opened 2023-10-25
[DC-3550] cleaning rule updates for pdr tables and other rdr stage changes

ratuagga:
pdr cr and rdr import related changes.


nishanthpp93 (approved):
Looks good!

PR #1798, opened 2023-10-31
[DC-3333] Python 3.11 upgrade

hiro-mishima:
Highlight
- python is upgraded from 3.7 to 3.11
- All the necessary packages are also upgraded to the latest except the following: `pandas`
- Pandas is upgraded to 1.5.3 (the latest of 1.x), not to 2.1.2. Pandas 2.x no longer supports `df.append()` command. We have to use `df.concat()` instead. Our legacy QC notebooks reference `df.append()` a lot, so upgrading it to Pandas 2.x will disrupt our CDR generation quite a lot. It is safer to keep it Pandas 1.x. https://pypi.org/project/pandas/#history

---

Updated `bigquery.SchemaField()`

For the following, the internal function `_to_standard_sql_type()` uses `bigquery.StandardSqlDataTypes()` which is no longer supported by the package. All it does is get the column type, so I removed it and used `resources.get_bq_col_type()` instead. Also, `bigquery.SchemaField()` has a different parameter order than the latest version. The following needed an update on how they use `bigquery.SchemaField()`.
- `data_steward/gcloud/bq/__init__.py`
- `data_steward/utils/bq.py`
- `tests/unit_tests/data_steward/gcloud/bq/bq_test.py`
- `tests/unit_tests/data_steward/utils/bq_test.py`
- `tests/unit_tests/data_steward/cdr_cleaner/cleaning_rules/ppi_branching_test.py`

The following had to be also updated because of `bigquery.SchemaField()`:
- `data_steward/tools/load_vocab.py`

---

Deprecated `moz-sql-parser` in `ehr_union_test.py`
- `tests/integration_tests/data_steward/validation/ehr_union_test.py` was using `moz-sql-parser` but `moz-sql-parser` is deprecated and was causing errors with Python 3.11. I had to update the test case.
https://github.com/mozilla/moz-sql-parser

---

Updated `pylint`
- pylint is upgraded to the latest version. Some scripts needed updates to align with the latest pylint's linting guidelines.


ksdkalluri (approved):
@hiro-mishima This is good to merge but, we will hold off on merging to develop till the day we are ready to deploy and migrate.

PR #1804, opened 2023-11-14
[DC-3485] Update ehr_upload_pids

hiro-mishima:


data_steward/tools/generate_ehr_upload_pids.py
```diff
@@ -15,76 +17,88 @@
...
+CREATE OR REPLACE VIEW `{{project_id}}.{{operations_dataset_id}}.{{ehr_pids_view}}` 
+OPTIONS(description='A participant-level view of when ehr data was sent. NOTE: the RDR calls this view to support HealthPro (1/27/21)')
+AS {% for hpo_site in hpo_sites %}
+SELECT 
+    person_id,
+    current_datetime() AS report_run_time,
+    Org_ID as org_id,
+    HPO_ID as hpo_id,
+    Site_Name as site_name, 
+    TIMESTAMP_MICROS(t.last_modified_time * 1000) AS latest_upload_time
```
nishanthpp93, line 39: @hiro-mishima Our biggest problem with the ehr_upload_pids was the "Query too large" error, which prevented us from doing this. This feels like it is making the query more complicated. Let me know if you've already tested this and it works fine.

nishanthpp93 (commented):
Looks good, but one concern. Let me know if already addressed.

PR #1806, opened 2023-11-16
[DC-3596] Update generate_wear_study_table.py to use the latest wear …

brendagutman:
…consent records


tests/integration_tests/data_steward/cdr_cleaner/cleaning_rules/generate_wear_study_table_test.py
```diff
@@ -115,12 +115,14 @@ def test_queries(self):
...
-                (2, 'Yes', self.date_one, None),
-                # two yes records and one no. earliest dates recorded
-                (5, 'Yes', self.date_one, self.date_three),
+                # two yes records. latest yes recorded
+                (2, 'Yes', self.date_two, None),
+                # two no records then a yes. latest yes recorded
+                (4, 'Yes', self.date_three, None),
+                # two yes records and one no. first no record after the latest yes
+                (5, 'Yes', self.date_two, self.date_three),
                 # yes, then no, then yes. earliest dates recorded
```
hiro-mishima, line 124: This comment should be updated too to something like:
`# yes, then no, then yes. latest dates recorded`

hiro-mishima (approved):
One comment in the test needs un update but everything else looks good!!

PR #1807, opened 2023-11-17
[DC-3597] Add Indian Health Services responses to Registered Tier Generalization Rules

hiro-mishima:


brendagutman (approved):
This addition looks perfect Hiro. Thanks for all of the test cases and clear descriptions!

PR #1808, opened 2023-11-20
[DC-3605] [DC-3600] Exclude 'person_ext' from being removed by cleaning Rule

ratuagga:
Updates made to ensure person_ext is created, populated, and kept at CT base.


nishanthpp93 (approved):
Looks good!

PR #1809, opened 2023-11-20
[DC-3593] 'aian_list' table does not exist in rdr sandbox

ratuagga:
updated aian lookup to have table in sandbox.


nishanthpp93 (approved):
Looks good!

PR #1810, opened 2023-11-20
[DC-3606] Fix bug in union script

ratuagga:
updated ehr scripts as per ticket.

### DIFF
--- a/data_steward/tools/generate_unioned_ehr_dataset.py
+++ b/data_steward/tools/generate_unioned_ehr_dataset.py
@@ -83,7 +83,8 @@ def copy_mapping_tables(client, source_dataset, destination_dataset):
     ]:
         source_table = f'{source_dataset}.{table}'
         destination_table = f'{destination_dataset}.{table}'
-        if client.table_exists(source_table, source_dataset):
+        LOGGER.info(f'Copying Table: {source_table}')
+        if client.table_exists(table, source_dataset):
             job = client.copy_table(source_table, destination_table)
             job.result()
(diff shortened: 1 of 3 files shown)


data_steward/validation/ehr_union.py
```diff
@@ -918,6 +919,7 @@ def create_load_aou_death(bq_client, project_id, input_dataset_id,
 def main(input_dataset_id,
          output_dataset_id,
          project_id,
+         run_as,
```
nishanthpp93, line 922: We cannot add impersonation here. This is run by app engine which cannot use impersonation

nishanthpp93 (changes requested):
The changes to generate_unioned script look good, but the ehr_union and test changes need to be reverted.

data_steward/tools/generate_unioned_ehr_dataset.py
```diff
@@ -83,7 +83,8 @@ def copy_mapping_tables(client, source_dataset, destination_dataset):
     ]:
         source_table = f'{source_dataset}.{table}'
         destination_table = f'{destination_dataset}.{table}'
-        if client.table_exists(source_table, source_dataset):
+        LOGGER.info(f'Copying Table: {source_table}')
```
nishanthpp93, line 86: This needs to be moved inside the if condition

nishanthpp93 (commented):
Looks good but added a suggestion.

nishanthpp93 (approved):
Looks good!

PR #1813, opened 2023-11-27
[DC-3611] Update QC notebook for Indian Health Services responses generalization CR

MichaelSchmidt82:
Description:

Participants answer multiple-choice insurance questions on multiple surveys.  One of the available insurance selections is “Indian Health Services” (or a variant).  This option does not necessarily identify a participant as a self-identifying AI/AN participant.  This check ensures this data is indeed removed via this check.

See related ticket [DC-3597](https://precisionmedicineinitiative.atlassian.net/browse/DC-3597)

Note: some partial re-formatting occurred, resulting in 5 superfluous diff's.

[DC-3597]: https://precisionmedicineinitiative.atlassian.net/browse/DC-3597?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/data_steward/analytics/cdr_ops/rt_cdr_qc/cdr_deid_base_qa_report1.py
+++ b/data_steward/analytics/cdr_ops/rt_cdr_qc/cdr_deid_base_qa_report1.py
@@ -48,11 +48,12 @@ df = pd.DataFrame(columns=['query', 'result'])
 
 # ## Query 1.0
 #
-# Even though this option does not overtly identify a participant as AI/AN, it suggests so.  Therefore, these responses are generalized to "Other" and later deduplicate the rows that were created as a result
+# Participants answer multiple choice insurance questions on multiple surveys.  One of the available insurance selections is “Indian Health Services” (or a variant).  This option does not necessarily identify a participant as a self identifying AI/AN participant.  Even though this option does not overtly identify a participant as AI/AN, it suggests so.  Hence, we generalize these responses to the category "Other" and subsequently eliminate duplicate rows generated during this process.
 #
-# We need a check for each part: We update both the value_source_concept_id and value_as_concept_id. Therefore, it is essential to include a validation check to confirm the success of the update.
-#
-# See [DC-3597](https://precisionmedicineinitiative.atlassian.net/browse/DC-3597)
+# This check verifies records were suppressed. The observation_concept_id's are:
+# - `40766241`
+# - `1585389`
+# - `43528428`
 
 # +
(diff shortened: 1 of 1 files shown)


data_steward/analytics/cdr_ops/rt_cdr_qc/cdr_deid_base_qa_report1.py
```diff
@@ -54,14 +54,71 @@
...
+# The observation_concept_id's are:
+# - `40766241`
+#
+# - `1585389`
+#
+# - `43528428`
+
+# +
+query = JINJA_ENV.from_string("""
+SELECT observation_id FROM `{{project_id}}.{{deid_base_cdr}}.observation`
```
ksdkalluri, line 79: @MichaelSchmidt82 this check looks for if there are any duplicates caused by the generalization which is good, but there should also be a check which finds out if we can find old(i.e Indian Health related) responses in the tables as well. I suggest you add another check which checks if the Indian health related(i.e old) value_source_concept_id and value_as_concept_id still exists in the table, if exists the check should fail and if you get no result the check should pass.

ksdkalluri (changes requested):

data_steward/analytics/cdr_ops/rt_cdr_qc/cdr_deid_base_qa_report1.py
```diff
@@ -39,29 +39,142 @@
...
+#   - value_as_string
+#   - value_source_value
+#
+# has been done in first sql for deid, can be skipped here
+
+# ## Query 1.0
+#
+# Even though this option does not overtly identify a participant as AI/AN, it suggests so.  Therefore, these responses are generalized to "Other" and later deduplicate the rows that were created as a result
+#
+# We need a check for each part: We update both the value_source_concept_id and value_as_concept_id. Therefore, it is essential to include a validation check to confirm the success of the update.
```
ksdkalluri, line 53: @MichaelSchmidt82 the following is added in the query 1.1 but that description belongs in here.
```
This check verifies records were suppressed. The observation_concept_id's are:
# - `40766241`
# - `1585389`
# - `43528428`
```

ksdkalluri (changes requested):
@MichaelSchmidt82 Checks look good, please take a look at my comments about the descriptions for each check.

ksdkalluri (approved):
Looks Good!

PR #1820, opened 2023-12-06
[DC-3614] Person_ext Table is empty in RT Deid Stage

ratuagga:
copy ext table before deid to run deid on person_ext.


nishanthpp93 (approved):
Looks good!

PR #1821, opened 2023-12-07
[DC-3513] Update the raw_fitbit_qc.py - sleep_level - `level` check

lledezma:


brendagutman (approved):
Looks good Luis. Thanks for the update!

PR #1823, opened 2023-12-11
[DC-3598] Update `convert_pre_post_coordinated_concepts` to create an INSERT sandbox as well as the DELETE sandbox

brendagutman:


data_steward/cdr_cleaner/cleaning_rules/convert_pre_post_coordinated_concepts.py
```diff
@@ -136,13 +135,20 @@
     o.value_source_concept_id,
     o.value_source_value, 
     o.questionnaire_response_id
-FROM `{{project}}.{{sandbox_dataset}}.{{sandbox_table}}` o
+FROM `{{project}}.{{sandbox_dataset}}.{{sandbox_table}}_deleted` o
```
nishanthpp93, line 138: So this uses the deleted obs ids and creates a new one using the concept_ids? This isn't your logic but it feels odd. I guess it's ok as long as there are no id collisions.

nishanthpp93 (commented):
Added a comment for clarification but otherwise looks good

nishanthpp93 (approved):
There may be collisions in the future but we may disable this CR since it is covered by RDR

PR #1828, opened 2023-12-15
[DC-3649] Update CT notebook 10 to use updated site_maskings location

ratuagga:
CT notebook update for site_maskings.


nishanthpp93 (approved):
Looks good!

PR #1829, opened 2023-12-15
[DC-3646] Update git tag function to sort by version correctly

ratuagga:
updated git tag fetch function.


nishanthpp93 (approved):
Looks good!

PR #1830, opened 2023-12-15
[DC-3645] Update Union QC notebook with fixes for column name and moving notes check

ratuagga:
updated QC notebook for ehr_union and moved 1 check to combined.


nishanthpp93 (approved):
Looks good!

PR #1832, opened 2023-12-15
[DC-3647] Update repopulate person post deid CR to account for custom concept

ratuagga:
updated repopulate_person query for aouDRC indicator.


nishanthpp93 (approved):
Looks good!

PR #1834, opened 2023-12-18
[DC-3635] Remove portion of query checking primary consent

MichaelSchmidt82:
This PR remove the portion of checks where wear_consent participants must have primary consent.

[Ticket](https://precisionmedicineinitiative.atlassian.net/browse/DC-3635)
[Investigation](https://precisionmedicineinitiative.atlassian.net/browse/DC-3626)

### DIFF
--- a/data_steward/analytics/cdr_ops/rt_cdr_qc/cdr_deid_qa_report10_extra.py
+++ b/data_steward/analytics/cdr_ops/rt_cdr_qc/cdr_deid_qa_report10_extra.py
@@ -866,26 +866,6 @@ WHERE person_id not in ( -- person table --
   SELECT person_id
   FROM `{{project_id}}.{{rt_cdr_deid}}.person` o
   )
-  
-UNION ALL
-
-SELECT
-  'no primary consent' as issue,
-  COUNT(person_id) as bad_rows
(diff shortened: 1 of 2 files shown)


ksdkalluri (changes requested):
@MichaelSchmidt82 there is also a similar check in `check_controlled_tier_part2.py` that checks for primary consent for wear participants that needs to be removed as well as per the scope of the ticket.

MichaelSchmidt82:
> @MichaelSchmidt82 there is also a similar check in `check_controlled_tier_part2.py` that checks for primary consent for wear participants that needs to be removed as well as per the scope of the ticket.

Good catch!

data_steward/analytics/cdr_ops/rt_cdr_qc/cdr_deid_qa_report10_extra.py
```diff
@@ -866,26 +866,6 @@ def my_sql(table_name, column_name):
   SELECT person_id
   FROM `{{project_id}}.{{rt_cdr_deid}}.person` o
   )
-
```
ksdkalluri, line 869: @MichaelSchmidt82 you would also need to update the comment for this check to remove primary_consent related  info.

ksdkalluri (approved):

PR #1838, opened 2023-12-20
[DC-3660] Update Combined QC notebook to include CE

ratuagga:
Updated combined QC check.


nishanthpp93 (changes requested):
Linting check is failing due to ticket number, so I would suggest pushing another commit so it uses the PR title instead of the commit message

PR #1839, opened 2023-12-20
[DC-3659] Update check controlled tier part 2 to include standard classifications

ratuagga:
Updated CT Part2 QC check.


nishanthpp93 (approved):
Looks good!

PR #1840, opened 2023-12-20
[DC-3658] Update Participant Validation QC check for excluded sites

ratuagga:
Updated Participant validation QC check.


nishanthpp93 (approved):
Looks good!

PR #1841, opened 2024-01-03
[DC-3632] Remove irrelevant bq_utils import statements

MichaelSchmidt82:
This PR removes several superfluous `bq_utils` imports.  It removes 22 import statements where they are no longer needed.  Since some files contain trailing whitespace, it may be hard to determine the actual, important change.

Luckily, most import statements occur at the beginning of each file; therefore, one may need to `check only the first few lines!` to see a removal using diff.


lledezma (approved):
The title of the PR does not match the title of the ticket. I would update the title of the PR before merging.

lledezma (changes requested):
The title of the PR does not match the title of the ticket. I would update the title of the PR before merging.

lledezma (approved):

PR #1846, opened 2024-01-11
[DC-3668] Ignore race/ethnicity sub categories in ct notebook

brendagutman:


data_steward/analytics/cdr_ops/controlled_tier_qc/sql/query_templates.py
```diff
@@ -116,6 +116,8 @@
     SELECT {{ new_id }}
     {% if mapping_table == 'site_maskings' %}
     FROM `{{ project_id }}.{{ pipeline_dataset }}.{{ mapping_table }}`
+    {% elif mapping_table == '_deid_questionnaire_response_map' %}
+    FROM `{{ project_id }}.{{ questionnaire_response_dataset }}.{{ mapping_table }}`
```
brendagutman, line 120: The dataset containing the table `_questionnaire_response_map`  is given as a notebook parameter. This check should use the dataset defined by the parameter instead of falling to the else clause and therefore the 'mapping_dataset'.

data_steward/analytics/cdr_ops/controlled_tier_qc/utils/helpers.py
```diff
@@ -157,25 +157,29 @@ def run_check_by_row(df,
...
+                data_type=data_type,
+                primary_key=primary_key,
+                new_id=new_id,
+                mapping_dataset=mapping_dataset,
+                mapping_table=mapping_table,
+                pipeline_dataset=PIPELINE_TABLES,
+                zip_table_name=ZIP_CODE_AGGREGATION_MAP)
+            result_df = pd.read_gbq(query, dialect="standard")
+            result_df['query'] = str(query)
+            results.append(result_df)
```
brendagutman, line 182: This function builds queries using multiple csvs. The csvs don't always have the same fields. The `if` statement will skip a check if the field `is_suppressed` is present and is flagged 'NO'. Otherwise the function will run as it has previously.

data_steward/analytics/cdr_ops/controlled_tier_qc/csv/Controlled_Tier_Concept_Level.csv
```diff
@@ -218,45 +218,45 @@ observation,observation_source_value,,NHPI_NHPISpecific,YES,DC-1366
 observation,observation_source_value,,White_WhiteSpecific,YES,DC-1366
 observation,observation_source_value,,Hispanic_HispanicSpecific,YES,DC-1366
 observation,observation_source_value,,RaceEthnicityNoneOfThese_RaceEthnicityFreeTextBox,YES,DC-1366
-observation,observation_source_concept_id,1586151,,YES,DC-1366
```
nishanthpp93, line 221: Let's get rid of these so we can shrink this convoluted notebook

nishanthpp93 (changes requested):

nishanthpp93 (approved):
Looks good!

PR #1852, opened 2024-01-22
[DC-3731] [DC-3740] [DC3741] Updates to the combined cross tier and controlled tier notebooks 

brendagutman:


nishanthpp93 (approved):
Looks good!

PR #1857, opened 2024-03-14
[DC-3769] Use rewrite instead of copy

hiro-mishima:


data_steward/validation/main.py
```diff
@@ -1010,7 +1010,18 @@ def process_hpo_copy(hpo_id):
...
+            rewrite_token = False
+            while True:
+                rewrite_token, bytes_rewritten, bytes_to_rewrite = dest_blob.rewrite(
+                    hpo_blob, token=rewrite_token)
+                logging.info(
+                    f"{full_name}: Copied: {bytes_rewritten}/{bytes_to_rewrite} bytes."
+                )
+
+                if not rewrite_token:
+                    break
```
nishanthpp93, line 1024: Can this be in the gcloud/gcs.py client as a custom function? We have it just for this purpose. Also, maybe we can use gsutil rsync as long term since it only copies if the checksum is different.

nishanthpp93 (commented):
Overall looks good and works! Added some suggestions

data_steward/validation/main.py
hiro-mishima: Thanks for the comment! It all makes sense. Regarding `rsync`, I created an enhancement ticket, DC-3771, where we can implement it as a long-term solution.

nishanthpp93 (approved):
Looks good! As a sanity check before deployment, I would suggest locally running this in a python script to copy the large file to make sure there is no other issue.

PR #1862, opened 2024-05-01
[DC-3786] [DC-3788] Add self reported population columns

nishanthpp93:
* [DC-3786] [DC-3788] Add self reported population


brendagutman (approved):
Looks good thanks Nishanth!

PR #1874, opened 2024-12-23
[DC-3858] Fix id columns not found for yob suppression

ratuagga:
person_ext schema update


nishanthpp93:
@ratuagga `self_reported_category_concept_id` is now permanently added to the CDR. I think we removed these for V7 and earlier fixes, but they should not be removed for V8 or OC.

PR #1896, opened 2025-07-23
[DC-3928] Add 'DL' in test script to allow Lakehouse ticket tags to pass CircleCi tests

ratuagga:
Added 'DL' in test script.


nishanthpp93 (changes requested):
Looks like it also needs to be changed in https://github.com/all-of-us/curation/blob/2f55817aff312ecf20494cf7a946f72f0be03cbf/docker/develop/scripts/run-tests/lint_00_validate_commit_message.sh#L5

nishanthpp93 (approved):
Looks good!

PR #1929, opened 2026-08-04
[DL-2445] Add CT+ ID regeneration script

hiro-mishima:
What

- Two standalone tools, mirroring the RT pair: regenerate_ct_plus_ids.py (CDM tables, forked from regenerate_rt_ids.py) and regenerate_etm_ct_plus_ids.py (the four EtM tables, modelled on regenerate_etm_rt_ids.py). Run in sequence after deid_clean, last step before publication.

Where to look

- Row IDs are a random permutation, not ORDER BY src_<table>_id. The RT form leaks CT ordering, so anyone holding both tiers realigns rows by sorting. RAND() is drawn outside the DISTINCT; inside it, duplicate source IDs survive.
- The src_table_id stamp is the constant ct_plus, not the input dataset name. Both the append filter and the load-side join match on it, so a per-release value would silently reassign every ID and defeat persistence.
- Two guards abort the run before writing: a person mapping from a different run, and any already-published participant who would be issued a new CT+ ID.

Refinement from validation

- update_ext_table now resolves _ext IDs against the loaded base table as well as the mapping, so _ext rows stay in step with a base table the load has filtered.

Notes

- The person mapping has been exercised against a mock rdr_participant_research_ids_view, using real CT person_id values with placeholder CT+ ids. The real values arrive with DL-2396, after which the pipeline can run end to end.
- Expect row count differences on two tables: visit_detail (9.65%) and fact_relationship (92%), where source rows carry references these scripts have no target to map to. The filters are intentional and the source of those rows is upstream of this work.
- 56 unit tests cover the query builders for both scripts. Validated against preprod scratch datasets only; nothing live was written to.

CI

- integration_test fails at gcloud auth activate-service-account with invalid_grant: Invalid JWT Signature. The CircleCI service-account key is stale or revoked, zero tests ran, unrelated to these commits. unit_test and Codacy pass.

### DIFF
--- a/data_steward/tools/regenerate_ct_plus_ids.py
+++ b/data_steward/tools/regenerate_ct_plus_ids.py
@@ -17,8 +17,10 @@ four ways:
    sorting. The random value is materialized in an inner subquery so no
    nondeterministic function is evaluated inside a window ORDER BY. The mapping table
    is persisted, so the draw does not need to be reproducible and no salt or long
-   lived secret is introduced. Mappings are re-cut per release rather than appended,
-   since offsetting by a running maximum reintroduces ordering.
+   lived secret is introduced. Mappings are appended to across releases rather than
+   re-cut, so a row published once keeps its CT+ id for good. New ids sit above the
+   running maximum, which orders releases relative to each other, but the releases
+   ship as separate datasets so that boundary is already public.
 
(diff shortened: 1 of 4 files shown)


ksdkalluri (approved):
Tested the generated datasets and Looks good to me!

PR #1934, opened 2026-08-19
[DL-2458] CT+ version of the off-cycle HealthKit dataset (Curation, V9)

hiro-mishima:
To reviewer:
- The dataset is already created. And I logged what I ran in a SQL file in this PR. Please review the SQL file.
- See the comment in JIRA's DL-2458 for the actual project and dataset names.
- I manually ran the SQL statements for Curation. But for Warehouse, we want to have it with dbt, if possible.


nishanthpp93 (approved):
This is helpful for keeping track

PR #1937, opened 2026-08-26
[DL-2427] [DL-2428] [DL-2429] CT+ concept suppression

hiro-mishima:


data_steward/cdr_cleaner/cleaning_rules/deid/ct_additional_privacy_suppression.py
```diff
@@ -110,6 +120,27 @@ def setup_validation(self, client, *args, **keyword_args):
...
+    """
+
+    def __init__(self,
+                 project_id,
+                 dataset_id,
+                 sandbox_dataset_id,
+                 table_namer=None):
+        super().__init__(project_id, dataset_id, sandbox_dataset_id,
+                         table_namer)
+        self._concept_suppression_lookup_table += CT_PLUS_LOOKUP_SUFFIX
```
nishanthpp93: Very rare edge case: This gets a separate lookup table name, but the inherited load appends to that table. Cleanup removes it only when empty. If a concept’s CSV flag changes from `true` to `false` and the stage reruns against the same sandbox dataset, its old lookup row remains and the expanded concept is still deleted. Ideally the CT+ lookup needs to be replaced on each run.

nishanthpp93 (approved):
Not a requirement, but just to note.

data_steward/cdr_cleaner/cleaning_rules/deid/ct_additional_privacy_suppression.py
hiro-mishima: Thanks for this comment. I think it's better to fix the issue you raised now, since we might need to rerun the process a few times because of privacy review, etc..
I have updated the codes accordingly.

nishanthpp93 (approved):

PR #1938, opened 2026-08-31
[DL-2481] Coverage guard and rerun safety in `regenerate_ct_plus_ids.py`

hiro-mishima:


data_steward/tools/regenerate_ct_plus_ids.py
```diff
@@ -1265,13 +1408,22 @@ def main(input_dataset_id,
     # Load Fitbit tables if they exist
     LOGGER.info('Processing Fitbit tables (if they exist)...')
     for fitbit_table in FITBIT_TABLES:
```
nishanthpp93: Edge case: With --allow_replace, the script clears only CDM tables. If a previous attempt wrote a Fitbit table that is now absent from the input, this skips it, leaving the old output table in place while reporting success. The rerun should remove that stale table.

tests/unit_tests/data_steward/tools/regenerate_ct_plus_ids_test.py
```diff
@@ -449,3 +451,210 @@ def test_missing_person_mapping_is_accepted(self):
...
+        self.assertIn(f'{self.input_dataset_id}.{PERSON}', q)
+
+    def test_the_coverage_check_runs_before_any_output_table_is_touched(self):
+        """A run that fails it has to leave the output dataset as it found it."""
+        source = inspect.getsource(ct.main)
+
+        self.assertLess(source.index('assert_person_mapping_covers_input'),
+                        source.index('create_empty_output_table'))
+
+    def _run_main(self, discovery_tables=(), fitbit_in_input=True):
```
nishanthpp93: This mocks assert_output_dataset_is_safe and always passes allow_replace=True. The existing test checks the assert in isolation, so a regression that stops main() from refusing a non-empty output without the flag would go undetected.

nishanthpp93 (commented):
Small comments

data_steward/tools/regenerate_ct_plus_ids.py
hiro-mishima: Thanks for the comments! Both are valid. I fixed them accordingly.

PR #1939, opened 2026-09-01
[DL-2477] Source the CT+ deid map from the research IDs view

hiro-mishima:


data_steward/cdr_cleaner/cleaning_rules/deid/ct_plus_pid_rid_map.py
```diff
@@ -0,0 +1,198 @@
...
+                    RDR_PARTICIPANT_RESEARCH_IDS_VIEW)
+
+LOGGER = logging.getLogger(__name__)
+
+# A participant carrying two rows that disagree on the ID or the shift. SELECT DISTINCT
+# below collapses only byte identical duplicates, which is the shape the view has been
+# observed carrying; a disagreeing pair survives it and would leave PIDtoRID's
+# UPDATE ... FROM matching two source rows for one target row, which BigQuery rejects
+# mid-run, after output tables have already been written.
+CONFLICTING_PARTICIPANTS_TMPL = JINJA_ENV.from_string("""
```
nishanthpp93: The map query filters NULL participant_id and controlled_tier_id, but not NULL registered_tier_date_shift, contrary to the requested unit assertion. The conflict check also considers those rows, so a participant with one NULL-shift row and one valid row would abort the run instead of using the valid row. Apply the shift filter in both queries.

tests/integration_tests/data_steward/cdr_cleaner/cleaning_rules/deid/ct_plus_pid_rid_map_test.py
```diff
@@ -0,0 +1,291 @@
...
+        self.load_ids_view()
+        self.load_test_data([stale_map_query, person_query, co_query])
+
+    def read_deid_map(self):
+        """Return the rebuilt map as a sorted list of (person_id, research_id, shift)."""
+        query = f'SELECT person_id, research_id, shift FROM `{self.fq_deid_map_table}`'
+        return sorted((row['person_id'], row['research_id'], row['shift'])
+                      for row in self.client.query(query).result())
+
+    def test_deid_map_is_rebuilt_from_the_ids_view(self):
```
nishanthpp93: There is no test of the rendered setup SQL or the inherited sandbox → update → delete query order. The integration tests cover several behaviors, but leave those explicit scope items unchecked.

tests/unit_tests/data_steward/cdr_cleaner/clean_cdr_test.py
```diff
@@ -93,6 +96,25 @@ def test_controlled_tier_lists_carry_no_ct_plus_variant(self):
...
+        Position and length are covered by the copy-policy test above, through the
+        substitution declared at the top of this module.
+        """
+        ct_plus = cc.CONTROLLED_TIER_PLUS_DEID_CLEANING_CLASSES
+        ct = cc.CONTROLLED_TIER_DEID_CLEANING_CLASSES
+
+        self.assertIn((CtPlusPIDtoRID,), ct_plus)
+        self.assertNotIn((RtCtPIDtoRID,), ct_plus)
+
+        # The registered and controlled tiers keep the original rule and its source.
```
nishanthpp93: The new list assertion checks controlled tier only. In the current registered-tier cleaning list, RtCtPIDtoRID is commented out, so the stated requirement that both registered and controlled tier lists run it is not true of this baseline. These commits did not introduce that mismatch.

nishanthpp93 (commented):
Added comments

tests/integration_tests/data_steward/cdr_cleaner/cleaning_rules/deid/ct_plus_pid_rid_map_test.py
hiro-mishima: Thanks, I added a test for it.

tests/unit_tests/data_steward/cdr_cleaner/clean_cdr_test.py
hiro-mishima: You are right, I corrected the test accordingly.

data_steward/cdr_cleaner/cleaning_rules/deid/ct_plus_pid_rid_map.py
hiro-mishima: I want to keep NULL registered_tier_date_shift because of the following reasons:
1. this is for CT+ and CT+ does not need date shift
2. RDR_PARTICIPANT_RESEARCH_IDS_VIEW is relatively new for RDR too and there is a chance they might miss registered_tier_date_shift for some records. We should be able to move forward even so because of (1).

PR #1942, opened 2026-09-02
[DL-2483] Emit observation records on the adult person_id for guardian-about pediatric items

hiro-mishima:


data_steward/cdr_cleaner/cleaning_rules/route_pediatric_adult_observations.py
```diff
@@ -0,0 +1,489 @@
...
+    ROW_NUMBER() OVER (ORDER BY r.person_id, r.observation_id)
+      + (SELECT MAX(observation_id) FROM `{{project}}.{{dataset}}.{{obs}}`)
+      AS observation_id,
+    l.adult_person_id AS person_id,
+    /* The export already resolves observation_concept_id, including the cases where
+       a refusal carries 903102 PMI instead of the item's own standard concept.
+       Copying it keeps that shape; the declared standard concept only fills in when
+       the export left the row unmapped, which pediatric rows have been observed
+       doing. */
+    IF(r.observation_concept_id = 0, r.standard_concept_id, r.observation_concept_id)
```
ksdkalluri: Since all 12 are known standard observations for the adult, we should just say r.standard_concept_id AS observation_concept_id?

ksdkalluri (changes requested):
@hiro-mishima I have one small suggestion and the integration test is failing.

data_steward/cdr_cleaner/cleaning_rules/route_pediatric_adult_observations.py
hiro-mishima: @ksdkalluri Thanks, you are right. We can simply do that here instead.
And while checking if it works using the RDR dataset, I found the mapping for maritalstatus_currentmaritalstatus was incorrect. It has to be 1585892 -> 3046344 instead of 1332833 -> 1332833. I updated that as well.

As for integration test failures, it was because I rebased multiple PRs at the same time and they fought for the same table. I will make sure to not run rebase on multiple PRs at the same time going forward.

hiro-mishima: Sorry I noticed the integration test is failing with different ones now. I am fixing it now.

hiro-mishima: @ksdkalluri Integration tests are fixed, too.

ksdkalluri (approved):
Thanks for fixing Hiro, this is good to merge!

